PR #2, opened 2018-05-24
Revised Network Metrics

robgil:
Inbound and outbound are ambiguous with regard to netflow data. When
using src and destination, the direction is explicit. `inbound` and
`outbound` don't map back to `src` and `dest` and becomes ambiguous. I don't
see a use case for `inbound` and `outbound` as this is already
defined by `src` and `dest`. This removes the `inbound` and `outbound`
sections and replaces them with a single transfer field which lines up with
netflow, gcp flow logs, aws flowlogs, and so on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/elastic/ecs/compare/0.1.0...master)
 
 ### Breaking changes
+- Replaced network.inbound/outbound with less ambigious network.transfer.
 
 ### Bugfixes
 
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -10,16 +10,6 @@
       description: >
         Network protocol name.
       example: http
-    - name: direction
-      type: keyword
-      description: >
-        Direction of the network traffic.
-
-        The recommended values are:
-          * inbound
-          * outbound
-          * unknown
-      example: inbound
 
     - name: forwarded_ip
       type: ip
@@ -29,23 +19,17 @@
       example: 192.1.1.2
 
     # Metrics
-    - name: inbound.bytes
+    - name: transfer.bytes
       type: long
       description: >
-        Network inbound bytes.
+        Network bytes transferred.
+        
+        Typically used with netflow, sflow, or cloud provider flow logs
       example: 184
-    - name: inbound.packets
+    - name: transfer.packets
       type: long
       description: >
-        Network inbound packets.
-      example: 12
-    - name: outbound.bytes
-      type: long
-      description: >
-        Network outbound bytes.
-      example: 184
-    - name: outbound.packets
-      type: long
-      description: >
-        Network outbound packets.
+        Network packets transferred.
+        
+        Typically used with netflow, sflow, or cloud provider flow logs
       example: 12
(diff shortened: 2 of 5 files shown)


ruflin:
+1 on this change. Would like to get some feedback from @MikePaquette on this one.

Could you rebase this one on master to get the proper CI checks and update the CHANGELOG file?

CHANGELOG.md
```diff
@@ -11,3 +11,4 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 ### Deprecated
+- Replaced network.inbound/outbound with less ambigious network.transfer.
```
ruflin: Can you put this under `Breaking Changes`?

schemas/network.yml
```diff
@@ -29,23 +19,15 @@
...
-      description: >
-        Network inbound packets.
-      example: 12
-    - name: outbound.bytes
+    - name: transfer.bytes
       type: long
       description: >
-        Network outbound bytes.
+        Network bytes transferred. Typically used with netflow, sflow, or 
+        cloud provider flow logs
```
ruflin: We often use the first line as a quick description and then more details below. Can you reformat this as below and add a dot at the end?
```
Network bytes transferred. 

Typically used with netflow, sflow, or cloud provider flow logs.
```

Same applies to the description below.

schemas/network.yml
```diff
@@ -10,16 +10,6 @@
       description: >
         Network protocol name.
       example: http
-    - name: direction
```
ruflin: I think this field we still need?

robgil: @ruflin src and dest are explicit in the direction. This is redundant imo.

ruflin: It's a convention field to filter on. We for example use it here in auditbeat: https://github.com/elastic/beats/blob/master/auditbeat/module/auditd/_meta/fields.yml#L162

@andrewkroh can perhaps share some more background here?

robgil:
@MikePaquette I think I partially understand the reasoning. Conceptually though inbound and outbound is still ambigious as to the direction, but I think security zones is more analogous to the intent. For my own edification and for that of others, can you provide specific use cases and/or vendor products that utilize an inbound and outbound designation? I have not personally come across any vendor products which utilize this notation with the exception of perhaps firewall rulesets, but even then its a prefix list which utilizes sources and destinations. 

Another thought is that inbound and outbound can refer to RFC1918 address ranges vs. public ranges. This however breaks when doing private peering, forcing one to fall back on src/dest or src-zone to dst-zone. 

I could see the inbound/outbound concepts evolving in to security zones which could be arbitrarily defined by an operator for their particular environment and derived based on importance to the operator. 

Ex.
security.source.zone: public (loosely defined by the operator as public internet)
security.destination.zone: DMZ (loosely defined by the operator as their DMZ)

Another example with private peering..

network.source.zone: radianz
network.destination.zone: big-bank-1

Example of public peering

network.source.zone: level3 (upstream peered network)
network.destination.zone: dc5 (datacenter in dc5)

Example with ASNs (which could be considered inbound/outbound)
network.source.asn: 7224
network.destination.asn: <my public asn>

The ambiguity for me is what specifically derives inbound and outbound. Inbound and outbound to _what_? The value for `direction` is only `inbound` and `outbound`, but this could mean anything. `src` and `destination` are explicit.

ruflin:
We have an example here in beats for docker where we use inbound and outbound in the same event: https://github.com/elastic/beats/pull/7253

robgil:
@MikePaquette oooo so you're referring to aggregate inbound/outbound on a specific device/endpoint/host. That makes sense now. 

For the netflow case, I think we'll want to add the `network.total_packets` and `network.total_bytes` since netflow is not an aggregate. It might also be worth noting that inbound/outbound is an aggregation for a specific device/server/container/router/switch. 

I can clarify that this is an aggregation for the device and add the `network.total_*` stuff if that makes sense?

PR #3, opened 2018-05-28
Add log.message

ruflin:
The field `log.message` contains the full log message before splitting it up in multiple parts.

In contrast to the `message` field which can contain an extracted part of the log message, this field contains the original, full log message. It can have already some modifications applied like encoding or new lines removed to clean up the log message.

This field is not indexed and doc_values are disabled so it can't be
queried but the value can be retrieved from `_source`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,5 +13,6 @@ All notable changes to this project will be documented in this file based on the
 * Adds cloud.account.id for top level organizational level. #11
 * Add `http.response.status_code` and `http.response.body` fields. #4
 * Add fields for Operating System data. #5
+* Add `log.message`. #3
 
 ### Deprecated
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ Fields which are specific to log events.
 | <a name="log.level"></a>`log.level`  | Log level of the log event.<br/>Some examples are `WARN`, `ERR`, `INFO`.  | keyword  |   | `ERR`  |
 | <a name="log.line"></a>`log.line`  | Line number the log event was collected from.  | long  |   | `18`  |
 | <a name="log.offset"></a>`log.offset`  | Offset of the beginning of the log event.  | long  |   | `12`  |
+| <a name="log.message"></a>`log.message`  | This is the log message and contains the full log message before splitting it up in multiple parts.<br/>In contrast to the `message` field which can contain an extracted part of the log message, this field contains the original, full log message. It can have already some modifications applied like encoding or new lines removed to clean up the log message.<br/>This field is not indexed and doc_values are disabled so it can't be queried but the value can be retrieved from `_source`.  | keyword  |   | `Sep 19 08:26:10 localhost My log`  |
 
 
 ## <a name="network"></a> Network fields
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -21,3 +21,20 @@
       description: >
         Offset of the beginning of the log event.
       example: 12
+    - name: message
+      type: keyword
+      phase: 1
+      example: "Sep 19 08:26:10 localhost My log"
+      index: false
+      doc_values: false
+      description: >
+        This is the log message and contains the full log message before
+        splitting it up in multiple parts.
+
+        In contrast to the `message` field which can contain an extracted part
+        of the log message, this field contains the original, full log message.
+        It can have already some modifications applied like encoding or new
+        lines removed to clean up the log message.
+
+        This field is not indexed and doc_values are disabled so it can't be
+        queried but the value can be retrieved from `_source`.
--- a/template.json
+++ b/template.json
@@ -459,6 +459,12 @@
             "line": {
               "type": "long"
             },
+            "message": {
+              "doc_values": false,
+              "ignore_above": 1024,
+              "index": false,
+              "type": "keyword"
+            },
             "offset": {
               "type": "long"
             }
--- a/schema.csv
+++ b/schema.csv
@@ -88,6 +88,7 @@ kubernetes.namespace,keyword,0,
 kubernetes.pod.name,keyword,0,
 log.level,keyword,0,ERR
 log.line,long,0,18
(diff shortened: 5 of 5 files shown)


ruflin:
Thanks a lot for the feedback, that is super useful. Currently the way I differentiate between logs and metrics is that metrics are pulled on a predefined period and logs are pushed, but this line is very blurry. Based on this the above would also be logs I'm thinking. But I see that a user is in this case not going to look for the "raw event" under `log`. 

The initial reason I wanted to get it out of event prefix was that it felt like the only object in there that does not contain meta information but actual data. I'm also ok with leaving where it is.

@webmat Out of curiosity: Would you categorise operational events under logs, metrics or it's own category?

ruflin:
@webmat Working on https://github.com/elastic/beats/pull/7207 I realised `event.raw` and `log.message` might be even two different fields with different purpose. In https://github.com/elastic/beats/pull/7207 I do some processing on the field like encoding, multiline and strip new lines as the target is to make it possible to reprocess the event. In `event.raw` I would expect a completely unprocessed event. Perhaps we ned both?

webmat:
>? @webmat Out of curiosity: Would you categorise operational events under logs, metrics or it's own category?
For t type of data I was dealing with, I would definitely say "events", in the sense that it was not just numerical data. I was processing email addresses, IP addresses, email headers and email categories.  So full on events with lots of juicy data ;-)

On the other hand, I don't have a strong attachment to the "event" part of event.raw. It could totally be something else equally generic, like `raw_payload` or the like. The part I was reacting to was losing the "generic"-ness of "event" and going to "log", when not all operational events come from logs\*.

\* Well, depending on how broad you define log. But I think most people think of text files, when they hear "log". And "most people" is our audience for this ;-)

ave19:
We have some data sources that aren't logs.  Some are data base scrapes and some are API calls we make to a service and then index what comes out.  Not sure `log.message` would apply as it was never in a log and was really never a message either.  They do describe things that happened, or at least facts that were true at a particular time, though, so `event` seems ok.

Indexing the original data (before it is parsed and mutilated) as `*.raw` would allow us to re-parse events later if we find a bug or gain some other new insight.  We wouldn't have to go back to the original source, which may or may not have the information anymore.

ruflin:
Based on the feedback above I reverted the change to `event.raw`. The idea for this one is to really contain the raw initial event.

For `log.message` I think it fits the logging purpose and is potentially already modified a bit like encoding applied, new lines removed, multiline applied so it does not full fit `event.raw`. 

My suggestion is now to only introduce `log.message`.

webmat:
Yes that part I agree with. Makes total sense, especially wrt reconstructed messages (like multiline).

schemas/log.yml
```diff
@@ -21,3 +21,14 @@
...
+    - name: message
+      type: keyword
+      phase: 1
+      example: "Sep 19 08:26:10 localhost My log"
+      description: >
+        This is the raw log message and contains the full log message before processing.
+
+        In contrast to the `message` field which can contain an extracted part
+        of the log message, this field contains the raw log message and should
+        not be processed. It can have already some modifications like encoding
```
praseodym: This sounds like a contractiction: 'raw log message' and 'not be processed' vs. 'can have already some modifications'

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Breaking changes
 
+* Rename `event.raw` to `log.message`. #3
```
praseodym: This change should be removed now that `event.raw` stays as is.

schemas/log.yml
ruflin: Agree. Any suggestion for a better description instead of raw we could use here?

CHANGELOG.md
ruflin: fixed, thanks for the review.

README.md
```diff
@@ -275,6 +275,7 @@ Fields which are specific to log events.
 | <a name="log.level"></a>`log.level`  | Log level of the log event.<br/>Some examples are `WARN`, `ERR`, `INFO`.  | keyword  |   | `ERR`  |
 | <a name="log.line"></a>`log.line`  | Line number the log event was collected from.  | long  |   | `18`  |
 | <a name="log.offset"></a>`log.offset`  | Offset of the beginning of the log event.  | long  |   | `12`  |
+| <a name="log.message"></a>`log.message`  | This is the log message and contains the full log message before splitting it up in multiple parts.<br/>In contrast to the `message` field which can contain an extracted part of the log message, this field contains the raw log message and should not be processed. It can have already some modifications like encoding applied or new lines removed to clean up the log message.<br/>This field is not index and doc_values are disabled so it can't be queried but the value can be retrieved from `_source`.  | keyword  |   | `Sep 19 08:26:10 localhost My log`  |
```
praseodym: "not index" -> "not indexed"

praseodym: Also, w/ regards to [my previous comment](https://github.com/elastic/ecs/pull/3#discussion_r192311911): "[..] this field contains the raw log message and should not be processed." -> "[..] this field contains the original, full log message."

ruflin: applyed

ruflin:
PR rebased, commit and PR message updated, fixes applied. Read for an other review.

template.json
```diff
@@ -459,6 +459,12 @@
             "line": {
               "type": "long"
             },
+            "message": {
+              "doc_values": false,
+              "ignore_above": 1024,
```
andrewkroh: For logs, `ignore_above: 1024` will be too small.

ruflin: I checked the docs for this and run some tests: https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html

It seems like `ignore_above` does not play any role here as it's not index anyway. `_source` is not affected by `ignore_above`.

andrewkroh: Good. Thanks for testing.

Maybe we shouldn't even set an `ignore_above` when `index: false`? I imagine others having a similar reaction to mine.

PR #5, opened 2018-05-29
Add fields for Operating System data

ruflin:
There were some previous discussions around the version field and if it should split up into major, minor, bugfix, patch. I chose to only go with `version` based on the assumption that if someone searches for a specific minor version he will know the major version. So searching for it can be done with a prefix query based on version keyword.


ruflin:
@andrewkroh Rebased and https://github.com/elastic/ecs/pull/15 opened to reduce conflicts in the CHANGELOG.

PR #6, opened 2018-05-29
Proposal: TLS related fields

ph:
While working on the TCP inputs for beats I am able to extract information about the TLS envelope,
this might be useful information for post-mortems or audit. Also, this was a feature requested on
the beats input on the Logstash side.

We are also extracting some information in [packetbeat](https://github.com/elastic/beats/blob/master/packetbeat/protos/tls/_meta/fields.yml) from the work of @adriansr.

Concerning the TCP or the beats input, I think we can extract the following field and provide some
useful information.

- TLS version
- Remote Certificates (mutual auth)
- ServerName
- Cipher used.

Where OR should this information be in the ECS schema? Maybe we could add a `transport` top level
key and have something like this.

| Field  | Description  | Type  | Example  |
|---|---|---|---|
tls.version| string representation of the tls version| keyword|	TLSv1.1 |
tls.certificates| array of certificates used by the client.|text| raw certificate|
tls.servername| server name requested by client|keyword| localhost[1]
tls.ciphersuite| Name of the cipher used| keyword|  ECDHE-ECDSA-AES-128-CBC-SHA

[1]:https://en.wikipedia.org/wiki/Server_Name_Indication

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ ECS defines these fields.
  * [Process fields](#process)
  * [Service fields](#service)
  * [Source fields](#source)
+ * [TLS fields](#tls)
  * [URL fields](#url)
  * [User fields](#user)
  * [User agent fields](#user_agent)
@@ -356,6 +357,21 @@ Source fields describe details about the source of the event.
 | <a name="source.mac"></a>`source.mac`  | MAC address of the source.  | keyword  |   |   |
 | <a name="source.domain"></a>`source.domain`  | Source domain.  | keyword  |   |   |
 | <a name="source.subdomain"></a>`source.subdomain`  | Source subdomain.  | keyword  |   |   |
+
+
+## <a name="tls"></a> TLS fields
+
+The tls fields contains the TLS related data about a specific connection.
+
+As an example in the case of Filebeat and the TCP input, the `version` field would be the version of the TLS protocol in use, the `certificates` would be the chain of certificates provided by the client and the `ciphersuite` is the encryption algorithm used for the communication.
+
+
+| Field  | Description  | Type  | Multi Field  | Example  |
+|---|---|---|---|---|
+| <a name="tls.version"></a>`tls.version`  | TLS version  | keyword  |   | `TLSv1.2`  |
+| <a name="tls.certificates"></a>`tls.certificates`  | An array of certificates.  | keyword  |   |   |
+| <a name="tls.servername"></a>`tls.servername`  | Server name requested by the client.  | keyword  |   | `localhost`  |
+| <a name="tls.ciphersuite"></a>`tls.ciphersuite`  | Name of the cipher used for the communication.  | keyword  |   | `ECDHE-ECDSA-AES-128-CBC-SHA`  |
 
 
 ## <a name="url"></a> URL fields
--- a/README.md
+++ b/README.md
@@ -361,17 +361,18 @@ Source fields describe details about the source of the event.
 
 ## <a name="tls"></a> TLS fields
 
-The tls fields contains the TLS related data about a specific connection.
-
-As an example in the case of Filebeat and the TCP input, the `version` field would be the version of the TLS protocol in use, the `certificates` would be the chain of certificates provided by the client and the `ciphersuite` is the encryption algorithm used for the communication.
+The tls fields contain the TLS related data about a specific connection.
 
 
 | Field  | Description  | Type  | Multi Field  | Example  |
 |---|---|---|---|---|
-| <a name="tls.version"></a>`tls.version`  | TLS version  | keyword  |   | `TLSv1.2`  |
+| <a name="tls.version"></a>`tls.version`  | TLS version.  | keyword  |   | `TLSv1.2`  |
 | <a name="tls.certificates"></a>`tls.certificates`  | An array of certificates.  | keyword  |   |   |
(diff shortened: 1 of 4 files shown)


ph:
@ruflin Good job, the tooling for this repository works great. 🎉

schemas/tls.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: version
+      type: keyword
+      description: >
+        TLS version
+
+      example: TLSv1.2
+    - name: certificates
+      type: text
+      description: >
+        An array of certificates.
```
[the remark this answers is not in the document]
ph: You are absolutely right for the keyword, I’ll update the PR.

ph:
@ruflin to move the certificate type from a text field to a keyword I will need to make a small change on the template generator, we have the `ignore_above` value hardcoded, I think we should allow users to override that from the schema definition. I need to do some checks on the openssl documentation but it can be over 1024.

I don't expect any problems since the ES default is much higher.

```
Do not index any string longer than this value. Defaults to 2147483647 so that all values would be accepted. 
```

ruflin:
@ph We should add support for `ignore_above` config in field.go: https://github.com/elastic/beats/blob/master/libbeat/common/field.go#L36

ph:
@ruflin The required support for `ignore_above` was merged into master, I've rebased and regenerated the files in this PR. I've gone with a 4096 upper bound limit on the base64 certificate; I think it should cover the maximum size, look at the commit message for more details.

praseodym:
Is there a reason to specify a value for `ignore_above`? [Current documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) doesn't list any advantages.

It is very common for certificates to be >4096 bytes, e.g. check out [this certificate for elastic.co](https://crt.sh/?id=517752464) at 4885 bytes or [this certificate for microsoft.com](https://crt.sh/?id=391432857) at 5697 bytes. There is no set limit on a certificate's size; e.g. [the number of subject alternative names is unbounded](https://tools.ietf.org/html/rfc5280#section-4.2.1.6).

ph:
Right, thanks for enlightening me,  I am OK to remove it relies on ES [keyword default](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html) which will be in this case  `2147483647`, I will need to make changes on the template generator on libbeat, since there is no way to opt out of it.

@ruflin what you think if we make it an explicit choice, so users have to use -1 in the fields.yaml?

ph:
@ruflin Also, defining this kind of size limits should be in another document we generated like the CSV, I don't think this information should only be kept in the template. This could be part of a new discussion or issue.

ruflin:
We started to set for some fields index: false and `doc_values: false` (see https://github.com/elastic/ecs/blob/master/schemas/event.yml#L73) I wonder if this also applies to certificate as it will not be a field that anyone queries on I would assume?

@ph +1 on having the option in the generator to disable it. For the discussion on where these things should show up, let's move it to an other Github issue as proposed.

ph:
@ruflin I think it would make sense in that context to set `doc_values` to false, we won't sort or aggregate on that field. So that would be a space saver.

ph:
Setting `index: false` will make it [not queryable](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-index.html), in the context of certificates, querying it can be useful when doing forensic, no?

What is your reasoning for not indexing it?

ph:
Created https://github.com/elastic/ecs/issues/23 for exposing keyword limit.

ruflin:
My assumption was that there is not going to be a query on the certificate field but as you mentioned above, there are potential queries. So index should be set to true.

ph:
@ruflin updated the PR with the latest review.

README.md
```diff
@@ -358,6 +359,21 @@ Source fields describe details about the source of the event.
 | <a name="source.subdomain"></a>`source.subdomain`  | Source subdomain.  | keyword  |   |   |
 
 
+## <a name="tls"></a> TLS fields
+
+The tls fields contains the TLS related data about a specific connection.
```
ruflin: s/contains/contain/

schemas/tls.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    As an example in the case of Filebeat and the TCP input, the `version` field would be the
+    version of the TLS protocol in use, the `certificates` would be the chain of certificates
+    provided by the client and the `ciphersuite` is the encryption algorithm used for the
+    communication.
+  fields:
+    - name: version
+      type: keyword
+      description: >
+        TLS version
```
ruflin: Can you add a `.` at the end?

README.md
```diff
@@ -358,6 +359,21 @@ Source fields describe details about the source of the event.
...
+
+As an example in the case of Filebeat and the TCP input, the `version` field would be the version of the TLS protocol in use, the `certificates` would be the chain of certificates provided by the client and the `ciphersuite` is the encryption algorithm used for the communication.
```
karenzone: Consider tagging as `footnote: >`to push this example to after the table

ruflin:
@ph I think you need to run `make` again.

ph:
@ruflin @karenzone added footnote and run make again, I didn't know about footnote.

ruflin:
@ph `footnote` is something we just introduced a few days ago.

PR #7, opened 2018-05-30
Change structure of URL

ruflin:
So far the url structure was heavily inspired by https://github.com/whatwg/url/issues/337. I initially only wanted to make some tweaks to it to improve querying but I realised I never fully felt comfortable with the field names used here. So I started to look at the url parser of different languages like Go, Ruby, Python and the output they provide are surprisingly similar but not consistent with whatwg. The change made here brings the field names closer to what most url parsers output.


README.md
```diff
@@ -346,26 +346,23 @@ A complete URL, with scheme, host, and path.
...
+| <a name="url.scheme"></a>`url.scheme`  | The scheme of the request, e.g. "https".<br/>Note: The `:` is not part of the scheme.  | keyword  |   | `https`  |
+| <a name="url.hostname"></a>`url.hostname`  | The hostname of the request, e.g. "example.com".<br/>For correlation the this field can be copied into the `host.name` field.  | keyword  |   | `elastic.co`  |
+| <a name="url.port"></a>`url.port`  | The port of the request, e.g. 443.  | long  |   | `443`  |
+| <a name="url.path"></a>`url.path`  | The path of the request, e.g. "/search".  | text  |   |   |
+| <a name="url.path.raw"></a>`url.path.raw`  | The url path. This is a non-analyzed field that is useful for aggregations.  | keyword  | 1  |   |
+| <a name="url.query"></a>`url.query`  | The search describes the query string of the request, e.g. "q=elasticsearch".<br/>The `?` is excluded from the query string. In case an URL contains no `?` it is expected that the query field is left out. In case there is a `?` but no query, the query field is expected to exist with an empty string. Like this the `exists` query can be used to differentiate between the two cases.  | text  |   |   |
```
praseodym: “The search” still references the old field name.

README.md
```diff
@@ -346,26 +346,23 @@ A complete URL, with scheme, host, and path.
...
-| <a name="url.pathname"></a>`url.pathname`  | The path of the request, e.g. "/search".  | text  |   |   |
-| <a name="url.pathname.raw"></a>`url.pathname.raw`  | The url path. This is a non-analyzed field that is useful for aggregations.  | keyword  | 1  |   |
-| <a name="url.search"></a>`url.search`  | The search describes the query string of the request, e.g. "q=elasticsearch".  | text  |   |   |
-| <a name="url.search.raw"></a>`url.search.raw`  | The url search part. This is a non-analyzed field that is useful for aggregations.  | keyword  | 1  |   |
-| <a name="url.hash"></a>`url.hash`  | The hash of the request URL, e.g. "top".  | keyword  |   |   |
+| <a name="url.scheme"></a>`url.scheme`  | The scheme of the request, e.g. "https".<br/>Note: The `:` is not part of the scheme.  | keyword  |   | `https`  |
+| <a name="url.hostname"></a>`url.hostname`  | The hostname of the request, e.g. "example.com".<br/>For correlation the this field can be copied into the `host.name` field.  | keyword  |   | `elastic.co`  |
+| <a name="url.port"></a>`url.port`  | The port of the request, e.g. 443.  | long  |   | `443`  |
```
praseodym: The `integer` type should be large enough, is there any reason for going with `long` instead?

README.md
ruflin: fixed

README.md
ruflin: In ES it does not really matter from a compression perspective if it is set as integer or long so we default to long. Agree that it is not needed here but it also does not make a difference.

schemas/url.yml
```diff
@@ -8,41 +8,46 @@
...
       type: keyword
       description: >
         The hostname of the request, e.g. "example.com".
 
         For correlation the this field can be copied into the `host.name`
         field.
+      example: elastic.co
     - name: port
-      type: keyword
+      type: long
```
Mpdreamz: `integer` since port numbers are unsigned 16-bit integers? Nitpicking :)

Mpdreamz: And again already covered in a previous comment, I'll see myself out :)

README.md
Mpdreamz: Still makes sense IMO to go with integer to signal to anyone implementing ECS the appropriate type to use in their implementation.

PR #11, opened 2018-05-31
Adds cloud.account.id

robgil:
The cloud.account.id can be a common field. All cloud providers have a
high level account id, org id, or subscription id that can be used in
this field. Anything more specific such as GCP's project id should go in
an extended schema for the specific cloud provider. Its important to put
GCP's project id in to a separate extended schema, because it does not
overlap conceptually with other cloud providers' structure.

For example. In GCP, you'll have an `organization id`. Under that
`organization id`, you can have multiple projects. This diverges quite
substantially from the flat AWS structure which is just a bunch of
linked accounts. For AWS, there is no subtree like GCP has for projects
or folders.

Azure has a top level organization as well. Under that there are
[subscriptions](https://docs.microsoft.com/en-us/office365/enterprise/subscriptions-licenses-accounts-and-tenants-for-microsoft-cloud-offerings).

For security analytics use cases it will be difficult to normalize
across all cloud providers, but account id could be the most basic
start. Since high level orgs are pretty much consistent across all cloud
providers, this could be in the common schema.

_I'm open to calling this organization.id also_

### DIFF
--- a/schemas/cloud.yml
+++ b/schemas/cloud.yml
@@ -55,5 +55,4 @@
         The cloud account or organization id
 
         This could be the AWS account id or Google Cloud ORG Id. This should be
-        organizational. Other things like GCP project id should be in an
-        extended schema for the specific cloud provider.
+        organizational. 
--- a/schemas/cloud.yml
+++ b/schemas/cloud.yml
@@ -47,3 +47,13 @@
       example: t2.medium
       description: >
         Machine type of the host machine.
+
+    - name: account.id
+      type: keyword
+      example: 666777888999
+      description: >
+        The cloud account or organization id
+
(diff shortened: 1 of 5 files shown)


schemas/cloud.yml
```diff
@@ -47,3 +47,13 @@
...
       description: >
         Machine type of the host machine.
+
+    - name: account.id
+      type: keyword
+      example: 666777888999
+      description: >
+        The cloud account or organization id
+
+        This could be the AWS account id or Google Cloud ORG Id. This should be
```
ruflin: Let's remove the `should be in an extended schema` part here as we don't have that yet.

PR #16, opened 2018-06-06
Disable doc_values and indexing for event.raw

ruflin:
This should reduce the storage need for the event.raw field.


ruflin:
@andrewkroh The same should be applied to the `log.message` field.

PR #17, opened 2018-06-06
Update scripts for Python 3 compatibility

praseodym:
Fixes `make` on systems where Python 3 is default.


ruflin:
@praseodym Thanks, appreciate the fix.

PR #18, opened 2018-06-06
Rename url.href multi_field for consistency

praseodym:
Non-analyzed `keyword` fields are currently in `.raw`, except for `url.href` where there is a `url.href.analyzed` field. This PR fixes that.


ruflin:
Agree, probably good to standardise on a common behaviour here.

Could you update the CHANGELOG under breaking changes?

praseodym:
@ruflin I've update the CHANGELOG

PR #22, opened 2018-06-11
[DISCUSS]  Edits to schema YAML files

karenzone:
Schema file edits (YAML)
Contains `schema.csv` and `template.json` which I did not mean to include.  Thought I had unstaged them, but it didn't work.

### DIFF
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -19,5 +19,22 @@
     - name: offset
       type: long
       description: >
-        Offset from the beginning of the log event.
+        Offset of the beginning of the log event.
       example: 12
+    - name: message
+      type: keyword
+      phase: 1
+      example: "Sep 19 08:26:10 localhost My log"
+      index: false
+      doc_values: false
+      description: >
+        This is the log message and contains the full log message before
+        splitting it up in multiple parts.
+
+        In contrast to the `message` field which can contain an extracted part
+        of the log message, this field contains the original, full log message.
+        It can have already some modifications applied like encoding or new
+        lines removed to clean up the log message.
+
+        This field is not indexed and doc_values are disabled so it can't be
+        queried but the value can be retrieved from `_source`.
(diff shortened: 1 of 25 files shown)


schemas/log.yml
```diff
@@ -19,22 +19,5 @@
     - name: offset
       type: long
       description: >
-        Offset of the beginning of the log event.
+        Offset from the beginning of the log event.
       example: 12
-    - name: message
```
ruflin: Not sure why this one is removed but it shouldn't be. Can you revert that part? Probably happened during rebasing.

schemas/os.yml
```diff
@@ -2,14 +2,13 @@
...
-
-    The os fields are often used inside other prefixes like `host.os.*` or `user_agent.os.*`.
+    The OS fields contain information about the operating system.
+    These fields are often used inside other prefixes, such as `host.os.*` or `user_agent.os.*`.
   fields:
     - name: platform
       type: keyword
       description: >
-        Operating system platform (e.g. centos, ubuntu, windows).
+        Operating system platform (such centos, ubuntu, windows).
```
ruflin: Should this be `such as` like below?

schemas/service.yml
```diff
@@ -17,9 +14,10 @@
         Unique identifier of the running service.
 
         This id should uniquely identify this service. This makes it possible
-        to correlate logs and metrics for one specific service. For example
-        in case of issues with one redis instance, it's possible to filter on
-        the id to see metrics and logs for this single instance.
+        to correlate logs and metrics for one specific service. 
+        
+        Example: If you are experiencing issues with one redis instance, you
```
ruflin: Redis should probably be upper case.

schemas/user_agent.yml
```diff
@@ -15,45 +13,41 @@
...
     - name: os.major
       type: long
       description: >
-        The major version of the operating system.
+        Major version of the operating system.
     - name: os.minor
       type: long
       description: >
-        The minor version of the operating system.
-    - name: os.name
```
ruflin: Nice find. I think I should add a check to the script that throws an error if we have a field twice.

PR #28, opened 2018-06-26
Editing first pass for CONTRIBUTING.md

karenzone:
Summary:
- Edited text
- Chunked instructions into steps and sub-steps
- Added "Guidelines" section for naming fields, writing fields, etc.  (Pulled from README.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ happen through Git and Pull Requests.
 
 ## Requirements
 
-To contribute to ECS the following tools are expected to be running on your machine:
+You need these tools to contribute to ECS:
 
 * [Git](https://git-scm.com/)
 * [Python](https://www.python.org/)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,43 +6,57 @@ happen through Git and Pull Requests.
...
+Here are the steps for contributing to ECS.
+
+* Set up your git environment.
+  - Create [your own fork](https://help.github.com/articles/fork-a-repo/) of the ECS repo.
+  - Clone your fork to your machine.
+* Create a working branch on your machine.
+  - Run `git checkout -b branch-name`, where `branch-name` is the name you want to give your local branch
+* Do your work. 
+  - Make changes to the `.yml` files as needed
+* Run `make` to update generated files such as `schema.csv` and `schema.md`
```
ruflin: There is no `schema.md` anymore. Perhaps we can just mention here that this will update the generated files.

This is probably also a good place to mention that the README is fully generated and data has to changed under `docs`.

CONTRIBUTING.md
```diff
@@ -6,43 +6,57 @@ happen through Git and Pull Requests.
...
+* Commit your changes locally.
```
ruflin: I wonder if we really still need that? The reason I put it here is that I hope not only people with Git experience Contribute. Perhaps we can mention here instead that contributions can also be done through opening a Github issue?

CONTRIBUTING.md
```diff
@@ -52,22 +66,47 @@ The structure of the of each document looks as following:
       phase: 0
 ```
 
-Each namespace has it's own file to keep the files itself small. Each namespace contains a list of fields which has all the fields inside. `title` and `description`  are used to describe the namespace. `level` is for pure sorting purpose in the documentation output.
+Each namespace has its own file. The fields for each namespace are stored in the file. `title` and `description` describe the namespace. `level` is for sorting in the documentation output.
```
ruflin: We should remove all mentions of `namespace` here and replace it by `prefix`.

CONTRIBUTING.md
```diff
@@ -52,22 +66,47 @@ The structure of the of each document looks as following:
...
 
-Each field under `fields` has first the field `name`. The `type` is the [Elasticsearch field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html). `description` is used to add details about the field itself. With `example` an example value can be provided. The `phase` field is used to indicate in which `phase` the current field is (more about this below). In case `phase` is left out, it defaults to 0.
+In the `fields` section, the `name` of the field is the first entry. 
+The `type` is the [Elasticsearch field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html). 
+`description` adds details about the field. 
+`example` adds an sample value.
+The `phase` field indicates the development status of the field. If `phase` is left out, the default is 0. See
```
ruflin: We now also added `footnote`. We should mention that here too.

CONTRIBUTING.md
karenzone: Git can be intimidating. Unless this is misleading or wrong, it's good to have it in there--especially if you want to invite new people to contribute.  Experienced users can scan the list and keep going.  

I added an "Other ways to contribute" section based on your comment about inviting contributors who are not experienced with Git.  Let me know what you think.

CONTRIBUTING.md
```diff
@@ -1,48 +1,68 @@
...
+  
+**Important:** Be sure to push changes only to your own fork. Changes must be approved before they are merged into the main repository.
+
+### Other ways to contribute
+
+You can contribute even if you are not an experienced Git user. You'll need a github.com account.
+* Go to the ECS repo: https://github.com/elastic/ecs
+* Click `Issues` in the nav bar under the repo name.
+* Click `New issue`.
+* Add your text, and click `Submit new issue`.
```
karenzone: @ruflin I added an "Other ways to contribute" section based on your comment about encouraging inexperienced Git users to contribute.  Let me know what you think. 

Are there any other guidelines or requirements (other than what are in this document) that we want to suggest for submitting issues?

CONTRIBUTING.md
```diff
@@ -1,48 +1,68 @@
...
```
ruflin: Let's add a note here that users should provide as many details as possible in the description of an issue. Sounds obvious but ...

CONTRIBUTING.md
```diff
@@ -1,48 +1,68 @@
...
 ## Requirements
 
-To contribute to ECS the following tools are expected to be running on your machine:
+You need these tools to contribute to ECS:
 
 * [Git](https://git-scm.com/)
 * [Python](https://www.python.org/)
 * [Go](https://golang.org/)
 
 ## Steps to contribute
```
ruflin: Should we have a main title `Contribute` and below the two ways. Not sure yet how we should call the two. Ideas?

karenzone: I was thinking of "Steps to contribute" as the *preferred* (main) path for people who are capable of contributing through a PR, with "Other ways to contribute" as an alternate (if...else) path. That's why I formatted "Other ways" as a subheading to "Steps...".
If you feel strongly that contributing through a PR and contributing through an issue should be at the same level heading, I will make that happen.

ruflin: I'm on the fence about this one. For code parts I like PR's directly as they make discussing easier. But this is not really code and often an issue is the right part to get started as it's more question based initially. 

But in the end I don't think it will make a real difference on what we call it here. So let's move forward with the way it is.

karenzone: Thanks for taking time to explain.  Your insights help me understand the project better, and will help me add more value going forward.

PR #30, opened 2018-06-28
Add network.name to track network names with human language :-)

### LINKED ISSUE #25: Add network.name
Some operators take care to identify their networks in their monitoring pipelines. I think `network.name` should be the right place to put this.

I'd like to submit my first PR to ECS for this. I've already identified a few places where I should add the field:

* README.md
* schemas/network.yml
* schema.csv

## Questions

* Do you agree we should add `network.name`?
* Where else do I need to add the field? Any gotchas?
* In schema.csv, what's the meaning of the "Phase" column? What value would you suggest be put there. I suspect `network.name` will **not** be the most used field.

webmat:
Closes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 * Add `network.total.packets` and `network.total.bytes` field. PR#2
 * Add `event.action` field. #21
-* Add `network.name`, to track network names right in the monitoring pipeline. #25
+* Add `network.name`, to track network names in the monitoring pipeline. #25
 * Adds cloud.account.id for top level organizational level. #11
 * Add `http.response.status_code` and `http.response.body` fields. #4
 * Add fields for Operating System data. #5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 * Add `network.total.packets` and `network.total.bytes` field. PR#2
 * Add `event.action` field. #21
+* Add `network.name`, to track network names right in the monitoring pipeline. #25
 * Adds cloud.account.id for top level organizational level. #11
 * Add `http.response.status_code` and `http.response.body` fields. #4
 * Add fields for Operating System data. #5
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -5,6 +5,12 @@
   description: >
     Fields related to network data.
   fields:
+    - name: name
+      type: text
+      description: >
+        Name given by operators to sections of their network.
+      example: Guest Wifi
+
     - name: protocol
       type: keyword
       description: >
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 * Add `event.action` field. #21
+* Add `network.name`, to track network names right in the monitoring pipeline. #25
```
andrewkroh: I suggest striking `right` from the sentence.

schemas/network.yml
```diff
@@ -5,6 +5,12 @@
   description: >
     Fields related to network data.
   fields:
+    - name: name
+      type: text
```
andrewkroh: I would have chosen keyword here because I can see myself wanting to aggregate on this term. I personally wouldn't need to have this value analyzed/searchable, but maybe someone with lots of networks would.

So perhaps we should make this a keyword and reserve the option of making it a multi-field in the future if someone wants it to be searchable.

webmat: Yes I totally agree we need to be able to aggregate on this field.

However this is absolutely a field I would want analyzed as well. E.g. Search for all "Guest" or "DMZ" networks. You don't even need such a big organization for this to make sense. As soon as you have more than one physical location you'll have multiple guest networks, as soon as you have more than one application (if you're doing things right), you'll have more than one DMZ, application, private networks. Having this field analyzed lets people filter on those in a simple manner.

I haven't looked actively, but is ECS trying to steer away from having fields indexed both ways? I'm personally not a huge fan of how things ended up, with the "field" and "field.keyword" convention. But I find I often want both types of indexing.

ruflin: What we ended up with for now is having indexed by default and have the `.raw` convention instead. See for example: https://github.com/elastic/ecs/blob/master/schemas/url.yml#L26

More then happy to get some alternative suggestions on how to deal with it.

+1 on having this field as keyword and text.

webmat: Ok thanks @ruflin, I see how multi-fields is done in the schemas yml files now.

Side note: I notice that the description of field `href` seems to say the opposite of what the actual schema is.

Description says:

* `url.href` is keyword
* `url.href.analyzed` is text

Schema says:

* `url.href` is text
* `url.href.raw` is keyword

;-)

praseodym: Looks like the changes from #18 were overridden by #22, which was opened before that change got merged.

ruflin: @webmat @praseodym  Thanks, that was not intentional. 

@karenzone Could you have a look at that?

ruflin: @webmat Do you plan to make it multi field in this PR or should merge as is and add multi field later?

webmat: Well I wanted to wait for us to have this discussion next Monday. I want to discuss the  naming convention for multi field and confusion I've seen it create in a past life.

webmat:
One could say this was a long time coming, but this is ready for another review, and should be ready to go now ;-)

Issue #37, opened 2018-07-07
Session object


willemdh:
So I need quite a few session related fields for my F5 project. Do I place it in my new f5 object or should session deserve it's own tlf (top level field ;) )

For example:

```
        "f5": {
            "session": {
              "properties": {
                "id": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "key": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "value": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "bytes_in": {
                  "type": "long"
                },
                "bytes_out": {
                  "type": "long"
                },
                "client_ip": {
                  "type": "ip"
                },
                "deleted_reason": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "listener": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "location": {
                  "ignore_above": 1024,
                  "type": "keyword"
                },
                "vip_ip": {
                  "type": "ip"
                }
              }
            }
```

ruflin:
I wonder if quite a few of the above fields already map to ECS under `network.*` for the bytes part. Do you have some other use cases then F5 in mind that could use the session prefix?

willemdh:
For now I have no other use cases then F5. The most important field for session info, is f5.session.id which contains the id tha allows for F5 event correlation. I could place this in event.id. I think I can grok location into geoip.* fields.
bytes_in, bytes_out and client_ip could indeed get mapped to network.* fields
But what about deleted_reason, listener and reputation, vip ip..

willemdh:
@ruflin Hmm just noticed `event.id` documentation says: `Unique ID to describe the event.`

But the session is not unique for an event. Multiple events ould have the same session id..  So maybe this isn't so ideal after all..

MikePaquette:
@willemdh yes, event.id is not meant to contain the session ID.  Our intent was that the `network.*` fields would include common fields relating to sessions/connections.  It would appear that a session or connection id should be added, for example `network.session_id`.  Also a `network.vip_ip` could also make sense.

Can you point to F5 doc that specs out these fields with their descriptions and any recommended values, so we can better understand `deleted_reason`, `listener`, and `reputation`?

ruflin:
@willemdh I opened https://github.com/elastic/ecs/issues/52 to discuss on how we could best share things like an f5 example data structure without indicating it's part of ECS.

[...]

PR #43, opened 2018-07-11
New network fields

MikePaquette:
Added total.packets and total.bytes per discussion in PR #2 
Added session_id and virtual_ip per discussion in Issue #37 
Total of 4 new fields added.

### DIFF
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -27,6 +27,22 @@
         Host IP address when the source IP address is the proxy.
       example: 192.1.1.2
 
+    - name: session_id
+      type: keyword
+      description: >
+        This is the session ID or connection ID,
+        a unique identifier for the session /connection.
+        Bro conn.log uid field gets mapped here.
+      example: 00beeff00d1234
+
+    - name: virtual_ip
+      type: ip
+      description: >
+        The virtual IP associated with a session or connection.
+        Useful when load balancers, firewalls, and routers use a
+        virtual IP to "advertise" a service.
+      example: 192.168.1.1
+
     # Metrics
     - name: inbound.bytes
       type: long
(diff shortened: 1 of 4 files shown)


schemas/network.yml
```diff
@@ -27,6 +27,22 @@
         Host IP address when the source IP address is the proxy.
       example: 192.1.1.2
 
+    - name: session_id
+      type: keyword
+      description: >
+        This is the session ID or connection ID,
```
ruflin: Reading in this field and the next one about connection, I wonder if we should introduce `connection.* as mentioned in an other thread instead of uptting it under network.

schemas/network.yml
```diff
@@ -48,3 +64,13 @@
...
+    - name: total.bytes
+      type: long
+      description: >
+        Network Total bytes: Usually sum (inbound.bytes, outbound.bytes)
+      example: 368
+    - name: total.packets
+      type: long
+      description: >
+        Network Total packets: Usually sum (inbound.packets, outbound.packets)
+      example: 24
```
webmat: I would phrase both of these more directly, e.g. "The sum of inbound.packets + outbound.packets", same for bytes.

MikePaquette:
Closing this out to split into two PR's as requested.

schemas/network.yml
MikePaquette: Sorry @ruflin I forgot to reply before closing this PR out for splitting.  Are you just talking about a name change i.e. s/network/connection?  The network.* field set is intended to pick up flow and connection-based fields, and also network events that are not flow/connection related. I think this requires a bit more thought before making a decision.

ruflin: I'm actually thinking if we could use a network and a connection prefix. I think a big chunk of the info we have right now in `network` is in the right place, but there a things like `forwarded_ip` which probably fit better into connection and also these fields here.

Definitively needs more discussions, just an idea.

PR #48, opened 2018-07-12
Added Network Session Objects - From Split of #43

MikePaquette:
Add network.session_id and network.virtual_ip in response to issue #37 
This PR was from the split-out of #43

### DIFF
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -27,6 +27,21 @@
         Host IP address when the source IP address is the proxy.
       example: 192.1.1.2
 
+    - name: session_id
+      type: keyword
+      description: >
+        This is the session ID or connection ID,
+        which is a unique identifier for the session /connection.
+        Bro conn.log uid field gets mapped here.
+      example: 0bbeef000d1234
+    - name: virtual_ip
+      type: ip
+      description: >
+        The virtual IP associated with a session or connection.
+        Useful when load balancers, firewalls, and routers use a
+        virtual IP to "advertise" a service.
+      example: 192.168.1.1
+
     # Metrics
     - name: inbound.bytes
       type: long
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 
 ### Added
+* Add `network.session_id` and `network.virtual_ip` field. #37
 * Add `event.action` field. #21
 * Adds cloud.account.id for top level organizational level. #11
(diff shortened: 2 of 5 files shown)


schemas/network.yml
```diff
@@ -27,6 +27,21 @@
         Host IP address when the source IP address is the proxy.
       example: 192.1.1.2
 
+    - name: session_id
+      type: keyword
+      description: >
+        This is the session ID or connection ID,
```
ruflin: Based on the description I'm wonder if we should put it into a connection object? See other discussions around connection.

I think there is value for `network` info and connection info.

krisATelastic:
Hey guys, it looks like this PR was not merged because of conflicts. Implementing `session_id` actually fixes a heap of use cases that i'm helping with, especially F5's. Are we going to re-push this or was it superceded? cc. @MikePaquette

webmat:
Hey @krisATelastic, thanks for the ping. We'll look at it again.

In the meantime if you or a customer needs to store session ID, please create custom fields in the meantime.

You should never feel blocked by ECS not supporting your use case yet.

MikePaquette:
The `network.community_id` field should be used for this purpose.

Docs at:  https://www.elastic.co/guide/en/ecs/current/ecs-network.html

andrewthad:
`network.community_id` is not the same thing as a session ID. According to the spec published by Corelight, a community ID is just a hash of the source ip, dest ip, source port, dest port, and network protocol. If a client opens a TCP connection to the same server twice and uses the same ephemeral port as the source port both times, the community ID will be the same for both sessions. By contrast, a session ID (as provided by Cisco, F5, Fortinet, and possibly other vendors that I'm not aware of) uniquely identifies the TCP session. I would like to see this included in ECS. Alternatively, one of the tracing fields (transaction.id`, `span.id`, or `trace.id`) could be used for this purpose, but this doesn't seem to fit their expressed purpose:

> Distributed tracing makes it possible to analyze performance throughout a microservice architecture all in one view. This is accomplished by tracing all of the requests - from the initial web request in the front-end service - to queries made through multiple back-end services.

ebeahan:
Hi @andrewthad. A proposal for a top-level `session` field set, including `session.id`, is actively going through the RFC process. Here are the links to the merged stage 0 proposal and the opened stage 2 PR if you have any feedback:

* [Session stage 0 proposal](https://github.com/elastic/ecs/blob/master/rfcs/text/0004-session.md)
* [Session stage 2 PR](https://github.com/elastic/ecs/pull/935)

PR #55, opened 2018-07-25
Make fields in uses cases which are not in ECS italic

ruflin:
Some of the existing use cases contain fields which are not in ECS. Also we had several fields commented to make it clear they are not in ECS. In recent discussions it popped up that it could be useful for others to share their schema they use for specific use cases + the fields which are not in ECS. As we now show the fields different this is possible.

I plan to do follow up PR's to also clean up the use cases. Some are outdated and should be updated to use ECS fields, others can be useful for inspiration on new fields. Reviewing use cases is now much easier as it is directly visible which fields come from ECS and which ones not.

Additional changes

* Remove links for non ecs fields.
* Remove ticks from name to make it nicer and support italic
* Add fields to uses cases which were commented out previously

### DIFF
--- a/use-cases/auditbeat.yml
+++ b/use-cases/auditbeat.yml
@@ -73,75 +73,74 @@ fields:
     type: date
     description: The last change time of the file (time when metadata was changed).
 
-# TODO (@ruflin 2018-05-01): These fields are not in ECS. Needs decision or removal.
-#
-#- name: hash
-#  group: 3
-#  description: >
-#    Hash fields used in Auditbeat.
-#
-#    The hash field contains cryptographic hashes of data associated with the event
-#    (such as a file). The keys are names of cryptographic algorithms. The values
-#    are encoded as hexidecimal (lower-case).
-#
-#    All fields in user can have one or multiple entries.
-#  fields:
-#    - name: md5
-#      type: keyword
-#      description: >
-#        MD5 hash.
-#
-#    - name: sha1
-#      type: keyword
-#      description: >
-#        SHA-1 hash.
-#
-#    - name: sha224
-#      type: keyword
-#      description: >
-#        SHA-224 hash (SHA-2 family).
-#
-#    - name: sha256
-#      type: keyword
-#      description: >
-#        SHA-256 hash (SHA-2 family).
-#
-#    - name: sha384
-#      type: keyword
-#      description: >
-#        SHA-384 hash (SHA-2 family).
-#
-#    - name: sha512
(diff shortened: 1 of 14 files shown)


use-cases/metricbeat.md
```diff
@@ -7,24 +7,25 @@ ECS fields used Metricbeat.
...
+| [service.id](https://github.com/elastic/ecs#service.id)  | Unique identifier of the running service.<br/>This id should uniquely identify this service. This makes it possible to correlate logs and metrics for one specific service. For example in case of issues with one redis instance, it's possible to filter on the id to see metrics and logs for this single instance.  | keyword  |   | `d37e5ebfe0ae6c4972dbe9f0174a1637bb8247f6`  |
+| [service.name](https://github.com/elastic/ecs#service.name)  | Name of the service data is collected from.<br/>The name is normally the same as the module name.  | keyword  |   | `elasticsearch`  |
+| [service.version](https://github.com/elastic/ecs#service.version)  | Version of the service the data was collected from.<br/>This allows to look at a data set only for a specific version of a service.  | keyword  |   | `3.2.4`  |
+| <a name="service.host"></a>*service.host*  | *Host address that is used to connect to the service.<br/>This normally contains hostname + port.<br/>REVIEW: Should this be service.uri instead, sometimes it's more then just the host? It could also include a path or the protocol.*  | keyword  |   | `elasticsearch:9200`  |
```
andrewkroh: Given that it's more than a hostname, `service.uri` SGTM.

use-cases/auditbeat.yml
```diff
@@ -73,75 +73,74 @@ fields:
...
-#      type: keyword
-#      description: >
-#        SHA3-384 hash (SHA-3 family).
-#
-#    - name: sha3_512
-#      type: keyword
-#      description: >
-#        SHA3-512 hash (SHA-3 family).
+
+- name: hash
```
andrewkroh: There are at least 2 more hash types in Auditbeat now - blake2b and xxhash. Would be good to add them so that people use the same names.

https://github.com/elastic/beats/blob/master/auditbeat/module/file_integrity/_meta/fields.yml#L12-L22

https://github.com/elastic/beats/blob/master/auditbeat/module/file_integrity/_meta/fields.yml#L72-L74

ruflin:
@andrewkroh Would you be ok to address the above comments in follow up PR's? The goal of this PR is mainly to get the "infrastructure" in place not to change existing definitions or add new ones.

PR #57, opened 2018-07-27
More use cases cleanup

ruflin:
* Add better description on how use cases are used
* Complete use cases table with ECS type, example, description if none is provided
* Adjust apm use case to use `@timestamp`
* Clean up generation script to not link `prefix.*` entries
* Refactor fields cleanup part

### DIFF
--- a/docs/use-cases-header.md
+++ b/docs/use-cases-header.md
@@ -2,6 +2,6 @@
 
 These are example on how ECS fields can be used in different use cases. Most use
 cases not only contain ECS fields but additional fields which are not in ECS to
-describe the full use case. The fields which are not in ECS are italic.
+describe the full use case. The fields which are not in ECS are in italic.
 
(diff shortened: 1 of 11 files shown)


docs/use-cases-header.md
```diff
@@ -1,3 +1,7 @@
 # <a name="use-cases"></a>Use cases
 
-ECS fields can be applied to common use cases.
+These are example on how ECS fields can be used in different use cases. Most use
+cases not only contain ECS fields but additional fields which are not in ECS to
+describe the full use case. The fields which are not in ECS are italic.
```
ruflin: @andrewkroh `are italic` sounds strange but not sure how to put it in english.

andrewkroh: I suggest "are **in** italic".

ruflin: thanks, updated.

PR #59, opened 2018-07-30
f5 use case poc

willemdh:
@ruflin Tried to run `python use-cases.py` but it doesn't seems to do  generate the markdown? Please review this as a PoC for now and feedback, I need to add a lot more fields, but I'm out of time for now.

### DIFF
--- a/use-cases/f5.yml
+++ b/use-cases/f5.yml
@@ -1,0 +1,103 @@
+title: F5 BIG-IP Load Balancer Syslog
+name: syslog-f5
+description:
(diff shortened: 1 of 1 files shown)


ruflin:
Try to run `make` as this should generate all files. Not sure if your PR is already on top of https://github.com/elastic/ecs/pull/57 but in any case make sure to rebase on the latest of master.

ruflin:
@willemdh Bare with us, things should get much more stable in the next 2-3 weeks.

PR #66, opened 2018-08-03
Add blake2b and xxhash to Auditbeat use-case fields

andrewkroh:
There are a few new hash algorithms under the `hash` namespace used by Auditbeat.

- blake2b_256
- blake2b_384
- blake2b_512
- xxh64


ruflin:
Thanks for the Makefile cleanup.

Issue #62, opened 2018-07-31
host.name vs hostname, host.ip vs ip inconsistencies


vbohata:
There are inconsistencies about host.name vs hostname, ip. vs host.ip naming. For example device.hostname, device.ip, host.name, url.host.name. Is there any reason for that?

ruflin:
Mainly for historical reason but I agree it's something we should cleanup up. Also ties into the discussion we have here: https://github.com/elastic/ecs/issues/40

ruflin:
I opened https://github.com/elastic/ecs/pull/74 to address the host.name issue. For `ip` I'm starting to think we should make it a top level field like `message` and device or host are just using it, so it could be in both objects.

webmat:
Not sure if the question around IPs is still relevant here.

If not, then #141 finally addresses the `host.name` vs `hostname` question. We've decided to go with the industry convention of using `hostname`, even if there is repetition in one case (`host.hostname`). This will ensure it's consistent with all other places where we have `hostname`.

vbohata:
Industry convention usually allows hostname to contain IP if host name is unknown ... so it could mix with ip ...

webmat:
Ah yeah if you don't have a hostname set for a device or host, you can totally cram an IP in there.

webmat:
Note that to get the full use of indexing IP addresses properly, you would still want to save the host's IP(s) in the field `host.ip`, which is indexed with type `ip`.

Given the variety of possible values in a `hostname`, this field is indexed as `keyword`.

jamiehynds:
@ebeahan are we still planning on deprecating host.hostname in 8.0? If so can we close this issue?

ebeahan:
@jamiehynds No, we're not deprecating `host.hostname` in 8.0.

Issue #71, opened 2018-08-08
Compose ECS objects vs reuse objects


ruflin:
In several open issues / PRs (for example https://github.com/elastic/ecs/pull/51) the `host` objects  are reused in different places. To reuse objects we can either repeat the name of the object or we could use composition. Examples:

Composition: Field `c` is composed out of `host` and `geo`, means it contains all the fields in host and geo. Example:
```
c.ip
c.mac
c.location
```

Reusing the object:
```
c.host.ip
c.host.mac
c.geo.location
```

Composition has the advantage that the names are shorter but it's not directly know out of what `c` is composed, meaning which fields it contains. Reusing the objects reduces the number of objects in ECS that look different. `c` could have a field `d` that is specific to `c` and still have all host and and geo fields. As soon as `geo` or `host` are used in any place it is clear which fields it will contain. This should simplify the mental model around ECS. To handle the longer field names with reusing objects it would be nice if Kibana would "understand" ECS and could potentially add some magic shortening here.

Most important is that we are consistent across ECS and use one or the other and don't mix them as we do at the moment.

MikePaquette:
@ruflin ECS how would this composition be implemented?  My first reaction is that the composition strings are actually additional things to remember, and might make ECS harder to understand.

ruflin:
We already do a bit of composition at the moment for example in `source.*` and `destination.*` and use the host fields. From an "implementation" perspective using composition could be as simple as defining the following in yaml:

```
- name: source
  composed_of: geo, host
```

Then the generator would do all the magic.

My bigger concern is how we make that understandable and visible to the user. This is where I think the reusing of objects becomes nice. We can mention where we recommend to use the `host` object but we don't have to repeat all the fields every time. And users adding non ECS fields on the top level can inside still reuse for example host.

webmat:
I would go for simplicity and have the reused objects present and visible, so I would not go for composition, personally.

ruflin:
If we make `ip` also a top level field like `message`, would this solve the issue of having too many nesting? This means for the above it could be `c.ip` or `c.host.ip`, meaning `host.ip` is just reusing the field.

vbohata:
I think composing to c.ip, ... could lead to a lower number of reusable objects. Yes, for this use case it is clear and for the first point of view nice but I vote for the longer names. Everyone can see here from which parts it is composed from.

vbohata:
New example. For some logs from the switch I have following fields:
- source.host.mac
- client.host.mac
[...]

PR #74, opened 2018-08-10
Rename hostname to host.name fields to be consistent

ruflin:
`destination.hostname`, `device.hostname` and `source.hostname` were renamed to `*.host.name` to be consitent with the host object.

Related to https://github.com/elastic/ecs/issues/62


webmat:
LGTM

MikePaquette:
@ruflin I think "hostname" is a better ECS name for this field, as we are looking to populate this with the internet hostname which is commonly known as "hostname."  See [wikipedia hostname](https://en.wikipedia.org/wiki/Hostname).  

This too would preserve a consistent use of "hostname" in `source.hostname`, `destination.hostname`, and `device.hostname`.  But for consistency, this also means that we would have `host.hostname`, which has been criticized previously as being redundant, and could seem to run afoul of our _Avoid Repetition_ [naming guideline ](https://github.com/elastic/ecs/blob/master/README.md#guidelines). Nonetheless, since "hostname" is the actual name of the entity stored in the field, I think this justifies keeping it.

Also, in case we decide to keep `host.*` as a top-level namespace/object/prefix only, the use of "hostname" keeps us clear of conflicts.

Finally, this also allows us to continue to have only one dot in our commonly used "ECS-compatible" fields, an original ECS goal to assist with ease-of-recall and predictability. 

I vote for no further changes from `*.hostname` to `*.host.name` and for reverting `host.name` to `host.hostname` for consistency.

ruflin:
Renaming `host.name` to `host.hostname` is an other way we could go. My current preference is still with `host.name` but it ties into the discussion here: https://github.com/elastic/ecs/issues/71 It would mean `hostname` is a reusable field.

webmat:
Yeah I think I'm also in favor of sticking with the widely used naming, in cases like this.

webmat:
As discussed, we're closing this and we'll be sticking with `hostname` instead.

ruflin:
Would be good to add here a comment with reasoning so if we get back to it we know why.

PR #79, opened 2018-08-13
 Adds network.protocol.name and network.protocol.id 

robgil:
With most logs coming in, the protocol names are not resolved to their respective full name and instead rely on the IDs from [IANA](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). This adds names and ids. The regular `network.protocol` would use name, but if you're not planning to use resolution, `network.protocol.id` would be available.

### DIFF
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -9,7 +9,7 @@
       type: keyword
       description: >
         Network protocol name.
-      example: http
+      example: tcp
     - name: direction
       type: keyword
       description: >
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -10,6 +10,16 @@
       description: >
         Network protocol name.
       example: tcp
+    - name: protocol.name
+      type: keyword
+      description: >
+        Network protocol name.
+      example: TCP
+    - name: protocol.id
+      type: keyword
+      description: >
+        Network protocol id https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
+      example: 6 
     - name: direction
       type: keyword
       description: >
--- a/README.md
+++ b/README.md
@@ -285,7 +285,9 @@ Fields related to network data.
 
 | Field  | Description  | Type  | Multi Field  | Example  |
 |---|---|---|---|---|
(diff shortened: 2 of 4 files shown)


schemas/network.yml
```diff
@@ -9,7 +9,17 @@
       type: keyword
       description: >
         Network protocol name.
-      example: http
+      example: tcp
+    - name: protocol.name
```
ruflin: It is not possible to have `protocol` and `protocol.name` as otherwise ingestion in ES will not work (object vs keyword type). I assume we could remove `protocol`?

robgil: @ruflin ah that makes sense. I guess we can just stay with `network.protocol` if we plan to add `network.application.protocol`. That means we should be able to keep `network.protocol.id`.

ruflin: `network.protocol` and `network.protocol.id` can also not coexist.

webmat: Yeah I think the confusion comes from ElasticSearch' multi-fields. So it's possible to have `foo` and `foo.bar` because it's actually one field (one value) indexed in different ways. By default/convention, the top level is `analyzed` for full text search, whereas the nested field is indexed another way, in this case `keyword` indexing, for aggregations and exact match searches. You could even have more than one nested field, but the common thread is that it's always exactly the same value being indexed in different ways.

On the other hand, you cannot have `foo` and `foo.bar` be two different values, because as @ruflin said, the top level needs to be of type `object`, to be able to nest arbitrary stuff under it.

webmat: Well actually, looking again at the actual file, it seems like `protocol` is just a dupe of `protocol.name`.

So the following should do it (we don't need to explicitly define the top level object):

```yaml
   fields:
    - name: protocol.name
      type: keyword
      description: >
        Network protocol name.
      example: TCP
    - name: protocol.id
      type: keyword
      description: >
        Network protocol id https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
      example: 6
```

webmat:
Closing this issue, as #81 fixes this use case, and more.

PR #80, opened 2018-08-13
Revises event.created and expands with event.start and event.end

robgil:
While trying to map [VPC FlowLogs](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/flow-logs.html#flow-log-records) I was finding that `event.created` didn't really align with the `start` and `end` of a flow. This PR adds `event.start` and `event.end`. 

`event.created` could still be used with regard to `events` in general. So for example if there is a threat feed match, we could create an event, and the time the event was created would relate to the event, but not the source of the event itself. So for example, VPC FlowLogs and Netflow for that matter only ever log after the flow is complete. This means that the `event.created` would land after the start of the flow. While this may not be ideal, this is what I'm trying to delineate between - the time it was determined it was a security event, and when the event actually happened. 

A more concrete example.

I have a netflow flow that lasted 36 hours. The time at which it was considered an event was after the flow completed, but the `event.start` and `event.end` would map to when the flow started and ended. `event.duration` would be the following..

`event.end - event.start = event.duration`.


schemas/event.yml
```diff
@@ -128,6 +128,27 @@
...
+    - name: start
+      type: date
+      description: >
+        event.start contains the date when the event started.
+
+        This timestamp is distinct from @timestamp in that @timestamp contains
+        the processed timestamp. For logs these two timestamps can be different
+        as the timestamp in the log line and when the event is read for example
+        by Filebeat are not identical. `@timestamp` must contain the timestamp
+        extracted from the log line, event.start when the log line is read.
```
ruflin: For me that is `event.created`.

schemas/event.yml
```diff
@@ -128,6 +128,27 @@
...
+        The same could apply to package capturing where @timestamp contains the
+        timestamp extracted from the network package and event.created when the
+        event was created.
+
+        In case the two timestamps are identical, @timestamp should be used.
+
+    - name: end
+      type: date
+      description: >
+        event.end contains the date when the event ended.
```
ruflin: You have a nice description in the PR of this with an example. Could you add this here?

schemas/event.yml
```diff
@@ -128,6 +128,27 @@
...
+      type: date
+      description: >
+        event.start contains the date when the event started.
+
+        This timestamp is distinct from @timestamp in that @timestamp contains
+        the processed timestamp. For logs these two timestamps can be different
+        as the timestamp in the log line and when the event is read for example
+        by Filebeat are not identical. `@timestamp` must contain the timestamp
+        extracted from the log line, event.start when the log line is read.
+        The same could apply to package capturing where @timestamp contains the
```
webmat: "package capturing" => "packet capture" (same correction on line below as well)

schemas/event.yml
webmat: I would not explain `event.start` in terms of the phenomenon of reading log files.

I think it would be more beneficial to specifically explain `start`/`end` specifically in terms of the observation of long running phenomena, like flows.

Some points I'd try to hit in this description and/or `end`:

* `start` and `end` usually go hand in hand. Although there may be exceptions (missing field) in the case of partial flows.
* These are to describe potentially long running events. `start` can actually be hours before `@timestamp`, if the observed phenomenon started a long time ago and was just reported.
* I would relate this description to `event.duration`. 

To @ruflin's point, having 2/3 fields out of `start`, `end` and `duration` is enough to derive the 3rd fields. I'd rather have 3/3, but we may (or may not) want to mention that 2/3 can be enough?

PR #81, opened 2018-08-13
Revises network.* to cover more use cases

robgil:
This is to help delineate between `network.protocol` like `tcp` and an
`network.application.protocol` such as `http`.

### DIFF
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -18,7 +18,7 @@
       level: core
       type: keyword
       description: >
-        In the OSI Model this would be the Network Layer.
+        In the OSI Model this would be the Network Layer. IPv4, IPV6, IPSec, PIM, etc
       example: IPV4
 
     - name: iana_number
--- a/schemas/network.yml
+++ b/schemas/network.yml
@@ -14,11 +14,45 @@
         Name given by operators to sections of their network.
       example: Guest Wifi
 
+    - name: type
+      level: core
+      type: keyword
+      description: >
+        In the OSI Model this would be the Network Layer.
+      example: IPV4
+
+    - name: iana_number
+      level: extended
+      type: keyword
+      description: >
+          IANA Protocol Number (https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml).
+          Standardized list of protocols. This aligns well with NetFlow and
+          sFlow related logs which log using the IANA Protocol Number.
+      example: 6
+
+    - name: transport
+      level: core
+      type: keyword
+      description: >
+        Same as network.iana_number, but instead using the Keyword name.
+      example: IPv6-ICMP
+
+    - name: application
+      level: extended
+      type: keyword
+      description: >
+        A name given to an application. This can be arbitrarily assigned for
+        things like microservices, but also apply to things like skype, icq,
+        facebook, twitter. This would be used in situations where the vendor
+        or service can be decoded such as from the source/dest IP owners,
+        ports, or wire format.
+      example: AIM
+
     - name: protocol
       level: core
       type: keyword
       description: >
-        Network protocol name.
+        L7 Network protocol name. ex. http, lumberjack, transport protocol
       example: http
 
     - name: direction
--- a/README.md
+++ b/README.md
@@ -282,7 +282,11 @@ Fields related to network data.
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
 | <a name="network.name"></a>network.name | Name given by operators to sections of their network. | extended | keyword | `Guest Wifi` |
-| <a name="network.protocol"></a>network.protocol | Network protocol name. | core | keyword | `http` |
+| <a name="network.type"></a>network.type | In the OSI Model this would be the Network Layer. | core | keyword | `IPV4` |
+| <a name="network.iana_number"></a>network.iana_number | IANA Protocol Number (https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). Standardized list of protocols. This aligns well with NetFlow and sFlow related logs which log using the IANA Protocol Number. | extended | keyword | `6` |
+| <a name="network.transport"></a>network.transport | Same as network.iana_number, but instead using the Keyword name. | core | keyword | `IPv6-ICMP` |
+| <a name="network.application"></a>network.application | A name given to an application. This can be arbitrarily assigned for things like microservices, but also apply to things like skype, icq, facebook, twitter. This would be used in situations where the vendor or service can be decoded such as from the source/dest IP owners, ports, or wire format. | extended | keyword | `AIM` |
+| <a name="network.protocol"></a>network.protocol | L7 Network protocol name. ex. http, lumberjack, transport protocol | core | keyword | `http` |
 | <a name="network.direction"></a>network.direction | Direction of the network traffic.<br/>Recommended values are:<br/>  * inbound<br/>  * outbound<br/>  * unknown | core | keyword | `inbound` |
 | <a name="network.forwarded_ip"></a>network.forwarded_ip | Host IP address when the source IP address is the proxy. | core | ip | `192.1.1.2` |
 | <a name="network.inbound.bytes"></a>network.inbound.bytes | Network inbound bytes. | core | long | `184` |
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ Fields related to network data.
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
 | <a name="network.name"></a>network.name | Name given by operators to sections of their network. | extended | keyword | `Guest Wifi` |
-| <a name="network.type"></a>network.type | In the OSI Model this would be the Network Layer. | core | keyword | `IPV4` |
+| <a name="network.type"></a>network.type | In the OSI Model this would be the Network Layer. IPv4, IPV6, IPSec, PIM, etc | core | keyword | `IPV4` |
 | <a name="network.iana_number"></a>network.iana_number | IANA Protocol Number (https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). Standardized list of protocols. This aligns well with NetFlow and sFlow related logs which log using the IANA Protocol Number. | extended | keyword | `6` |
 | <a name="network.transport"></a>network.transport | Same as network.iana_number, but instead using the Keyword name. | core | keyword | `IPv6-ICMP` |
 | <a name="network.application"></a>network.application | A name given to an application. This can be arbitrarily assigned for things like microservices, but also apply to things like skype, icq, facebook, twitter. This would be used in situations where the vendor or service can be decoded such as from the source/dest IP owners, ports, or wire format. | extended | keyword | `AIM` |
--- a/fields.yml
+++ b/fields.yml
@@ -830,7 +830,7 @@
           level: core
           type: keyword
           description: >
-            In the OSI Model this would be the Network Layer.
+            In the OSI Model this would be the Network Layer. IPv4, IPV6, IPSec, PIM, etc
           example: IPV4
     
         - name: iana_number
--- a/fields.yml
+++ b/fields.yml
@@ -826,11 +826,45 @@
             Name given by operators to sections of their network.
           example: Guest Wifi
     
+        - name: type
+          level: core
+          type: keyword
+          description: >
+            In the OSI Model this would be the Network Layer.
+          example: IPV4
+    
+        - name: iana_number
+          level: extended
+          type: keyword
+          description: >
+              IANA Protocol Number (https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml).
+              Standardized list of protocols. This aligns well with NetFlow and
+              sFlow related logs which log using the IANA Protocol Number.
+          example: 6
+    
+        - name: transport
+          level: core
+          type: keyword
+          description: >
+            Same as network.iana_number, but instead using the Keyword name.
+          example: IPv6-ICMP
+    
+        - name: application
+          level: extended
+          type: keyword
+          description: >
+            A name given to an application. This can be arbitrarily assigned for
+            things like microservices, but also apply to things like skype, icq,
+            facebook, twitter. This would be used in situations where the vendor
+            or service can be decoded such as from the source/dest IP owners,
+            ports, or wire format.
+          example: AIM
+    
         - name: protocol
           level: core
           type: keyword
           description: >
-            Network protocol name.
+            L7 Network protocol name. ex. http, lumberjack, transport protocol
           example: http
     
         - name: direction
--- a/schema.csv
+++ b/schema.csv
@@ -86,8 +86,10 @@ http.response.status_code,long,extended,404
 http.version,keyword,extended,1.1
 log.level,keyword,core,ERR
 log.original,keyword,core,Sep 19 08:26:10 localhost My log
+network.application,keyword,extended,AIM
 network.direction,keyword,core,inbound
 network.forwarded_ip,ip,core,192.1.1.2
+network.iana_number,keyword,extended,6
 network.inbound.bytes,long,core,184
 network.inbound.packets,long,core,12
 network.name,keyword,extended,Guest Wifi
@@ -96,6 +98,8 @@ network.outbound.packets,long,core,12
 network.protocol,keyword,core,http
 network.total.bytes,long,core,368
 network.total.packets,long,core,24
+network.transport,keyword,core,IPv6-ICMP
+network.type,keyword,core,IPV4
 organization.id,keyword,extended,
 organization.name,keyword,extended,
 os.family,keyword,extended,debian
--- a/template.json
+++ b/template.json
@@ -429,12 +429,20 @@
         },
         "network": {
           "properties": {
+            "application": {
+              "ignore_above": 1024,
+              "type": "keyword"
+            },
             "direction": {
               "ignore_above": 1024,
               "type": "keyword"
             },
             "forwarded_ip": {
               "type": "ip"
+            },
+            "iana_number": {
+              "ignore_above": 1024,
+              "type": "keyword"
             },
             "inbound": {
               "properties": {
@@ -473,6 +481,14 @@
                   "type": "long"
                 }
               }
+            },
+            "transport": {
+              "ignore_above": 1024,
+              "type": "keyword"
+            },
+            "type": {
+              "ignore_above": 1024,
+              "type": "keyword"
             }
           }
         },


ruflin:
As in https://github.com/elastic/ecs/pull/79 you introduce an additional level to protocol, what about using `network.protocol.type: application`? I would prefer if all protocols stay in the same field.

robgil:
@ruflin I'd prefer having them separate and delineate between application protocol and network protocol. The example we used in slack was TCP, UDP, and HTTPS DNS lookups.

webmat:
I totally agree we need two distinct fields for the transport vs application protocols. But I would like to group them, however.

Instead of this

```YAML
network.application.protocol: DNS
network.protocol: UDP
```

I would prefer that

```YAML
network.protocol.application: DNS
network.protocol.transport: UDP
```

The former looks like what Suricata is doing (`app_proto` & `proto`), but this means the fields are scattered at two ends of the same event, when you're drilling down :-) It's a minor gripe, but I think the grouping makes it easier to work with the data day to day.

Plus if we ever want to support protocols at other levels (e.g. ethernet/wifi, or even "userland" protocols built on top of application layer), having a protocol object with sub-fields will easily support this future growth.

MikePaquette:
@robgil @webmat I agree with your points, but I'm concerned that the proposed names may confuse users due to mixing layers, and they could be shortened.  Could we instead have `network.protocol: "ipv4"` `network.transport: "udp"` and  `network.application: "dns"` as the fields under network?  We already have `source.port` and `destination.port` to capture the transport port numbers.

vbohata:
+1 for network.protocol.application and network.protocol.transport. It is easy to see what is related to what layer.

webmat:
I like the shortened names suggested by @MikePaquette. And sorry for mixing up layers :-)

Actually this mixing up made me wonder if `network.protocol` was actually the right naming for the layer where we'd find ipv4/6. I ended up on the trusty Wikipedia, where this layer is actually called the "Internet Layer". So, partially quoting the [Internet Layer page](https://en.wikipedia.org/wiki/Internet_layer):

> **Application Layer**
> ..., DNS, FTP, HTTP ...
> **Transport Layer**
> TCP, UDP, ...
> **Internet Layer**
> IP (IPv4 IPv6), ICMP, ...
> **Link Layer**
> ARP, ..., MAC (Ethernet, DSL, ...

Perhaps we could follow this nomenclature?

```YAML
network.internet: ipv4 or icmp
network.transport: tcp
network.application: http
```

And if it comes up as necessary/desirable, `network.link` as well. Personally I think `network.internet` sounds a bit awkward, but I think it's better than `network.protocol` because each of these layers are actually composed of "protocols"...

robgil:
@webmat to make it more complicated the "Internet Layer" is actually Layer 3 in the OSI model which is called the "Network Layer". So it would actually be `network.network` 😭 . Or perhaps `network.type` but that's confusing also.

```
network.network: ip or ipx (L3)
network.transport: tcp or udp (L4)
network.application: http or transport protocol (L7)
```

robgil:
Updated and expands/replaces `network.protocol` with `network.network`, `network.transport`, and `network.application` to cover Layers 3, 4 and 7 respectively.

schemas/network.yml
```diff
@@ -16,11 +16,20 @@
           type: keyword
           description: >
             Name given by operators to sections of their network.
-
-    - name: protocol
+    - name: network
```
ruflin: This would be `network.network`. Not sure if that make sense. Any alternatives we could use?

schemas/network.yml
```diff
@@ -16,11 +16,20 @@
...
+      type: keyword
+      description: >
+          OSI Layer 3 Network Layer. Examples - IP, ICMP
```
ruflin: Could you add a bit more details to all descriptions. For example add/describe a use case from which you pick the fields.

schemas/network.yml
robgil: Yeah, I'm not a huge fan, but it aligns with the OSI model which is a standard with regard to naming these things. Does anyone else have any thoughts? @webmat @MikePaquette ?

webmat: I think I'd rather stick with `network.internet`, which is also weird, but at least is not a repetition like `network.network`...

Or perhaps `network.layer3`? Not a super fan of that either because it's inconsistent vs using the actual names for all the others. But I wanted to mention that option.

vbohata: Why not to use network.protocol.* naming which is not confusing?

ruflin: Putting OSI aside: What will the common user look for naming wise? What about the suggestion from @vbohata 

What is still not fully clear to me why we can't mix them all in one field. Are the aggregations we can't do then?

One more option: What about `network.layer: layer_name` together with `network.protocol`?

vbohata: Putting all into one field can not work as expected. Some logs contains for example both network.protocol.application and network.protocol.transport and users want to do term aggs per application protocol, or ...

webmat: I agree with @vbohata, people will want to aggregate on these. You may want to see ratio of UDP vs TCP, but if you then have HTTP, SMTP, etc intermingled, that doesn't help. Same thing for aggregating around the application level protocol, you don't want to see TCP and UDP pop up there.

Let's make @vbohata's proposal explicit, to see it in action and make sure we all understand the same thing:

```
network.protocol.link: arp, mac...
network.protocol.network: ip, ipx, icmp
network.protocol.transport: udp, tcp...
network.protocol.application: http, ftp, dns...
```

@vbohata is that what you had in mind?

I do think it helps reduce the weirdness of seeing `network.network`, that's for sure :-)

Another thing to consider is that the top two layers in the list above will be used less often than the bottom two, in many cases. So a bit of naming weirdness there may not be the end of the world. I would expect to see `link` and `network` in some of the networking logs for people managing data centers. I would expect to see it less often in cloud environments, where usually the networking layer is managed by the provider / taken for granted.

Also, these fields being in ECS doesn't mean all logs using ECS must fill them, IMO. I don't think web server logs should include the `link` and `network` layer, as their value would absolutely always be the same. I could hardly imagine a situation where I need to compare web server logs vs BGP logs directly (as in: aggregation to see which one you have more of). So the logs centered around managing the network will have the lower level ones, and the logs centered around the applications will have only `application` (e.g. web logs) or `application` and `transport` (e.g. dns, IDS).

dainperkins: Couple thoughts:
a specific section for wireless info might be nice, unless that fits into another bucket (e.g. ssid # / name, not sure how likely it would be to be able to tell physical network info from anything besides e.g. wireless alerts, but having a way to normalize e.g. WIPS alert info / RADIUS WPA2-ENT would be nice)

Something like this might eliminate some of the understanding issues:
network.protocol  (IP / GRE / EIGRP, etc)
network.transport (tcp/udp)
network.application (http/telnet/ssh/smtp/tls)
network.application.id (L7 id - twitter, facebook, etc)

but adding in e.g. COS / DSCP, MPLS tags, etc and some of the other items might be useful from an APM standpoint . IPFIX / Netflow9 would probably be a decent place to start adding relevant network data.

https://www.iana.org/assignments/ipfix/ipfix.xhtml
some of the fields may well be superfluous (delta counts, aggregate counts, etc)

and layer 3 is definitely the NETWORK layer :)

dainperkins: and 1 question - as I look thru the various models are the ymls meant to support a given type of logging source (e.g. network = netflow / network sensor / basic fw network telemetry?)  If so it would be good to discretely identify the expected sources & target info (e.g. NPM people want QOS/DSCP for monitoring, APM people should too... but a normalized netflow target is likely to look significantly different than a normalized FW flow/connection target data type)

webmat: Hi @dainperkins! ECS is meant to document a common event schema that would help make different event/log sources look more similar to one another. This will make it easier to find the basics in any event stream. The use cases we are mostly focusing on at the moment is operational monitoring as well as security.

By its nature, though, ECS will never encompass all use cases for all sources. Any given stream is likely to have fields that don't fit in ECS. You can use your custom fields around the documented ECS fields.

If you see existing fields that aren't quite defined correctly to fit some needs, or if you see generic fields that are missing, it's totally fair to open the discussion around it, though. So feel free to open issues or PRs with specific suggestions.

dcode: from an NSM perspective where we might have knowledge of more than one layer of protocol in a given event, this is fantastic:

> network.protocol (IP / GRE / EIGRP, etc)
> network.transport (tcp/udp)
> network.application (http/telnet/ssh/smtp/tls)
> network.application.id (L7 id - twitter, facebook, etc)

Namely, HTTP is absolutely an OSI layer 7 protocol, but it adds value if we have signatures or behavior models that are picking up the next layer application model of twitter, facebook, etc.

webmat: Thanks Derek!

Note that we can't have both `network.application` and `network.application.id` at the same time in the field mapping. The following would work, though: `network.application` and `network.application_id`.

I like @dainperkins's suggestion of adding `network.application_id`, but for now I think we should focus on finding sensible names for each network layer. We can add `_id` later. I'll try to close on this sooner than later. Based on the conversations here I would be inclined to go with:

* `network.protocol`: e.g. IP, GRE, EIGRP
* `network.transport`: TCP, UDP
* `network.application`: HTTP, HTTPS, SSH, TLS, SMTP

Let me know if there's any big disagreement on this.

vbohata: Pro: shorter, con: not self describing enough
Example: User wants to filter all events related to program "named" and "UDP" protocol. Until deeper look at logged data it seems network.application will be "named" and network.protocol will be "udp". In longer variant everyone immediately see "network.protocol.transport" is the correct field and network.protocol.application will probably not contain program name ...

dainperkins:
How about this (more accurate, possibly more annoying)


   - network.protocol.version:    4/6 (ipv4, ipv6)
   - network.protocol.number:   1/6/17
   - network.protocol.name:       ICMP, TCP, UDP, GRE, EIGRP, etc
   - network.transport.port:        80/443/25/5061, etc .  **could be
   network.protocol.port, but seemed like a better option to split into
   transport for tcp/udp details
   - network.transport.protocol:  http/sip/tls (may or may not be derived
   from port - e.g. you can run TLS over 8443...)
   - network.application.id:        Twitter, Gmail, Facebook, Sharepoint,
   Outlook (e.g. open slot for vendor specific e.g. cisco nbar, Palo AppID or
   whatever they call it)

/d

schemas/network.yml
dcode: My concern is that `network.protocol` in network sensor applications (bro, suricata, etc) is almost always the transport layer protocol (in the OSI sense). ICMP, GRE, and EIGRP, and of course TCP and UDP are all IP-encapsulated protocols, in that they all require a valid IP header first. Arp is probably one of the common protocols that you might log that sits at the network layer. 

All that said, there are certainly times where one wouldn't be using IP as the network protocol, but in general it's at least a 95% solution. I'm not inclined to accrue the cost of adding a field to simply mark all my log entries as `IP`. And after I do that, I have to rename all the `network.protocol` fields to `network.protocol.transport`. 

I will adopt the `.network.application` and `network.application_id` annotation, however. Thanks for that suggestion @webmat

dcode:
>>  - network.protocol.name:       ICMP, TCP, UDP, GRE, EIGRP, etc
>>   - network.transport.port:        80/443/25/5061, etc .  **could be
>>   network.protocol.port, but seemed like a better option to split into
>>   transport for tcp/udp details
>>   - network.transport.protocol:  http/sip/tls (may or may not be derived
>>   from port - e.g. you can run TLS over 8443...)

@dainperkins the issue here is that at the transport layer, there's always two ends of the connection, so there are two ports. If you're not logging the network connection itself, but rather just application data, you may not know the port(s) of the other end of the connections, however. Also, I think `network.transport.protocol` may be a bit confusing, as "transport" is often used for protocols operating just above the Internet layer (e.g. TCP, UDP, SCTP, etc)

robgil:
Updated proposal for new `network.*` fields

We met internally to try to cover as many use cases as possible and filter/collate the comments and suggestions above. We came up with the following proposal. 

## network.type
Describes essentially the `Network Layer` of the OSI model. `network.type` seemed the most logical name over something like `network.network` to describe this layer of the OSI model. 
```
network.type: ipv4, ipv6, ipx
```

## network.iana_number
This is the most helpful aspect of this schema proposal. Aligning to the [IANA Protocol Numbers](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) lets us cover a great many more use cases. This is also standardized and well understood. There are situations where you need to classify specialized traffic such as IPIP, and this provides a way to do that. Other use cases such as DNS (which can be either UDP, TCP, or now TCP/HTTPS) are strong reasons to cover every layer of the OSI model.
```
network.iana_number: 1/6/17 IANA Protocol Numbers
```

## network.transport
`network.transport` specifically refers to the Keywords in the [IANA Protocol Numbers](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) standard.
```
network.transport: TCP, UDP, IPv6-ICMP, ICMP 
```

## network.application_protocol
This brings us up to L7 protocols.  It would be nice to refer to a public standard list, but obviously there are cases such as DNS over HTTPS and others. The application protocols can typically be picked up  based on the known protocols in wireshark or other full packet inspection tools.
```
network.application_protocol: http, dns, smtp, ftp, etc
```

## network.application
Again at L7, but this time specific to the vendor. So for example, if a capture device captures HTTP traffic it can potentially determine which vendor its in relation to (ex. Facebook, Twitter, LinkedIn requests). There was some debate as to whether this is a generic field or very specific to security, but this field could also be used to map out microservices. For example, if I have an auth service, an api service, and an upload service. I could map each of these to its own `network.application` value (ie `my-auth`, `my-api`, `my-upload`) for the purposes of identifying flows based on application/service.
```
network.application: skype, icq, aim, dns (dns over https)
```

# Fields to be removed
```
network.protocol
```

vbohata:
Some feedback from our usage. In our company we are using following field names which are for now covering all of our use cases (OS logs, application logs, network devices logs, ...):
companyprefix.application.name (not related to network field, but for demonstration)
companyprefix.network.protocol.application
companyprefix.network.protocol.transport
in future maybe: companyprefix.network.protocol.internet

For network.type the logical values are also WAN, LAN ... (which is type of network, not type of network/internet layer).

webmat:
Ok, I like how `network.type` works around weird naming issues :-D

So the nitty gritty of the proposal would be the following, correct?

| Field | Type | Level |
|--|--|--|
| network.type | keyword | core |
| network.iana_number | long | extended |
| network.transport | keyword | core |
| network.application_protocol | keyword | core |
| network.application | keyword | extended |

And network.protocol would be removed.


What if instead of `network.application_protocol` we kept `network.protocol` for the L7 protocol? It wouldn't have a similar name to its sibling `network.application`, but on the upside the short and most canonical field name would be the home of the highest level protocol.

webmat:
@vbohata The example you're giving is less about protocol details, and more about free form field about your network topology. I maybe could see the need for this, but I wonder how it would work in practice. My understanding is: it's not a value you would determine based on each event's details, but rather a blanket field value you would put in place for every log coming out of a device meant to service the "LAN", for example, correct?

In any case, I would like it if you opened an issue about this. I would prefer if we kept this PR strictly about how we describe the protocol stack details of each event.

vbohata:
OK, it was just an example "from end-users point of view".

webmat:
@vbohata Just making sure we're on the same page: I do see the value in your suggestion. All I'm saying is that this PR is more about mapping the network layer details (think OSI) to ECS. Your proposition is more about mapping a user-centric view of the network on ECS. Also valuable, but we should work on one chunk at a time :-)

dainperkins:
FWIW - seems like the latest round is mixing up layers...
1) type does sound more like a physical layer thing (1gbt, wireless, etc)
2) iana_number seems excessively specific & network.protocol is going to
mean something to a lot of people (and a lot of logs...) - e.g. the ip
protocol (tcp/udp/icmp, etc) and will typically be the IANA decimal.
(netflow, firewall logs, sensors, etc.)
I could like with type for IP4, ip6, ipx, but network.protocol should stay

then I get a little ambivalent on where to split the layers, e.g. TCP, HTTP

  network =
    IP4, IP6, IPX,
  network.protocol           = IANA Protocol #                 1/6/17
  network.protocol.name = IANA Protocol name .        TCP

  network.transport =          IANA Protocol Name .       TCP
  network.transport.port . = TCP Service Port .            80
  network.transport.protocol = well known services     HTTP

  network.application.protocol    =                                HTTP .
(well known services)
  network.application.type  ??                                      WEB
(dpi / service  visibility required)
  network.application.name .  =                                    Vendor /
App specific (e.g. sharepoint vs WEB, DPI or service knowledge required))

Would it make more sense to potentially use an application space for
network.application fields (a'la source / destination potentially being
mixed in with network for a full event)?  realistically beyond the
transport.protocol (tls, http, smtp) and port, we're not really talking
about network anymore - dns over TLS, sharepoint, git, twitter, etc..

robgil:
@dainperkins to add some more context to some of your points (which are all valid). Keep in mind, myself and others look at this from a network engineer perspective but also from a generic naming convention that can map may types of data. Obviously its challenging to map all security devices and network devices on to one common schema, but the goal here is to map the most common fields that are consistent between devices for the top level fields. 

- type does sound more like a physical layer thing (1gbt, wireless, etc)
In this particular field the alternative would have been `network.network` for the `Network Layer` in the OSI model. This is obviously ugly. `network.type` came out of discussions around what sFlow and NetFlow did. So in this case we used [sFlow as a reference](https://sflow.org/SFLOW-DATAGRAM5.txt) with regard to sFlow's `address_type`. We just shortened it to type. @vbohata this should help with your question also. LAN/WAN could be mapped in part by `network.direction`.

- iana_number seems excessively specific & network.protocol is going to mean something to a lot of people (and a lot of logs...) - e.g. the ip protocol (tcp/udp/icmp, etc) and will typically be the IANA decimal. (netflow, firewall logs, sensors, etc.)
Yes, this is very specific, however it is also a standard reference to pull from. Add to that, it is also the field returned from NetFlow. Per @webmat, this would be an _extended_ field and `network.transport` would be the keyword name from the IANA Protocols standard. Think of this more as enumeration over a free-form field. 

- then I get a little ambivalent on where to split the layers, e.g. TCP, HTTP
Those are two different layers in the OSI model. That's what I've been trying to delineate in this PR from the start. The prime example of this is DNS. DNS can be UDP and TCP, but also TLS and HTTPS based. 

- network.transport.port
Ports are all moved to `source.port` and `destination.port`. I could see this being a service listening on a specific port however, but we have not identified a way to represent `listening ports` yet. This should be opened as a new Issue or PR to track things like listening ports with regards to inventory type use cases.

- realistically beyond the transport.protocol (tls, http, smtp) and port, we're not really talking about network anymore - dns over TLS, sharepoint, git, twitter, etc..
Agreed on this to the extent it goes beyond wire formats. I still think wire formats should be captured in `network.*`. For example, http, memcache (binary), transport protocol, lumberjack, mysql (binary), etc. I think application protocol decode is pretty common in both network probes and devices that do full packet inspection. 

@dainperkins @vbohata what do you think about providing some sample log strings to map? The ones that are top of mind for me are Netflow, sFlow, AWS VPC FlowLogs, and GCP FlowLogs. We obviously need some vendor examples (Cisco ASA, Palo Alto, Fortinet, etc). We'd need IDS (snort, etc) and other TAP related device log samples as well. If we can map 90% of the _common_ fields, I think we'll be in a good spot for v1.

dainperkins:
I should be able to provide netflow, bro, asa, Palo pretty easily, I think
thats a good way to look at it...

VPC is pretty much useless but i can add those in easily as well


/d

dainperkins:
I probably screwed a bunch of stuff up, but here are some examples to chew
on (and some thought on some other things that might be good to add)

   - vpc flow (straight from AWS definitions)
   - uni direction netflow 9 . (ubiquiti)
   - bidirectional netflow 9 .  (qradar from meraki)
   - palo traffic log from definitions
   - firepower from logs (grok + kv to split)

/d

webmat:
@dainperkins Haha yeah, email to GitHub doesn't support attachments, it seems :-) For this you need to attach directly from the GitHub web form.

dainperkins:
doh!

Issue #87, opened 2018-08-16
Inconsistent usage of .raw, and discussion about .raw vs .keyword naming


webmat:
This issue is about two distinct problems that I think are closely related, hence the creation of a single issue.

**Inconsistent usage of .raw**

We currently have 7 fields ending in `.raw` in ECS. Here they are, listed with their type, and the "meaning" of `.raw` in their context:

Field Name | Type | Meaning of `.raw`
----------- | ----- | -------------------
`event.raw` | keyword | *Original value, prior to parsing*
`file.path.raw` | keyword | Nested field of a multi-field
`file.target_path.raw` | keyword | Nested field of a multi-field
`url.href.raw` | keyword | Nested field of a multi-field
`url.path.raw` | keyword | Nested field of a multi-field
`url.query.raw` | keyword | Nested field of a multi-field
`user_agent.raw` | **text** | *Original value, prior to parsing*

In this list, the middle 5 are following one of the conventions for [multi-field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html): `text` indexing for the top level field (e.g. `file.path`) and `keyword` indexing for the nested field (e.g. `file.path.raw`).

The other two are not following that convention:

* `user_agent.raw` is the one breaking from the convention the most. It's not part of a multi-field, and `user_agent.raw` is actually of type `text`, not type `keyword`. Meaning a user could not use this field for aggregations, as opposed to what the `.raw` convention establishes. It's named `.raw` because it's the full user agent string, prior to breaking it up into name, OS, version fields & so on.
* `event.raw` happens to be of type `keyword` which is good. But it's not actually part of a multi-field. It just means the *original* message is stored there.

**Naming of the nested field for multi-field**

I'm not 100% clear on the exact timeline. But my experience with ElasticSearch in monitoring pipelines from time immemorial, has been using the `.raw` nomenclature for the name of a sub-field of type `keyword`. Since about Stack v5 (or perhaps ES 2.x and Kibana 4.x), I've seen the naming shift to using `.keyword` for the nested field, instead of `.raw`.

Given the inconsistency I've outlined in the first part of this issue, I wonder if we shouldn't move to the new naming convention of using `.keyword` for multi-field, and having the freedom to use `.raw` for fields where we actually mean an original value.

[...]

PR #89, opened 2018-08-16
Create a section to explain some conventions used in ECS. Explain 2 conventions.

webmat:
Explaining the multi-fields convention will help keep each field description more to the point. Right now there's an inconsistent quick explainer in many of the multi-fields, which is redundant. Having this section to explain it a bit better will let us clean these field descriptions up later.

I also like that we're already using `keyword` for most IDs (I haven't checked
them all out). But it's something I would have pushed for if that hadn't
been the case already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -510,8 +510,8 @@ convention used is the following:
   The nested field has suffix `.raw` and is what you will use for aggregations.
   * Performance tip: when filtering your stream in Kibana (or elsewhere), if you
     are filtering for an exact match or doing a prefix search,
-    both `text` and `keyword` field can be used, but doing so on the `raw`
-    field will be much faster and less memory intensive.
+    both `text` and `keyword` field can be used, but doing so on the `keyword`
+    field (named `.raw`) will be much faster and less memory intensive.
 
 **Keyword only fields**
 
--- a/README.md
+++ b/README.md
@@ -487,14 +487,16 @@ ElasticSearch can index text multiple ways:
 
 * `text` indexing allows for full text search, or searching arbitrary words that
   are part of the field.
-* `keyword` indexing allows for exact match search (much faster) and allows for
-  aggregations (what Kibana visualizations are built on).
+* `keyword` indexing allows for much faster
+  [exact match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html)
+  and [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
+  and allows for [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html)
+  (what Kibana visualizations are built on).
 
-In some cases only one type of indexing makes sense for a field. E.g. no need to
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -479,6 +479,48 @@ Contributions of additional uses cases on top of ECS are welcome.
 * *Use prefixes.* Fields must be prefixed except for the base fields. For example all `host` fields are prefixed with `host.`. See `dot` notation in FAQ for more details.
 * Do not use abbreviations. (A few exceptions like `ip` exist.)
 
+## Understanding ECS conventions
+
+### Multi-fields text indexing
+
+ElasticSearch can index text multiple ways:
+
+* `text` indexing allows for full text search, or searching arbitrary words that
```
ruflin: Do you want to link to the ES docs here?

README.md
```diff
@@ -479,6 +479,48 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+  are part of the field.
+* `keyword` indexing allows for exact match search (much faster) and allows for
```
ruflin: And prefix which is pretty nice for version fields for example: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html

README.md
```diff
@@ -479,6 +479,48 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+  aggregations (what Kibana visualizations are built on).
+
+In some cases only one type of indexing makes sense for a field. E.g. no need to
+do full text search on an id, and nobody needs to do an exact match search on
```
ruflin: `normally there is no need for an exact match search on 2 kb stack trace`. Nobody is a bit strong :-)

README.md
```diff
@@ -479,6 +479,48 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+
+However there are cases where both types of indexing can be useful.
+As an example, log messages can sometimes be short enough that it makes sense
+to sort them by frequency (that's an aggregation). They can also be long and
+varied enough that full text search can be useful on them.
+
+Whenever both types of indexing are helpful, we use multi-fields indexing. The
+convention used is the following:
+
+* `foo`: `text` indexing. The top level of the field (its plain name) is used
```
ruflin: Will this convention change any of the existing fields we have? Mainly curious.

In general +1 on the convention.

webmat: Well most of the non-multi-fields text fields are actually `keyword`. That's why I follow all this immediately with a mention to that effect.

But any time we do multi-field, we need to follow this convention, otherwise it will get confusing :-)

Also, this PR is a side-effect of the thinking/discussion that's happening around https://github.com/elastic/ecs/issues/87. And obviously we will need to tweak the wording here if we go with `.keyword` instead of `.raw`, but I'm fine with that :-)

README.md
webmat: Yes I'll link to the docs in a few places, good point

webmat:
This is ready to be reviewed again

ruflin:
Thanks for adding the convention. Let's see what #87 will have as an affect on this.

Issue #102, opened 2018-08-23
Rename two field suffixes to ".original"


webmat:
In issue #87 I brought up the inconsistent usage of `.raw` between situations that meant "original/unparsed" and `.raw` meaning the keyword indexing of a multi-field.

We've decided to rename the fields with the unambiguous suffix `.original` instead:

- [ ] `event.raw` => `event.original`
- [ ] `user_agent.raw ` => `user_agent.original`

If you've noticed other field names that could be named `.original` please point them out here. This is a small cleanup I'll do next week.

ruflin:
@webmat As I already was working on the raw fields I opened https://github.com/elastic/ecs/pull/107

PR #106, opened 2018-08-28
Rename log.message to log.original

ruflin:
log.message was confused with message. To bring it in line with https://github.com/elastic/ecs/issues/102 the field log.message was renamed to log.original.

### DIFF
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -26,8 +26,6 @@
       phase: 1
       example: "Sep 19 08:26:10 localhost My log"
       index: false
-      # TODO: Best would be to disable ignore_above completely
-      ignore_above: 32766
       doc_values: false
       description: >
 
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -21,15 +21,18 @@
       description: >
         Offset of the beginning of the log event.
       example: 12
-    - name: message
+    - name: original
       type: keyword
       phase: 1
       example: "Sep 19 08:26:10 localhost My log"
       index: false
+      # TODO: Best would be to disable ignore_above completely
+      ignore_above: 32766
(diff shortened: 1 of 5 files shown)


webmat:
Love this wording instead of `log.message` indeed. It's less ambiguous.

schemas/log.yml
```diff
@@ -21,15 +21,18 @@
...
         Offset of the beginning of the log event.
       example: 12
-    - name: message
+    - name: original
       type: keyword
       phase: 1
       example: "Sep 19 08:26:10 localhost My log"
       index: false
+      # TODO: Best would be to disable ignore_above completely
+      ignore_above: 32766
```
webmat: I'm hesitant about the value of `ignore_above`, however. `ignore_above` does not truncate the value itself, it's a safety valve to avoid indexing values that are unexpectedly long in this field.

In other words, I don't think we would be likely to do exact match search or aggregations on an original log message that's this long. Log messages that are really short would be much more likely to be used this way, and as such I think 1024 would be enough.

ruflin: Agree with that. Removed the `ignore_above`.

schemas/log.yml
```diff
@@ -21,15 +21,16 @@
       description: >
         Offset of the beginning of the log event.
       example: 12
-    - name: message
+    - name: original
       type: keyword
       phase: 1
       example: "Sep 19 08:26:10 localhost My log"
       index: false
```
ruflin: @webmat As we have `index: false` ignore_above should not play a role anyways?

ruflin:
I don't expect this field to be used for aggregation at all.

I would prefer to keep here the merging as we had, so please merge :-)

Issue #104, opened 2018-08-23
Pore over all existing text fields to identify more that should be multi-fields


webmat:
Moving from `keyword` only indexing to multi-fields is a breaking change, so we need to make sure we're consistent with this.

- [ ] Review all current textual fields with an open mind and consider if they could benefit from being multi-field.
  * Since we currently do `keyword` by default, the question will mostly be: would we benefit from the ability to do partial searches on these values? E.g. filter for "guest" or "wifi" in field `network.name` or "prod" in `host.name`.
- [ ] Document this as an ECS convention: we consider doing multi-fields on any non-trivial text field.

vbohata:
Not just text fields should be multifields. For example file.uid, gid, http response_code is good to be integer (need to do comparison) but also very useful to be keyword (useful in visualisations, machine learning, ...).

webmat:
Actually discrete numbers like integers can be used for aggregations / visualizations. I don't think that's the case for floats, but integers definitely so. I often do visualizations on port fields, saved only as integer.

vbohata:
I know there are some limits I hit in the past. For example for multi metric machine learning job you can not split data per integer value. So I had to make http response code multi-field to be able to use it for split.

webmat:
Ah, this is great information! I'm not that familiar with ML yet. Thanks for pointing out, @vbohata

ruflin:
I'm starting to think if we should make decisions about `text` or `keyword` as the default per field. For example for `host.name` my take is it should be keyword by default as it's often used for filtering and aggregation and I see this as one of the main goals of ECS. For me in this case `text` is and addition and nice to have. This means if it is a multi field it would be `host.name.text`. 

If we would add `host.name.text` to the default fields and someone just does a query without specifying a field, he would also get results for this field.

webmat:
We're moving to almost exclusively `keyword` indexing at the canonical field name. Two exceptions are `message` and `error.message` which remain `text` on the canonical field name.

A rewrite of the section about text indexing is coming up, where we'll suggest people who need text indexing on other fields add `text` indexing as a sub-field instead, which will not be a conflict with ECS, and will not cause a breaking change.

So closing this issue.

PR #107, opened 2018-08-29
Rename .raw to .original fields

ruflin:
* Rename `user_agent.raw` to `user_agent.original` and make it a keyword. I think this always should have been a keyword.
* Rename `event.raw` to `event.original`.

Closes https://github.com/elastic/ecs/issues/102

### DIFF
--- a/schemas/user_agent.yml
+++ b/schemas/user_agent.yml
@@ -7,7 +7,7 @@
     show up in web service logs coming from the parsed user agent string.
   fields:
     - name: original
-      type: keyword
+      type: text
       description: >
         Unparsed version of the user_agent.
     - name: device
--- a/schemas/user_agent.yml
+++ b/schemas/user_agent.yml
@@ -6,8 +6,8 @@
     The user_agent fields normally come from a browser request. They often
     show up in web service logs coming from the parsed user agent string.
   fields:
-    - name: raw
-      type: text
+    - name: original
+      type: keyword
       description: >
         Unparsed version of the user_agent.
     - name: device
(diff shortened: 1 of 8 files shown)


schemas/user_agent.yml
```diff
@@ -6,8 +6,8 @@
     The user_agent fields normally come from a browser request. They often
     show up in web service logs coming from the parsed user agent string.
   fields:
-    - name: raw
-      type: text
+    - name: original
+      type: keyword
```
webmat: I would actually make this one multi-field.

Full text search will be useful for weird user agents that aren't recognized by the US parsers (various language HTTP libs, obscure crawlers, etc). Still useful to filter for everything that contains "Python", for example. So for this we need to retain full text search.

So I would go with FTS at the top + `.keyword` for the nested field.

webmat: Or you can consider this out of scope for now, and wait for the full review of fields in #104 to go MF. In that case we should leave it at `text` indexing.

use-cases/filebeat-apache-access.md
```diff
@@ -21,7 +21,7 @@ ECS fields used in Filebeat for the apache module.
 | <a name="http.response.body_sent.bytes"></a>*http.response.body_sent.bytes*  | *Http response body bytes sent, currently apache.access.body_sent.bytes*  | long  |   | `117`  |
 | <a name="http.referer"></a>*http.referer*  | *Http referrer code, currently apache.access.referrer<br/>NOTE: In the RFC its misspell as referer and has become accepted standard*  | keyword  |   | `http://elastic.co/`  |
 | <a name="user_agent.&ast;"></a>*user_agent.&ast;*  | *User agent fields as in schema. Currently under apache.access.user_agent.*<br/>*  |   |   |   |
-| [user_agent.raw](https://github.com/elastic/ecs#user_agent.raw)  | Raw user agent. Currently apache.access.agent  | text  |   | `http://elastic.co/`  |
+| <a name="user_agent.raw"></a>*user_agent.raw*  | *Raw user agent. Currently apache.access.agent*  | text  |   | `http://elastic.co/`  |
```
webmat: Ha, good catch. You've looked for `event.raw` in the examples as well, I assume?

schemas/user_agent.yml
ruflin: I consider it out of scope for now :-)

use-cases/filebeat-apache-access.md
ruflin: you mean the use cases? Hm, I didn't :-(

schemas/user_agent.yml
webmat: Agreed. So you'll revert back to `text` indexing?

ruflin: Can do. So far I was thinking it's a bug that it's text :-)

ruflin:
@webmat New version pushed and use cases updated.

PR #108, opened 2018-08-30
Add Beats compatible fields.yml for ECS

ruflin:
Add `type: group` to each prefix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-generate: schemas readme template
+generate: schemas readme template fields
 
 schemas:
 	python scripts/schemas.py
--- a/Makefile
+++ b/Makefile
@@ -41,4 +41,13 @@ template:
 	go get github.com/elastic/beats/libbeat/template
 	go run scripts/template.go > ./template.json
 
-.PHONY: generate schemas fmt check setup clean readme template
+fields:
+	cat schemas/*.yml > fields.tmp.yml
+	sed -i.bak 's/^/  /g' fields.tmp.yml
+	sed -i.bak 's/---//g' fields.tmp.yml
+	cat scripts/fields_header.yml > fields.yml
+	cat fields.tmp.yml >> fields.yml
+	rm fields.tmp.yml.bak
+	rm fields.tmp.yml
+
+.PHONY: generate schemas fmt check setup clean readme template fields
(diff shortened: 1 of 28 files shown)


Makefile
```diff
@@ -41,4 +41,13 @@ template:
...
 	go run scripts/template.go > ./template.json
 
-.PHONY: generate schemas fmt check setup clean readme template
+fields:
+	cat schemas/*.yml > fields.tmp.yml
+	sed -i.bak 's/^/  /g' fields.tmp.yml
+	sed -i.bak 's/---//g' fields.tmp.yml
+	cat scripts/fields_header.yml > fields.yml
+	cat fields.tmp.yml >> fields.yml
+	rm fields.tmp.yml.bak
```
andrewkroh: Where does `fields.tmp.yml.bak` come from?

fields.yml
```diff
@@ -0,0 +1,1215 @@
+---
+- key: ecs
+  title: ECS
+  description: >
+    ECS fields.
+  fields:
+  
+  - name: agent
```
andrewkroh: Shouldn't this have `type: group`?

andrewkroh: Would it be easy to move this over another 2 spaces? My goal is to have the it look like this

```
fields:
  - blah
```

rather than

```
fields:
- blah
```

which is consistent with indentations found in other areas of the file.

Makefile
andrewkroh: `make check` needs to updated to call `fields`. 

It looks like it could be simplified by making it depend on the targets that it invokes by calling `$(MAKE)` (e.g. change it to `check: generate fmt fields`).

ruflin:
Thanks @andrewkroh I rebased on master and pushed again. Somehow build had a diff, not sure why.

andrewkroh:
With the rebase it looks like it pulled in the recent .original changes so ‘make fields’ needs to run again.

ruflin:
The problem was that the fields target was missing for generate, so only check failed but it was not updated when running just make. I assume updating it always is the behaviour we want?

scripts/fields_header.yml
```diff
@@ -0,0 +1,6 @@
+---
+- key: ecs
+  title: ECS
+  description: >
+    ECS fields.
+  fields:
```
webmat: Not familiar with the automation we have in Beats that will consume this. I won't nest everything under `ecs.*`, will it?

ruflin: It will not. There is quite a bit of history on why this additional thingy is here but in short it has to do with docs and how the fields.yml was organised in the past.

PR #109, opened 2018-09-03
Rename url.href to url.original

ruflin:
Recently we changed several fields which contain the initial original value to .original. `href` now follows the same example.

There is an issue with original in the description as it is text by default and we have .raw. What is our recommended approach here?


webmat:
I would keep the keyword indexing but rename the nested field to `url.original.keyword`.

I think full text search on an original URL can be useful for ad hoc filtering on parts of the URL. So I keep it around.

ruflin:
@webmat Quite a few things changed since the initial PR. Could you have a look again?

Issue #86, opened 2018-08-15
Add network.address


vbohata:
I often need to somehow store the network name and network address. I propose to add network.address field for values like 192.168.0.0/24 ...

ruflin:
@webmat @robgil Would be great to get your feedback on this one.

@vbohata Can you share a bit of background on what queries you would run against the above field?

vbohata:
Currently for firewall logs I filter documents by exact match of some network address value like "192.168.50.0/24". It would be handy to be able to do some network vs. ip comparison natively in ES, but ES do not have a data type for network address.

robgil:
@ruflin @vbohata Yes, this makes sense. We definitely need a field for this. How about `network.cidr`? This would avoid confusion with other addresses like MAC or IP. We should probably also add `network.netmask` to be complete here.

vbohata:
network.cidr sounds reasonable. Netmask is also useful, but if there will be netmask, you should add also network base address. The network object will be maybe even more complicated because I can imagine someone will need to store both IPv4 and IPv6 addresses. Also there is missing interface object (maybe network.interface.name) which will be probably handy for metricbeat network module.

ruflin:
++ on cidr. Do you want to open a PR?

vbohata:
Yes, please.

vbohata:
Btw. I just found ES knows ip_range type which can be entered as CIDR ... good to know :)

ruflin:
Good point, should we make the field an ip_range then?

Also nice to see that `ip` addresses can be queried with a cidr notation: https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html

vbohata:
AFAIK for term aggregations it should be keyword, for ip comparisons it should be ip_range.
So probably network.cidr should be ip_range type and using fields network.cidr.keyword for keyword data type.

MikePaquette:
@vbohata @rgil Quick question:  From where would this field be populated?  For example is this a field that would normally be found in an event?  Can you provide examples?  Or is this a kind of enrichment field that you might populate by doing a lookup from an external db?

robgil:
[...]

PR #111, opened 2018-09-04
Adds .cidr and .netmask

robgil:
Ref: https://github.com/elastic/ecs/issues/86

For use with DHCP logs, Firewall logs, or BGP logs.

Adds `source.cidr`, `source.netmask`, `destination.cidr`,`destination.netmask`, `network.cidr`, and `network.netmask`.


webmat:
@ruflin:

> where do I search if I want to search across all of them?

A: Someone who uses `.cidr` and `.netmask` in multiple places should also add all these values in `related.cidr` and `related.netmask`, to simplify searching across all of them at the same time.

fields.yml
```diff
@@ -203,6 +203,14 @@
             IP address of the destination.
     
             Can be one or multiple IPv4 or IPv6 addresses.
+        - name: cidr
+          type: ip
```
webmat: Each `.cidr` field should actually be of type `ip_range` [1], not `ip` [2]. You can search an `ip` field with a CIDR range, but you can't store a range directly in it.

If you want to play with these in a Kibana console:

```JSON
PUT ip_test
{ "mappings": {
    "_doc": {
      "properties": {
        "address": {
          "type": "ip"
        },
        "range": {
          "type": "ip_range"
}}}}}

PUT ip_test/_doc/1
{ "address": "127.0.0.1" }
PUT ip_test/_doc/2
{ "address": "::1" }
PUT ip_test/_doc/3
{ "address": "255.255.255.0" }
# Fails:
PUT ip_test/_doc/4
{ "address": "10.0.0.0/8" }
PUT ip_test/_doc/4
{ "range": "10.0.0.0/8" }
# Fails:
PUT ip_test/_doc/5
{ "range": "255.255.255.0" }

GET ip_test/_search
DELETE ip_test
```



* [1] https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html
* [2] https://www.elastic.co/guide/en/elasticsearch/reference/current/range.html

MikePaquette:
I am closing this PR as it is outdated and the contributor is no longer active.

PR #116, opened 2018-09-18
Fix example when processed as asciidoc

adriansr:
The `fields` entry example consists of a dictionary surrounded by quotes. This causes problems when the example is used for documentation in asciidoc format, as it is interpreted as an escaped attribute, and the docs generation fails because it cannot find a "key1" attribute.

Fixed by removing the quotes around the example, which isn't perfect because it looks a bit ugly in the generated README.md and CSV formats.

Suggestions welcome.


webmat:
I'd like to wait for Nic to chime in after his vacation, though. He's obviously more well versed with the details here.

ruflin:
I'm pretty sure we will hit other issues down the road with these but then can use other methods in yaml like using `>`. Thanks for the fix.

PR #118, opened 2018-09-21
Rename all nested fields mentioned in #103 to .keyword

### LINKED ISSUE #103: Move to the ElasticSearch convention of using ".keyword" to denote nested fields indexed with type "keyword"
The list of fields at the time of writing this:

- [x] `file.path.raw` => `file.path.keyword`
- [x] `file.target_path.raw` => `file.target_path.keyword`
- [x] `url.href.raw` => `url.href.keyword`
- [x] `url.path.raw` => `url.path.keyword`
- [x] `url.query.raw` => `url.query.keyword`

webmat:
Closes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file based on the
 * Rename log.message to log.original. #106
 * Rename `event.raw` to `event.original`.
 * Rename `user_agent.raw` to `user_agent.original` and make it a keyword.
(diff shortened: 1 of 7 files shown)


andrewkroh:
I think the Auditbeat use-case document if affected by this change so probably it needs updated even if Auditbeat itself has not been updated.

ruflin:
@andrewkroh Merged it already but all the use cases will need updating. Also I think we don't verify well there multifields yet.

PR #119, opened 2018-09-21
Make cloud.instance.name and kubernetes.instance.name multi-fields

webmat:
Not ready for review. I want to discuss this with @ruflin after his vacation.

In order to make `cloud.instance.name` multi-fields, we need to make
`instance` into an object explicitly. But the automation doesn't seem to pick up
the multi-fields I'm adding below `name`.

I've done a pass at all text fields for #104. `kubernetes.container.name` will
also likely have the same problem.

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -133,9 +133,12 @@
             Instance ID of the host machine.
     
         - name: instance.name
-          type: keyword
+          type: text
           description: >
             Instance name of the host machine.
+          multi_fields:
+            - name: keyword
+              type: keyword
     
         - name: machine.type
           type: keyword
@@ -704,10 +707,13 @@
             Kubernetes annotations map
     
         - name: container.name
-          type: keyword
+          type: text
           description: >
             Kubernetes container name. This name is unique within the pod only.
             It is different from the underlying `container.name` field.
+          multi_fields:
+            - name: keyword
+              type: keyword
     
     - name: log
       title: Log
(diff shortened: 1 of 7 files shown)


fields.yml
```diff
@@ -126,16 +126,31 @@
           description: >
             Region in which this host is running.
     
-        - name: instance.id
-          type: keyword
-          example: i-1234567890abcdef0
-          description: >
-            Instance ID of the host machine.
+        - name: instance
+          type: group
```
ruflin: The group feature does not work in ECS. You have to use `instance.id` and `instance.name`.

webmat:
Haha indeed. I had assumed that a nested field named with a shorthand like `container.name` would need to be exploded into an explicit object, if we wanted to add multi-fields.

But your comment made me realize that this was an assumption on my part. I tested it out and we can indeed declare a multi-field directly with `container.name.keyword`. No need to explode the object. Will follow up with the updated code.

webmat:
@ruflin This is ready for review again

ruflin:
Before we merge this, let's have a discussion on how we could reduce the impact this has on all the add_metata processors in Beats because this is a breaking change.

webmat:
This PR is no longer relevant, given the move to `keyword` everywhere, and people having the option to add `text` indexing wherever they want as the nested multi-field (#140).

PR #121, opened 2018-09-24
Make 4 more text fields multi-field.

webmat:
For issue #104, I identified 6 fields I think should be made multi-field.

This PR addresses 4 of them:

* `device.vendor`
* `error.message`
* `file.path`
* `file.target_path`

The other two are `cloud.instance.name` and `kubernetes.container.name`. However I was hitting problems with the automation, so I'm submitting the 4 straightforward ones now and will look into the other ones in #119.

Note that this PR was added on top of #118 (commits from Sept 19th).

### DIFF
--- a/schemas/device.yml
+++ b/schemas/device.yml
@@ -23,6 +23,9 @@
       type: text
       description: >
         Device vendor information.
+      multi_fields:
+        - name: keyword
+          type: keyword
     - name: version
       type: keyword
       description: >
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file based on the
 * Rename `file.path.raw` to `file.path.keyword`, `file.target_path.raw` to `file.target_path.keyword`,
   `url.href.raw` to `url.href.keyword`, `url.path.raw` to `url.path.keyword`,
   `url.query.raw` to `url.query.keyword`, and `network.name.raw` to `network.name.keyword`.
+* Make `host.name` multi-fields. This is a breaking change because the top
+  field went from datatype `keyword` to `text`.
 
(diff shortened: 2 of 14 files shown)


schemas/device.yml
```diff
@@ -23,6 +23,9 @@
       type: text
       description: >
         Device vendor information.
+      multi_fields:
```
ruflin: This change would affect every single event we have in Beats at the moment and I'm very hesitant to make this change.

webmat:
I think we should talk in person about the following, I'd like to see if I'm missing any context:

> From my perspective in the optimal case we would potentially remove all text fields we have and only go with keyword

This would mean anyone who disagrees and makes some of their fields multi-field would no longer conform to ECS. This in turn would mean that any code that expects to interact with these fields would have problems doing so (the keyword field moves to the sub-field).

I do agree that all of this is painful. But "using text indexing" is not the pain point per se, IMO. The pain point is that we are using a variable standard of having `keyword` indexing on the canonical field by default, and moving to `text` at canonical field + `keyword` in nested field, when we want to do multi-fields indexing.

This is precisely the point I was raising at the beginning of the summer. For monitoring, should we consider always having `keyword` on the canonical field, and only when needed, define `text` on the nested field. Which is a non breaking change.

If we go this route, we could indeed just worry about the canonical field for now, and let people add `text` indexing only where they feel the need, in the nested field, aka **in a non-breaking change** :-)

webmat:
@ruflin The above was also essentially your suggestion in [this comment](https://github.com/elastic/ecs/issues/104#issuecomment-415717532)

ruflin:
Closing this as we decided to go the opposite direction. See https://github.com/elastic/ecs/pull/137

PR #122, opened 2018-09-24
added space between a and keyword

0xflotus:


webmat:
@0xflotus Thanks for your pull request!

`README.md` is a generated file. I've incorporated your proposed fix in PR #123. This PR also adds an HTML comment at the top of the README to indicate it's generated (to avoid situations like this :-) )

Since it's being addressed elsewhere, I'll close this PR. Thanks again, and you're welcome to jump into any other discussions you see in the issues, or submit other PRs.

PR #130, opened 2018-09-28
WIP Documentation on how to add non-ECS fields in one's schema

### LINKED ISSUE #114: Document various options to adding non-ECS fields to a stream that wants to conform to ECS
There have been some discussions about how to add non-ECS fields to a stream that users want to conform to ECS. ECS being a general purpose schema, this situation will happen often.

The current thinking is outlined in [this comment of issue 101](https://github.com/elastic/ecs/issues/101#issuecomment-421425708).

We need to document these guidelines in the repo. The gist:

1. You're free to add other top level objects or even keys under ECS objects, at the risk of having a breaking change eventually, if 1) ECS adds the exact same field and 2) we pick a different datatype (e.g. `keyword` vs `long`).
2. If you want to ensure you never experience a breaking change because of a clash with ECS, you can nest your custom data under `custom`, `cust` or `c` at any level (top level or under existing ECS objects). We want ECS to reserve `custom` and its shorthands for user's custom needs.
  * Note that we cannot guarantee that various third party integrations using ECS will not clash between one another under the `custom` namespace. In that sense, custom integrations may actually prefer to go with option 1 and use their product name for their custom nesting.

In the process of documenting this, we may also publish a temporary short list of top level objects that are coming to ECS but haven't made it to the repo yet. Here's a list of actual pending TLOs coming to ECS:

- `related`
- `threat`
- `vulnerability`

@MikePaquette @ruflin Please don't hesitate to add to this, if I forgot anything

webmat:
This is not ready for final review. This PR is just to facilitate collaboration
on the redaction of this doc.

Closes #114


Makefile
```diff
@@ -25,12 +25,12 @@ clean:
 	find ./use-cases -type f -name '*.md' -not -name 'README.md' -print0 | xargs -0 rm --
 
 readme:
-	cat docs/intro.md > README.md
+	cat docs/1-intro.md > README.md
```
ruflin: What is the number for? Is it for sorting? If I decide to move about on top of implemented do I have to rename the files? If yes, does not make sense to me.

docs/3-implementing.md
```diff
@@ -20,7 +20,14 @@
 * *General to specific.* Organise the prefixes from general to specific to allow grouping fields into objects with a prefix like `host.*`.
 * *Avoid repetition.* Avoid stuttering of words. If part of the field name is already in the prefix, do not repeat it. Example: `host.host_ip` should be `host.ip`.
 * *Use prefixes.* Fields must be prefixed except for the base fields. For example all `host` fields are prefixed with `host.`. See `dot` notation in FAQ for more details.
-* Do not use abbreviations. (A few exceptions like `ip` exist.)
+* Do not use abbreviations or word contractions, use full words.
+
+**Exceptions**
+
+In some cases, ECS naming conventions are overridden by common parlance, if
```
ruflin: Learned a new word: `parlance`.

docs/3-implementing.md
```diff
@@ -66,3 +73,55 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
 Despite the fact that IDs are often integers in various systems, this is not
 always the case. Since we want to make it possible to map as many data sources
 to ECS as possible, we default to using the `keyword` type for IDs.
+
+## Using non-ECS fields in your event stream
+
+The goal of ECS is to define a common set of fields to make working across various
+streams easier. Each stream will have their own particularities, however, and ECS
```
ruflin: What do you mean by streams here?

docs/3-implementing.md
```diff
@@ -66,3 +73,55 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+cannot account for all possibilities. It's therefore important that implementers
+feel at ease adding fields and top level objects in their schema that are
```
ruflin: Why differentiate here between top level objects and fields? It's all fields.

docs/3-implementing.md
```diff
@@ -66,3 +73,55 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+
+Before going into strategies, let's define what constitutes a conflict.
+
+#### Conflicts with ECS
+
+* A custom field has the same name as a new ECS field, but an incompatible type:
+  * `float` vs `long`
+  * `text` vs `keyword`
+  * `keyword` vs numeric (`integer`, `long`, etc.)
+  * discrete field vs an object (e.g. a `keyword` field vs a field that contains other fields)
```
ruflin: Do you mean here core datatypes? https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html

docs/3-implementing.md
```diff
@@ -66,3 +73,55 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+Let's review some strategies to reduce the risk of conflicts happening.
+
+### Conflicts
+
+Before going into strategies, let's define what constitutes a conflict.
+
+#### Conflicts with ECS
+
+* A custom field has the same name as a new ECS field, but an incompatible type:
+  * `float` vs `long`
```
ruflin: You list here only a small set of examples with conflicting types. I think it should be made clear that this are some examples and above could be just stated `have a different type`. This is correct in most cases as very few types are compatible.

docs/3-implementing.md
```diff
@@ -66,3 +73,55 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+* A new field is added to ECS, and matches the name and type of an existing
+  custom field.
+  * E.g. you define `process.xpid` as a `keyword` field, and ECS adds
+    `process.xpid` as a `keyword` field.
+* A custom `keyword` field gets defined with the same purpose, with a different
+  `ignore_above` value than a new ECS field.
+
+#### Conflicts with Third Parties
+
+Elastic is developing ECS and is adjusting their solutions to leverage ECS.
```
ruflin: I think we put it a bit too broad here. We should care about not conflicting with ECS, third parties will be a problem of it's own but not worth mention here as otherwise it gets too complex.

docs/3-implementing.md
webmat: Yay!

Makefile
webmat: Yes. I was just trying to make it slightly faster to go edit the right file right away.

If this makes your skin crawl, I'll just roll this back LOL. No big deal :-)

docs/3-implementing.md
webmat: Event streams. An abstract\* concept of a whole bunch of events of a given kind. 

\* I say abstract because it's not necessarily tied to an ElasticSearch index. For example, the way beats does indexes, all event streams (e.g. Apache logs, system logs, etc.) picked up by Filebeat would be in the same index.

docs/3-implementing.md
webmat: I think it's worthwhile to be explicit here. Otherwise I'm sure we'll be asked either question:

* "But can I create my own top level objects?"
* "But can I add fields to other ECS objects?"

I'm trying to make it clear that both are expected and ok from the get go.

I don't mind rewording a bit, though.

docs/3-implementing.md
webmat: Almost. But non-core data types could also cause this problem. IP and geo point come to mind :-) What I mean is "not complex" to "complex" data type.

docs/3-implementing.md
webmat: I guess this comment supersedes the other. Yeah I'll word it this way, it will be much simpler. Something along the lines of:

> * A custom field has the same name as a new ECS field, but an incompatible type (most fields are incompatible).

Makefile
ruflin: It does, reminds me of people trying to organise word document and excel sheets 😆

karenzone: @ruflin  Sounds good.  Happy to do another review of all recent changes. Please ping me when you're ready for me to take a look.

webmat: Here's how I organize my ECS excel sheets 😆 

```
C:\> dir
ecs.zip
ecs-backup.zip
ecs-final.zip
ecs-final-(bobs-corrections).zip
ecs-true-final.zip
ecs-last-minute-corrections-really-final.zip
```

docs/3-implementing.md
webmat: Yeah I read your comment too fast. I will reformulate similar to what I wrote here. But I'll also keep the examples, and make it clearer that these are just examples, as you suggest.

docs/implementing.md
```diff
@@ -66,3 +73,69 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+Before going into strategies, let's define what constitutes a conflict.
+
+#### Conflicts with ECS
+
+* A custom field has the same name as a new ECS field, but a different type
+  (most field types are incompatible). Here are concrete examples:
+  * `float` vs `long`
+  * `text` vs `keyword`
+  * `keyword` vs any numeric (`integer`, `long`, etc.)
+  * discrete field vs an object (e.g. a `keyword` field vs a field that contains other fields)
```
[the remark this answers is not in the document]
webmat: What you mean is to remove the "discrete vs object" example?

I agree it's not as straightforward as the other ones. I wanted to include it because it'll speak very clearly to people who are about to go through the `source` field transition... But perhaps this is not necessary.

docs/implementing.md
```diff
@@ -66,3 +73,69 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+
+When indexing more than one type of event in the same index, a conflict will
+be experienced as a mapping conflict. This means all events that don't conform
+with the current mapping will be rejected by ElasticSearch.
+
+When ingesting into separate indices per event source, a conflict will not affect ingestion.
+Each event stream will be ingested in their respective index.
+The problems will be experienced at query time.
+If you have a Kibana index pattern to query all of your ECS indices at once
+(e.g. `*-ecs-*`), you will not be able to query on the conflicting fields.
```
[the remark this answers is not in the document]
webmat: Yeah good point, this should be considered before being dropped in the doc like this ;-)

docs/implementing.md
```diff
@@ -66,3 +73,69 @@ plain field (`foo`) will be of type `keyword`, with no nested field.
...
+when thinking about avoiding conflicts, it's important to consider the
+broader ecosystem, not just "will I conflict with Elastic?".
+
+### Consequences of a Conflict
+
+The consequences of a conflict will vary depending on whether you use an index
+per event stream or if you're indexing more than one stream in the same index.
+
+When indexing more than one type of event in the same index, a conflict will
+be experienced as a mapping conflict. This means all events that don't conform
```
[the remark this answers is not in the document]
ruflin: I would keep the content but still link to the mapping page. It explains what mappings actually are.

docs/implementing.md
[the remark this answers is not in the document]
ruflin: What I mean use a specific example `keyword` vs `object` here. From the context I understand what you mean by `discrete` but for me it's not something I would directly understand.

jacob87o2:
So, any idea how to reduce the risk of conflict so far? How to log data related to business domain, like `productId`, `customerId`, `orderStatus`... or groups like `product.`?

Maybe fields and groups not defined by ECS should be prefixed with some character, like `_`? Ex. `_productId` or `_product.id`?
Or maybe there should be a group that will never have field any field defined in ECS - ex. `non_ecs.`?

webmat:
Hey @jacob87o2 thanks for reaching out. You've found an old abandoned PR, but in fact I just redid that and merged it to master here https://github.com/elastic/ecs/pull/713 :-)

You can see this new page on the "master" section of the official docs here: https://www.elastic.co/guide/en/ecs/master/ecs-custom-fields-in-ecs.html.

This page only addresses various approaches one can take to add custom fields to their ECS events. But this page doesn't address the consequences of getting into a conflict, as this is too dependent on how the system is set up.

I'll close this PR now. I consider this has been resolved by #713.

PR #131, opened 2018-10-01
Remove log.offset and log.line from ECS

ruflin:
These fields were too specific and does not fit ECS well. It is still fields tthat are used by filebeat and other logging tools but this does not mean they have to be in ECS.


webmat:
LGTM. Merging.

PR #132, opened 2018-10-01
Remove two top level objects: kubernetes and tls.

webmat:
As discussed in last week's meeting, we're removing these two top level objects.
They may re-appear later as a third level object (whatever comes after Core and
Extended) or as more generic / less product-centric objects.


ruflin:
Can we put these two into a kubernetes use case and a tls use case? This will make it very easy to revive the fields if needed.

ruflin:
@ph @exekias FYI

webmat:
I've added them both back as use cases. Same exact field list, I've added more concrete examples for fields for kubernetes, but that's all.

First time I spend time in the use cases. Love how the existing ECS field default values get pulled in automatically :-)

Ready for review again.

ruflin:
Can you update the CHANGELOG?

template.json
```diff
@@ -436,36 +436,6 @@
             }
           }
         },
-        "kubernetes": {
```
ruflin: @adriansr This might have an effect on the ecs.yml we use in Beats.

webmat:
Ah yes, sorry about the changelog :-)

exekias:
Does this mean we should rename any of the kubernetes fields in Filebeat / Metricbeat?

ruflin:
@exekias No. They will probably show up in ECS again in the near future it's just not clear yet what we will call this. They are still in the use cases covered.

Issue #93, opened 2018-08-17
ECS Core and ECS Extended


webmat:
We've discussed creating a separation between the most stable and well understood fields (ECS Core) and the more experimental ones (ECS Extended). This issue is meant to have a discussion on how we should do this, and track the subtasks.

- [x] Write documentation explaining the difference between the two
- [x] Assign each field to one of the two levels (#133)
  - [x] Add the appropriate markings in all of the `schema/*.yml` files
- [x] Modify the scripts to support this new metadata about each field
  - [x] Output Level instead of Phase in the csv
  - [x] Output Level in a new column in the HTML
- Later, only if needed
  - Option to generate core only csv & template?
- ... (comment below for any other suggestion)

webmat:
We've discussed two options about how to display Core vs Extended fields:

- Add a column in the tables that says "Core" or "Extended"
- Split the tables in two. So each section (e.g. [event](https://github.com/elastic/ecs#-event-fields)) would have a table of all "Core" fields, followed by a table of all "Extended" fields.

webmat:
Another note: a field being marked as "Extended" is not a permanent thing. Rather it should be seen as a more experimental section of ECS, where things may change a bit more. When the community gels around the usage of a field or set of fields, they could be promoted to "Core", and be considered more stable.

webmat:
Related to #54 as well. After Core & Extended, users of ECS have the option to add their custom fields for their use cases (without them being part of ECS).

ruflin:
I started to play around a view options on how we could visualise it. I think the challenge is mainly on the UI side as in the fields.yml it would be just a flag. I took agent fields only as an example, does not mean it has to be split up.

Currently I prefer Option 2 and 3 over Option 1. One of the reasons to have Core and Extended is to give users new to ECS a smaller set of fields. If we mix them together under one title they see directly all to get started (this also might be a benefit). Also I assume there are going to be fields under prefixes like `agent` that are going to be extended but also full prefixes like `hash.*`. If we put it in under the same title, where would the full prefixes go?

The reason I like Option 3 is that it is similar to 1 but also brings benefits of 2. If we go with 3 we should have an indication in the table which of the fields are core (like bold or italic?).


# Option 1: Core and Extended under 1 title

## <a name="agent"></a> Agent fields

The agent fields contain the data about the agent/client/shipper that created the event.

### Core

| Field  | Description  | Type  | Multi Field  | Example  |
|---|---|---|---|---|
[...]

PR #133, opened 2018-10-01
WIP add level (Core, Extended) on all fields.

webmat:
Subtasks:
- [x] Assign all fields either value
- [x] Remove all mentions of `phase`

This issue is only meant to cover assigning a level to the fields for #93. No modifications to the scripts in this PR.


ruflin:
@webmat We could even split this up further. In a first PR only to the fields.yml modification and follow up with scripts etc. in parallel.

Let me know in case this is ready for review. It seems to have all the fields.yml changes inside?

webmat:
@ruflin Yes that's true. I'll modify the scripts in a separate PR.

All fields should be assigned. If one is not, or if it's not == to the voting doc, it's a bug :-)

MikePaquette:
LGTM. Thanks @webmat !

webmat:
Thanks folks!

PR #137, opened 2018-10-19
Switch most fields to keyword as the type

ruflin:
We started to use multi fields in ECS and if we used a multi field, the base field was text as is the default in Elasticsearch. The problem with this is we decide in the future to make a field multi field, it would be breaking change as what was before a keyword now becomes text. To prevent this all fields except `message` are by default a keyword. Making a multifield out of a field is then a non breaking change.

On the ECS side we still need to figure out what our recommendation is for naming multiple fields like `.analyzed, .text` or others.

This change has also an affect on Beats as in ttps://github.com/elastic/beats/pull/8313 the fields.yml from ECS was added to Beats. There was even a breaking change I think we missed when switch to ECS there as the `http.response.body` in packetbeat was text and in Metricbeat keyword.

The following fields were changed to keyword and multifield removed for now. We can add it later again when we figure out the convention:

* device.vendor
* file.path
* file.target_path
* http.response.body
* network.name
* organization.name
* url.href
* url.path
* url.query
* user_agent.original

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file based on the
 * Remove `log.offset` and `log.line` as too specific for ECS. #131
 * Remove top level objects `kubernetes` and `tls`. #132
 * Remove `*.timezone.offset.sec` fields as too specific for ECS at the moment. #134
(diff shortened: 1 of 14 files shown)


ruflin:
@adriansr After this is merged we should update the ECS fields.yml in Beats again and mention the breaking change that happened in Packetbeat.

PR #138, opened 2018-10-19
Disable indexing for user_agent.original

ruflin:
All the original fields are not indexed. For consistency the same should be true for user_agent.original.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file based on the
 * Remove `log.offset` and `log.line` as too specific for ECS. #131
 * Remove top level objects `kubernetes` and `tls`. #132
 * Remove `*.timezone.offset.sec` fields as too specific for ECS at the moment. #134
+* Set indexing to false for `user_agent.original`.
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file based on the
 * Remove `log.offset` and `log.line` as too specific for ECS. #131
 * Remove top level objects `kubernetes` and `tls`. #132
 * Remove `*.timezone.offset.sec` fields as too specific for ECS at the moment. #134
+* Set indexing to false for `user_agent.original`.
```
webmat: Why turn indexing off here?

Sure it may not be the most often used field, but I can still see use cases where kw indexing can be useful.

* exact match filtering on the whole user agent string
* aggregations on typically shorter UA strings from various programming languages, which are often not well parsed by UA parsers

ruflin: I think if we index it by default, it should be text instead of keyword. Not indexing it by default I would hope it saves on space as the user_agent can be pretty extensive.

We can do the same trick here as we do for the keywords. Not index it by default and when someone needs it, he can add `.text` as a multifield.

webmat: I think turning off indexing here is premature optimization and will cause problems down the line.

At the very least, though, I would change the disabled type to `keyword`, not `text`. We need to stop the `text` indexing on the canonical field.

ruflin:
Closing in favor of https://github.com/elastic/ecs/pull/240

PR #139, opened 2018-10-19
Add check that field level is set on all fields

ruflin:


webmat:
@ruflin Not actually sure why the build is failing. All generated files are up to date, and the script doesn't raise an exception now, as there's no conflict.

Not sure what is causing these 2 newlines to be detected in the script but `git diff`.

Perhaps I merged too fast :-)

ruflin:
I think master is fine it was just a flaky CI instance.

PR #140, opened 2018-10-19
Introduce the new convention for multi-fields text indexing to the README.

webmat:
The nested field name is open for discussion. I started with the name `*.text`
because I think it will be intuitive even for less technical users. If they want
to search the text, they need to targed `fieldname.text`. To me it sounds
less technical than `*.analyzed` or `*.fts`.

I'm however open to discussing this point before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,7 +458,7 @@ Contributions of additional uses cases on top of ECS are welcome.
 
 ### Multi-fields text indexing
 
-ElasticSearch can index text multiple ways:
+Elasticsearch can index text multiple ways:
 
 * [text](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html)
   indexing allows for full text search, or searching arbitrary words that
@@ -470,35 +470,38 @@ ElasticSearch can index text multiple ways:
   and allows for [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html)
   (what Kibana visualizations are built on).
 
-By default, unless your index mapping specifies otherwise, ElasticSearch indexes
-text field as `text` at the canonical field name, and indexes as second time
-as `keyword` in a nested field:
+By default, unless your index mapping or index template specifies otherwise
+(as the ECS index template does),
+Elasticsearch indexes text field as `text` at the canonical field name,
+and indexes a second time as `keyword`, nested in a multi-field.
+
+Default Elasticsearch convention:
 
 * Canonical field: `myfield` is `text`
-* Nested field: `myfield.keyword` is `keyword`
+* Multi-field: `myfield.keyword` is `keyword`
 
-For monitoring use cases, we need almost exclusively `keyword` indexing, with
-full text search on very few field fields. Given this premise, ECS defaults
-all text indexing to `keyword` at the top level (with only two exceptions).
+For monitoring use cases, `keyword` indexing is needed almost exclusively, with
+full text search on very few fields. Given this premise, ECS defaults
+all text indexing to `keyword` at the top level (with very few exceptions).
 Any use case that requires full text search indexing on additional fields
-can simply add a nested field for full text search.
-Doing so does not conflict with ECS, as the canonical field name will remain
-`keyword` indexed.
+can simply add a [multi-field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html)
+for full text search. Doing so does not conflict with ECS,
+as the canonical field name will remain `keyword` indexed.
 
 ECS multi-field convention for text:
 
 * Canonical field: `myfield` is `keyword`
-* Nested field: `myfield.text` is `text`
+* Multi-field: `myfield.text` is `text`
 
 #### Exceptions
 
-The only two exceptions to this convention are fields `message` and `error.message`,
-which are indexed for full text search only, with no nested field.
+The only exceptions to this convention are fields `message` and `error.message`,
+which are indexed for full text search only, with no multi-field.
 These two fields don't follow the new convention because they are deemed too big
 of a breaking change with these two widely used fields in Beats.
 
 Any future field that will be indexed for full text search in ECS will however
-follow the multi-field convention where `text` indexing is the nested field.
+follow the multi-field convention where `text` indexing is nested in the multi-field.
 
 ### IDs are keywords not integers
 
--- a/README.md
+++ b/README.md
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
 
 ElasticSearch can index text multiple ways:
 
-* [text](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html) indexing allows for full text search, or searching arbitrary words that
+* [text](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html)
+  indexing allows for full text search, or searching arbitrary words that
   are part of the field.
-* [keyword](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html) indexing allows for much faster
-  [exact match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html)
-  and [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
+* [keyword](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html)
+  indexing allows for much faster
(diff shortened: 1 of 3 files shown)


webmat:
If there's ever a change or a concept introduced by ECS that may rain 🔥on us, this is this new convention 😄 

I'd like the opinions of a few people on this one please :-)

I think doing the nesting differently (with full text search nested instead of keyword nested) is something we don't have a choice, if we want to avoid future breaking changes as much as possible. It's probably the most controversial part, though.

Then there's the question of "what should we name the nested field?", which is less critical. I think `*.text` is the least technical word, so that's what I went with in this initial writing of the documentation. But I'm open to suggestions.

webmat:
Note that the build fails because of an unrelated change introduced in #139

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
-  and [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
+* [keyword](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html)
+  indexing allows for much faster
+  [exact match filtering](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html),
+  [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
   and allows for [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html)
   (what Kibana visualizations are built on).
 
-In some cases, only one type of indexing makes sense for a field.
+By default, unless your index mapping specifies otherwise, ElasticSearch indexes
```
ruflin: `Elasticsearch`

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+text field as `text` at the canonical field name, and indexes as second time
```
ruflin: Note here: The ECS templates has keyword as default specified: https://github.com/elastic/ecs/blob/master/template.json#L13

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+as `keyword` in a nested field:
 
-However there are cases where both types of indexing can be useful, and we want
-to index both ways.
-As an example, log messages can sometimes be short enough that it makes sense
-to sort them by frequency (that's an aggregation). They can also be long and
-varied enough that full text search can be useful on them.
+* Canonical field: `myfield` is `text`
+* Nested field: `myfield.keyword` is `keyword`
```
ruflin: Lets not used  `nested field` here normally people think of nested datatype in Eleasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
 
-Whenever both types of indexing are helpful, we use multi-fields indexing. The
-convention used is the following:
+For monitoring use cases, we need almost exclusively `keyword` indexing, with
```
ruflin:
```suggestion
For monitoring use cases, almost exclusively `keyword` indexing is needed, with
```

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+full text search on very few field fields. Given this premise, ECS defaults
+all text indexing to `keyword` at the top level (with only two exceptions).
```
ruflin:
```suggestion
all text indexing to `keyword` at the top level (with only few exceptions).
```

I don't like the docs to become outdated just because we add or remove one.

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+Any use case that requires full text search indexing on additional fields
+can simply add a nested field for full text search.
```
ruflin: It is a multi field not a nested field. Best also link here to the multi field docs.

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
-    field (named `.raw`) will be much faster and less memory intensive.
+ECS multi-field convention for text:
 
-**Keyword only fields**
+* Canonical field: `myfield` is `keyword`
+* Nested field: `myfield.text` is `text`
 
-The fields that only make sense as type `keyword` are not named `foo.raw`, the
-plain field (`foo`) will be of type `keyword`, with no nested field.
+#### Exceptions
```
ruflin: I don't think we need to document this.

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+
+The only two exceptions to this convention are fields `message` and `error.message`,
+which are indexed for full text search only, with no nested field.
+These two fields don't follow the new convention because they are deemed too big
```
ruflin: I think the purpose of `message` is to be index so even if we would not have it as text today I think it should be text.

README.md
```diff
@@ -460,38 +460,45 @@ Contributions of additional uses cases on top of ECS are welcome.
...
+of a breaking change with these two widely used fields in Beats.
+
+Any future field that will be indexed for full text search in ECS will however
```
ruflin: Let's really skip the Exception part here as I think we need to discuss when we encounter other fields with text purpose what we should do with it and not get ahead of us in the docs here.

README.md
webmat: Yes. In our case, the index template *does* specify otherwise. Perhaps I should indeed reword this a bit, to avoid confusion.

I was referring to the broader use case when Elasticsearch, not while using the ECS index template.

README.md
webmat: Ok I can see how that can cause confusion. Looking at the documentation for [multi-fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html), there is no specific wording to reference that other field.

Do you have a suggestion on what we should call that nested field?

A "multi-field" sounds to me like it refers to all the fields pointing at the same data, not specifically the nested field.

README.md
webmat: Good catch, forgot to drop that old text.

webmat: Sorry, misread that. I thought we still had a reference to `.raw` fields.

I think it's worthwhile to document the break from this new convention. These are widely used fields and they will behave exactly the reverse of this new convention we're introducing. I think it's helpful to make sure it's clear.

Or alternately, I would make them follow this new convention, but declare them right away as a multi-field. E.g. `message` is `keyword` and `message.text` is `text`. This would avoid the uncomfortable explanation of explaining two exceptions, and would let people do fast exact match filtering based on the `message` field without having to reintroduce `message.keyword` ;-)

ruflin: As mentioned before, these are not exceptions for me.

README.md
ruflin: Instead of canonical fields, I would use core data types (coming from the es docs) and use multi fields here. Something like `the field inside the multi field`?

README.md
```diff
@@ -458,40 +458,50 @@ Contributions of additional uses cases on top of ECS are welcome.
...
-  and [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
+* [keyword](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html)
+  indexing allows for much faster
+  [exact match filtering](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html),
+  [prefix search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html),
   and allows for [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html)
   (what Kibana visualizations are built on).
 
-In some cases, only one type of indexing makes sense for a field.
+By default, unless your index mapping or index template specifies otherwise
```
ruflin: Instead of writing this here, could we link to the ECS docs? We should if possible not explain how Elasticsearch works in ECS.

webmat:
Rebased on top of the autopep8 fix.

README.md
webmat: Not sure what you mean by this. `message` is indexed as `text` right now.

Are you mixing this up with my comment here https://github.com/elastic/ecs/pull/138#pullrequestreview-166647391 😄 ?

webmat:
Since it's been approved 2 days ago, I'll merge and we can indeed tweak later.

PR #141, opened 2018-10-22
Harmonize hostname, and move to domain + full_domain

### LINKED ISSUE #84: Clarify use of hostname, subdomain, domain in source/destination
It's not clear to me how to populate the `hostname`, `subdomain`, and `domain` fields of `source` / `destination`. More detailed descriptions of each field are needed with examples. 

It would probably be helpful to establish some terminology that could be used in clarifying the descriptions.

Terms:
- FQDN: fully qualified domain name
- TLD: top level domain (e.g. `.com`, `.net`, `.bmw`, `.us`) [[list of TLDs](http://data.iana.org/TLD/tlds-alpha-by-domain.txt)]
- eTLD: effective top level domain (e.g. `.com`, `.co.uk` and `pvt.k12.wy.us`) [these get determined with the help of the [public suffix list](https://publicsuffix.org/)]
- eTLD+1: effective top level domain plus one level (e.g. `example.com`, `example.co.uk`)
- SLD: second level domain (e.g. `co` is the SLD of `www.example.co.uk`)

Examples showing the mappings of these FQDNs to ECS would probably be sufficient to clarify the topic for me.

- `example.com`
- `www.example.com`
- `www.example.co.uk`

Logstash has a [TLD filter](https://github.com/logstash-plugins/logstash-filter-tld/blob/16f6f911ec0c8c1bbcfe48268af41a68e24bad03/spec/filters/tld_spec.rb#L44-L50) that uses similar field names, possibly(?) with different meanings.

webmat:
- Gets rid of the last instances of `host.name`
- Moves to the new standard of `domain` + `full_domain` everywhere, including
  under `url.`
- I added a comment on source & destination's `.hostname` field. A local agent
  is unlikely to be able to populate this reliably (at best its side of the
  connection). It's expected to be added in as an enrichment. It's also a cue
  that it may be omitted in cases where operational monitoring is the goal,
  rather than security.

Closes #84


webmat:
We're moving from `full_domain` to `fqdn`

webmat:
@ruflin Ok, `host.name` commit has been extracted out to #144 :-)

schema.csv
```diff
@@ -114,13 +114,14 @@ service.state,keyword,0,
...
 source.domain,keyword,0,
+source.full_domain,keyword,0,
 source.hostname,keyword,0,
 source.ip,ip,0,
 source.mac,keyword,0,
 source.port,long,0,
-source.subdomain,keyword,0,
+url.domain,keyword,0,elastic.co
 url.fragment,keyword,0,
-url.host.name,keyword,0,elastic.co
```
ruflin: We can't remove the hostname here but should rename it to `url.hostname`. These entries are pretty standard for url_parsers across various languages.

webmat: I agree that the field `url.host.name` needs to change to something else. But I disagree that it should become `url.hostname`.

Please read https://github.com/elastic/ecs/issues/84#issuecomment-431951431 to understand why I'd rather introduce `domain` here instead of `hostname`.

ruflin: The url.hostname is what shows up in the url. It could also be an ip address.

webmat: For the record, @ruflin won me over, at least for `url.hostname`: #147.

The deciding argument was that a URL can either have a domain name or an IP there. So using the word `domain` was not suitable.

webmat:
Closing this. `host.hostname` and `url.hostname` have been dealt with in #144 and #147 respectively. Also latest discussions have made clear that `domain` & `full_domain` were not necessarily the best naming for these fields. Let's wait for consensus to emerge in #84.

Issue #143, opened 2018-10-23
agent.environment and service.environment


ruflin:
An agent and a service are often run in a certain environment like testing, staging, production. To group services and agents together by environment I'm proposing the following fields to be added to ECS:

* service.environment
* agent.environment

ruflin:
@graphaelli Would this be helpful for APM?

graphaelli:
Certainly, we actually already use `service.environment` for this purpose.  We capture environment variables in `env`.

ruflin:
I wonder if one event is always from max 1 environment. So it can be that the service we monitor is in production but the agent is in testing? If yes, we could unify the two fields in event.environment.

webmat:
I would rather add `environment` once, in a general enough place, though.

I feel like needing two instances of environments (e.g. in the example of pre-release agent you mention) is an edge case. I'd rather track a beta release of an "off the shelf" tool like a third party monitoring agent via its version.

I've always tracked which environment my stuff was in with free form labels in previous lives. Be it in Chef, some competing monitoring systems, or in my Elastic Stack logging pipeline. So I like the idea of adding it explicitly to ECS.

ruflin:
Sounds like we should go with `event.environment` then?

To prevent any confusion: I expect this field NOT to be for environment variables but to be set to the environment the service / agent / device / host is running. Values could be `testing`, `staging` etc.

graphaelli:
I see environment is part of the name/version/environment triple that uniquely identifies a service so `service.environment` makes more sense to me.  Also, `The event fields are used for context information about the data itself` - I think this is about the service not the event itself. 

APM has to move this either way we go (service is nested under context right now).

ruflin:
@graphaelli I think you make a good point here: +1 on having both then.

webmat:
In ECS I would much rather have one definition of environment, at the top level, so `environment`. 

When thinking about services, `service.environment` makes sense, of course. But my problem with this is when processing logs and metrics across the board (a lot of which won't be using `service.` specifically), then setting `service.environment` will seem out of place.

Conversely, I do think it will make sense to people if their APM events have `environment`, `service.name` and `service.version` (), even if `environment` is not nested under `service.

ruflin:
I was more thinking to have also `device.environment`, meaning adding it to all objects where it makes sense.

webmat:
But why have multiple places to store `environment`? Do you see instances of having one event with a different value for `environment` for a device, host, service etc. at the same time?

[...]

Issue #142, opened 2018-10-23
service.type and service.name


ruflin:
Currently ECS has a `service.name` and the expected content when for example talking to a MySQL service is that it contains `mysql`. APM on the other hand uses `service.name` as the manual set name by the user for the service in which the agent is running and if the service talks to MySQL, `service.type` is set as `mysql`. The following proposal is to add `service.type` and define the usage of both.

A few examples to explain the usage

## Metricbeat monitoring MySQL and service `foo`

Metricbeat monitors with the MySQL module a MySQL host and with the HTTP module a user built service `foo`. The events collected from these will look as following:

MySQL events:

```
service.name:
service.type: mysql
```

Foo events:

```
service.name:
service.type: foo
```

It will be up to user to define the field `service.type` in the HTTP module. We should make it easier on our end.

### Questions

* Should be default service.name be filled with `service.type` if it's empty?

## Metricbeat monitoring 2 MySQL servers

Metricbeat is monitoring two different MySQL servers on the same host. One server stores user data the other one is for tracking. The events will look as following:

User data MySQL instance:

```
service.name: user-data-mysql
service.type: mysql
```

Tracking data MySQL instance

```
service.name: mysql-tracking
service.type: mysql
```

The user can set a name for each service he is monitoring. This will allow later to differentitate between the two services. We should recommend to prefix the name with the service.type.

## Metricbeat monitoring an Elasticsearch cluster

Metricbeat monitors and Elasticsearch cluster across different hosts. The cluster name is `baz`.

```
service.name: baz
service.type: elasticsearch
```

The name of the cluster is put in as `service.name`. This allows to group together all nodes related to a cluster. Currently `elasticsearch.cluster.id` or `elasticsearch.cluster.name` are use for this. One of these fields should be copied to the service.name.

## APM agent running in service foo 

An apm agent running inside the service `foo` will report it's data as following. The service is built on top of node.js:

```
service.name: foo
service.type: node.js
```

The `service.name` is set by the user inside the agent. It could also be `foo-prod1` and `foo-prod2`. In this case these would be two different services.

## APM agent running in service foo talking to mysql service

An apm agent running inside the service `foo` is talking to a MySQL instance.

```
service.name: mysql
service.type: mysql
```

The `service.type` will be set to MySQL as the info is about a connection / query to MySQL. If not name of the service it connects to can be found, it will fill in service.type.

### Question

* Is the above correct?
* Is the service.name needed in thsi context?

roncohen:
[...]

PR #144, opened 2018-10-23
Rename `host.name` to `host.hostname`.

webmat:
This is to align with the industry's convention of using the word "hostname".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,10 @@ All notable changes to this project will be documented in this file based on the
 * Remove top level objects `kubernetes` and `tls`. #132
 * Remove `*.timezone.offset.sec` fields as too specific for ECS at the moment. #134
 * Make the following fields keyword: device.vendor, file.path, file.target_path, http.response.body, network.name, organization.name, url.href, url.path, url.query, user_agent.original
-* Rename `url.host.name` to `url.hostname` to better align with industry convention.
+* Rename `url.host.name` to `url.hostname` to better align with industry convention. #147
 * Make the following fields keyword: device.vendor, file.path, file.target_path, http.response.body, network.name, organization.name, url.href, url.path, url.query, user_agent.original. #137
   * Only two fields using `text` indexing at this time are `message` and `error.message`.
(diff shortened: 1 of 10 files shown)


ruflin:
Can you add a changelog? Also CI does not seem to be happy. My assumption is it's related to an autopep version issue. We had the same recently in Beats. I can open a separate PR for it.

I'm thinking if `host.name` becomes also valuable beside `host.hostname`. `host.name` would contain a name that can be manually set. We can discuss this in an other issue.

webmat:
@ruflin Yeah the fix for the autopep8 code formatting issue is here: #145.

Changelog incoming.

ruflin:
@tsg @jsvd As this is a pretty major on for LS and Beats and want to get your eyes on this one. It is a case we can handle pretty well with the alias feature as it's a 1-1 mapping.

ruflin:
Based on https://github.com/elastic/ecs/issues/143 and https://github.com/elastic/ecs/issues/142 I think we need `host.name` and `host.hostname`. Let me explain. The `host.hostname` is what we described at the moment as `host.name` so this PR is valid as it is. `host.name` is a user given name to the host. In some cases this could be just `my-test-host` or also `fqdn`. Is a request we saw frequently in Beats and is the reason `beat.name` is configurable.

I suggest we move forward with this PR and do a follow up PR to reintroduce `host.name`. In general `*.name` is something user configurable.

ppf2:
Some of our customers who have hit the breaking change in 6.3 (from hostname to host.name) have since switched to host.name.  Re-introducing host.name as an additional "host name" field can be nice so there is 1 less breaking change when they get to 7 :)  @acchen97

webmat:
@ppf2 Yeah the convention in ECS will be that `.name` will be purely a customizable name that people can override. Soon we'll have both of them in the schema.

- `host.hostname` == what `hostname` command returns on this host
- `host.name` == whatever override name the admin wants to give this machine

This is also true of other places where we have `.name`.

- `agent.type=mysql` this agent monitors a MySQL thing
- `agent.name=finance-db` this agent monitors the finance dept's MySQL thing

So in short, customers will be free to continue using `host.name` however they want.

PR #149, opened 2018-10-25
Add column 'level' to display which field is core/extended

webmat:
Also in this PR

- Get rid of the concept of `phase`, which was no longer used
- Some variable renaming in use-cases.py `write_stdout()`, to improve readability
- Fix a few use case fields that should now be `keyword`
- Fix a few use cases for `host.hostname` rename

**Caveat**:

- [ ] As of #9ee4ceb, the code to generate use cases completely breaks
  if there's a `multi_fields` anywhere. This is true whether the mf is only
  in the use case (e.g. a customization) or in cases where the field is both in
  ECS and in the use case.
  - Note that since #137, there are no longer any multi-field in ECS, so at this
  time this isn't a blocker. We'll need to fix this bug before adding any multi-fields,
  however.

Partially addresses #93


webmat:
Folks, here's the URL to browse this PR as if it was pushed to master, to look at the readme and use cases: https://github.com/webmat/ecs/blob/level-scripts

webmat:
About the caveat, I'd be inclined to open a GH issue and get this resolved when we get back to having a use case or an ECS field that's multi_fields. WDYT?

webmat:
Link to browse above is incorrect. Use this one to view the rendered readme: https://github.com/webmat/ecs/tree/level-scripts

scripts/helper.py
```diff
@@ -137,7 +134,7 @@ def get_markdown_table(namespace, title_prefix="##", link=False):
     # Replaces one newlines with two as otherwise double newlines do not show up in markdown
     output += namespace["description"].replace("\n", "\n\n") + "\n"
 
-    titles = ["Field", "Description", "Type", "Multi Field", "Example"]
+    titles = ["Field", "Level", "Description", "Type", "Multi Field", "Example"]
```
ruflin: I would put the level after the Type.

As we don't have multi fields anymore, we can remove the multi field part? Or should we keep it just in case. Not related to this PR but stumbled over it.

scripts/schemas.py
```diff
@@ -95,12 +95,14 @@ def check_fields(fields):
 
     check_fields(sortedNamespaces)
 
+    # Generates html for README
```
ruflin: It creates Markdown not HTML :-)

webmat:
@ruflin This is ready for a review. And I'd like your opinion on the caveat for the PR as it stands (see PR description)

scripts/helper.py
webmat: I agree level at column 2 is not ideal.

However putting level after type puts it between type and multi-field, which are closely related. I'll try with level just before type instead, and let's see how this looks.

scripts/schemas.py
webmat: Haha true :-)

webmat:
You're right about the examples, damn! Not sure how I missed that :-)

- [ ] Actually display example column content in readme, again
- [ ] Move "level" column between description and type

webmat:
@ruflin Ready for another review. The two issues (and incorrect comment) are fixed:

https://github.com/webmat/ecs/blob/level-scripts/README.md
https://github.com/webmat/ecs/tree/level-scripts/use-cases

However here's an experiment I made at the same time:

- Removed the "multi field" column (right now there's no MF in ECS, and when there is, they are displayed on their own line anyway)
- Move columns of predictable width before description (level and type)

I find this to be much more readable, check it out here: 

https://github.com/webmat/ecs/blob/level-tbl-display-experiment/README.md

Unless you have reservations about this, I would really like to move to this representation.

ruflin:
I agree variable column width is not optimal but as already the first column is variable I would rather focus on the importance of the information in ordering it. For me priority is:

* field
* description
* type

Then somewhere much later comes

* level
* example

So I would definitively keep the order of the first 3.

PR #150, opened 2018-10-25
Add agent.type field and redefine agent.name

ruflin:
This is also inspired by the discussion in https://github.com/elastic/ecs/issues/142


ruflin:
@roncohen @graphaelli Would be good to get your take on this one from an APM view.

ruflin:
@ycombinator This will also be relevant for the Elasticsearch modules as we use `service.name` at the moment.

ruflin:
Let's not block this PR by the above follow up discussion. I would even suggest to not touch the other type fields for now and reevaluate in a few months if we hit issues.

PR #152, opened 2018-10-25
WIP Reuseable objects

### LINKED ISSUE #94: Document how / whether objects should be reused in different places in ECS
I'll come back and flesh out this issue later. Just jotting down a bunch of things that have been floating in the air for a bit.

webmat:
This PR aims at documenting which sets of fields are expected to be reused, and
where. It's not complete yet, but it's taking shape:

- [x] Add attributes in the YML files declaring which objects are meant to be
  reused and where.
- [x] Output text to that effect in the README. You can check it out on
  [my branch here](https://github.com/webmat/ecs/tree/reuseable-objects#-geo-fields)
- [ ] Generate `template.json` according to where these objects are meant to be
  reused.
  - Nest everywhere it's expected.
  - Remove from the top level for field sets where `top_level: false`
- [ ] Maybe add a paragraph to `implementing.md` explaining the concept of object reuse?

Closes #94


webmat:
@ruflin looking at the code of `template.go`, the `Field struct` declared in libbeat, then ucfg, and provided that I'm a total Go noob, I can say with confidence I not be able to modify `template.go` in a timely manner :-)

We need to insert the reuseable objects at the expected places in `template.json`. I've added new fields in the YAML to describe the reuseability (e.g. [on geo](https://github.com/webmat/ecs/blob/reuseable-objects/schemas/geo.yml#L8-L14)). However ucfg is reading the YAML and shoving it directly in the Field struct declared in libbeat, which of course doesn't know about my new fields (and it shouldn't, either).

So I'm not sure how to approach this. Perhaps I'm missing a simple thing I can do myself in Go, to work around this. What do you think makes most sense?

* I try to copy over the Field struct locally and add my custom stuff, then work with this.
* I do some light pre-processing of the YAML files in Python to shove the reuseable objects where they're expected, in a new set of YAML files, and process those with ucfg instead.
* I re-implement the whole index template generation in Python.
  * This is my favourite option, because it would simplify maintaining ECS to only have Python scripts. But I may be missing some important functionality that libbeat/ucfg are providing.

WDYT?

ruflin:
I suggest we first focus on getting it into the docs and YAML properly and only in a follow up discuss what it means for the template as I think there are a few different solutions here.

README.md
```diff
@@ -212,16 +212,20 @@ File fields provide details about each file.
 
 ## <a name="geo"></a> Geo fields
 
-Geo fields can carry data about a specific location related to an event or geo information for an IP field.
+Geo fields can carry data about a specific location related to an event or geo information derived from an IP field.
 
 
+The `geo` fields are expected to be nested at: `destination.geo`, `device.geo`, `host.geo`, `source.geo`.
+
+Note also that the `geo` fields are **not** expected to be used directly at the top level.
```
ruflin: Why do we make this bold? I see `geo` as a reserved prefix in ECS so even if it is used on the top level I would expect everything to work.

README.md
```diff
@@ -301,8 +305,12 @@ The organization fields enrich data with information about the company or entity
 
 ## <a name="os"></a> Operating System fields
 
-The OS fields contain information about the operating system. These fields are often used inside other prefixes, such as `host.os.*` or `user_agent.os.*`.
+The OS fields contain information about the operating system.
+
 
+The `os` fields are expected to be nested at: `device.os`, `host.os`, `user_agent.os`.
+
+Note also that the `os` fields are **not** expected to be used directly at the top level.
```
ruflin: See above.

schema.csv
```diff
@@ -64,11 +64,11 @@ file.size,long,extended,
 file.target_path,keyword,extended,
 file.type,keyword,extended,
 file.uid,keyword,extended,
-geo.city_name,keyword,core,
-geo.continent_name,keyword,core,
-geo.country_iso_code,keyword,core,
-geo.location,geo_point,core,
-geo.region_name,keyword,core,
+geo.city_name,keyword,core,Montreal
```
ruflin: Thanks for adding the example. Perhaps in the future make separate PR's for these changes as then we can get them in quickly and it becomes more obvious which parts in the auto generated files are affected by this PR.

scripts/schemas.py
```diff
@@ -34,32 +34,19 @@ def create_csv(fields, file):
                 schema_writer.writerow([field["name"], field["type"], field["level"], field["example"]])
 
 
-def create_markdown(fields, file):
```
ruflin: Trying to understand how this removal is related to this PR? Just a left over?

schema.csv
webmat: I agree in principle. Although there are so many changes to do still, that I'm trying to lump some related things together, to speed up the overall process (even though the examples in this case will be delayed slightly).

README.md
webmat: The bold is to make sure the difference between the two versions of this canned sentence sticks out.

And in the meeting about reuseable objects, we agreed which object was expected where, including whether or not it was expected at the top level. We also agreed on the "expectation" part, and it not being an absolute requirement.

So the phrasing "not expected" allows for someone to do it if needed (as they can already do what they want around ECS anyway).

I can get rid of the bold, though. I did end up writing the reverse sentence in a way that doesn't look too much like this one (see [#user fields](https://github.com/webmat/ecs/tree/reuseable-objects#-user-fields)).

scripts/schemas.py
webmat: Just old dead code. As I mentioned above ^ I'm just trying to lump more changes together, since we have a lot to do.

But I agree cleanups should normally be their own PR as well.

Issue #154, opened 2018-10-26
Using ECS to track the source of errors or exceptions


webmat:
I'd like to discuss the use case of tracking the source of program errors / exceptions (line numbers, function names, etc) with ECS. These fields may end up under the field sets `file` and/or `error`.

I'd like to get feedback on how we could approach this. Information we may like to track about where the error was raised:

* error source file
* line number
* error message
* function/method name
* class / module (or other structure) inside which that function/method lives, when appropriate
* error class / standardized name (e.g. literally the class name in an OO language)
* error fields, if any
* error stack trace

Please let me know if there's anything I'm missing in terms of relevant data to support.

Here's a suggestion of a mapping for the above:

* error source file: [file.path](https://github.com/elastic/ecs#user-content-file.path)
* `error.line` (`long`)
* `error.message` (`keyword`)
* `error.function` (`keyword`)
  * the word function is a bit more broad than `method`, IMO, so can serve both use cases.
* Error class: [error.id](https://github.com/elastic/ecs#user-content-error.id)
* Error fields
  * Either create a field similar to [labels](https://github.com/elastic/ecs#user-content-labels), where people nest their kv attributes or we don't add it to ECS at first.
* stack trace: `error.stack_trace` (non indexed)

* @graphaelli or @roncohen I'd like your POV on how you're doing it currently in APM
* @ruflin I'd like your opinion on this as well

graphaelli:
A fully filled (though contrived) error example can be found [here](https://github.com/elastic/apm-server/blob/b11687ab1e10f44baa122bb05aacc63429822b2d/docs/data/elasticsearch/error.json) with corresponding spec [here](https://github.com/elastic/apm-server/blob/b11687ab1e10f44baa122bb05aacc63429822b2d/docs/spec/errors/common_error.json).  We're not thrilled about the error.exception / error.log model but haven't made concrete plans to change that.

Some notable differences:

* `error.message` is a text field in APM.  In the proposed schema, how is this different from the existing `error.code` keyword field?
* `error.code` is in APM, as a long
* `error.function` is `error.culprit` in APM
* `error.stack_trace` is well defined and in use across a number of languages
* `error.fields` appears to be `error.attributes` in APM
* `error.handled` is a thing in APM, indicating  whether the error was caught somewhere in the code

Have you considered an error level?  `log.level` might be sufficient.

webmat:
Excellent, thanks @graphaelli for the detailed answer and the example log :-)

[...]

PR #157, opened 2018-10-29
Tweak note about 'IDs' being keyword to also mention 'codes'

webmat:


README.md
```diff
@@ -503,11 +503,12 @@ of a breaking change with these two widely used fields in Beats.
 Any future field that will be indexed for full text search in ECS will however
 follow the multi-field convention where `text` indexing is nested in the multi-field.
 
-### IDs are keywords not integers
+### IDs and codes are keywords, not integers
```
ruflin: Are you thinking of some specific fields here?

webmat: Yes, error codes. Came up during a discussion with @graphaelli over on #154

ruflin:
The reason I'm hesitant to merge this is because there are some instances like status_code where things are long and as I see this as a good thing. Perhaps it can be rephrased to "in most cases".

webmat:
Gotcha. I'll tweak accordingly.

webmat:
Reworded. Although I find this a bit wordy.

webmat:
I agree README is getting verbose. But my plan is definitely to split off some of these discussions into secondary documentation files, to streamline the readme again.

PR #162, opened 2018-10-31
Use word reusable rather than reuseable.

webmat:
The shorter word appears to be used 40x more than the longer one
(according to Google) :-)


MikePaquette:
+1

PR #163, opened 2018-11-01
Use domain and registered_domain for the domain breakdown.

### LINKED ISSUE #84: Clarify use of hostname, subdomain, domain in source/destination

webmat:
This PR removes fields `subdomain` and `hostname` from `destination` and `source`.

View the readme [on my branch](https://github.com/webmat/ecs/tree/domain_breakdown)

Closes #84

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,12 +25,7 @@ All notable changes to this project will be documented in this file based on the
 * Update definition of `service.type` and `service.name`.
 * Redefine purpose of `agent.name` field to be user defined field.
 * Rename `url.href` to `url.original`.
-* Remove `source.subdomain` and `destination.subdomain` fields.
-* Remove `source.hostname`, `destination.hostname`, `source.subdomain`,
-  `destination.subdomain` and `url.hostname`. #163
-* Remove `source.hostname`, `destination.hostname`, `source.subdomain` and
-  `destination.subdomain`. #163
-* Remove `source.subdomain` and `destination.subdomain`. #163
+* Remove `source.subdomain` and `destination.subdomain` fields. #165
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ All notable changes to this project will be documented in this file based on the
 * Redefine purpose of `agent.name` field to be user defined field.
 * Rename `url.href` to `url.original`.
 * Remove `source.subdomain` and `destination.subdomain` fields.
+* Remove `source.hostname`, `destination.hostname`, `source.subdomain`,
+  `destination.subdomain` and `url.hostname`.
 
 ### Bugfixes
 
@@ -41,5 +43,7 @@ All notable changes to this project will be documented in this file based on the
 * Add `host.os.kernel` containing the OS kernel version. #60
 * Add `agent.type` field.
 * Add `http.request.referrer` field. #164
+* Add `source.registered_domain`, `destination.registered_domain`,
+  `url.domain` and `url.registered_domain`.
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ All notable changes to this project will be documented in this file based on the
 * Remove `source.subdomain` and `destination.subdomain` fields.
 * Remove `source.hostname`, `destination.hostname`, `source.subdomain`,
   `destination.subdomain` and `url.hostname`. #163
+* Remove `source.hostname`, `destination.hostname`, `source.subdomain` and
+  `destination.subdomain`. #163
 
 ### Bugfixes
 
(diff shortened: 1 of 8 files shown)


schemas/url.yml
```diff
@@ -24,15 +24,25 @@
         Note: The `:` is not part of the scheme.
       example: https
 
-    - name: hostname
-      level: extended
+    - name: domain
```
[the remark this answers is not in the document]
webmat: Since a URL's host can either be a domain name or an IP, I agree that `domain` may not work here. Is that what you're referring to?

The correct wording to represent this would actually be `host`, not `hostname`.

So if we go to `url.host`, I would be fine to word it like:

> If 'host' is a domain name, it should be copied to source.domain or destination.domain, whichever is appropriate.

So the domain breakdown could happen there, and domains would be expected in two places instead of 3 (a plus, in my mind).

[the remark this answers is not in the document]
ruflin: host would contain the port, but it doesn't. hostname is correct here. 

Just leave it `hostname` and we are good.

webmat:
I undid the changes under `url.` for now. This PR is now only about `source` and `destination`.

I still think if we're parsing URLs `hostname` is not the correct terminology (it should be `host`, which can be an IP or a domain name), so there's still work to be done in `url`.

fields.yml
```diff
@@ -273,14 +267,22 @@
...
     
-        - name: subdomain
-          level: core
+        - name: registered_domain
+          level: extended
           type: keyword
+          example: example.com
           description: >
-            Destination subdomain.
+            Destination domain, with the subdomain stripped out.
```
andrewkroh: Perhaps we could also include that this is the “effective top level domain plus one additional level” as secondary description to remove any ambiguity.

webmat:
Ok I will remove the mentions of `hostname` in this PR, and only keep the `domain` and `registered_domain` fields. Let's move the discussion to #166 for the `hostname` field.

Nic's point that in some of the web server logs, it's possible for the server to perform a reverse lookup on the source IP of each request, and log a hostname instead of an IP. So I'm now convinced that we do need a field for this value that can either be a hostname or an IP. I still think that `hostname` is not the right name here, after reading RFCs 2396, 3986 and 7230. So let's hash this out in #166 ;-)

fields.yml
webmat: Done

CHANGELOG.md
```diff
@@ -25,7 +25,7 @@ All notable changes to this project will be documented in this file based on the
 * Update definition of `service.type` and `service.name`.
 * Redefine purpose of `agent.name` field to be user defined field.
 * Rename `url.href` to `url.original`.
-* Remove `source.subdomain` and `destination.subdomain` fields.
+* Remove `source.subdomain` and `destination.subdomain` fields. #165
```
webmat: This is from another PR, I'm just adding the PR link here.

ruflin:
@webmat I wonder if we go with `source.name` in https://github.com/elastic/ecs/issues/166 if we can use this also for what we use `source.domain` here and then `source.registred_domain` becomes `source.domain`.

webmat:
@ruflin I'd be worried that `.name` is too generic to represent the domain name of a source / destination.

As I've said elsewhere, I would expect the registered_domain (with subdomain stripped out) value to not be populated as often by integrations, since it requires additional processing based on a TLD list & so on.

This would leave people seeing their domain names in most cases only in `source.name` and `destination.name`, which will feel strange, IMO. 

I'm totally open to changing the name `registered_name` to something else, if you feel the name is clumsy. After the RFC review leading to #166, I no longer feel like it's a perfect name. "Registered name", afaik, could mean a short machine name that's in an internal DNS server. It doesn't actually mean the highest level of domain you can purchase (TLD+1 level).

Actually, based on the above paragraph, I'd be tempted to keep this PR for after beta1. This would mean the usage of the existing field `domain` is still ambiguous for the release. And we can take the time to find the perfect pair of names for this breakdown after beta1. I could go with that.

webmat:
Closing this. This is based on a version of ECS that's too outdated.

We will start fresh on this later, in #267.

PR #165, opened 2018-11-02
Remove subdomain from fields

ruflin:
We decided that for now the usage of subdomain is not fully clear yet so we removed it from ECS. The field can still be used but it is not in the ECS definition.


ruflin:
@webmat I opened a separate PR for this as I do not want to mix it with the other discussion.

PR #168, opened 2018-11-05
Remove duplicate definitions of `os` from `host.os` and `user_agent.os`...

webmat:
With `os` now a reuseable field set, it's now expected in these locations. The field set should only be defined once.

Having it (and other reuseable field sets) appear correctly in `template.json` is still work to be done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,6 @@ All notable changes to this project will be documented in this file based on the
 * Redefine purpose of `agent.name` field to be user defined field.
 * Rename `url.href` to `url.original`.
 * Remove `source.subdomain` and `destination.subdomain` fields.
-* Remove duplicate definitions of the `os` field set from `host.os` and
-  `user_agent.os`. #168
 
 ### Bugfixes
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ All notable changes to this project will be documented in this file based on the
 * Redefine purpose of `agent.name` field to be user defined field.
 * Rename `url.href` to `url.original`.
 * Remove `source.subdomain` and `destination.subdomain` fields.
+* Remove duplicate definitions of the `os` field set from `host.os` and
```
ruflin: This is not really a breaking change.

ruflin:
As discussed on Slack, I would suggest to list it under `Improvements` in the changelog.

CHANGELOG.md
webmat: Done

webmat:
@ruflin You were asking if the stuff removed was actually in line with the `os` field set. Well both of these dupe instances were actually missing one field already :-)

PR #169, opened 2018-11-05
Rename `event.version` to `ecs.version`.

webmat:
- This opens up the possibility for data sources to use event.version for the version of their event source.
- This opens up the possibility of tracking more ECS-specific meta-data in the ECS field set in the future

cc @MikePaquette
cc @ruflin (as discussed)


webmat:
Sorry, when reviewing the changes you'll want to tweak your diff settings to "ignore whitespace". My editor automatically strips out trailing whitespace :-)

MikePaquette:
LGTM.

Issue #167, opened 2018-11-05
Adjust the readme intro for the Beta1 release


webmat:
We'll need to adjust the readme for this first release:
 
- [ ] Tone down the chances for breaking changes. The point of Beta1 is to start guaranteeing more backwards compatibility. But since it's Beta1, we shouldn't guarantee 100% necessarily yet.
- [ ] We should link to the specific git tag right in the intro, to make it easy to browse the repo as of the Beta1 release.
  - [ ] We should also ensure all internal links in the repo (e.g. towards use cases) are relative, and don't throw people back to master :-)
- [ ] Warning that in cases of discrepancies between the readme and schema.csv or template.json, the readme should prevail.
  - This is a temporary situation. Ideally all would be in line, but they won't be yet, at the time of Beta1

webmat:
I haven't seen mentions of breaking changes that needed to be toned down (referring to my first subtask above).

Other than this point, #174 is a first attempt at what this release could look like, from a readme/changelog point of view.

#173 will also be important for people to be able to navigate the markdown renderings directly on Github. Otherwise clicking some links brings them out of a git tag and back to the master repository, which is constantly in flux.

PR #173, opened 2018-11-06
Make Markdown links relative, so browsing a git ref stays in the same git ref

webmat:
This is important for Beta1, as people visiting elastic/ecs will soon be directed
to past versions (e.g. `v1.0.0-beta1`), if they're looking for official ECS
releases.

This fix ensures that navigating the links from the Readme to the use cases and
the links from use cases back to the readme keeps the person in the same branch
or tag. Prior to this, they would get sent back to the latest version of ECS.

In this PR:

- Make links to use cases relative
- Make links to field defs (from use cases) relative
- Fix last two instances of documentation links pointing to a specific version.
  They now link to the "current" stack version.

This addresses one element of #167.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,15 +403,15 @@ describe the full use case. The fields which are not in ECS are in italic.
 
 Contributions of additional uses cases on top of ECS are welcome.
 
- * [APM](https://github.com/elastic/ecs/blob/master/use-cases/apm.md)
- * [Auditbeat](https://github.com/elastic/ecs/blob/master/use-cases/auditbeat.md)
- * [Beats](https://github.com/elastic/ecs/blob/master/use-cases/beats.md)
- * [Filebeat Apache](https://github.com/elastic/ecs/blob/master/use-cases/filebeat-apache-access.md)
- * [Kubernetes](https://github.com/elastic/ecs/blob/master/use-cases/kubernetes.md)
- * [Logging](https://github.com/elastic/ecs/blob/master/use-cases/logging.md)
- * [Metricbeat](https://github.com/elastic/ecs/blob/master/use-cases/metricbeat.md)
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -437,14 +437,14 @@ describe the full use case. The fields which are not in ECS are in italic.
...
 
- * [APM](https://github.com/elastic/ecs/blob/master/use-cases/apm.md)
- * [Auditbeat](https://github.com/elastic/ecs/blob/master/use-cases/auditbeat.md)
- * [Beats](https://github.com/elastic/ecs/blob/master/use-cases/beats.md)
- * [Filebeat Apache](https://github.com/elastic/ecs/blob/master/use-cases/filebeat-apache-access.md)
- * [Kubernetes](https://github.com/elastic/ecs/blob/master/use-cases/kubernetes.md)
- * [Logging](https://github.com/elastic/ecs/blob/master/use-cases/logging.md)
- * [Metricbeat](https://github.com/elastic/ecs/blob/master/use-cases/metricbeat.md)
- * [TLS](https://github.com/elastic/ecs/blob/master/use-cases/tls.md)
+ * [APM](use-cases/apm.md)
```
ruflin: I'm wondering if these links still work if someone looks at the README directly from the landing page? Because then the link changes to `blob/master` so the relative link is not correct.

We should make sure these links keep working as I assume that are the ones most used.

webmat: I tested all combinations, and it works in all cases:

- From a branch, clicking in root page rendering of readme towards use cases
- From a branch, clicking in /readme.md towards use cases
- From a branch, clicking from use cases back to field defs in the readme
- From root page of the repo (temporarily merged to master in my fork) to the readme and back

All of these cases

PR #175, opened 2018-11-07
Remove src/dst.hostname, rename url.hostname to url.domain.

### LINKED ISSUE #166: Finding a better name for `url.hostname`, `source.hostname` and `destination.hostname`
As most of you have realized, using the word "hostname" for two different purposes in ECS is bugging me. I'm very happy of the resistance I've faced when bringing this up so far, because it has forced me to dig deeper and figure this out more precisely. Hopefully this helps me make a better case for why in one of these two cases, "hostname" is incorrect.

## Two usages of hostname in ECS

Let's review the two different uses of "hostname" currently in ECS:

* "hostname" as a server or device name, as given by administrators. Or the value you get when you run the command `hostname` on a host/device.
  * This use I agree 100% with.
  * Currently used this way in `device.hostname` and `host.hostname`.
  * To complicate the following discussion, some network devices default to returning their main IP address, when they don't have a textual hostname configured.
* "hostname" as a network address, which may be an IP or a domain name.
  * This is the use I object with, and what I want to address here. From most sources I've seen, this usage of "hostname" is incorrect, it should be "host".
  * Currently used this way in `source.hostname`, `destination.hostname` and `url.hostname`
  * If some of us thought this was meant to be a place to populate with the hostname of a host under management (e.g. via enrichment):
    1. This confirms that this is an ambiguous name, because that's not the actual purpose. The purpose is to store an address prior to having determined whether it's an IP or a name.
      * This concept comes from the common web log format. See [Apache's "%h" log format string](http://httpd.apache.org/docs/current/mod/mod_log_config.html#formats) and its [HostnameLookups directive](http://httpd.apache.org/docs/current/mod/core.html#hostnamelookups), or Traefik's [ClientHost and RequestHost](https://docs.traefik.io/configuration/logs/#list-of-all-available-fields)
    2. Enrichment with known host details -- if someone wanted to do that -- would be done more accurately by nesting a subset of the `host` fields in there instead anyway (e.g. `host.id`, `host.hostname`).

## What's wrong with `hostname`?

In the RFCs I'm aware of, when mentioning "a network address, which may be an IP or a domain name", they all refer to that as a "host", not a "hostname". They sometimes use the word "hostname" (but not always) to refer to an address that's a registered name (such as a domain name or a local DNS entry).

This is not a recent change either, the oldest RFC I'm linking to below is from 1994, and is using "host" to mean the "ip or name" concept. Here's a few excerpts:

* [1738 - Uniform Resource Locators (URL) - 3.1. Common Internet Scheme Syntax](https://tools.ietf.org/html/rfc1738#section-3.1)
  * Quoting
    * "**host**: The fully qualified domain name of a network host, or its IP address as a set of four decimal digit groups separated by '.'."
    * "http://\<**host**>:\<port>/\<path>?\<searchpart>"
* [2396 - Uniform Resource Identifiers (URI): Generic Syntax - 3.2.2. Server-based Naming Authority](https://tools.ietf.org/html/rfc2396#section-3.2.2)
  * Quoting
    * "**host** = hostname | IPv4address"
    * "\<userinfo>@\<**host**>:\<port>"
    * "server = [ [ userinfo "@" ] **hostport** ]"
* [3986 - URI Generic Syntax - 3.  Syntax Components](https://tools.ietf.org/html/rfc3986#section-3)
  * (Check out the section intro for a nice visual of the URI parts, then scroll down to [3.2.2. Host](https://tools.ietf.org/html/rfc2396#section-3.2.2))
  * Quoting: "**host** = IP-literal / IPv4address / reg-name"
* [7230 - HTTP/1.1 Message Syntax and Routing - 5.4. Host](https://tools.ietf.org/html/rfc7230#section-5.4)
  * Quoting: "**Host** = uri-host [ ":" port ]"

The fact that they are sometimes mentioned in the same places or even in some cases used interchangeably in tool documentation (we're not the only ones facing this mixup) helps explain why the two are often mixed up. I haven't looked farther back in the RFCs, so that mixup may indeed come from there.

However as @MikePaquette pointed out, using the name "host" in these places would conflict with our top level field set "host". Even if "host" is not currently a reuseable object, I do think it could become one, and the most obvious places I would expect to nest it is at `source.host` and `destination.host`.

So I agree we should not rename these 3 fields to `*.host`.

Here are some suggestions for new field names, based on what I've seen in the various server documentation tools and the RFCs. Remember that these proposed renames apply only to `source`, `destination` and `url`, **not** to `host` and `device`:

* `host_address`
* `uri_host` or `url_host`
* Drop `hostname` from source, destination and url.
  * In most cases people will want to store the value in their proper field (ip or domain), after determining which type it is. Them wanting to keep the ambiguous field anyway could be considered a use case, and they're free to name this field however they want, if it's not in ECS.
  * Independent from subpoint above, we could also work on finding the right name for this field after Beta1, and just drop the field out of ECS temporarily for the Beta1 release.

I'm open to other suggestions, of course.

webmat:
`domain` is now the place to store the host address under `source`, `destination` and `url`.

Also in this PR:

- Remove ambiguous mention in `url` description talking about reuse. Since we
  haven't officially made `url` reuseable yet, I took it out for now. People
  are free to reuse everything everywhere, as it's just them defining new fields
  around the official ECS fields. But the spec making this statement about it
  in a different manner than the official reuseable objects is needlessly confusing.
- Tweak the url example a bit.
- Bring `source` description in line with `destination`'s

Closes #166, and addresses part of #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ URL fields provide a complete URL, with scheme, host, and path. The URL object c
 |---|---|---|---|---|
 | <a name="url.original"></a>url.original | Full original url. The field is stored as keyword. | extended | keyword | `https://elastic.co:443/search?q=elasticsearch#top` |
 | <a name="url.scheme"></a>url.scheme | Scheme of the request, such as "https".<br/>Note: The `:` is not part of the scheme. | extended | keyword | `https` |
-| <a name="url.domain"></a>url.domain | Hostname of the request, such as "elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `hostname` field. | extended | keyword | `elastic.co` |
+| <a name="url.domain"></a>url.domain | Domain of the request, such as "elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `domain` field. | extended | keyword | `elastic.co` |
 | <a name="url.port"></a>url.port | Port of the request, such as 443. | extended | integer | `443` |
 | <a name="url.path"></a>url.path | Path of the request, such as "/search". | extended | keyword |  |
 | <a name="url.query"></a>url.query | The query field describes the query string of the request, such as "q=elasticsearch".<br/>The `?` is excluded from the query string. If a URL contains no `?`, there is no query field. If there is a `?` but no query, the query field exists with an empty string. The `exists` query can be used to differentiate between the two cases. | extended | keyword |  |
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ URL fields provide a complete URL, with scheme, host, and path. The URL object c
 |---|---|---|---|---|
 | <a name="url.original"></a>url.original | Full original url. The field is stored as keyword. | extended | keyword | `https://elastic.co:443/search?q=elasticsearch#top` |
 | <a name="url.scheme"></a>url.scheme | Scheme of the request, such as "https".<br/>Note: The `:` is not part of the scheme. | extended | keyword | `https` |
-| <a name="url.hostname"></a>url.hostname | Hostname of the request, such as "elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `hostname` field. | extended | keyword | `elastic.co` |
+| <a name="url.domain"></a>url.domain | Hostname of the request, such as "elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `hostname` field. | extended | keyword | `elastic.co` |
 | <a name="url.port"></a>url.port | Port of the request, such as 443. | extended | integer | `443` |
 | <a name="url.path"></a>url.path | Path of the request, such as "/search". | extended | keyword |  |
 | <a name="url.query"></a>url.query | The query field describes the query string of the request, such as "q=elasticsearch".<br/>The `?` is excluded from the query string. If a URL contains no `?`, there is no query field. If there is a `?` but no query, the query field exists with an empty string. The `exists` query can be used to differentiate between the two cases. | extended | keyword |  |
--- a/README.md
+++ b/README.md
@@ -129,7 +129,6 @@ Destination fields describe details about the destination of a packet/event.
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
 | <a name="destination.ip"></a>destination.ip | IP address of the destination.<br/>Can be one or multiple IPv4 or IPv6 addresses. | core | ip |  |
-| <a name="destination.hostname"></a>destination.hostname | Hostname of the destination. | core | keyword |  |
 | <a name="destination.port"></a>destination.port | Port of the destination. | core | long |  |
 | <a name="destination.mac"></a>destination.mac | MAC address of the destination. | core | keyword |  |
 | <a name="destination.domain"></a>destination.domain | Destination domain. | core | keyword |  |
@@ -353,7 +352,6 @@ Source fields describe details about the destination of a packet/event.
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
 | <a name="source.ip"></a>source.ip | IP address of the source.<br/>Can be one or multiple IPv4 or IPv6 addresses. | core | ip |  |
-| <a name="source.hostname"></a>source.hostname | Hostname of the source. | core | keyword |  |
 | <a name="source.port"></a>source.port | Port of the source. | core | long |  |
 | <a name="source.mac"></a>source.mac | MAC address of the source. | core | keyword |  |
 | <a name="source.domain"></a>source.domain | Source domain. | core | keyword |  |
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ The service fields describe the service for or from which the data was collected
 
 ## <a name="source"></a> Source fields
 
-Source fields describe details about the source of the event.
+Source fields describe details about the destination of a packet/event.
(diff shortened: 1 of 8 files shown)


webmat:
If you folks want me to separate the hostname work from the various cleanups included here, I can do that as well. The cleanups seemed straightforward, though.

README.md
```diff
@@ -347,28 +346,27 @@ The service fields describe the service for or from which the data was collected
...
 
 
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
-| <a name="url.original"></a>url.original | Full original url. The field is stored as keyword. | extended | keyword | `https://elastic.co:443/search?q=elasticsearch#top` |
+| <a name="url.original"></a>url.original | Full original url. The field is stored as keyword. | extended | keyword | `https://www.elastic.co:443/search?q=elasticsearch#top` |
 | <a name="url.scheme"></a>url.scheme | Scheme of the request, such as "https".<br/>Note: The `:` is not part of the scheme. | extended | keyword | `https` |
-| <a name="url.hostname"></a>url.hostname | Hostname of the request, such as "elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `hostname` field. | extended | keyword | `elastic.co` |
+| <a name="url.domain"></a>url.domain | Domain of the request, such as "www.elastic.co".<br/>In some cases a URL may refer to an IP and/or port directly, without a domain name. In this case, the IP address would go to the `domain` field. | extended | keyword | `www.elastic.co` |
```
jsoriano: Is this still the case? should IP addresses be stored in the `domain` field?
I think that having sources that don't differentiate names from IPs is quite frequent when parsing logs, having IPs stored in a field called `domain` can be a bit confusing.

MikePaquette: @jsoriano the `url.domain` field should generally *not* be populated with an IP address, however, when an event or log indicates that an IP address was used *in* a URL as the host/hostname, such as in `http://15.73.192.108/index.html`, then the string `15.73.192.108` would be entered into the `url.domain` field.  This is useful, for example, for a security analyst to know that a user or system is attempting to bypass the DNS infrastructure to access a web resource.  In some cases, this could be an indicator of malware activity.  So searching through all values of `url.domain` and looking for values that have the form of an IP address can be a useful analysis.

webmat: @jsoriano We were in a tough spot and couldn't actually call it `host`, as this concept of "name or IP" is generally called. In ECS, `host` is a field set that can be reused in some places.

Just to be clear, the concept of "hostname" is often misused in exactly the same way. The definition of a "host" is "IP address or hostname" (check out the RFC survey I summarized in #166).

An IP address doesn't belong in a "hostname" either. Using "hostname" this way is a mistake that's more common, though, so it doesn't stick out quite as much when people see it.

jsoriano: Thanks for the explanations, I see the problematics and the reasoning, but using `domain` to store IPs still sounds a bit weird to me. In any case I don't have a better proposal if we cannot use `host`, but I'm afraid this can be a bit confusing also for users.
Maybe after moving some beat modules to ECS and playing a bit more with it I see it in a different way :)

Will it be the same for source and destination? Maybe we should add some explanations about this also there.

webmat: @jsoriano Under source and destination, we actually have both `domain` and `ip`. Having started one transition for the Traefik module here https://github.com/elastic/beats/pull/9005 outlines a difference in approach.

Most of the modules dealt with this indirectly and named a `keyword` field something like `remote_ip` and attempted to do GeoIP on it. In other words, the duality of this value being either an IP or a hostname was not addressed. Hopefully IN's GeoIP just fails gracefully when it encounters a hostname (I haven't checked). So I think in most of the modules right now, the lie is the other way around  ;-) The `remote_ip` field can contain either an IP or a hostname, the field is `keyword` to support this, which ultimately means you can't do CIDR searches on it & so on.

The approach I took in the PR linked above is to add a tiny bit of logic to the module in this conversion. The ambiguous value is stored in a custom field, then via Grok I try to grab an IP out of the field, and store that in `source.ip`. If that grok fails, the fallback is to grab whatever's there and store it in `source.domain`.  This happens [here](https://github.com/elastic/beats/pull/9005/files#diff-5f2f0d705013b636262f7085a1bd369cR60)

jsoriano: I see, but having hosts in two fields can be a source of problems:
* Queries on aggregations of both fields will be more complicated, and they can be a common use case (for example to do a visualization that shows top N hosts, including ips and domains).
* Every module is obligated to implement a logic like this one and use more complicated queries to comply with ECS, even if the author doesn't care so much if a host is an IP or a domain. Or they can use a custom field, but I think this is going to be a common use case, so it'd be nice to be a field covered by ECS.

We could have a field for "ambiguous" hosts in addition to `domain` and `ip`, so all cases can be covered. Modules that try to parse the host as `domain` or `ip` could still keep the original in the ambiguous field, modules that don't care can use the less-featured ambiguous keyword field.
Even if ambiguous and not fully compliant with standards this field can still be common and useful.

webmat: Yes, I agree that having a field for the ambiguous information would be useful.

It used to be `hostname`, which was an incorrect name, as demonstrated in #166 (see the RFC survey in the issue body).

And we're back to square 1: the "correct" name according to the RFCs is "host" which is defined as either an IP or a hostname. And we don't want to use "host" because it will conflict with the current `host` field set, if people want to enrich their stream by nesting known host information there.

My best name for the ambiguous field for now is `host_address` at this time. But we haven't gotten together yet to plan the next steps towards ECS Beta 2.

Note also that the simplest way to avoid this problem now is to store the ambiguous "host" in a custom field. To get back to my Traefik module example, I'm leaving the incorrectly named `traefik.access.remote_ip` field there, with the exact same meaning as before.

PR #183, opened 2018-11-20
Update reference to user_agent to agent

Mpdreamz:
Since I don't see `user_agent` mentioned elsewhere I presumed this to be `agent`.


webmat:
Haha we missed removing this mention of `user_agent` :-) As Nic said, it's been removed temporarily.

I'll close this PR as it will most likely come back as `user_agent`. We've just taken it out for this first official release, with the goal of reviewing the field names in the `user_agent` field set.

PR #185, opened 2018-11-21
Add event.start and event.end

andrewkroh:
`event.start` and `event.end` are date fields that demarcate the beginning and end, respectively,
of an activity. For example in a network flow the event.start is time time of the first packet in the
flow and `event.end` time the time of the last observed packet in the flow.

The `event.duration` value is then computed as the difference between end and start times.


fields.yml
```diff
@@ -491,6 +494,20 @@
     
             In case the two timestamps are identical, @timestamp should be used.
     
+        - name: start
+          level: core
```
ruflin: I would suggest to put both these values in extended for now

andrewkroh:
Updated with:
- Fields are now `extended`.
- CHANGELOG entry added.

PR #187, opened 2018-11-23
Add host.name field

ruflin:


ruflin:
@roncohen FYI

schemas/host.yml
```diff
@@ -17,6 +17,14 @@
       description: >
         Hostname of the host.
 
+        It normally contains what the hostname command returns on the host machine.
```
webmat: Nit: quote `hostname` like in the `.name` description

ruflin: applied

webmat:
Good to go 👍

PR #193, opened 2018-11-28
Clarify device definition

MikePaquette:
Improve the definition of Device to clarify it is the observer during collection of events and metrics.


webmat:
We discussed that `device` was going to become `observer`, right?

I would be inclined to close this PR, and introduce this definition directly as the `observer` field set.

MikePaquette:
OK, Closing this PR - will generate new PR for the observer change.

PR #194, opened 2018-11-28
Clarify event definition - add metrics

MikePaquette:
Improve the definition of the event fields at the start of the event section.  Also mention that metrics details are to be included in the event fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ These fields can represent errors of any kind. Use them for errors that happen w
 
 ## <a name="event"></a> Event fields
 
-The event fields are used for context information about the event or metric itself. An event is defined as a log message containing details of something that happened. Events must include the time at which the thing happened. Examples of events include a process starting on a host, a network packet being sent from a source to a destination, or a network connection between a client and a server being initiated or closed. A metric is defined as a log message containing one or more numerical or categorical measurements and the time at which the measurement was taken. Examples include memory pressure measured on a host, or vulnerabilities measured on a scanned host.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -186,7 +186,7 @@ These fields can represent errors of any kind. Use them for errors that happen w
 
 ## <a name="event"></a> Event fields
 
-The event fields are used for context information about the data itself.
+The event fields are used for context information about the event or metric itself. An event is defined as a log message containing details of something that happened. Events must include the time at which the thing happened. Examples of events include a process starting on a host, a network packet being sent from a source to a destination, or a network connection between a client and a server being initiated or closed. A metric is defined as a log message containing one or more numerical or categorical measurements and the time at which the measurement was taken. Examples include memory pressure measured on a host, or vulnerabilities measured on a scanned host.
```
ruflin: `log message or metric`?

From a metricbeat perspective it's not really a log message. I think we mean the same thing but to me it sounds like a metric would be collected from a log message and it's not the case for metricbeat.

MikePaquette: Good point. ECS also needs to define `log message`.
  
This definition proposes a very simple model:

log message - anything sent to centralized collection (Elastic Stack) with timestamp and more
- event - log message telling about something that happened
- metric - log message telling about something that was measured

I'd like to keep this this very simple and inclusive.
Certainly open to other ways to define these concepts.

ruflin: What you define as `log message` here is for me an ECS event :-) a metric is also an event. So in the above:

* log - event telling about something that happened
* metric - event telling about something that was measured

MikePaquette: Yes, I had this wrong.  I've re-defined as you've suggested, removing "message" (now just using "log") to not confuse things with the `message` field.

PR #195, opened 2018-11-28
Clarify host definition

MikePaquette:
Updated the definition of host at the start of the host section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,9 +264,7 @@ The group fields are meant to represent groups that are relevant to the event.
 
 ## <a name="host"></a> Host fields
 
-Host fields provide information related to a host. A host can be a physical machine, a virtual machine, or a Docker container.
-
-Normally the host information is related to the machine on which the event was generated/collected, but they can be used differently if needed.
+A host is defined as a general computing instance that is not a device. ECS host.* fields should be populated with details about the host on which the event happened, or on which the measurement was taken. Hosts types include hardware, virtual machines, docker containers, and kubernetes nodes.
 
 
 | Field  | Description  | Level  | Type  | Example  |
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ The group fields are meant to represent groups that are relevant to the event.
 
 ## <a name="host"></a> Host fields
 
-A host is defined as a general computing instance that is not a device. ECS host.* fields should be populated with details about the host on which the event happened, or on which the measurement was taken. Hosts types include hardware, virtual machines, docker containers, and kubernetes nodes.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -249,9 +249,7 @@ Note also that the `geo` fields are not expected to be used directly at the top
 
 ## <a name="host"></a> Host fields
 
-Host fields provide information related to a host. A host can be a physical machine, a virtual machine, or a Docker container.
-
-Normally the host information is related to the machine on which the event was generated/collected, but they can be used differently if needed.
+A host is defined as a general computing instance that is not a device. ECS host.* fields should be populated with details about the host on which the event happened, or on which the measurement was taken. Hosts types include hardware, virtual machines, docker containers, and kubernetes nodes.
```
ruflin: Could we for now exclude the part `is not a device` until we figured the device one out? Like this we could merge the PR and clarify this later.

MikePaquette: @ruflin thanks, I would prefer to keep the `is not a device` in unless/until we find a case where it is not true, but I agree we should re-address after we resolve the `device` definition.

fields.yml
```diff
@@ -647,11 +647,7 @@
       title: Host
       group: 2
       description: >
-        Host fields provide information related to a host. A host can be a
-        physical machine, a virtual machine, or a Docker container.
-    
-        Normally the host information is related to the machine on which the event
-        was generated/collected, but they can be used differently if needed.
+        A host is defined as a general computing instance that is not a device. ECS host.* fields should be populated with details about the host on which the event happened, or on which the measurement was taken. Hosts types include hardware, virtual machines, docker containers, and kubernetes nodes.
```
tsg:
```suggestion
        A host is defined as a general computing instance that is not a device. ECS host.* fields should be populated with details about the host on which the event happened, or on which the measurement was taken. Hosts types include hardware, virtual machines, Docker containers, and Kubernetes nodes.
```

Capitalized Docker and Kubernetes.

README.md
webmat: Typo: "Host types", no "s" on "Host"

webmat: I actually agree that the part "that is not a device" seems superfluous. The device/observer description already states that it shouldn't be considered a host.

Once device is renamed observer, there's not going to be any ambiguity, I think. People are not going to be tempted to go to `observer` to record some type of host information.

MikePaquette: Typos fixed, thanks.

MikePaquette: Ok, removing `is not a device`

MikePaquette:
I unintentionally corrupted this PR with additional device changes.  Going to close it and start this one over.

PR #197, opened 2018-11-28
Clarify network definition

MikePaquette:
Added detail about network fields, specifically that these fields can be populated for both network and host-based events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Fields which are specific to log events.
 
 ## <a name="network"></a> Network fields
 
-The network is defined as the communication path over which a host or network event happens. Network events do not include the transmission of an event from a host or device to the Elastic Stack. The network.* fields should be populated with details about the network activity associated with an event.
+The network is defined as the communication path over which a host or network event happens. The network.* fields should be populated with details about the network activity associated with an event.
 
 
 | Field  | Description  | Level  | Type  | Example  |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -278,7 +278,7 @@ Fields which are specific to log events.
 
 ## <a name="network"></a> Network fields
 
-Fields related to network data.
+The network is defined as the communication path over which a host or network event happens. Network events do not include the transmission of an event from a host or device to the Elastic Stack. The network.* fields should be populated with details about the network activity associated with an event.
```
ruflin: Do we need to specify the `to Elastic Stack` part? It could be that someone monitors exactly that with a tool and then these events would be part of it.

MikePaquette: Thanks @ruflin good question.  

This sentence was added to clarify some early scoping questions I received about whether details about the networking gear that was simply transporting the event to the log collection entity were to be included in ECS. They are not expected to be included.

Your point about networking gear in general (as ECS devices) is accurate, and this sentence is not attempting to comment about them.  I'll try to clarify, give an example, or just remove.

MikePaquette: Decided to remove the offending sentence :-)

webmat: Perhaps another way of phrasing it could simply have been that this is not about the pipeline, but about the activity being observed or monitored.

If you're fine without that clarification, we can go with that as well.

PR #198, opened 2018-11-29
Use type for queryable object types for labels and annotations

jsoriano:
There is a new [ES datatype](https://github.com/elastic/elasticsearch/issues/33003) in development that will allow to store json objects with any field name, what will allow to have fields inside these objects with dots. This helps to avoid mapping conflicts as the ones we have seen with docker and kubernetes labels in beats.

I propose that ECS recommends the use of this type for this sort of values.

Not to be merged before the type is available in ES.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## Unreleased
 
 ### Breaking changes
+* Change type of labels to `json`.
 
 ### Bugfixes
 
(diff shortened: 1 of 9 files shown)


ruflin:
The challenge here is that aggregations on it will not be possible, at lease in the short term. I wonder if we should still recommend it?

jsoriano:
Yeah, if it lacks terms aggregation then we shouldn't recommend it, I wonder if any other aggregation could be needed for these fields.
I am closing this by now.

PR #201, opened 2018-11-30
Add user.full_name

webmat:
To be populated when a full name is available.

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -1352,10 +1352,18 @@
         - name: name
           level: core
           type: keyword
+          example: albert
           description: >
-            Name of the user.
+            Programmatic name / login of the user.
     
             The field is a keyword, and will not be tokenized.
+    
+        - name: full_name
+          level: extended
+          type: keyword
+          example: Albert Einstein
+          description: >
+            User's full name, if available.
     
         - name: email
           level: extended
--- a/fields.yml
+++ b/fields.yml
@@ -1356,8 +1356,6 @@
           description: >
             Programmatic name / login of the user.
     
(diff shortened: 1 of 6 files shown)


fields.yml
```diff
@@ -1223,10 +1223,16 @@
         - name: name
           level: core
           type: keyword
+          example: albert
           description: >
-            Name of the user.
+            Programmatic name / login of the user.
     
-            The field is a keyword, and will not be tokenized.
+        - name: full_name
```
ruflin: I initially hoped that `user.name` is what accounts for both of these. But I see that's not necessarily the case. Do we need `full_name` in ECS? 

In most other cases `*.name` is the user defined part. With the above we change the logic here a bit.

ruflin: My proposal is to go with `1` but specify user.name broader. It does not have to be the programmatic one but can also be a user specified one. `full_name` in contrast is very specific and from my perspective means `firstname` + `lastname` (not sure about order and middle names ...).

webmat: I would never recommend being specific on the content of a full name. It is indeed going to be different from my expectations in some cultures. People should enter whatever makes sense to them in there.

However if we instruct people to add both names in the same fields, all events that have a full name specified in addition to the programmatic name will appear twice in aggregations. I don't think it's a good idea to go in that direction.

ruflin: Let's just soften up logic / programmatic name phrasing a bit in the description, then I'm good to go.

webmat: Updated. Does `Short name or login of the user.` work?

PR #203, opened 2018-11-30
Add minimal field set to represent groups.

webmat:
We may want more fields under `group.` eventually, but I figured we could start with the two most obvious fields.

Note that we're already starting to use them in the Beats transition to ECS. E.g. https://github.com/elastic/beats/pull/9138

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -670,13 +670,13 @@
       fields:
     
         - name: id
-          level: core
+          level: extended
           type: keyword
           description: >
             Unique identifier for the group on the system/platform.
     
         - name: name
-          level: core
+          level: extended
           type: keyword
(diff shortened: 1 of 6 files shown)


fields.yml
```diff
@@ -660,6 +660,27 @@
...
+      title: Group
+      group: 2
+      description: >
+        The group fields are meant to represent groups that are relevant to the
+        event.
+      type: group
+      fields:
+    
+        - name: id
+          level: core
```
ruflin: Perhaps approved too quick. Can we put these under `extended` for now? We can always "migrate" them to `core` later.

webmat: Sure thing

PR #204, opened 2018-11-30
Add user.groups

webmat:
Question: singular or plural?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file based on the
 * Add `host.name` field and clarify usage of `host.hostname`.
 * Add `event.start` and `event.end` date fields.
 * Create new `related` field set with `related.ip`. #206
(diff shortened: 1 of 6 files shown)


ruflin:
I remember we had a discussion around this in the past. As in lucene single and multi values are the same, we should always stick to singular I think. For example `user.id` can also be a group but it's still not `user.ids`. So I would go with `user.group` but as in this case it can often be plural, mention in the description that it can also be a list.

webmat:
@MikePaquette Does that make sense to you as well? `user.group` is singular, even if in some cases we will want to have multiple groups in this field.

MikePaquette:
Yes @webmat singular makes sense, even if it might contain multiple value.  In ECS, we've been using plural only when the field "always" (or virtually aways) contains multiple.  (e.g. *.bytes).

webmat:
Great! I'll rebase and merge, in this case.

PR #205, opened 2018-11-30
Add user details commonly seen in user management audit logs.

webmat:


ruflin:
Not sure if we should put this into scope of ECS 1.0. Do we have these in the auditbeat already?

webmat:
Yes these are fields we will see in any typical user auditing solution. There are more, but these 3 fields are the most obvious ones, IMO.

I haven't touched Filebeat's `auditd` mapping yet, but I saw a need for these in Filebeat's `system.auth` fileset ([useradd log](https://github.com/elastic/beats/pull/9138/files#diff-d89539fc1f780fc038a94210a3c70df5L143)).

ruflin:
What about auditbeat?

ebeahan:
This PR has been stale for a while, and we're going to close it.

PR #207, opened 2018-11-30
Introduce url.full

webmat:
`url.original` is meant to record the unmodified URL, as it was observed.
It's typically only a path in access logs, whereas it may be a full URL in
network monitoring or APM use cases.

This PR introduces a field that should always be populated with a full URL,
whether it's populated at the source or reconstructed after the fact.

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -1180,7 +1180,27 @@
           level: extended
           type: keyword
           description: >
-            Full original url. The field is stored as keyword.
+            Unmodified original url as seen in the event source.
+    
+            Note that in network monitoring, the observed URL may be a full URL,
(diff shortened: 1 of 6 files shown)


fields.yml
```diff
@@ -1130,7 +1130,27 @@
...
+            This field is meant to represent the URL as it was observed, complete
+            or not.
+          example: >
+            https://www.elastic.co:443/search?q=elasticsearch#top
+    
+            or
+    
+            /search?q=elasticsearch
+    
+        - name: full
```
ruflin: Not too happy with the name but don't have a good alternative.

webmat:
Yes, I got this from APM.

PR #208, opened 2018-11-30
Introduce network.community_id

webmat:
I'm introducing this with a dot before the ID, rather than an underscore. The
standard is defined such as there can be new versions in the future, and when
that happens, we may want to introduce additional fields under `community.`,
to help make sense of this.

Let me know if that sounds reasonable.

[Community ID spec](https://github.com/corelight/community-id-spec)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file based on the
 * Add `process.working_directory` and `process.start`. #215
 * Reintroduce `http`. #237
 * Add `user.full_name` field. #201
+* Add `network.community.id` field. #208
 
 ### Improvements
 * Improved the definition of the file fields #196
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,6 @@ All notable changes to this project will be documented in this file based on the
 * Add `process.working_directory` and `process.start`. #215
 * Reintroduce `http`. #237
 * Add `user.full_name` field. #201
-* Add `network.community.id` field. #208
 * Add `network.community_id` field. #208
 
 ### Improvements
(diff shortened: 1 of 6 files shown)


ruflin:
I would like to get a review on this from @MikePaquette or @robgil

webmat:
As the spec mentions, the community ID starts with the community ID version and a colon (e.g. `1:`) and is followed by the hash.

When version `2:` is introduced, we may want to be able to normalize or add fields to help deal with the transition, no? Perhaps I'm solving a problem that won't exist, though.

MikePaquette:
@webmat OK, since we don't use the colon representation in ECS, maybe we should just define two fields now?   Either:
- `network.community.id` and `network.community.version` (a bit funny, since it's supposed to be the **id** version, not the community version)
- `network.community_id` and `network.community_id_version`

webmat:
I'm also of the opinion that we shouldn't add `.version` there yet. We're not sure what the changes to the spec will look like, and whether that would be useful. Let's wait until it happens. Using the dot just gives us "room to grow", when the growing needs to happen.

@MikePaquette Can we go with only `network.community.id` as is, for now?

webmat:
Ping @dcode

dcode:
I agree with @andrewkroh, the version is encoded in the value itself. There won't be any concern about comparing v1 vs v2 ID values. The only reason you'd need to parse that out is if you wanted to ask "What are all records with v1 community IDs?"

I also agree with @MikePaquette with the "too many dots radar". I'm trying to imagine other useful contexts of `network.community`, but failing to come up with something. That said, I could live with it.

webmat:
Yeah I'm ok with going to `community_id`. It may indeed be a problem that won't manifest.

And to compare 1: vs 2:, we people can always do a prefix filter on the hash and compare both anyways.

I'll update the PR

PR #209, opened 2018-11-30
Introduce process.executable.

webmat:
Note that in this PR, I also take the opportunity to move the more important
fields to the top. You may want to review one commit at a time.

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -985,6 +985,28 @@
       type: group
       fields:
     
+        - name: pid
+          level: core
+          type: long
+          description: >
+            Process id.
+          exmple: ssh
+    
+        - name: name
+          level: extended
+          type: keyword
+          description: >
+            Process name.
+    
+            Sometimes called program name or similar.
+          example: ssh
+    
+        - name: ppid
+          level: extended
+          type: long
+          description: >
+            Process parent id.
+    
         - name: args
           level: extended
           type: keyword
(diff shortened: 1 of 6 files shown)


fields.yml
```diff
@@ -982,19 +980,28 @@
...
             Process parent id.
     
+        - name: args
+          level: extended
+          type: keyword
+          description: >
+            Process arguments.
+    
+            May be filtered to protect sensitive information.
+          example: ["-l", "user", "10.0.0.16"]
```
andrewkroh: This example was wrong because `args` should include all the arguments. It's missing `args[0]` which is the thing that was executed.

fields.yml
```diff
@@ -982,19 +980,28 @@
...
+    
+        - name: executable
+          level: extended
+          type: keyword
+          description: >
+            Full path to the process executable.
```
andrewkroh: How about using the word "Absolute" rather than "Full"? They're the same thing, but IMO "absolute path" is even less ambiguous.

andrewkroh:
There are a few places in Beats that will need changed because we have `process.exe`. IIRC they are

- auditbeat auditd fields
- add_process_metadata processor
- packetbeat

fields.yml
webmat: Good point.

fields.yml
webmat: Good catch

PR #215, opened 2018-12-04
Add `process.working_directory` and `process.start`.

### LINKED ISSUE #210: Add working directory and start time to process field set
These are two more straightforward bits of information we'll want to support about processes in the near term. I suggest we use:

- `process.working_directory`
- `process.start_time`

webmat:
Closes #210


webmat:
Rebased.

PR #216, opened 2018-12-04
Re-introduce HTTP protocol breakdown at `network.http`

webmat:
We're considering the idea of nesting protocol breakdowns under `network.`. HTTP is the first one we're incorporating.

Doing so will make it clear where all protocol breakdowns belong, whether or not they are officially in ECS.

If we go this route, we'll likely need to change the documentation format a bit, to be able to link to each protocol breakdown defined in ECS, and also split the yml files to define each protocol breakdown in their own file. Although this work would be done as a separate PR.


webmat:
This may be controversial.

Ping @robgil @graphaelli.

graphaelli:
APM stores request/response headers and parses out cookies into their own objects as well.

webmat:
HTTP will stay at the top level for now the initial GA release of ECS.

Will re-introduce as a different PR.

PR #237, opened 2018-12-05
Re-introduce `http` precisely as it was, as of #d8cece6

webmat:
The goal of this PR is strictly to re-introduce the \`http\` field set as it was, not to work on enhancements to the field set. Those enhancements belong into separate PRs.

We've decided to leave http at the top level for the inaugural release of ECS.  We are still considering nesting all protocol breakdowns eventually, but this will come at a later date.

This reverts #171

### DIFF
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -3,7 +3,7 @@
   title: HTTP
   group: 2
   description: >
-    Fields related to HTTP requests and responses.
+    Fields related to HTTP activity.
   type: group
   fields:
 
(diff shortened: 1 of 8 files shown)


schemas/http.yml
```diff
@@ -0,0 +1,43 @@
+---
+- name: http
+  title: HTTP
+  group: 2
+  description: >
+    Fields related to HTTP requests and responses.
```
MikePaquette: A minor issue, but I'd suggest broadening the definition to something like.  Fields related to HTTP activity.  (Since, for example, `http.version` is not technically a request or a response.)

Otherwise, LGTM

PR #238, opened 2018-12-05
Change `device` to `observer`

MikePaquette:
This changes the `device.*` fields to `observer.*` fields, and clarifies the definition.

The `device` name continued to cause significant confusion amongst new users of ECS, so we're proposing this change before beta2, even though it is a breaking change from beta1.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,15 +84,6 @@ The `type` is the [Elasticsearch field type](https://www.elastic.co/guide/en/ela
 `description` adds details about the field.
 `example` adds an sample value.
 
-## Phases
-
-A field's `phase` indicates where it is in its development. Different phases exist to make it easy for you to contribute new fields, while allowing for more iteration. Here are the phases:
-
-* 0 (alpha): The field is new and is up for discussion as to whether or not it should be added. The field might be removed at any time.
-* 1 (beta): It's clear that there is value of having the field in ECS and discussions about prefixes/naming have started. It's unlikely that the field will be removed, but naming might change at any time.
-* 2 (rc): The field has been accepted and is unlikely to change. It is now tested in the field.
-* 3 (GA): The field is part of ECS. Any breaking changes will happen only at a major release.
-
 ## Guidelines for implementing ECS
 
 * The document MUST have the `@timestamp` field.


README.md
```diff
@@ -314,6 +298,22 @@ The network is defined as the communication path over which a host or network ev
 | <a name="network.total.packets"></a>network.total.packets | Network outbound packets. The sum of inbound.packets + outbound.packets | core | long | `24` |
 
 
+## <a name="observer"></a> Observer fields
+
+An observer is defined as a special network, security, or application device used to detect, observe, or create network, security, or application-related events and metrics. This could be a custom hardware appliance or a server that has been configured to run special network, security, or application software. Examples include firewalls, intrusion detection/prevention systems, network monitoring sensors, web application firewalls, data loss prevention systems, and APM servers. The observer.* fields shall be populated with details of the system, if any, that detects, observes and/or creates a network, security, or application event or metric. Message queues and ETL components used in processing events or metrics are not considered observers in ECS.
```
ruflin: @graphaelli Can you leave your comment on this one? APM-Server is in here :-)

graphaelli: Elastic APM server typically operates as more as a message queue using this definition.  Would logstash be considered an observer?

CONTRIBUTING.md
```diff
@@ -80,15 +80,6 @@ The `type` is the [Elasticsearch field type](https://www.elastic.co/guide/en/ela
...
 
-## Phases
-
-A field's `phase` indicates where it is in its development. Different phases exist to make it easy for you to contribute new fields, while allowing for more iteration. Here are the phases:
-
-* 0 (alpha): The field is new and is up for discussion as to whether or not it should be added. The field might be removed at any time.
-* 1 (beta): It's clear that there is value of having the field in ECS and discussions about prefixes/naming have started. It's unlikely that the field will be removed, but naming might change at any time.
-* 2 (rc): The field has been accepted and is unlikely to change. It is now tested in the field.
-* 3 (GA): The field is part of ECS. Any breaking changes will happen only at a major release.
-
```
webmat: Haha you always try to sneak this one in unrelated PRs. Let's remove it now and be done with it ;-)

README.md
```diff
@@ -238,7 +222,7 @@ A file is defined as a set of information that has been created on, or has exist
 Geo fields can carry data about a specific location related to an event or geo information derived from an IP field.
 
 
-The `geo` fields are expected to be nested at: `destination.geo`, `device.geo`, `host.geo`, `source.geo`.
+The `geo` fields are expected to be nested at: `destination.geo`, `host.geo`, `observer.geo`, `source.geo`.
```
webmat: 👍🏼

README.md
webmat: I would say even if APM server doesn't do a lot of work on the payload, compared to the library installed in each app, it could still be considered as the observer.

graphaelli: Perhaps it makes sense for an observer simply to be an entity that works on behalf of an agent to report its data.

MikePaquette: @graphaelli I think our definition of observer needs to extend beyond what you suggest to include firewalls and network IDS, etc. that don't have an agent at all.

CONTRIBUTING.md
MikePaquette: just backed out the CONTRIBUTING.md chages in bce0b83.

CONTRIBUTING.md
```diff
@@ -84,6 +84,15 @@ The `type` is the [Elasticsearch field type](https://www.elastic.co/guide/en/ela
...
 
+## Phases
+
+A field's `phase` indicates where it is in its development. Different phases exist to make it easy for you to contribute new fields, while allowing for more iteration. Here are the phases:
+
+* 0 (alpha): The field is new and is up for discussion as to whether or not it should be added. The field might be removed at any time.
+* 1 (beta): It's clear that there is value of having the field in ECS and discussions about prefixes/naming have started. It's unlikely that the field will be removed, but naming might change at any time.
+* 2 (rc): The field has been accepted and is unlikely to change. It is now tested in the field.
+* 3 (GA): The field is part of ECS. Any breaking changes will happen only at a major release.
+
```
webmat: So yeah, please make sure to remove this section before merging ;-)

README.md
ruflin: @graphaelli Yes, LS is in many cases also the observer.

simitt: I agree that compared to the agent the server doesn't do a lot with the payload currently. However, I am not sure that this is a general enough criteria for not considering the server an ETL component. There are some additions and changes the server does, and going forward with ECS, the server will change the structure of events even more in the future, as there need to be a mapping from the incoming events to the ES ECS structure. 

My point is, I assume that 

> Message queues and ETL components used in processing events or metrics are not considered observers in ECS.

should clarify what to consider an observer and what not, but for me it introduces more confusion.

PR #240, opened 2018-12-05
Re-introduce a simplified version of `user_agent`.

### LINKED ISSUE #235: Bringing back and cleaning up `user_agent`
Early versions of ECS had `user_agent` more or less as the Ingest Node UA parser happens to output it's parsing, slightly renamed ([see it here](https://github.com/elastic/ecs/blob/d8cece6fbb1f178d77ce91972d1097c0cbc75bfd/schemas/user_agent.yml)).

We've taken it out for Beta 1, to give ourselves some time to review its semantics and re-introduce it the right way.

Previously, the field set covered:

- user agent / browser information: name and version (as 3 fields major/minor/patch)
- device information: name and `version`
  - Note that according to the description of `user_agent.version`, the field was for the device version, not the user agent version. This is unintuitive. Also, it doesn't look like the IN UA parser actually emits this value, at this time.
- operating system information: name and version fields

Looking at https://www.whatsmyua.info/ or https://www.neutrinoapi.com/api/user-agent-info, we can see the result of parsing user agents by a few different libraries. We don't need to support all of this, but it's a good review of some of the concepts encapsulated in a user agent:

- Agent type (desktop vs mobile vs feed reader, robot, etc)
- Agent name and version
- Device name and version
- OS name and version
- CPU architecture (e.g. 64 bits)
- Rendering engine name and version (e.g. WebKit/537.36)
- Producer of the user agent (e.g. Google, Mozilla)
- Screen dimensions

I'm proposing we re-introduce `user_agent` as basic as it was, but a bit better structured. This will give us "room to grow", so we can support more categories of information eventually.

Proposal:

```
user_agent:
  type: broad category of the agent (see above definition)
  name: name of the agent
  version: full version string of the agent
  device:
    name: name of the device
    version: version of the device
  os: (the reusable field set)
```

Caveats:

- Currently ingest node / user_agent doesn't give us full version strings, we'll have to reconstruct from major/minor/patch, which will be imperfect in some situations

Future:

The above proposal leaves room to add other sections such as "engine", "screen" and "architecture", if we deem it necessary.

webmat:
This is an implementation of my proposal in #235.

You'll notice the version breakdown fields are gone for now. I get the feeling that people either don't care about version breakdowns, or they want to see all version fields they see broken down (not just the UA version). So I'm holding off on adding them for now, unless someone feels strongly about this. This will let us discuss ideas about generalizing the concept of breaking down arbitrary version fields in a separate place.

In line with the discussion in #235, for now this is a simple field set, but may grow to cover more concepts:
- agent type (desktop/mobile browser, email client, feed reader, robot)
- client architecture
- rendering engine details
- producer of the user agent (Chrome, Mozilla)
- screen dimensions

This kind of reverts #172 and closes #235.


fields.yml
```diff
@@ -1335,3 +1335,42 @@
...
+          example: Chrome
+          description: >
+            Name of the user agent.
+        - name: version
+          level: extended
+          type: keyword
+          description: >
+            Version of the user agent.
+    
+        - name: device.name
```
ruflin: How does this correlate with https://github.com/elastic/ecs/pull/238 ?

webmat: It correlates in absolutely no way. Here `user_agent.device.name` would contain things like "iPad", "iPhone"...

webmat:
@ruflin No, it does not match what IN's user_agent parser gives us. Here's the format it outputs, and the ECS format, side by side:

IN's user_agent result:

```JSON
"user_agent" : {
  "major" : "12",
  "minor" : "0",
  "os" : "iOS 12.1",
  "os_minor" : "1",
  "os_major" : "12",
  "name" : "Mobile Safari",
  "os_name" : "iOS",
  "device" : "iPhone"
}
```

If we include ECS' reusable `os` field in, the ECS user_agent from this proposal looks like this:

```JSON
"user_agent" : {
  "name" : "Mobile Safari",
  "version": "12.0",
  "device": {
    "name": "iPhone",
    "version": "???"
  },
  "os": {
    "name": "iOS",
    "version": "12.1"
  },
  "original": "Mozilla/5.0 (iPhone; CPU iPhone OS 12_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Mobile/15E148 Safari/604.1"
}
```

Takeaways:

- This just made me realize that I'm not sure if user agents typically give a `device.version` (different from the OS version). **I will remove it for now**
- IN's UA parser gives us a full OS version field, and a broken up agent version field. Initially we may have to reconstruct it from that, to have `user_agent.version`. This is not perfect.
- Most of the other differences I'm already working around with a bunch of field renames, for now. See any of the Beats access log modules: https://github.com/elastic/beats/pull/9245, https://github.com/elastic/beats/pull/9084, https://github.com/elastic/beats/pull/9081, https://github.com/elastic/beats/pull/9005.

ruflin:
@webmat As a follow up we should open an Issue in the Github repo so the processor gets adjusted to ECS format.

fields.yml
```diff
@@ -1423,3 +1423,37 @@
...
+    - name: user_agent
+      title: User agent
+      group: 2
+      description: >
+        The user_agent fields normally come from a browser request. They often
+        show up in web service logs coming from the parsed user agent string.
+      type: group
+      fields:
+    
+        - name: original
```
ruflin: @webmat I still think we should turn indexing off here: https://github.com/elastic/ecs/pull/138

webmat: I don't mind turning it off, actually. But didn't we get a request in Beats for precisely the reverse?

webmat: Of course people can override that, so not a big issue

ruflin: I think that request was for having it as a `text` field? Having is a keyword would still not make ti searchable. I wonder now if not index if a user can still add a `.text` multi-field which is indexed?

webmat: I don't think it's possible, no. But if that person wants indexing, they can totally re-enable keyword indexing on the canonical field, and add a multi-field for text indexing.

If ECS makes it non-indexed, it means that no application that will depend on the official ECS spec can ever perform searches in that field. So whatever users do with it (re-enabling indexing) will not conflict. Because ECS will say it should not be indexed, so the applications can't depend on this being a `keyword` or a `text` field.

What I want to avoid at all cost is to make this canonical field `text`. If we do `text` here, it must be a multi-field.

webmat:
`user_agent.original` is now marked as non indexed. This gives us absolute freedom to change it in the future, if we decide to do so. Whereas going from indexed to not is a breaking change.

PR #242, opened 2018-12-06
Event categorisation fields

tsg:
This PR is a minimal change to the current ECS fields that allows us to go in the planned path for categorization without requiring breaking changes later on. The hope is that we'll get this in for Beta2. It does the following:

* Updates the docs for `event.action` and `event.category` to be closer to what we are planning, but keeps their definition fairly abstract. The docs don't go as far as showing a list of examples.
* Changes the `event.type` definition to be only reserved for now.
* Adds `event.kind` and `event.outcome`, which are fairly non-contentious. I can remove them from the PR if needed to get this in before Beta2.

I plan to follow up with a more complete PR that adds more details.


fields.yml
```diff
@@ -381,35 +381,55 @@
...
     
+        - name: kind
+          level: core
+          type: keyword
+          description: >
+            The kind of the event.
+    
+            This gives information about what type of information the event
+            contains, without being specific to the contents of the event.  Examples
+            are `event`, `state`, `alarm`.
```
ruflin: Instead of `event`, perhaps we should use `log` here? For me all 3 kind are an `event` (based on ECS).

tsg: @cwurm Also mentioned `action` here as an option.

tsg: `log` is not great, btw, in Auditbeat we have events like "process start" that we get from a netlink socket, not from a "log".

.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 *.pyc
 env
+*.swp
```
webmat: LOL I have this in my global gitignore, so it never came up.

Please make it `*.sw?`, though 🙏🏼

fields.yml
webmat: Yes, I agree "event" is better than "log" for this.

webmat: I think the definition should be a bit more forceful in that this field **must** contain one of these values. The goal here is super broad categorization of how this document/event came about. So the values have to be predictable:

- event: just reporting something that was observed in real-time (e.g. syslog message, log entry, event on a bus such as ebpf).
- state: generating an update on the state of something on a regular schedule (list of running processes, installed packages, current CPU usage).
  * So for example, metrics from metricbeat should all use `event.kind:state`.
- alarm: alerting on the fact that something unwanted happened, or a state changed enough to warrant attention from someone. Could be triggered by a static threshold, ML, a rules engine, etc.

What we want to avoid here is to have people map arbitrary values from their event sources to this field, and we end up with lots of unpredictable values in `event.kind`.

webmat: Of course the corollary is that if people determine that their stuff doesn't fall into one of these 3, we want them to open a GH issue to discuss it, and potentially grow that list.

fields.yml
```diff
@@ -381,35 +381,55 @@
...
           type: keyword
           description: >
             Event category.
     
-            This can be a user defined category.
-          example: metrics
+            This contains high-level information about the contents of the event. It
+            is more generic than `event.action`, in the sense that typically a
+            category contains multiple actions.
+          example: user-management
```
webmat: The plan with this field is to have a predetermined set of categories (a few tens of them). Having the field now in ECS sets us up for a sort of breaking change when we come up with that list.

Therefore I think the description can remain as is, but we should add a warning telling people that they're using this field at their own risk, because it will soon be a field that should be populated with prescribed values.

It won't be the end of the world, though. It's not going to break ingestion or anything. It's just that Elastic solutions will expect this field to contain certain values, so if some sources populate this differently, they won't have the best experience...

fields.yml
```diff
@@ -381,35 +381,55 @@
...
+            The action captured by the event.
     
-            This is normally defined by the user.
-          example: nginx-stats-metrics
+            This describes the information in the event. It is more specific than
+            `event.category`. The value is normally defined by the user.
+          example: user-password-change
     
-        - name: action
+        - name: outcome
```
webmat: `outcome` is also meant to be a field with prescribed content. We should add a similar warning as what I suggest for `event.category`. More or less "We're about to come up with a list of prescribed values, so use with caution."

tsg: warning added.

fields.yml
tsg: warning added.

fields.yml
tsg: I added a warning like for the others. I'm not sure we want to define `event`, `state`, `alarm`, etc just yet, there will be more than that and we'd do a poor job if we try to do it by the freeze date.

tsg:
@webmat I think I addressed your comments, please have a look again.

fields.yml
ruflin: I think the defined list of values later on should go into a different document and potentially we have 1 doc per use case and the values for each use case might be different. 

`log` for me does not mean it must come from a file. I'm more thinking of the historical definition `an official record of events during the voyage of a ship or aircraft`.

fields.yml
```diff
@@ -396,35 +396,64 @@
             Unique ID to describe the event.
           example: 8a4f500d
     
+        - name: kind
+          level: core
```
ruflin: Sorry, only spotted now but these fields should go into extended for now.

tsg: I moved `event.kind` and `event.outcome` to extended.

fields.yml
cwurm: > @cwurm Also mentioned `action` here as an option.

I think it was `state_restart`. I wanted a way to distinguish between a state that is triggered periodically (e.g. every 1h) and one triggered by a restart of (in this case) Auditbeat. One could imagine other reasons it was sent (e.g. a threshold was breached, or that too many changes accumulated). If we had something like an `event.trigger` field that captures _why_ a document was sent that would also work - but maybe that's one field too many as well.

ruflin:
@tsg I think you need to run `make` again locally to update the generated files.

tsg:
@ruflin, thanks, I pushed again.

fields.yml
MikePaquette: I'll be proposing pushing these to core sometime in the future :-)

PR #245, opened 2018-12-06
`http.request.method` must be lowercase.

webmat:
Without this, aggregation results will list the same method multiple times, when there's capitalization differences ('get' vs 'GET').


webmat:
In private conversation, we've agreed that actually the normalization here should also be lowercase, for consistency.

ruflin:
What if for now we change the language from `must` to `should` and be more strict later on when we can provide nice templates etc. to be used by our users?

webmat:
I will move the implementation of this to #253, to streamline the discussion in one place.

Issue #250, opened 2018-12-07
Non-indexed fields should show they are not indexed in the readme


webmat:
We just noticed that fields like `event.original`, which is set to `index: false` still renders as `type: keyword` in the readme.

We must fix this rendering issue for Beta 2.

MikePaquette:
+1 to making indexing status clear in the readme. Not sure I'd burn a column for this though, since there will be very few fields with this setting.  Are you thinking just a text note added to the description?

PR #254, opened 2018-12-07
Fix that non index field are shown with keyword in table

ruflin:
Closes https://github.com/elastic/ecs/issues/250


ruflin:
I like the idea of `(not indexed)`. Changed it.

PR #255, opened 2018-12-07
Fix url.original example

ruflin:
Currently examples do not allow multiple lines. This is a quick fix, we must find something nicer in the future.


ruflin:
me neither but couldn't find a better solution quickly.

PR #258, opened 2018-12-07
Add `event.timezone`.

### LINKED ISSUE #213: Adding timezone support in ECS
ECS is currently missing timezone support.

We used to have fields where the value should be defined as a differential in seconds, which we agreed to [remove](https://github.com/elastic/ecs/pull/134), as it was not very friendly (`device.timezone.offset.sec` and `host.timezone.offset.sec`).

In order to re-introduce the timezone, we need to pick a field name, and what values we support in there.

Fun facts:

- Beats can either output:
  - a time differential in hh:mm format (e.g. `-05:00`)
  - or in abbreviated form (e.g. `EST`).
- Ingest Node supports at least 3 input formats:
  - time differentials in hh:mm format
  - abbreviated timezones
  - canonical timezones (e.g. `Europe/Amsterdam`)
- Logstash's date filter supports 2 formats, according to the docs:
  - time differentials in hh:mm format
  - canonical timezones (e.g. `Europe/Amsterdam`)

Here are a few questions that could help shape the direction here:

- [ ] Can someone confirm whether Beats emits the correct time differential immediately after DST changes?
- [x] Should we limit which format goes into the timezone field to only one format, or let people use their preferred format, as long as it's one of the 3?
- [x] What field name(s) seem more appropriate?

@MikePaquette @ruflin @andrewkroh

webmat:
Closes #213.


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ All notable changes to this project will be documented in this file based on the
 * Add `client` and `server` objects and fields. #236
 * Reintroduce a streamlined `user_agent` field set. #240
 * Add `geo.name` for ad hoc location names. #248
+* Add `event.timezone`, to allow for proper interpretation of incomplete timestamps. #258
```
MikePaquette: not sure we need the `,` in the change line.

PR #259, opened 2018-12-07
Add `os.full`, to allow capturing full OS name string, including version

webmat:
I think having a field with the full OS name (including version) is a useful concept, in addition to the broken up `name` + `version` fields. It makes some types of analysis much simpler.

I was inspired to add this when I realized that our user agent parser gives us such a string, and ECS doesn't have a home for it.

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -1208,7 +1208,7 @@
         - name: version
           level: extended
           type: keyword
-          example: "10.12.6-rc2"
+          example: "10.15.1"
           description: >
             Operating system version as a raw string.
     
--- a/fields.yml
+++ b/fields.yml
@@ -1191,7 +1191,7 @@
           description: >
             Operating system name, without the version.
     
-        - name: full_name
+        - name: full
           level: extended
           type: keyword
           example: "Mac OS X 10.15.1"
--- a/fields.yml
+++ b/fields.yml
@@ -1189,7 +1189,14 @@
           type: keyword
           example: "Mac OS X"
           description: >
-            Operating system name.
+            Operating system name, without the version.
+    
+        - name: full_name
+          level: extended
+          type: keyword
+          example: "Mac OS X 10.15.1"
+          description: >
+            Operating system name, including the version.
     
         - name: family
           level: extended
--- a/fields.yml
+++ b/fields.yml
@@ -1194,9 +1194,9 @@
         - name: full
           level: extended
           type: keyword
-          example: "Mac OS X 10.15.1"
+          example: "Mac OS Mojave"
           description: >
-            Operating system name, including the version.
+            Operating system name, including the version or code name.
     
         - name: family
           level: extended
@@ -1208,7 +1208,7 @@
         - name: version
           level: extended
           type: keyword
-          example: "10.15.1"
+          example: "10.14.1"
           description: >
             Operating system version as a raw string.
     
--- a/schemas/os.yml
+++ b/schemas/os.yml
@@ -43,7 +43,7 @@
     - name: version
       level: extended
       type: keyword
-      example: "10.12.6-rc2"
+      example: "10.15.1"
       description: >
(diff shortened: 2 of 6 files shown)


schemas/os.yml
```diff
@@ -24,7 +24,14 @@
       type: keyword
       example: "Mac OS X"
       description: >
-        Operating system name.
+        Operating system name, without the version.
```
ruflin: I think an OS name is always without the version? I see the context you are coming from but I don't think we should mention that here.

schemas/os.yml
```diff
@@ -24,7 +24,14 @@
...
+
+    - name: full_name
```
ruflin: Is that really the `full_name`. This looks to strongly inspired by `user.full_name` but I don't think it's the same. Otherwise it would be `ruflin 18` ;-) 

Do we need this in ECS?

schemas/os.yml
webmat: The user agent parser gives us the following. Check out `os`:

```JSON
          "user_agent" : {
            "patch" : "3538",
            "major" : "70",
            "minor" : "0",
            "os" : "Mac OS X 10.14.0",
            "os_minor" : "14",
            "os_major" : "10",
            "name" : "Chrome",
            "os_name" : "Mac OS X",
            "device" : "Other"
          }
```

I do think that's useful to have this value to compare overall OS usage. The broken down os name and os version in two fields is useful for broader analysis (os name) and then filtering down one OS at a time to see the version breakdowns. But it's not obvious how to use these two fields to get a sense of the overall usage of each specific OS version, all at the same time.

schemas/os.yml
webmat: Open to suggestions for other names. Could be `os.description`, perhaps. Or just `os.full`. I want the concept, but I have no strong attachment to the name `full_name`, I agree it's not perfect.

schemas/os.yml
ruflin: Wouldn't it be that there is first a filter on the OS and then an aggregation on the version to get a correct result?

webmat: If you want to look at the breakdown within one OS, yes, filtering first works, and then the versions you're looking at are consistent and make sense within the context of that OS.

If you want to look at the overall "population" however (e.g. looking at all the OSes accessing your website), then having the full name with version is helpful.

Then you can see a ranking like:

- Mac OS X 10.12: 43345
- Windows 10: 40300
- Mac OS X 10.11: 20000
- Windows 7: 12000

and so on...

webmat: > I think an OS name is always without the version? I see the context you are coming from but I don't think we should mention that here.

Actually this brings up an interesting case to mind. Do people consider "Windows 7" to be the OS name, or is it "Windows"? I think if we don't keep this clarification, we'll get "Windows 7" and "Windows 10" in this field (just without the minor & patch versions).

schemas/os.yml
webmat: Changed to `.full`. I'm good with that.

I think `full_name` was a bit more clear on what it was, but I can see your point vs `user.full_name`. There's some patterns we can use in many places (like the customizable `.name`), but I wouldn't have put `.full_name` in those patterns, so I didn't really see a conflict there :-)

fields.yml
```diff
@@ -1189,7 +1189,14 @@
...
           description: >
-            Operating system name.
+            Operating system name, without the version.
+    
+        - name: full
+          level: extended
+          type: keyword
+          example: "Mac OS X 10.15.1"
+          description: >
+            Operating system name, including the version.
```
ruflin: I would make it more generic, so it could also contain for example `El Capitan` etc.

webmat: Ah, like "including the version or code name"?

webmat: done

PR #261, opened 2018-12-07
Adds a script that generates a JSON representation of the ECS schema

andrew-goldstein:
Adds a script that generates a JSON representation of the ECS schema in the following format:

* The root-level keys are ECS namespace `name`s
* Every root-level value is an object containing the properties of a namespace
* Every namespace object contains a `fields` object, where each key is the `name` of the field, and the value is an object containing the properties of the field

Formatted output as screenshots:

- root-level (collapsed):
root-level
- an expanded root-level object:
expanded

A sample root-level object for the `agent` namespace:

```
{
    "description": "The agent fields contain the data about the software entity, if any, that collects, detects, or observes events on a host, or takes measurements on a host. Examples include Beats. Agents may also run on observers. ECS agent.* fields shall be populated with details of the agent running on the host or observer where the event happened or the measurement was taken.\n",
    "fields": {
      "agent.ephemeral_id": {
        "description": "Ephemeral identifier of this agent (if one exists).\nThis id normally changes across restarts, but `agent.id` does not.",
        "example": "8a4f500f",
        "footnote": "",
        "group": 2,
        "level": "extended",
        "name": "agent.ephemeral_id",
        "required": false,
        "type": "keyword"
      },
      "agent.id": {
        "description": "Unique identifier of this agent (if one exists).\nExample: For Beats this would be beat.id.",
        "example": "8a4f500d",
        "footnote": "",
        "group": 2,
        "level": "core",
        "name": "agent.id",
        "required": false,
        "type": "keyword"
      },
      "agent.name": {
        "description": "Name of the agent.\nThis is a name that can be given to an agent. This can be helpful if for example two Filebeat instances are running on the same host but a human readable separation is needed on which Filebeat instance data is coming from.\nIf no name is given, the name is often left empty.",
        "example": "foo",
        "footnote": "",
        "group": 2,
        "level": "core",
        "name": "agent.name",
        "required": false,
        "type": "keyword"
      },
      "agent.type": {
        "description": "Type of the agent.\nThe agent type stays always the same and should be given by the agent used. In case of Filebeat the agent would always be Filebeat also if two Filebeat instances are run on the same machine.",
        "example": "filebeat",
        "footnote": "",
        "group": 2,
        "level": "core",
        "name": "agent.type",
        "required": false,
        "type": "keyword"
      },
      "agent.version": {
        "description": "Version of the agent.",
        "example": "6.0.0-rc2",
        "footnote": "",
        "group": 2,
        "level": "core",
        "name": "agent.version",
        "required": false,
        "type": "keyword"
      }
    },
    "group": 2,
    "name": "agent",
    "title": "Agent",
    "type": "group"
  }
```

https://github.com/elastic/ingest-dev/issues/131

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ check-license-headers:
 # Clean deletes all temporary and generated content.
 .PHONY: clean
 clean:
-	rm -rf schema.csv schema.md fields.yml build
+	rm -rf schema.csv schema.md schema.json fields.yml build
 	# Clean all markdown files for use-cases
 	find ./use-cases -type f -name '*.md' -not -name 'README.md' -print0 | xargs -0 rm --
 
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,2015 @@
+{
+  "agent": {
+    "description": "The agent fields contain the data about the software entity, if any, that collects, detects, or observes events on a host, or takes measurements on a host. Examples include Beats. Agents may also run on observers. ECS agent.* fields shall be populated with details of the agent running on the host or observer where the event happened or the measurement was taken.\n", 
+    "fields": {
+      "agent.ephemeral_id": {
+        "description": "Ephemeral identifier of this agent (if one exists).\nThis id normally changes across restarts, but `agent.id` does not.", 
+        "example": "8a4f500f", 
+        "footnote": "", 
+        "group": 2, 
+        "level": "extended", 
+        "name": "agent.ephemeral_id", 
+        "required": false, 
+        "type": "keyword"
+      }, 
+      "agent.id": {
+        "description": "Unique identifier of this agent (if one exists).\nExample: For Beats this would be beat.id.", 
+        "example": "8a4f500d", 
+        "footnote": "", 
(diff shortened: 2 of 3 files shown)


webmat:
This is great, thanks for submitting! We'll check this out in detail after we're done with the beta2 release :-)

ruflin:
@andrew-goldstein This is basically the `fields.yml` converted to json? Could you share how you will use this and what the benefits over the yml are?

andrew-goldstein:
> @ruflin  wrote: This is basically the fields.yml converted to json?

Yes

> Could you share how you will use this

The `JSON` representation of the schema would be used by SecOps to display categories and descriptions of fields to users.

> and what the benefits over the yml are?

Since it's `JSON`, it's native to JS, thus you can directly `import` or `require` the file and start working with `ECS`. I think this is an opportunity to serve a large community of JS developers who might want to get started with `ECS`, but would face the same impedance of needing to parse yml as part of a build chain or at runtime to gain access to `ECS` "natively" in JS. Does that make sense?

PR #263, opened 2018-12-07
Add `http.request.body`.

webmat:
If we allow for deep inspection of response body already, I don't see why we wouldn't have a field for the request body as well. I think this was simply overlooked?

This would address an issue we have with migrating the Filebeat elasticsearch log module: https://github.com/elastic/beats/pull/9293

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -851,12 +851,12 @@
             Referrer for this HTTP request.
           example: https://blog.example.com/
     
-        - name: response.status_code
+        - name: request.body
           level: extended
-          type: long
+          type: keyword
           description: >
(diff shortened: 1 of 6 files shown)


fields.yml
```diff
@@ -851,12 +851,12 @@
             Referrer for this HTTP request.
           example: https://blog.example.com/
     
-        - name: response.status_code
+        - name: request.body
```
ruflin: This seems to conflict with https://github.com/elastic/ecs/pull/239

webmat:
Will be implemented directly in #239, to introduce a consistent desired state all at once.

PR #266, opened 2018-12-10
Convert ECS doc from MD to asciidoc guide: initial setup

karenzone:
Convert ECS readme.md file to asciidoc to be published at https://www.elastic.co/guide/index.html.
https://ecs-simple.firebaseapp.com/

Issue #276 opened to track additional steps

### DIFF
--- a/docs/contributing.asciidoc
+++ b/docs/contributing.asciidoc
@@ -1,0 +1,8 @@
+[[ecs-contributing]]
+== Contributing to {ecs}
+
+All information related to ECS is versioned in the https://github.com/elastic/ecs[elastic/ecs] repository. All
+changes to ECS happen through Pull Requests submitted through Git.
+
+See the https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md[Contribution Guidelines].
+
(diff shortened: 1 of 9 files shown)


ruflin:
@karenzone Would it make sense to get this in master rather soonish so we could start to play around with the docs builds? Or can we do this also when it's in a branch?

karenzone:
> @karenzone Would it make sense to get this in master rather soonish so we could start to play around with the docs builds? Or can we do this also when it's in a branch?

What is there so far is in decent shape. It builds without error.  There's more I want to do, but moving v1 into Master wouldn't prevent that.  I'm good with move to master if you are.  Thanks for taking time to look at this.

ruflin:
Now that we branched out 1.0 this can go into master. What are the commands you run the build the docs? I'm thinking of adding a `make` command for it so it's easy for everyone to build it.

ruflin:
Just opened https://github.com/elastic/ecs/pull/275 so we can have an automated docs build in the future.

@karenzone Happy to get yours in first and then rebase mine on top. The firebase link you posted is also very useful.

karenzone:
@ruflin Some reformatting required. For example: 
* Link format is different between Markdown and asciidoc.  `[here](https://github.com/elastic/ecs/releases)` vs `https://github.com/elastic/ecs/releases[official releases]`
* I manually removed the </br> tags because they were rendering in finished doc
* I think I had to reformat lists.

PR #274, opened 2018-12-11
Change `url.port` to long type

praseodym:
This is consistent with other `port` field types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
+* Changed `url.port` field type to `long` for consistency with other `port` fields.
+
 ### Deprecated
 
 ## [1.0.0-beta2](https://github.com/elastic/ecs/compare/v1.0.0-beta1...v1.0.0-beta2)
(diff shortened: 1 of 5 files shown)


elasticcla:
Hi @praseodym, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ecs/pull/274.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

webmat:
@praseodym Could you list this change in the Unreleased section under "Improvements", please? Check out the format of the other entries and keep it simple :-)

ruflin:
We should list this at least in the changelog as Improvement? In the end it does not make really difference if it is int or long.

webmat:
It's a more "lax" requirement, so it satisfies the previous definition indeed. So an improvement, not a breaking change.

praseodym:
Done!

ruflin:
This needs a rebased on master.

webmat:
@praseodym Sorry for messing up here, and thanks for your contribution.

We're working on improving the process for this new repo, so these mistakes don't happen again. 

For simplicity's sake we merged #339 instead.

PR #279, opened 2018-12-17
Add VERSION variable to Makefile

andrewkroh:
Move the version variable to the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-OPEN_DOCS?="-open"
+VERSION   ?= 1.0.0-beta2
+OPEN_DOCS ?= "-open"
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
-OPEN_DOCS?="-open"
+VERSION   ?= 1.0.0-beta2
```
ruflin: An other idea instead of having it in the Makefile would be to feed it into a `.version` file to make it better accessible to other scripts. But this will do for now.

andrewkroh: That could be added in the future by changing the assignment to `$(shell cat .version)`.

PR #283, opened 2018-12-17
Use the correct legal entity for copyright attribution

webmat:


webmat:
Not sure why, but this PR's Travis CI build now points to PR #284's CI...

PR #287, opened 2018-12-18
Version Go deps and use Python virtualenv

andrewkroh:
This updates the Makefile to always use a Python virtualenv when executing any Python scripts or tools.

It also controls what Go dependency versions are used by the .go scripts. This is done by creating a
Go 1.11 module in the scripts directory (not in root to avoid polluting that directory with noise). The
project will continue to work even if checked out into the GOPATH because the Makefile explicitly sets
GO111MODULE=on to force the Go tooling to look for go.mod even if on the GOPATH (see 'go help modules' for details).

Renamed `schemas` target to `csv` because it generates schema.csv.

Added `yamllint` to check the syntax of schemas/*.yml. This is not enforced by `check` (hopefully we can do this in the future).

Added `makelint` to ensure that the targets stay sorted alphabetically.

Added `check-license-headers` to verify that .go files have proper Apache license headers.


Makefile
```diff
@@ -50,11 +57,63 @@ fields:
...
-docs:
-	if [ ! -d $(PWD)/build/docs ]; then \
-		git clone --depth=1 https://github.com/elastic/docs.git ./build/docs ; \
-	fi
+# Format code and files in the repo.
+.PHONY: fmt
+fmt: ve
+	$(FIND) -name '*.py' -exec build/ve/bin/autopep8 --in-place --max-line-length 120 {} \;
+	go get golang.org/x/tools/cmd/goimports
+	goimports -w -l -local github.com/elastic $(shell $(FIND) -name '*.go')
```
ruflin: Nice addition.

Makefile
```diff
@@ -50,11 +57,63 @@ fields:
...
+# Alias to generate everything.
+.PHONY: generate
+generate: csv readme template fields
+
+# Check Makefile format.
+.PHONY: makelint
+makelint: SHELL:=/bin/bash
+makelint:
+	@diff <(grep ^.PHONY Makefile | sort) <(grep ^.PHONY Makefile) \
+	  || echo Makefile targets need to be sorted.
```
ruflin: Someone doesn't like unsorted Makefiles? :-) Only downside to this is that we can't group together commands that are similar.

scripts/go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/elastic/ecs/scripts
```
ruflin: Great to see versioning on these. 

++ on moving the template.go und cmd above.

Issue #113, opened 2018-09-13
Additional groups (Vulnerability / Threat)


dainperkins:
Would it make sense to add additional groups for things like vulnerabilities and threats for correlating between various services and getting granular definitions of this type of data?

e.g.
vulnerability.id
vulnerability.type
vulnerability.first.scanned
vulnerability.last.scanned
vulenerability.status (e.g. open / closed as of current date)
vulnerability.description
vulnerability.cve
vulnerability.severity

------------------
threat.id
threat.name
threat.type
threat.description
threat.risk_score
threat.confidence

Thanks
/d

MikePaquette:
@dainperkins I think this makes sense.  One idea is that the `threat.*` namespace/object/prefix would be one of a small number of special re-usable type objects that allow for uses including `source.threat.*` , `destination.threat.*` and `file.threat.*` 

I had not thought of vulnerability.* in the same way, but perhaps we should?  Would you see the `vulnerability.*` fields always being associated with a host, so `host.vulnerability.*` ?  Or were you thinking of using this to import vulnerability scan results into Elasticsearch?  Or both :-)?  Or others?

dainperkins:
Basis was actually looking at my companies products logging capabilities - and how we would represent "Threats" within  the ecs framework (and frankly I am so sick of grokking, kv'ing, tweaking, and mutating stuff in logstash I sort of jumped into the idea of this type of standardization with both feet).    

Its threat detection, and I had started looking at mixing name spaces like so:

	
	 
event.id				some_id
event.type		        Threat
event.original			send agent url reference

agent.name				PtrX.VA
agent.version		

threat.id				some-id
threat.start			start time
threat.end			end time
threat.duration		duration
threat.type			product specific classification 
threat.tactic			(ATT&CK) e.g. Lateral Movement
threat.technique		Pass the Hash
threat.technique.reference    https://attack.mitre.org/wiki/Technique/T1075
threat.module			(Product specific detection module info)
threat.description		Host Ip attacking destination IP  utilizing Pass the Hash techniques
threat.severity		How bad?
threat.confidence		How sure?

source.ip				
source.port
source.hostname
source.domain

destination.ip				
destination.port
destination.hostname
destination.domain

user.id
user.name

[...]

PR #291, opened 2018-12-20
Add event.tactic and event.technique fields

MikePaquette:
To accommodate security analytics use cases, it is useful to be able to filter, aggregate, and visualize events by the tactics and techniques in use when the event was created.  For example the MITRE ATT&CK framework provides a reference list of tactics and techniques.

If an event or alert is generated by a rule that was written to detect certain tactics or techniques, then those tactics and/or techniques would be populated in these fields.

This PR partially addresses questions raised in https://github.com/elastic/ecs/issues/113


ruflin:
As these are very specific for the security analytics use case I'm starting to think we should group these somewhere outside of `event`?

MikePaquette:
@ruflin that's certainly possible, for example we explored creating a threat.* object in #113 where they could exist.  The argument for keeping them in the `event.*` space is that they'll fit in along with the other ECS categorization fields `event.kind`, `event.category`, `event.type`, and `event.outcome`...

Let me propose an alternative for consideration.

Anyone else have thoughts on this?

ruflin:
Agree that some of the above could actually fit better under thread. Also thinking of `outcome` for example.

MikePaquette:
@ruflin I think "outcome"  will be used quite broadly across many events, including system (successful vs. failed calls), file (successful vs. failed access attempts), and security (successful vs. failed login attempts) events, so I would try to keep it in the common fields as `event.outcome` rather than moving it under a new `threat.*` object.

ruflin:
What about the tactic and technique added here? Should go under  `threat`.

Issue #75, opened 2018-08-10
Missing host fields


willemdh:
`host.containerized` and `host.os.codename` seem to be missing from the host object, while they are added by the new:

```
- add_host_metadata:
    netinfo.enabled: true
```

processor.

ruflin:
+1 on adding these, do you want to open a PR?

Note: Not all Beats field necessarily have to end up in ECS. Main focus is on correlation and filtiering which definitively applies on the two above.

andrewvc:
I've opened a PR here: https://github.com/elastic/ecs/pull/294

PR #294, opened 2018-12-21
Add host.containerized and os.codename

andrewvc:
This came up in https://github.com/elastic/ecs/issues/75 and https://github.com/elastic/beats/issues/9154.

These fields seem useful. I notice no other fields in ECS are boolean, so this may be a first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,7 @@ A host is defined as a general computing instance. ECS host.* fields should be p
 | <a name="host.mac"></a>host.mac | Host mac address. | core | keyword |  |
 | <a name="host.type"></a>host.type | Type of host.<br/>For Cloud providers this can be the machine type like `t2.medium`. If vm, this could be the container, for example, or other information meaningful in your environment. | core | keyword |  |
 | <a name="host.architecture"></a>host.architecture | Operating system architecture. | core | keyword | `x86_64` |
+| <a name="host.containerized"></a>host.containerized | True if the host is running in a containerized environment. | core | keyword | `True` |
 
 
 ## <a name="http"></a> HTTP fields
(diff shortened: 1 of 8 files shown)


webmat:
Thanks @andrewvc!

Not exactly the same semantics, but check out the definition for the new field `os.full`, wrt OS code names: https://github.com/elastic/ecs#os. Would this work for your use case?

As for `host.containerized`, a few things:

- I applaud you on your courage to introduce the first `boolean`! However you can only get that distinction if you actually make the field `boolean`, right now it's `keyword` ;-)
- Out of curiosity, in this case you're using `host` to report information from the point of view of the containerized process, correct? Can you expand on the situation, and perhaps give an example event how you would envision this?
  - I wonder if the `.containerized` field shouldn't be elsewhere, like perhaps `process.containerized` or `service.containerized`.

webmat:
Oh that could work too, indeed. You're right that the current `host.type` example sounds redundant vs `cloud.machine.type`.

andrewvc:
Is there still interest in this issue? CC @ppf2

webmat:
We'll need to step back and have a broader discussion, for what happens here.

I could see some value in adding `os.codename`, although `os` already has a pretty good amount of fields. Not sure if adding `codename` in adition to `name` and `full` is all that useful.

As for `host.containerized`, we've discussed in other places that `host` was purely meant for a VM or a physical host. Not for a container. The container details belong in `container`. Do you know how `host.containerized` is being populated in Beats? I'd like to better understand the semantics.

The fix for what @ppf2 reported is actually a bugfix in Beats. If Beats populates these custom fields (at least custom for now), there should be a field definition added in `libbeat/_meta/fields.common.yml`.

andrewvc:
@webmat I'm +1 for fixing this in beats as custom fields. I'm going to close this issue given that.

ruflin:
@webmat We should sync up offline around if host is also mean for containers and others. Not sure if we are on the same page here.

PR #295, opened 2018-12-28
[WIP] Add script for base fields extraction

ruflin:
The first trial was as Go script but it became too cumbersome so I did it quickly in Python ...

Problems with the generation of this fields.yml:

* The output is standardised but not very human readable. Is that ok?


ruflin:
Closing as this is not needed anymore.

PR #296, opened 2019-01-07
Generate process.args as []string in Go code

andrewkroh:
Add special handling for the process.args field to ensure that it's generated as a []string.


andrewkroh:
Having some kind of hint in the schema would be helpful. There are very likely other places where defining a slice would be helpful. I'll keep a list as I encounter them. But converting all of the generated code to slices would make the package very annoying to use.

webmat:
I created https://github.com/elastic/ecs/issues/297 for follow-up

Issue #36, opened 2018-07-06
syslog PRI


willemdh:
Is syslog program is meant for process.name, is there a field for the PRI syslog field (which I used to name syslog_pri). This allows for calculation of facility and severity:

https://gist.github.com/marvin/1017480

praseodym:
There is `event.severity`, but only as a `long`. I don't see a good way to map facility and severity labels from the [syslog_pri filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-syslog_pri.html).

willemdh:
Most syslog sources have a PRI field. A dedicated syslog object could also contain syslog.facility and syslog.severity.

ruflin:
I would prefer to not have service specific fields in the generic objects. I tempted to say at the moment because of it would belong under `syslog.pri` but not sure if it should be part of ECS. At the same time syslog is very popular.

willemdh:
A dedicated syslog object is ok for me. So is this mapping what we need?

```
"syslog": {
  "properties": {
    "pri": {
      "type": "long"
    },
    "severity": {
      "ignore_above": 1024,
      "type": "keyword"
    },
    "facility": {
      "ignore_above": 1024,
      "type": "keyword"
    }
  }
}
```

ruflin:
I'm working on extending our use cases here: https://github.com/elastic/ecs/pull/55 This would allow us to have a fields.yml for syslog in the use cases and then decide later if it makes sense in ECS directly or not but have it there for others to share (and comment).

willemdh:
This will soon be resolved, see https://github.com/elastic/ecs/pull/525
Thanks all!

PR #301, opened 2019-01-14
Add syslog protocol fields to event

kvch:
This PR adds the following fields to ECS:
* `event.facility`
* `event.priority`
* `event.facility_label`
* `event.priority_label`

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -603,6 +603,35 @@
             This is mainly useful if you use more than one system that assigns
             risk scores, and you want to see a normalized value across all systems.
     
+        - name: facility
+          level: core
+          type: long
+          example: 1
+          description: >
+            Value parsed from messages adhering to RFC 5424 or RFC 3164.
+            It represents the process the event has originated from.
+    
+        - name: facility_label
+          level: extended
+          type: keyword
+          example: kernel
+          description: >
+            Human readable format of `event.facility`.
+    
+        - name: priority
+          level: core
+          type: long
+          example: 1
+          description: >
+            Value parsed from messages adhering to RFC 5424 or RFC 3164.
+    
+        - name: priority_label
+          level: extended
+          type: keyword
+          example: Informational
+          description: >
+            Human readable format of `event.priority`.
+    
     - name: file
       group: 2
       title: File
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -195,3 +195,32 @@
 
         This is mainly useful if you use more than one system that assigns
         risk scores, and you want to see a normalized value across all systems.
+
+    - name: facility
+      level: core
+      type: long
+      example: 1
+      description: >
+        Value parsed from messages adhering to RFC 5424 or RFC 3164.
+        It represents the process the event has originated from.
+
+    - name: facility_label
+      level: extended
+      type: keyword
+      example: kernel
+      description: >
+        Human readable format of `event.facility`.
+
+    - name: priority
+      level: core
+      type: long
+      example: 1
+      description: >
+        Value parsed from messages adhering to RFC 5424 or RFC 3164.
+
+    - name: priority_label
+      level: extended
+      type: keyword
+      example: Informational
+      description: >
+        Human readable format of `event.priority`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ All notable changes to this project will be documented in this file based on the
   `server.address`. #247
 * Add `os.full` to capture full OS name, including version. #259
 * Add generated source code for Go. #249
+* Add syslog protocol fields to event namespace. #301
 
 ### Improvements
 * Improved the definition of the file fields #196
--- a/README.md
+++ b/README.md
@@ -220,6 +220,10 @@ The event fields are used for context information about the log or metric event 
 | <a name="event.end"></a>event.end | event.end contains the date when the event ended or when the activity was last observed. | extended | date |  |
 | <a name="event.risk_score"></a>event.risk_score | Risk score or priority of the event (e.g. security solutions). Use your system's original value here. | core | float |  |
 | <a name="event.risk_score_norm"></a>event.risk_score_norm | Normalized risk score or priority of the event, on a scale of 0 to 100.<br/>This is mainly useful if you use more than one system that assigns risk scores, and you want to see a normalized value across all systems. | extended | float |  |
+| <a name="event.facility"></a>event.facility | Value parsed from messages adhering to RFC 5424 or RFC 3164. It represents the process the event has originated from. | core | long | `1` |
+| <a name="event.facility_label"></a>event.facility_label | Human readable format of `event.facility`. | extended | keyword | `kernel` |
+| <a name="event.priority"></a>event.priority | Value parsed from messages adhering to RFC 5424 or RFC 3164. | core | long | `1` |
+| <a name="event.priority_label"></a>event.priority_label | Human readable format of `event.priority`. | extended | keyword | `Informational` |
 
 
 ## <a name="file"></a> File fields
--- a/code/go/ecs/event.go
+++ b/code/go/ecs/event.go
@@ -135,4 +135,17 @@ type Event struct {
 	// This is mainly useful if you use more than one system that assigns risk
 	// scores, and you want to see a normalized value across all systems.
 	RiskScoreNorm float64 `ecs:"risk_score_norm"`
+
+	// Value parsed from messages adhering to RFC 5424 or RFC 3164. It
+	// represents the process the event has originated from.
+	Facility int64 `ecs:"facility"`
+
+	// Human readable format of `event.facility`.
+	FacilityLabel string `ecs:"facility_label"`
+
+	// Value parsed from messages adhering to RFC 5424 or RFC 3164.
+	Priority int64 `ecs:"priority"`
+
+	// Human readable format of `event.priority`.
+	PriorityLabel string `ecs:"priority_label"`
 }
--- a/schema.csv
+++ b/schema.csv
@@ -45,12 +45,16 @@ event.created,date,core,
 event.dataset,keyword,core,stats
 event.duration,long,core,
 event.end,date,extended,
+event.facility,long,core,1
+event.facility_label,keyword,extended,kernel
 event.hash,keyword,extended,123456789012345678901234567890ABCD
 event.id,keyword,core,8a4f500d
 event.kind,keyword,extended,state
 event.module,keyword,core,mysql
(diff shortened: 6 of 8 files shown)


fields.yml
```diff
@@ -1521,6 +1521,42 @@
...
+          example: 1
+          description: >
+            Value parsed from syslog messages. It represents the process
+            the message is originated from.
+    
+        - name: facility_label
+          level: extended
+          type: keyword
+          description: >
+            Human readable format of `syslog.facility`.
```
webmat: Could you add an example here?

fields.yml
```diff
@@ -1521,6 +1521,42 @@
...
+          example: 1
+          description: >
+            Value parsed from syslog messages. It is calculated using the
+            formula `8 * syslog.facility + log.level`.
+    
+        - name: priority_label
+          level: extended
+          type: keyword
+          description: >
+            Human readable format of `syslog.priority`.
```
webmat: Could you add an example here as well?

schemas/syslog.yml
```diff
@@ -0,0 +1,36 @@
...
+      description: >
+        Human readable format of `syslog.facility`.
+
+    - name: priority
+      level: core
+      type: long
+      example: 1
+      description: >
+        Value parsed from syslog messages. It is calculated using the
+        formula `8 * syslog.facility + log.level`.
```
webmat: We could shorten this a bit. ECS doesn't really need to elaborate on the formula. We're just telling people where to stick their Syslog `priority`, they don't need to calculate it in their pipeline, I would assume :-)

schemas/syslog.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: syslog
+  title: Syslog
+  description: >
+    Source fields describe details about the source of a packet/event.
+    Source fields are usually populated in conjunction with destination fields.
```
webmat: Copy/pasta ;-)

schemas/syslog.yml
```diff
@@ -0,0 +1,36 @@
...
+  type: group
+  fields:
+
+    - name: facility
+      level: core
+      type: long
+      example: 1
+      description: >
+        Value parsed from syslog messages. It represents the process
+        the message is originated from.
```
webmat: is => has

webmat:
@kvch Oh, and your Hunglish is fine ;-)

kvch:
If you don't want to add a new top-level field, I suggest these are added to `log.*`. These concepts are strongly tied to logging in general.  Both values are syslog components specified by syslog RFCs. Also, they are added to log messages by syslog daemons.

willemdh:
Personally I'd love to see a spot for the syslog pri too, see https://github.com/elastic/ecs/issues/36

willemdh:
@ruflin Putting it under log.* is ok for me, as long as it has a place somewhere in ecs.

webmat:
I agree with adding syslog's priority as well. Any other attributes we should consider?

I prefer to see these syslog details under `log.` than under `service.`. We haven't discussed the semantics of how `service.` would work (at least I don't remember). So in that sense, nesting these details under `log.` seems much more straightforward.

willemdh:
Ok, seem most kind of agree it needs to be under log and we need facility, severity and priority. Only choice left is if we should place it directly under log like this:

```
log.facility
log.priority
log.severity
```

or, considering these are really syslog specific place them under log.syslog like this:

```
log.syslog.facility
log.syslog.priority
log.syslog.severity
```

To me the log.syslog prefix makes more sense. I think there might be other candidats with similar requirements in the future. (such as certain java or .net logging frameworks).

webmat:
I'd be curious to know which other tool use facility/priority/severity. And are the semantics close enough to Syslog's to map them to the same fields?

For now I have a slight preference for `log.syslog.*`

ph:
from my understanding, they are really close the Syslog reality, so I think would prefer to have it under `log.syslog.*`

MikePaquette:
Note: outlier position ahead :-)

 I think burying these syslog fields down under `log.*` or `syslog.*` defeats the purpose of trying to make a common schema. I think they would fit better under `event.*`

Reason: in ECS, the "event" is the top level entity.  "logs" and "metrics" are sub-entities (See [event field definition](https://github.com/elastic/ecs#-event-fields)).

Despite the name, "syslog" is just a transport mechanism for events.  In fact syslog can be used for any key-value pair, including metrics.  For example, Pivotal allows you to [send container metrics over the syslog protocol](https://docs.pivotal.io/pivotalcf/2-4/opsguide/logging-config-opsman.html#container-metrics). 

So nesting syslog fields below `log.*` does not seem correct, just as it would not be correct to nest it under `metric.*` if ECS had a "metric" object (which it doesn't).

So I'd propose that we create new fields for `event.priority` and `event.facility` and map as follows:

syslog severity -> `event.severity` (field already defined by ECS)
syslog priority -> `event.priority` (proposed new field)
syslog facility -> `event.facility` (proposed new field)

@webmat ArcSight CEF has a `deviceFacility` field that would also map to an ECS `event.facility`
Splunk CIM has several related fields including `message_priority` which would map to `event.priority` and `severity` would would map to `event.severity`

willemdh:
Personally I've wondered several times why there still is a log top level object. Isn't a log some sort of event? And why would log. Original not fit into event. Original? And log. Level under event. Level or event. Severity?

webmat:
@willemdh Yeah currently `log.` is used for lower level details specific to log files. Filebeat (in 7) will also be nesting a few more details here that aren't ECS specific, like the file offset.

@MikePaquette The reason I was proposing with `log.syslog.*` was because I want to avoid the `event` field set to become overcrowded. But you're right that these attributes don't necessarily have to be specific to log files, they are applicable to anything else.

I agree to go with `event.[facility|priority|severity]`.

Our description for these fields should mention the relationship Syslog, but shouldn't limit to Syslog. These values should be populated per event source, and only need to be consistent within that event source (they shouldn't be normalized vs other sources populating the same fields).

webmat:
@MikePaquette Revisiting this, I'd like to close this loop soon.

Another worry I have with adding them directly at `event.priority`, `event.priority_label`, `event.facility` and `event.facility_label` is that we've been moving in the direction of normalizing the content of some of the other event fields.

In that context, what do we do with the Syslog ones? Should we recommend people populate the fields with the Syslog values as is? And what about other event sources that have attributes similar to severity, priority and facility? Should they be recorded there as is, too?

MikePaquette:
> Another worry I have with adding them directly at event.priority, event.priority_label, event.facility and event.facility_label is that we've been moving in the direction of normalizing the content of some of the other event fields.

True for event categorization fields, but I think these don't fall strictly into that category.

> In that context, what do we do with the Syslog ones? Should we recommend people populate the fields with the Syslog values as is? 

Yes, I think that's fine.  Within a syslog transported event stream, the set should be consistent.

> And what about other event sources that have attributes similar to severity, priority and facility? Should they be recorded there as is, too?

Yes, but you raise a good question about `event.severity` however that is beyond the scope of this PR, as it is already an ECS field.

zlammers:
In addition to the syslog fields mentioned above, what about the syslog timestamp and syslog host?  

In many major vendors (Fortinet, Checkpoint, Palo Alto come to mind), there are log aggregators often in the mix, like a FortiAnalyzer, MLM, or Panorama.  Forwarded syslog often contains the TS and host of the log aggregator itself, not of the original [observer] host (or if it does, it contains both, where you'd see "TS HOST <PRI> TS HOST MESSAGE" where the first pair of TS HOST is a log fowarder).

We are currently working on syslog -> ECS parsers for major FW vendors (the above, and Cisco and Juniper), and as ECS was missing fields contained in this PR (plus others), we had decided to just 'store in labels' until we have a more official place (which sounds like it will be at least partially 'event.*' at this time):

`
labels.syslog_host (keep labels.syslog_host value even if that value is utilized for observer.hostname [notably Cisco, Juniper])
labels.syslog_timestamp (keep labels.syslog_timestamp value even if that value is utilized for @timestamp [notably Cisco, Juniper])
labels.syslog_priority
labels.syslog_facility 
labels.syslog_facility_label
labels.syslog_severity 
labels.syslog_severity_label
labels.syslog_forwarder_host
labels.syslog_forwarder_timestamp
`

micoq:
Since the facility, the priority are specific to syslog (in it's header), we use a separate `syslog` namespace:

```
syslog.pri: the complete priority code between `<` and `>`)
syslog.facility.code: the numeric identifier of the facility decoded from the priority code
syslog.facility.label: normalized label (`local0`,`kernel`...)
syslog.severity.code: the numeric identifier of the severity decoded from the priority code
syslog.severity.label: normalized label (`emerg`,`warning`...)
syslog.hostname: the hostname in the syslog header
syslog.rfc: which version of the RFC (5424 or 3164)
syslog.date.raw: the raw date/timestamp in the header
```
We can add some fields to store the `tag` (the process name):
```
syslog.tag.process: the name of the process (httpd)
syslog.tag.pid: the pid (if specified: `httpd[4242]`)
```
However, these fields can be redundant and are not already defined since they are considered as a part of the user message.

webmat:
@micoq I like your `.code` and `.label` layout :-) However since we already have `event.severity` defined in ECS, I'm not sure we should introduce a breaking change to move it to `event.severity.code`, unfortunately.

So I think adding the following fields is straightforward:

* `event.priority`
* `event.facility`
* `event.facility_label`

There's two issues I still see here:

1. I'm not sure I would add `event.severity_label`: values expected here are equivalent to what we have in `log.level` right now (debug, info, warn, error, etc). I think I would rather adjust the description of `event.severity` to say its corresponding label must be stored in `log.level`. Or perhaps we could instead deprecate `log.level` and move it under `event.` eventually?

2. This PR is adding `priority_label`, but I'm not sure there's such a thing as priority label? severity and facility both have a corresponding `_label`, but priority doesn't, afaik.

What do you folks think?

schemas/event.yml
```diff
@@ -195,3 +195,32 @@
...
+      example: 1
+      description: >
+        Value parsed from messages adhering to RFC 5424 or RFC 3164.
+
+    - name: priority_label
+      level: extended
+      type: keyword
+      example: Informational
+      description: >
+        Human readable format of `event.priority`.
```
MikePaquette: Rather than "Human readable", can we add "text-based" or something similar?  Humans can read numbers too :-)

webmat: The `Informational` example here is an example of a **severity** label (which maps to `log.level` in ECS).

I don't think there is such a thing as a priority label in Syslog. The priority number is the one made up of `8*facility + severity`.

I think we need to remove this field.

webmat:
Just opened a new PR to potentially replace this one. See details and rationale over there 👉#525 :-)

PR #308, opened 2019-01-23
Make `user.group` a nesting of the `group` field set

### LINKED ISSUE #304: `user.group` should not be a keyword field, but a place where we can nest the `group` field set
The current definition of the field `user.group` is problematic, in that it's a place to put a textual group name. This doesn't leave room for other bits of metadata related to a group (the most obvious one being group.id).

The `user.group` field actually goes against our principle of not reusing the name of a field set (`group`) as a field with a different meaning (`user.group` being a string, rather than the nested field set). 

This issue came up while working on the right way to represent the various users/groups that went into determining effective rights: https://github.com/elastic/beats/pull/10192, https://github.com/elastic/beats/pull/9963 and https://github.com/elastic/beats/issues/10111.

I would like to suggest we make this change for ECS 1.0.0 GA.

Discuss ;-)

cc @ruflin @MikePaquette @cwurm @andrewkroh

webmat:
An error slipped in, and instead of nesting the `group` field set at `user.group`, we originally made this a `keyword` field.

This PR introduces a breaking change, but we are aiming to introduce this fix in ECS 1.0.0 GA.

The initial implementation went against our principle of not reusing top level object names elsewhere in the hierarchy, using different semantics.

This PR closes #304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
+* Remove the `user.group` `keyword` field, introduced in #204. Instead,
+  the `group` field set can be nested at `user.group` #308
+
 ### Added
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,13 @@ All notable changes to this project will be documented in this file based on the
 ### Breaking changes
 
 * Remove the `user.group` `keyword` field, introduced in #204. Instead,
-  the `group` field set can be nested at `user.group` #308
+  the `group` field set can be nested at `user.group`. #308
 
 ### Bugfixes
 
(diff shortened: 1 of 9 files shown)


webmat:
Philosophical question: the changelog entry goes under "breaking", or "bugfixes"? :-)

webmat:
@cwurm for awareness

ruflin:
For the changelog, it should be mentioned in both places. It's a bug fix but a breaking bug fix.

webmat:
I just duplicated the whole changelog entry as is. Considered wording it differently (why it's a bugfix), but I figure if people care, they'll check out the PR and the issue.

ruflin:
@webmat I stumbled over this but was not sure if I should comment. If you can word it differently would appreciate it.

webmat:
Ok, will do ;-)

PR #329, opened 2019-02-18
Clarify the difference between @timestamp and event.created

### LINKED ISSUE #326: Clarify @timestamp vs event.created
To quote the current readme from `event.created`:

> event.created contains the date when the event was created.
> This timestamp is distinct from @timestamp in that @timestamp contains the processed timestamp.

If say we take a fake login pseudo-event:
`Jan 20 10:17:24 SERVER auth: username logged in`

The user *actually* logged in at 10:17:24 for all we know. Based on this comment by @ruflin https://github.com/elastic/ecs/issues/8#issuecomment-393418791 it would seem that event.created is the time that whatever tool (logstash, beats, etc) reads the file or received the syslog event. I do not think Filebeats works this way, and I know Winlogbeats does not. It will set `@timestamp` to the time the event occurred or when it landed in the local logging system, which I view as correct.

Can what "processed" means here be clarified? To me, when the event was created is when it occurred, and processed is when some tool (which could be many, Beats, LS, ES, etc) received the event. My real-world use case allows for events collected by a management console to make the time an event happened and the time when it was processed to be several minutes apart.

webmat:
Closes #326


schemas/base.yml
```diff
@@ -14,8 +14,11 @@
       description: >
         Date/time when the event originated.
 
-        For log events this is the date/time when the event was generated, and
-        not when it was read.
+        This is the date/time extracted from the event, representing when the
```
MikePaquette: should probably say "This is the date/time extracted from the event, typically representing when the"

schemas/base.yml
```diff
@@ -14,8 +14,11 @@
...
+        event was generated by the source.
+
+        If the event source has no original timestamp, this value must be
```
MikePaquette: "this value must be" seems too restrictive.  We've previously defined `must` and `should` semantics to indicate whether ECS will break or will cause unexpected results.   I suggest changing this to "this value is typically"

schemas/event.yml
```diff
@@ -153,16 +153,16 @@
...
-
-        This timestamp is distinct from @timestamp in that @timestamp contains
-        the processed timestamp. For logs these two timestamps can be different
-        as the timestamp in the log line and when the event is read for example
-        by Filebeat are not identical. `@timestamp` must contain the timestamp
-        extracted from the log line, event.created when the log line is read.
-        The same could apply to package capturing where @timestamp contains the
-        timestamp extracted from the network package and event.created when the
-        event was created.
+        event.created contains the time when the event was first read by an
```
MikePaquette: "event.created contains the time when the event was first read by an" should be
"event.created typically contains the date/time when the event was first read by an"

schemas/event.yml
```diff
@@ -153,16 +153,16 @@
...
+        agent, or by your pipeline.
+
+        This field is distinct from @timestamp in that @timestamp should contain
```
MikePaquette: "This field is distinct from @timestamp in that @timestamp should contain" should be
"This field is distinct from @timestamp in that @timestamp typically contains"

schemas/event.yml
```diff
@@ -153,16 +153,16 @@
...
+        the time extracted from the original event.
+
+        In most situations, these two timestamps will be slightly different.
+        The difference will represent the delay between your source
```
MikePaquette: "The difference will represent the delay between your source" could be
"The difference can be used to calculate the delay between your source"

schemas/event.yml
```diff
@@ -153,16 +153,16 @@
...
+        generating it, when your agent first processed it. This can be used to
```
MikePaquette: "generating it, when your agent first processed it. This can be used to" should be
"generating an event, and the time when your agent first processed it. This can be used to"

schemas/event.yml
```diff
@@ -153,16 +153,16 @@
...
+        monitor your agent's ability to keep up with your event source.
```
MikePaquette: "monitor your agent's ability to keep up with your event source." should be
"monitor your agent's or pipeline's ability to keep up with your event source."

PR #330, opened 2019-02-18
Introduction of an optional short field description

### LINKED ISSUE #317: cloud.provider example under Cloud fields
My understanding of cloud.provider should be the name of different cloud providers, like `aws`, `azure` or `gcp`. Under https://github.com/elastic/ecs#cloud, `ec2` is one of the examples for cloud.provider.

webmat:
This PR adds a `short:` description to the YAML format for defining fields. All fields who need a short description for easy display in cramped spaces now have one.

This PR does not address the outputs which would benefit from this new short description. This should be addressed in a separate PR per output. PRs welcome, BTW ;-)

In reviewing all of these field descriptions, some main descriptions were adjusted as well.

A few other concerns have been addressed also:

- Replace incorrect `ec2` example in `cloud.provider` with `aws`.
  - closes #317
- Add pointer in description of `http` field set to `url` field set.
- Split up all field set definitions in multiple paragraphs.

I can extract to their own PR, if needed.

## Questions

- [x] @andrewgoldstein Do you feel like we need a "short" description for "key" descriptions as well? (E.g. [client field set description](https://github.com/elastic/ecs/blob/v1.0.0-beta2/schemas/client.yml#L5-L6))
  - Yes we need it. Obvious when looking at fieldset headings in the asciidoc format. They're part of this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,18 @@ All notable changes to this project will be documented in this file based on the
 * Field set name "group" was being used as a leaf field at `user.group`, instead
  of being a nesting of the field set. This goes against a driving principle of ECS,
  and has been corrected. #308
+* Replaced incorrect `ec2` example in `cloud.provider` with `aws`. #330
 
 ### Added
+
+* Added pointer in description of `http` field set to `url` field set. #330
+* Added an optional short field description. #330
 
 ### Improvements
 * Clarified the definition of the host fields #325
 * Specify the `object_type` for field `labels`. #331
+
+* Clarified the definition of the host fields. #325
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,6 @@ All notable changes to this project will be documented in this file based on the
 * Clarified the definition of the host fields #325
 * Specify the `object_type` for field `labels`. #331
 
-* Clarified the definition of the host fields. #325
-
 ### Deprecated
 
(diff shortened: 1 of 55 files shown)


karenzone:
@dedemorton Here's @webmat 's work on toward giving us a shortdesc.

schemas/cloud.yml
```diff
@@ -20,7 +20,7 @@
       type: keyword
       short: Name of the cloud provider.
       description: >
-        Name of the cloud provider. Example values are aws, azure, gce, or
+        Name of the cloud provider. Example values are aws, azure, gcp, or
```
webmat: Good catch

CHANGELOG.md
```diff
@@ -14,11 +14,16 @@ All notable changes to this project will be documented in this file based on the
 * Field set name "group" was being used as a leaf field at `user.group`, instead
  of being a nesting of the field set. This goes against a driving principle of ECS,
  and has been corrected. #308
+* Replaced incorrect `ec2` example in `cloud.provider` with `aws`. #330
```
webmat:
```suggestion
* Replaced incorrect examples in `cloud.provider`. #330
```

schemas/cloud.yml
```diff
@@ -17,8 +18,9 @@
       level: extended
       example: ec2
```
kaiyan-sheng: @webmat Sorry I just saw this PR for some reason... Thanks for making the changes! I just found this `ec2` still in provider example. This probably should be `aws` instead.

webmat: Oh jeez, we fixed the example in the text, but not in the `example` field, haha! Thanks for pointing it out 👍

PR #332, opened 2019-02-18
Make mentions of lowercasing refer to a general section of the docs.

webmat:
The goal here is that the descriptions make sense no matter where they're seen (e.g. Kibana tooltip, as much as in the actual docs).


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 * Clarified the definition of the host fields #325
+* Make phrasing of lowercasing directive more relevant, no matter where its shown. #332
```
webmat:
```suggestion
* Make phrasing of lowercasing directive more relevant, no matter where it's shown. #332
```

PR #334, opened 2019-02-20
Phase 2 in conversion to asciidoc

karenzone:
More work toward the move to asciidoc:
- Added new files and content.
- Played with the content of *Base Fields* to demonstrate a three-column format.
- Watch for possible terminology change from "core and extended" to "core and custom" for fields.

Want to see this rendered? https://ecs-docs.firebaseapp.com/ecs2/


docs/index.asciidoc
```diff
@@ -16,58 +12,31 @@ data from sources like logs and metrics or IT operations
 analytics and security analytics.
 
 ECS is still under development and backward compatibility is not guaranteed. Any
-feedback on the general structure, missing fields, or existing fields is appreciated.
-For contributions please read the  https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md[Contribution Guidelines].
+feedback on the general structure, missing fields, or existing fields is
+appreciated. For contributions please read the
```
webmat: I think for 1.0, we should change the message here.

Here's what we're now shooting for:

- We are following Semantic Versioning, meaning:
  - Backwards compatibility during the 1.x lifetime
  - Possibility of breaking changes only at major versions
- Related: We will align ECS major version with Elastic Stack major versions

At least that's how I think we should approach it. This stability will be a big decision factor for people deciding whether or not to adopt ECS.

WDYT @MikePaquette?

cc @ruflin

docs/index.asciidoc
```diff
@@ -16,58 +12,31 @@ data from sources like logs and metrics or IT operations
...
-** Top-Level Namespaces/Objects
-** Reusable Namespaces/Objects
-** Assets and Asset Lists
-** Pseudonymized and Anonymized Data in ECS
-** Threat and Vulnerability Data in ECS
-** Revisit inbound, outbound bytes/packets
+* *Extended fields.* Any fields that are not a core field. 
+Extended fields may apply to more narrow use cases, or may be more open
+to interpretation depending on the use case. Extended fields are more likely to
+change over time.
```
webmat: Love the very direct and straightforward mention of what core and extended mean, right in the ECS index page ❤️

docs/guidelines.asciidoc
```diff
@@ -28,6 +27,8 @@ practices.
 * *Singular or plural.* Use singular and plural names properly to reflect the field content. For example, use `requests_per_sec` rather than `request_per_sec`.
 * *General to specific.* Organise the prefixes from general to specific to allow grouping fields into objects with a prefix like `host.*`.
 * *Avoid repetition.* Avoid stuttering of words. If part of the field name is already in the prefix, do not repeat it. Example: `host.host_ip` should be `host.ip`.
-* *Use prefixes.* Fields must be prefixed except for the base fields. For example all `host` fields are prefixed with `host.`. See `dot` notation in FAQ for more details.
+* *Use prefixes.* Fields must be prefixed except for the base fields. For example, all `host` fields are prefixed with `host.`. See `dot` notation in FAQ for more details.
```
webmat: Here's the perfect place to address one of the small clarifications I wanted to address in time for GA.
```suggestion
* *Use prefixes.* Fields must be prefixed except for the base fields. For example, all `host` fields are prefixed with `host.`. See `dot` notation in FAQ for more details.
  * Note that the document structure should be nested JSON objects. If you use Beats or Logstash, the nesting of JSON objects is done for you automatically. If you're ingesting to Elasticsearch via the API, make sure your fields are nested objects, not strings containing dots.
```

webmat: Trying to nest a bullet point under the "Use prefixes" guideline, not sure if this is the way in asciidoc.

This would address "Clarify on dots in field names" from the GA plan.

docs/glossary.asciidoc
```diff
@@ -2,11 +2,15 @@
...
+ECS::
 Elastic Common Schema. A common set of document fields, field names, and their respective entity
 relationships to be used in the storage of log messages and other data in
 Elasticsearch.
 
+ECS compatible::
+Describes an implementation with all existing fields mapped to ECS fields.
+
+ECS compliant:: Describes an implementation with all existing fields mapped to
+ECS fields, and with as many ECS core fields populated as is practical.
```
webmat: I think we wanted to remove these two distinctions, right @MikePaquette?

These concepts haven't been fully fleshed out, I would remove for now, and only introduce when we're ready to have these kinds of discussions.

docs/fields-gen.asciidoc
```diff
@@ -1,23 +1,28 @@
 [[ecs-base]]
 === Base fields
 
-The base set contains all fields which are on the top level. These fields are common across all types of events.
+The base set contains top level fields that are common across all types of events
```
webmat: We should always quote field names -- such as `base` above -- in backticks, to get the monospace visuals. Otherwise this can make for strange-sounding sentences.

docs/fields-gen.asciidoc
```diff
@@ -1,23 +1,28 @@
...
 |=======================================================================
-| Field  | Description  | Level  | Type  | Example 
-| [[@timestamp]] | Date/time when the event originated. 
+| Field  | Description  | Level 
+| [[@timestamp]] | `date`. Date/time when the event originated. 
 For log events this is the date/time when the event was generated, and not when it was read.
-Required field for all events. | core | date | `2016-05-23T08:05:34.853Z` 
-| tags | List of keywords used to tag each event. | core | keyword | `["production", "env2"]` 
-| labels | Key/value pairs.
+Required field for all events. Example: `2016-05-23T08:05:34.853Z` | core
```
webmat: Could we try introducing a line break, and mention the data type above or below the Level, in the cells?

Not sure how we'll name the column for this hybrid column, but I wouldn't worry about that just yet. I'm just curious to see if this renders well, and if it makes sense visually.

```suggestion
Required field for all events. Example: `2016-05-23T08:05:34.853Z` | core 
date
```

docs/conventions.asciidoc
```diff
@@ -1,7 +1,7 @@
 //[[ecs-conventions]]
 == {ecs} Conventions
 
-{ecs} is most effective when you understand and follow conventions.
+{ecs} is most effective when you understand and follow these guidelines and conventions.
```
webmat: ❤️ Much less patronizing! LOL, I shudder looking at previous wording. Oops 😆

docs/convert.asciidoc
```diff
@@ -0,0 +1,63 @@
+[[convert-to-ecs]]
+== Converting an implementation to ECS
+
+A common schema helps you correlate and use data from various sources. 
+
+Fields for most Elastic modules and solutions (version 7.1 and later) are mapped
```
webmat: Should be 7.0

```suggestion
Fields for most Elastic modules and solutions (version 7.0 and later) are mapped
```

webmat: Loving these suggestions :-) @karenzone plz let me know if they're useless to you :-)

docs/convert.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+to the Elastic Common Schema. You may want to map data from other
+implementations to ECS to help you correlate data across all of your products
+and solutions.
+
+Before you start a conversion, be sure that you understand the basic <link here>.
```
webmat: Let's keep a todo to finish this sentence. Or can we for now just say "the basics outlined below:"?

docs/convert.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+[float]
+[[ecs-comp]]
+=== ECS compatible or compliant?
+
+An implementation can be considered *compatible* with ECS as long it maps all
+existing fields to ECS fields.
+
+An implementation would be considered *compliant* with ECS only if it also
+populates as many of the ECS core fields as is practical. Duplication, adding
+metadata, or deriving values via manipulation and math are acceptable.
```
webmat: In line with my other comment on compatible/compliant, we should consider removing this section.

docs/faq.asciidoc
```diff
@@ -0,0 +1,99 @@
...
+
+[float]
+=== Why does ECS use a dot notation instead of an underline notation?
+
+There are two common key formats for ingesting data into Elasticsearch:
+
+* Dot notation: `user.firstname: Nicolas`, `user.lastname: Ruflin`
+* Underline notation: `user_firstname: Nicolas`, `user_lastname: Ruflin`
+
+For ECS we decided to use the dot notation. Here's some background on this decision.
```
webmat: Let's reinforce the meaning of dots here at the introduction of the concept. It's well laid out below for those who will pore over each section. Would this make sense?

```suggestion
For ECS we decided to use the dot notation, representing nested objects. Here's some background on this decision.
```

docs/index.asciidoc
karenzone: @MarkSettleES let me know that  "core" and "custom"  have been proposed instead of "core" and "extended". 
WDYT @webmat? @MikePaquette? @MarkSettleES? Others?

docs/fields-gen.asciidoc
karenzone: Noted. This particular example is populated by generated content, so we'll need make those adjustments in the schema files.

docs/guidelines.asciidoc
karenzone: I did something a little different. Take a look and let me know if you'd like to discuss.

docs/fields-gen.asciidoc
karenzone: See the new approach in the next commit.  I think you'll like it better than what we have here.

docs/fields-gen.asciidoc
webmat: Haha oh jeez. You can see I'm not used to reviewing the asciidoc yet 🤦‍♂️ 

I'll take care of this one :-)

docs/index.asciidoc
webmat: Also while not technically a beta anymore, perhaps we should still guard and say "this is the first official version, so consider accordingly".

docs/index.asciidoc
webmat: Core and extended are absolutely both the main levels we use to define ECS fields.

However Mark makes a good point. I have seen some of Mike's content list the 3 together, and I think that's a great idea. I think it could reinforce the idea that ECS is a standard that still embraces the custom that people need around it.

I agree we should add it as a third bullet point in there.

docs/guidelines.asciidoc
webmat: Yes, love the way you did it. 👍

docs/faq.asciidoc
```diff
@@ -16,33 +17,37 @@ The benefits to a user adopting these fields and names in their clusters are:
...
 There are two common key formats for ingesting data into Elasticsearch:
 
 * Dot notation: `user.firstname: Nicolas`, `user.lastname: Ruflin`
 * Underline notation: `user_firstname: Nicolas`, `user_lastname: Ruflin`
 
-For ECS we decided to use the dot notation. Here's some background on this decision.
+ECS uses the dot notation to represent nested objects. 
 
 [float]
+[[nptation-diff]]
```
webmat: I'd go for "notation-diff" here instead ;-)

docs/glossary.asciidoc
MikePaquette: @webmat @karenzone 
I agree to remove these definitions for now, but I want to make sure we include guidance like: "In order to increase the level of interoperability of your data with future analysis content, your implementation should attempt to populate as many of the Core fields as practical, even if these fields or values are metadata that are not present in your original event, or are already populated in another ECS field"

I think this guidance is fundamental to the success of a common schema, and one I've not been very successful at articulating.

Simple example: when converting a syslog firewall log event to ECS, there maybe no field in the event with the value "syslog" , but a good ECS implementation  will still populate `agent.type: syslog` so that future ECS-aware analysis content that filters events by `agent.type` will work properly.

docs/index.asciidoc
MikePaquette: @karenzone @MarkSettleES to close the loop here,  "Custom" can be included as a third level, but is not an alternative to "extended".

So I agree with @webmat let's go with three levels:
1. Core
2. Extended
3. Custom

docs/index.asciidoc
MikePaquette: @webmat I like the message you propose.  I would make sure to position these as goals, as there may be unforeseen situations that cause us to change these plans.

I would skip the "first official version" note - I the 1.0.0 version already makes that clear.

docs/glossary.asciidoc
webmat: I've taken an additional note to reinforce the docs this way, in the docs plan

PR #336, opened 2019-02-20
A new file generator

webmat:
This pull request introduces a simplified generator for files generated based on the ECS core files in `schemas/*`. The idea is to read everything, augment all fields with defaults, copy reusable fields to their intended destination, eventually lint for problems or add validations; then we save the intermediary in memory representation as a generated file (see `generated/ecs/fields_flat.yml`).

At this point, we can trigger a series of generators based on this. Either in Python, based on the in memory dictionary, or another language, based on this simplified and fully fleshed out intermediary file.

This PR introduces the following generators:

- The one that saves the intermediary YML representation
- The schema.csv file. It moves from the root to `generated/schema.csv`
- Elasticsearch 6 and 7 sample templates, at `generated/elasticsearch/*`

This PR also introduces a few other things:

- ECS version to be used in code generation is now saved in `version` at the root of the repo
- Some Python tests for the generator
- Python tests intended to spec ECS itself. The introductory test in this file ensures we don't introduce a bug where the base fields are nested under `base.*` :-) The file is `scripts/tests/test_ecs_spec.py`, and should be used for any high level truism we want to ensure about ECS itself. Not for typical corner cases and unit tests.
- These tests run as part of `make check` in Travis, or can be called specifically with `make test`.
- The new generator is automatically called in the global `make generate`, but you can run only the new generator with only `make generator`

**TODO** before merging

- [x] Generate different index template for v6 and v7
- [x] Remove template.json and schema.csv at the root (they're now in `generated/*/`)
- [x] Readme: point people to the generated files.
  - gist: they'll all gradually move to `generated/*/` from now on, except for docs
- [x] Set a default `object_type` for `type: object` fields? (only one without a default at this time)

Not in scope for this first PR, but should get attention soon:

- adding validations / linting rules
- generating asciidoc
- generating the readme (will be replaced by asciidoc soon anyway, so will not be ported)
- generating "perfect" beats yml defs
- generating the Kibana JSON or the Go library
- generating a sample Kibana index pattern in line with the ES templates
- generating a Kibana canvas workpad to explore ECS


webmat:
@MikePaquette Ok, the new generator is now able to generate the csv file and the template, including the reusable fields.

In the future I'd like to mostly output the generated files all in one place, in generated/*. But for the purpose of the pull request diffs, I'm also overwriting the old version of the two files, right at the root of the repo.

MikePaquette:
@webmat love the new CSV output with the nested fields!
Question: Should we / can we include a version column in the CSV output with the ECS release version?
Seems a bit redundant to have it in every row, but we'd be able to import each CSV file into Elasticsearch and keep statistics and visualizations of ECS over time.

webmat:
@MikePaquette I'm a bit hesitant about this, because as you say, it will be the same value for every single line.

I would rather recommend tweaking the import process or script to add this value to the destination, every time a new import is made of a new ECS version.

MikePaquette:
Thanks @webmat my preference is the CSV file should have some indication of what version of ECS it represents, regardless of any processes that consume it downstream.  Yes, putting version on every row seems a bit silly, but it works, and it less than 300 rows, so not a big deal from a space perspective.

webmat:
Alright, let's do it, then :-)

scripts/generators/csv_generator.py
```diff
@@ -0,0 +1,43 @@
...
+                                   lineterminator='\n')
+
+        schema_writer.writerow(["Field", "Type", "Level", "Example", "ECS version"])
+        for field in sorted_fields:
+            schema_writer.writerow([
+                field['flat_name'],
+                field['type'],
+                field['level'],
+                field.get('example', ''),
+                version
```
andrewkroh: If this were a `@since <version>` type of field that indicates when the field was first added to the spec this would be useful to anyone trying to write backwards compatible code.

webmat: Good idea. That's not currently the intent, but I'm noting as a future improvement.

The gist of this field specifically is that one can repeatedly import schema.csv in a spreadsheet, once for each ECS version, and then have them all at their disposal. A request from Mike.

But I love the idea of the introduction date for a field... We'll see how we can work that in.

PR #339, opened 2019-02-22
Change url.port datatype to long

webmat:
This discrepancy was noticed when I replaced the Go template generator with a
new one, in #336. Looks like the Beats generator was making the field `long`
even if the definition said `integer`.


webmat:
Let's wait for @ruflin's return on this one. Not sure if we should backport to 1.0.

webmat:
Tagging for backport anyway, to make sure we don't forget about it.

praseodym:
Duplicate of #274 🙂

webmat:
🤦‍♂️

PR #341, opened 2019-02-23
Add Autonomous System (AS) fields

andrewkroh:
This adds fields for Autonomous System (AS) information. It's common to enrich
events containing IP addresses with information about the AS controlling that
address.

It fields are in a reusable object that can be attached to source, destination, client,
and server. For example:

- `source.autonomous_system.number`
- `source.autonomous_system.organization.name`

Some future fields we might want in this object are (I'm considering them now just to make sure nothing we do now precludes future growth):

- `cidr` - the IP range advertised by the AS (e.g. 216.90.108.0/24)
- `organization.country_code_iso` - country code where the org is registered
- `registry` - what registry the AS is with (e.g. ARIN)
- `registration_date` - When the AS was first registered
- `last_update` - When the AS registration was last updated.

Some resources to see what kind of data is available.
- https://www.team-cymru.com/IP-ASN-mapping.html
- https://whois.arin.net/rest/net/NET-216-90-108-0-1/pft?s=216.90.108.1


fields.yml
```diff
@@ -76,6 +76,41 @@
...
+    
+        - name: number
+          level: extended
+          type: long
+          description: >
+            Unique number allocated to the autonomous system. The autonomous
+            system number (ASN) uniquely identifies each network on the Internet.
+          example: 15169
+    
+        - name: organization.name
```
praseodym: I think `organization_name` would be a better name, unless there will be more properties than just `name` under the `organization` property.

praseodym: Ah, I missed the suggestion of `organization.country_code`. Maybe that information is better suited for the `geoip` field group, though.

andrewkroh: I was thinking that the `source.geo` field might be populated with the location of the IP whereas the details under `source.autonomous_system.organization` would be about the organization (like perhaps the country where the organization is headquartered.)

praseodym: We could have something like `source.autonomous_system.organization.geo`, but maybe that's a bit over the top.

andrewkroh: `geo` is a “reusable object” so it could be used under `organization`. The current structure in this PR won’t preclude this from specified in the schema at a later time. This PR will get us the bare minimum fields for using the geoip processor with the AS database.

dainperkins:
couple of thoughts if its not too late:

1) AS is a specific routing term, its commonly referenced when talking about e.g. ARIN lookups, but   its also relevant for other routing protocols and is really only globally significant for public BGP (assuming no one started trading other protocols publicly while I wasn't looking)

2) We might want to consider AS for network as well for router logs, route changes etc. (network.as?) or possibly splitting up the taxonomies to be more technically accurate (or put it all under network and allow that to be nested under source / dest etc)

I have to consider a bit more to relearn my way down the bgp/routing stack, but I can immediately see NPM desires to track routing changes by AS (particularly in dual homed BGP setups) and usign a single organization header to describe a single AS country code (as opposed to src/dst AS & related info)...

andrewkroh:
I updated the PR to start address @webmat's actions:

- [x] We need to discuss whether we shorten the name to as
- [x] The PR needs to be rebased on top of master
- [ ] We should mark organization as reusable under as, rather than redefining the field explicitly there.

andrewkroh:
@webmat What's the right way to accomplish the last bullet?

I tried this, but it break the generator.
```
- name: organization
  reusable:
    top_level: true
    expected:
      - client.as
      - destination.as
      - server.as
      - source.as
```

```
Running generator. ECS version 1.1.0-dev
  File "scripts/generator.py", line 43, in <module>
    main()
  File "scripts/generator.py", line 17, in main
    (ecs_nested, ecs_flat) = schema_reader.load_ecs()
  File "/Users/akroh/go/src/github.com/elastic/ecs/scripts/schema_reader.py", line 175, in load_ecs
    finalize_schemas(fields_nested, fields_flat)
  File "/Users/akroh/go/src/github.com/elastic/ecs/scripts/schema_reader.py", line 169, in finalize_schemas
    duplicate_reusable_fieldsets(schema, fields_flat, fields_nested)
  File "/Users/akroh/go/src/github.com/elastic/ecs/scripts/schema_reader.py", line 130, in duplicate_reusable_fieldsets
    if 'nestings' not in fields_nested[new_nesting]:
KeyError: 'client.as'
make: *** [generator] Error 1
```

generated/csv/fields.csv
```diff
@@ -8,7 +8,11 @@ agent.id,keyword,core,8a4f500d,1.1.0-dev
 agent.name,keyword,core,foo,1.1.0-dev
 agent.type,keyword,core,filebeat,1.1.0-dev
 agent.version,keyword,core,6.0.0-rc2,1.1.0-dev
+as.number,long,extended,15169,1.1.0-dev
```
andrewkroh: @webmat Seems like `reusable.top_level: false` is not being honored. I assume that if it's `false` that this field would not exist?

andrewkroh: Created https://github.com/elastic/ecs/issues/495.

andrewkroh:
I reverted the change to make `oranization` reusable under `as` for the time being because it wasn't working as expected. For example the `organization` fields were missing in the places where `as` was reused (e.g. `source.as.organization.name`).

PR #347, opened 2019-02-27
Generator for asciidoc representation of all ECS fields

webmat:
This code generates both the index page for the fields section, and the detailed page for each fieldset.

**TODO**

- [x] Fix fieldset order: Base first, then alphabetic
- [x] Fix field order: preserve original order from schema files
- [x] Default `short` description for fieldsets to `description`. Raise if `description` has newlines.
- [x] Render table with complete field set details
- [x] Revert back to alphabetical sorting for fields
- [ ] Render the footnotes as well?
- [x] Mention where a field set is expected to be reused.
  - In fieldset, list acceptable destination fieldsets.
  - In destination fieldset, mention which fieldsets are expected here, as contractions linking to their canonical documentation.
- [x] Replace markdown in schemas with asciidoc markup (e.g. hyperlinks)

**How to visualize this PR**

Via GitHub's rendering of asciidoc, in my branch

- [Field Index Page](https://github.com/webmat/ecs/blob/asciidoc-fields/docs/fields.asciidoc)
- [Field Details Page](https://github.com/webmat/ecs/blob/asciidoc-fields/docs/field-details.asciidoc)  

Or in the ECS repo

```
make generate
make docs
```

http://localhost:8000/ecs-fields.html 👀


webmat:
@dedemorton Hey this uses the 3 column layout you were proposing in https://github.com/elastic/beats/issues/9519. I'm doing so with a twist: we're packing a bit of information vertically, for each row:

- Middle column is short description (1 line) and on line below, the example for the field
- Rightmost column is ECS level followed by the field's datatype.

I'm not in love with how exactly this first attempt looks, but I think it's a good start :-)

cc @karenzone

webmat:
Index page:

ecs-fields-index


Details page:

ecs-fields-details

webmat:
Updated screenshots.

**Index page**

Notice Base is now first.

ecs-fields-index

**Fieldset details page**

- Still has summary table at the beginning
- Below, each field is detailed in full
- Trimmed down the summary table, since we now have details below
- Fields are now displayed in the same order as they're listed in `schemas/*.yml`

ecs-fields-details

dedemorton:
@webmat I like the second layout is better (the table in the first iteration was a little confusing visually when you added the level and type info).  It would be nice for the entries in the table to link to the full description under field details.

@karenzone I pinged you about this in slack, but wanted to ask if you can make sure this content builds OK with the asciidoctor version of the doc build.

tl;dr for @webmat: we're eventually migrating to asciidoctor, which is more restrictive than asciidoc.

webmat:
@dedemorton Yes, linking to the detail section was next on my TODO :-) As for the table, I agree the second is better. But I'm, still experimenting to find just the right way to integrate the level, I'm still not satisfied. So ideas are welcome there as well. 

I'm more than happy to build using both versions of the tooling, to make sure things will remain ok when we switch over. Just point me in the right direction :-)

webmat:
Actually since ECS now has short descriptions everywhere, 4 columns renders pretty good. My favourite below is "**type last**". I'm also attaching a bunch of 3-col experiments as a zip. I really prefer 4 cols.

**level last**

level-last-base

level-last-http

 **type last**

type-last-base

type-last-http

**3 columns**

[3-col.zip](https://github.com/elastic/ecs/files/2919769/3-col.zip)

webmat:
Actually, here's another format that I think works best. Going with this for now.

I call this one **Mike** ;-)

mike-base

mike-http

webmat:
HTTP is really cramped, though 🤔

dedemorton:
>HTTP is really cramped, though
Yah the overall layout of tables in really bad. Unfortunately the knobs for adjusting cell padding in asciidoc don't work with our doc build. I haven't tested on asciidoctor. I've been wanting a css change for quite awhile, but that's not happened yet.

To build with asciidoctor, run the docker-based version of the doc build (build_docs instead of build_docs.ph) and include the --asciidoctor flag. Sometimes one problem can cause many messages, so don't lose hope. :-) Reach out to Nik E if you have questions about the asciidoctor build.

webmat:
I realize I haven't posted a full page rendering of the listing of field sets, yet. With the new short descriptions, the page looks great :-)

field-index-page


Also, moving back to the layout initially proposed in https://github.com/elastic/ecs/pull/334:

base

http

webmat:
First experiment displaying what field sets may be nested under `host`. Feedback welcome.

host-details-with-nesting

generated/ecs/fields_flat.yml
```diff
@@ -12,6 +12,7 @@
   flat_name: '@timestamp'
   level: core
   name: '@timestamp'
+  order: 0
```
ruflin: Why do we need to keep the order? Shouldn't we just sort alphabetically?

webmat: We can do either. I was preserving the order because we've been loosely sorting fields in order of importance, and grouping them as well (e.g. the metrics at the end).

Although now that you question this, I realize it makes more sense to sort alphabetically, and explain specifically the relationships between fields. Perhaps we can use the field set description for this.

ruflin: I would try with alphabetical ordering for now to see how well it works. And if we have some fields which need to be on the top (`@timestamp`?) we can still introduce a special param. If we have the order for each entry I worry we easily forget or have duplicates etc. If we introduce a new entry, potentially lots of other entries need update.

webmat:
@ruflin As of now, this PR doesn't render footnotes yet. Only `agent` and `cloud` have one, and I'm not sure what to do with it.

I wonder if we should drop this attribute altogether. If we need to flesh out details about a section, we should do it in the main description, I think.

What do you think?

webmat:
@karenzone Your input would be welcome on how best to display information about field reuse.

Where a field set is expected is currently two sentences (just like in the readme), whereas what's expected to be nested under a field set is a table.

Should we do all sentences, no tables? Should we do two tables? Only one table, with an added column that explains which direction the reuse goes?

See the screenshot above, where `user.*` is on both sides of the field reuse :-)

generated/ecs/fields_flat.yml
webmat: Now in alphabetical order. Haven't removed the metadata from intermediary files yet (the `order`key), in case we change our mind.

webmat:
Ok, this is ready to merge, IMO.

We can revisit how reusability is rendered as a separate PR. What we currently have there is functional, though.

ruflin:
I skimmed through the code and looked at the docs locally. What is most important for me in this PR is that we have the docs in asciidocs and can iterate on top of it. There are a few things I would change but lets discuss it later as it's just about pages and formatting, not the general idea.

PR #348, opened 2019-02-28
Fix the actual cloud.provider example.

webmat:
PR #330 only fixed the text in the description :-)

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -222,7 +222,7 @@
       fields:
         - name: provider
           level: extended
-          example: ec2
+          example: aws
           type: keyword
           short: Name of the cloud provider.
           description: >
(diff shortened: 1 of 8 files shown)


fields.yml
```diff
@@ -222,7 +222,7 @@
       fields:
         - name: provider
           level: extended
-          example: ec2
```
andrewkroh: The implementation in Beats populates this with `ec2`. It would be helpful to have a list of names in ECS to use for popular providers. Ironically one of the providers is using `ecs`. https://github.com/elastic/beats/blob/master/libbeat/docs/processors-using.asciidoc#add-cloud-metadata

MikePaquette: I think the document example is better using "aws" but there could be a case for using `service.name`  in cloud-provider relevant events to capture the service such as "ec2."   Agree we could expand the list of examples to eventually become a set of defined names.

webmat: Agreed on the enums. However they will come after 1.0 :-)

PR #350, opened 2019-02-28
Ecs version 1.0.0

webmat:
As is customary for each release, this is making '1.0.0' the example value for `ecs.version`. This PR is meant for the `1.0` branch only.

Notes:

- template.json was already being generated with version 1.0.0
- libbeat's template.New had to be adjusted to include the `migration` param. Not sure why, but this isn't a problem on master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Meta-information specific to ECS.
 
 | Field  | Description  | Level  | Type  | Example  |
 |---|---|---|---|---|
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -10,8 +10,8 @@ See CONTRIBUTING.md for more details on setting up.
 
 -->
 
-WARNING: This is the master branch. The current release v1.0.0-beta2
-can be found [here](https://github.com/elastic/ecs/tree/v1.0.0-beta2).
+WARNING: This is the master branch. The current release v1.0.0
```
ruflin: I assume this should be removed from the branch?

webmat: Well the readme as we know it will be gone soon, with the docs moving to the main site. So let's not sweat it too much. But I'll make this read "this is a development branch".

PR #351, opened 2019-02-28
Set version to 1.1.0-dev in master

webmat:
Notes:

- From now on, version on dev branches should contain the `-dev` label.
- Set master to 1.1.0-dev, as the 1.0 branch has been cut
- VERSION in Makefile now defaults to the content of `version` file
- Removes mention of what the "current version" is from the definition of
  `ecs.version`. The statement would go out of date very quickly.
- I'm setting the example for `ecs.version` to 1.0.0 and I think we should stop
  playing with it from now on, to reduce busywork. 1.0.0 is a good round number
  for an example :-)


Makefile
```diff
@@ -6,7 +6,7 @@ FIND             := find . -type f -not -path './build/*' -not -path './.git/*'
 FORCE_GO_MODULES := GO111MODULE=on
 OPEN_DOCS        ?= "-open"
 PYTHON           := build/ve/bin/python
-VERSION          ?= 1.0.0-beta2
+VERSION           = $(shell cat version)
```
ruflin: ++

webmat: You're good with the removal of the conditional assignment, right?

I think if people want to pretend things are a different version, they can do so by modifying their artifacts. I don't think they should do it, so I don't think the Makefile should support it, though ;-)

Merging now, but open to change back, if there's something I'm not seeing

ruflin: Yes

PR #366, opened 2019-03-05
Backport #336 to 1.0: A new file generator (#336)

webmat:
Backport of PR #336 to 1.0 branch. Original message:

This introduces a simplified generator for files, based on the ECS core files in `schemas/*`.

The idea is to read everything, augment all fields with defaults, copy reusable fields to their intended destination, eventually lint for problems or add validations; then we save the intermediary in memory representation as a generated file (see `generated/ecs/fields_flat.yml`).

At this point, we can trigger a series of generators based on this. Either in Python, based on the in memory dictionary, or another language, based on this simplified and fully fleshed out intermediary file.

This PR introduces the following generators:

- The one that saves the intermediary YML representation in `generated/ecs/`
- The schema.csv file. It moves from the root to `generated/schema.csv`
- Elasticsearch 6 and 7 sample templates, at `generated/elasticsearch/*`
- The old schema.csv and template.json have been moved to `generated/legacy/` for the time being (still generated).

This PR also introduces a few other things:

- ECS version to be used in code generation is now saved in he file `version` at the root of the repo
- Some Python tests for the generator
- Python tests intended to spec ECS itself. The introductory test in this file ensures we don't introduce a bug where the base fields are nested under `base.*` :-) The file is `scripts/tests/test_ecs_spec.py`, and should be used for any high level truism we want to ensure about ECS itself. Not for typical corner cases and unit tests.
- These tests run as part of `make check` in Travis, or can be called specifically with `make test`.
- The new generator is automatically called in the global `make generate`, but you can run only the new generator with `make generator` #naming-legend
- New section in the readme, pointing people to the generated files


version
```diff
@@ -0,0 +1 @@
+1.0.0-beta2
```
ruflin: needs updating (later?)

webmat: Yes, let's do straight backports and update this file after

PR #370, opened 2019-03-05
Backport #266 to 1.0: Convert ECS doc from MD to asciidoc guide: initial setup (#266)

webmat:
Backport of PR #266 to 1.0 branch. Original message:

* ECS doc conversion from md to asciidoc
* Add back uc-header file
* New asciidoc files

### DIFF
--- a/docs/conventions.asciidoc
+++ b/docs/conventions.asciidoc
@@ -1,0 +1,76 @@
+//[[ecs-conventions]]
+== {ecs} Conventions
+
+{ecs} is most effective when you understand and follow conventions.
+
+[float]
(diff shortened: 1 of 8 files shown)


docs/conventions.asciidoc
```diff
@@ -0,0 +1,76 @@
+//[[ecs-conventions]]
```
ruflin: Is this commented out on purpose? Probably in the original already like this.

webmat: It's like that in the original. We'll look at this on the final push for the docs, after this release.

PR #373, opened 2019-03-05
Backport #279 to 1.0: Add VERSION variable to Makefile (#279)

webmat:
Backport of PR #279 to 1.0 branch. Original message:

Move the version variable to the Makefile.

PR #374, opened 2019-03-05
Backport #287 to 1.0: Version Go deps and use Python virtualenv (#287)

webmat:
Backport of PR #287 to 1.0 branch. Original message:

This updates the Makefile to always use a Python virtualenv when executing
any Python scripts or tools.

It also controls what Go dependency versions are used by the .go scripts.
This is done by creating a Go 1.11 module in the scripts directory (not in
root to avoid polluting that directory with noise). The project will
continue to work even if checked out into the GOPATH because the Makefile
explicitly sets GO111MODULE=on to force the Go tooling to look for go.mod
even if on the GOPATH (see 'go help modules' for details).

Renamed `schemas` target to `csv` because it generates schema.csv.

Added `yamllint` to check the syntax of schemas/*.yml. This is not enforced
by `check` (hopefully we can do this in the future).

Added `makelint` to ensure that the targets stay sorted alphabetically.

Added `check-license-headers` to verify that .go files have proper Apache
license headers.


webmat:
This PR is built on top of #373 and shouldn't be merged before it.

PR #379, opened 2019-03-07
Generate full Beats field definitions, including nested fields

webmat:
This work has revealed a subtle bug in the generated files schema.csv and both ES templates, which were missing the `group` fieldset in all places where `user` is reused. (e.g. `host.user.group.*`)

**Notes**

- [ ] ECS has more keys to describe fields than Beats requires (e.g. `reusable`). This new generator will filter them, to let through only what matters. Question: should we let through the ECS `level`?


webmat:
@ruflin Could you check this out? A big difference between this file and the previous one we crafted manually (other than the YAML rendering of strings) is that fields no longer are in the original order.

If that's good with you for this, I think this is ready for final review.

scripts/generators/beats.py
```diff
@@ -0,0 +1,86 @@
+import yaml
+from collections import OrderedDict
+
+
+def generate(ecs_nested, ecs_version):
+    # base first
+    beats_fields = fieldset_field_array(ecs_nested['base']['fields'])
+
+    allowed_fieldset_keys = ['name', 'title', 'group', 'description', 'footnote', 'type']
```
andrewkroh: Is the `group` key used for anything in Beats? If not then we could drop this field.

webmat: Yes it comes from Beats.` group:1` fields are sorted before `group:2` fields, as far as I know.

Base is the only fieldset that has it set= to something other than 2 in ECS, and the semantics are different (base fields are not nested). So we could get rid of it eventually.

But group is present in Beats in some places, when it's not there, the default is 2.

webmat:
Fields listed under each `type:group` field set are now sorted alphabetically by name.

andrewkroh:
> Question: should we let through the ECS level?

If it's not being used then I'd drop it.

webmat:
It's a Rubyist's Python, so probably better like that 😆

webmat:
@ruflin Will ping you on the backport for review on Monday. Happy to adjust anything needed.

But I'm moving forward here to prepare the 1.0 backport here, then make https://github.com/elastic/beats/pull/11150 final, so it can hopefully be merged Monday.

PR #385, opened 2019-03-14
Define field formats that were defined in Beats but not yet in ECS.

webmat:
The 1.0.0 field definitions generated by ECS (https://github.com/elastic/beats/pull/11150)
removed a few field formats that had previously been defined on the Beats side,
but were not yet part of ECS.

This adds all of the missing these fields formats to ECS, and generates a Beats
field definition file that includes them all.


webmat:
@ruflin I don't think we should backport this to the 1.0 branch, as it wasn't part of ECS at the time we cut the 1.0.0 release. So this would officially be part of ECS as of 1.1.

I will however generate the file for Beats, in order to fix this issue for Beats 7.0.

cc @simitt

PR #393, opened 2019-03-20
Mat's review of static asciidoc files

webmat:
In this PR I'm going over the static asciidoc files, and getting things ready for the move to the main website.

Both the content of the files is being modified, and the overall structure of the documentation.
I'm trying to sequence sections in chronological order of learning/discovery of ECS. Please let me know what you think of this proposed structure. I'm also trying to address end users as much as implementors.

**New Nav Structure**

- Elastic Common Schema (ECS) Reference
- Using ECS
  - Guidelines and Best Practices
  - ECS Conventions
- ECS Field References
  - (One nested page per field set)
- Migrating to ECS
  - Solutions that Support ECS
  - Using ECS with Logstash
  - Converting a Custom Implementation
- Additional Information
  - Questions and Answers
  - Glossary of ECS Terms
  - Contributing to ECS

**TODO**

- [x] Converge on canonical ECS definition. Update the following accordingly:
  - glossary.asciidoc
  - index.asciidoc
  - using-guidelines.asciidoc
- [x] Reconcile using-guidelines.asciidoc and converting.asciidoc. Harmonize content, and question whether we need two distinct documents.
- [x] Add words to glossary


webmat:
@ruflin @karenzone @MikePaquette @MarkSettleES I'd like a preliminary look at this pass on the asciidoc for static pages.

This is my take on all everything other than the actual ECS field definition files, which was covered by #347.

I'd like your feedback on if you think this is a good direction, if you see any issues with this. I've moved things around and added a few paragraphs, but I haven't taken out anything.

docs/converting.asciidoc
```diff
@@ -12,41 +12,41 @@ Before you start a conversion, be sure that you understand the basics below.
 
 [float]
 [[core-or-ext]]
-=== Core and extended fields
+==== Core and extended fields
```
MikePaquette: "Core and extended fields" should be "Core and extended levels"

docs/converting.asciidoc
```diff
@@ -12,41 +12,41 @@ Before you start a conversion, be sure that you understand the basics below.
...
+Focus on populating these fields first.
 
-* *Extended fields.* Any field that is not a core field is called an *extended field*. 
+* *Extended fields.* Any field that is not a core field is called an *extended field*.
 Extended fields may apply to more narrow use cases, or may be more open
 to interpretation depending on the use case. Extended fields are more likely to
 change over time.
 
-Each {ecs} <<ecs-fields,field>> in a table is identified as core or extended.
+Each {ecs} <<ecs-field-reference,field>> in a table is identified as core or extended.
```
MikePaquette: add "in the Level column" at the end of this sentence.

docs/converting.asciidoc
```diff
@@ -12,41 +12,41 @@ Before you start a conversion, be sure that you understand the basics below.
...
-[[ecs-comv]]
-=== An approach to mapping an existing implementation
+[[ecs-conv]]
+==== An approach to mapping an existing implementation
 
 Here's the recommended approach for converting an existing implementation to {ecs}.
 
 . Start with Core fields.
 +
 Populate core fields first. Look at your set of event fields, and find
```
MikePaquette: Remove "Populate core fields first."  as it is redundant to the "Start with Core fields" above.

docs/converting.asciidoc
```diff
@@ -12,41 +12,41 @@ Before you start a conversion, be sure that you understand the basics below.
...
-the appropriate ECS field name for each one. 
+the appropriate ECS field name for each one.
```
MikePaquette: This section needs to be clarified/simplified here and wherever else it is used.  Suggestion:

1. Review each field in your original event and map it to the relevant ECS field.

- Start by mapping your field to the relevant ECS Core field.
- If a relevant ECS Core field does not exist, map your field to the relevant ECS Extended field.
- If no relevant ECS Extended field exists, consider keeping your field with its original details, or possibly renaming it using ECS naming guidelines and attempt to map one or more of your original event fields to it.

2. Review each ECS Core field, and attempt to populate it with data from the relevant field or fields from your original event.

docs/fields.asciidoc
```diff
@@ -1,6 +1,12 @@
 
-[[ecs-fields]]
-== {ecs} Fields
+[[ecs-field-reference]]
+== {ecs} Field Reference
+
+ECS defines multiple groups of related fields. They are called "field sets".
+The <<ecs-base,Base>> field set is the only one whose fields are defined
+at the top level of the events.
```
MikePaquette: "top level of the events."  would be clearer as  "top level of the ECS naming hierarchy."

docs/glossary.asciidoc
```diff
@@ -1,11 +1,8 @@
-//[[ecs-glossary]]
-== Glossary of {ecs} Terms
+[[ecs-glossary]]
+=== Glossary of {ecs} Terms
 
-[[glossary-ecs]] 
+[[ecs-glossary-ecs]]
 ECS::
 Elastic Common Schema. A common set of document fields, field names, and their respective entity
 relationships to be used in the storage of log messages and other data in
```
MikePaquette: "log messages and other data"  should be "event data such as logs and metrics"

docs/index.asciidoc
```diff
@@ -7,15 +7,38 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
...
-The Elastic Common Schema (ECS) defines a common set of fields for
-ingesting data into Elasticsearch. A common schema helps you correlate
-data from sources like logs and metrics or IT operations
-analytics and security analytics.
+[float]
+=== What is ECS?
+
+The Elastic Common Schema (ECS) defines a common set of fields,
+their datatype, and gives guidance on their correct usage.
+ECS is used to improve uniformity of event data ingested into Elasticsearch.
```
MikePaquette: Can we make sure this definition is the same as that used in the glossary?

docs/index.asciidoc
```diff
@@ -7,15 +7,38 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
...
+
+The ultimate goal of ECS is to help you correlate data from various sources
+like logs, metrics, IT operations analytics, and security analytics together.
+
+[float]
+=== My events don't map with ECS
 
-[[ecs-maturity]]
+ECS is a permissive schema. If your events have additional data that cannot be
+mapped to ECS, you can simply add them to your events, using non-ECS field names.
```
MikePaquette: "non-ECS" should be "custom" to be consistent with other definitions.

docs/solutions.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[ecs-solutions]]
+=== Solutions that Support ECS
+
+The following Elastic solutions support ECS out of the box:
+
+* {beats}
```
MikePaquette: Why are beats in the solution section?
Generally beats are a product or a module, but not a solution by itself.
Solutions include Site Search, App Search, Enterprise Search, SIEM app, Logging UI, etc.

docs/using-guidelines.asciidoc
```diff
@@ -14,20 +28,20 @@ practices.
 * Map as many fields as possible to ECS.
 
 [float]
-==== Guidelines for writing fields
+===== Guidelines for writing fields
```
MikePaquette: Why a distinction between "writing fields" and "naming fields"?  Would this be simpler in one list?

docs/using-guidelines.asciidoc
```diff
@@ -1,11 +1,25 @@
-//[[ecs-guidelines]]
-== Guidelines and Best Practices
+[[ecs-guidelines]]
+=== Guidelines and Best Practices
```
MikePaquette: This asciidoc seems to overlap with the converting doc.  This might be OK, but the information between the two needs to be reconciled.

docs/using-guidelines.asciidoc
```diff
@@ -14,20 +28,20 @@ practices.
...
 * No special characters except `_`
 
 [float]
-==== Guidelines for naming fields
+===== Guidelines for naming fields
 
 * *Present tense.* Use present tense unless field describes historical information.
 * *Singular or plural.* Use singular and plural names properly to reflect the field content. For example, use `requests_per_sec` rather than `request_per_sec`.
 * *General to specific.* Organise the prefixes from general to specific to allow grouping fields into objects with a prefix like `host.*`.
 * *Avoid repetition.* Avoid stuttering of words. If part of the field name is already in the prefix, do not repeat it. Example: `host.host_ip` should be `host.ip`.
```
MikePaquette: Should we delete this rule?  Seems like between our defined field sets and other naming guidelines, this will never come into effect, and in fact, we almost contradict it with `host.hostname`

docs/using-guidelines.asciidoc
```diff
@@ -36,6 +50,3 @@ ingesting to Elasticsearch using the API, your fields must be nested
 objects, not strings containing dots.
 
 * *Avoid abbreviations when possible*. A few exceptions like `ip` exist.
```
MikePaquette: This is repetitive and should be removed or made consistent by including `os` and `geo` etc.

docs/using-guidelines.asciidoc
```diff
@@ -1,11 +1,25 @@
...
+=== Guidelines and Best Practices
 
 The {ecs} schema serves best when you follow schema guidelines and best
 practices.
 
 [float]
-=== General guidelines
+==== Types of fields
+
+ECS defines "Core" and "Extended" fields.
```
MikePaquette: Again, should mention that these are "levels"  indicated in the levels column, and these should be consistent with those listed in the converting asciidoc file.

docs/converting.asciidoc
webmat: I would try to avoid talking about table structure too specifically in other documentation pages. I wouldn't want this to become obsolete (and we forget to update) if we adjust the table structure.

docs/converting.asciidoc
webmat: Really love this. I'm pushing this to the PR right away. I've used this and added a few small but important things.

I really love how much clearer the steps are becoming :-)

docs/fields.asciidoc
webmat: What about "whose fields are defined at the root of the event." ?

I would like it to be very clear at the conceptual level (it's just not nested).  I feel like this makes it clearer than talking about the ECS naming hierarchy. To someone new, it may not be obvious that the ECS hierarchy starts at the root of the document (it could be nested somewhere).

docs/glossary.asciidoc
webmat: Let's bring @karenzone's discussion about the definition of ECS here. We can use it in our glossary, and the top level glossary could then align with this one as well. So let's work on it here.

Karen, on March 18th:

> Elastic Common Schema (ECS)
>    A common set of document fields, field names, and their respective entity relationships to be used in the storage of log messages and other data in Elasticsearch.

My response, on the 19th:

> The Elastic Common Schema (ECS) defines a common set of fields, their datatype, and gives guidance on their correct usage. ECS is used to improve uniformity of event data ingested into Elasticsearch.

webmat: incorporating your feedback Mike

> The Elastic Common Schema (ECS) defines a common set of fields, their datatype, and gives guidance on their correct usage. ECS is used to improve uniformity of event data ingested into Elasticsearch, such as logs and metrics.

docs/index.asciidoc
webmat: Yes, agreed.

I'm this close to creating a YAML file that's not for the schema, but for text snippets we reuse often...

Let's iterate on the definition in the "glossary" conversation of this PR, and I'm adding a TODO to align it with index.asciidoc

docs/solutions.asciidoc
webmat: I'm using the layman term "solution". This wasn't intended to sound like Elastic solutions which have a specific meaning when talking about Elastic products.

Would the word "tool" be better for this? To me it sounded too low level / technical.

webmat: And "product" sounds a little bit commercial, given that our tools are all available as open source.

But I think I'll go with "product" for now, until I get feedback.

docs/using-guidelines.asciidoc
webmat: Just pushed out a rewrite of this section.

docs/using-guidelines.asciidoc
webmat: I've reworded the clause. The gist is: we want to avoid repetition, but it's ok to make exceptions if it would be confusing not to use the conventional name.

webmat:
I think the PR is ready for another round of review.

I've addressed a lot of @MikePaquette's feedback. His feedback also inspired me to rewrite and clarify more parts of the documentation.

I've removed links and sections that I don't think will be ready on Tuesday. We can add those back as soon as they become available.

I still see 3 todos to address for this to be ready to go, as listed in the body of this PR. Ideas welcome there :-)

karenzone:
@webmat  The doc still builds without blowing up, so that's half the battle!  
I like the user workflow-based approach. That was my original layout, but feedback was that we needed to have the fields at the beginning because that's what people cared about most. 
I'll keep digging.

docs/converting.asciidoc
MikePaquette: ++ the key is to mention that these are called levels.  which you've done.

docs/glossary.asciidoc
MikePaquette: Here is an expanded definition that came up while editing the "Overview" section. Not sure if this might be helpful with the glossary as well:

The Elastic Common Schema (ECS) is an open source specification, developed with support from the Elastic user community, that defines a common set of fields to be used when storing event data, such as logs and metrics, in Elasticsearch. 

ECS specifies field names and Elasticsearch datatypes for each field, and provides descriptions and example usage. ECS also groups fields into ECS levels, and provides a set of naming guidelines for adding custom fields. 

The goal of ECS is to enable and encourage users of Elasticsearch to normalize their event data, so that they can better analyze, visualize, and correlate the data represented in their events.  ECS has been scoped to accommodate a wide variety of events, spanning;

- Event sources: whether the source of your event is an Elastic product, a third- party product, or a custom application built by your organization.
- Ingestion architectures: whether the ingestion path for your events includes Beats processors, Logstash, Elasticsearch ingest node, all of the above, or none of the above.
- Consumers: whether consumed by API, Kibana queries, dashboards, apps, or other means.

webmat: Will update "What is ECS?" section with this

webmat: But this is too long for the glossary definitions, though.

This discussion is also happening here https://github.com/elastic/stack-docs/pull/254 for the glossary of the main docs site

webmat:
Ok, I've updated this PR in such a manner that I consider it ready for final review. I've addressed two of the todos in the body.

**Converge on canonical ECS definition**

I've updated it with my latest proposal, which I've also made in https://github.com/elastic/stack-docs/pull/254. I think it's high enough level to fit in a glossary. Being for the ECS glossary, I still kept quite a bit of details (which may get trimmed out for the main site glossary).

I've also taken Mike's proposal in this comment https://github.com/elastic/ecs/pull/393#discussion_r268217422 and made it the content of the "What is ECS?" section of the Overview. I think it's a great detailed definition.

** Reconcile using-guidelines.asciidoc and converting.asciidoc**

Both places used to mention the ECS levels. We had improved the wording in the "converting" page, which is meant to be informal pointers. I've therefore moved the improved definitions to the "guidelines" page, which is the official guideline on how we implement ECS.

**More glossary definitions**

I'm too into ECS to know if anything more is needed there. Let's add them as the need arises, and not block this PR for this.

**Other notable changes**

* I've started mentioning the ECS version directly in the text in two places. I'm aware there's a selector in the sidebar, but for a schema's documentation I think it's crucial to mention version explicitly in the text too (asciidoc can be compiled to pdf, right?). Note that in master, the version is `1.1.0-dev`, which is expected. When backporting to branch 1.0 for the release, this will become `1.0.0`.
* I've updated the "conventions" page a bit

PR #397, opened 2019-03-21
[DOCS] Adds book title

lcawl:
This PR adds a title for the book so that it matches other books in the library. 
If "Overview" isn't an appropriate replacement for the subsection, it can be changed.


webmat:
Thank you @lcawl :-)

PR #404, opened 2019-03-26
Fix version at the top of the docs.

webmat:
I knew I'd get bit by this not being generated.
I didn't know it would happen this fast :facepalm:


webmat:
@MikePaquette @ruflin Hey I'm merging this one right away. It's an important change of a single character.

`make check` is fine on my machine ;-)

PR #406, opened 2019-03-27
Fix links to reused fields at the bottom of field set detail pages.

webmat:
They didn't use asciidoc link helpers, so incorrectly pointed to to localhost:8000.


webmat:
Due to the nature of the fix, I'll push this all the way out right now

Issue #388, opened 2019-03-19
hashes ie: file hashes


neu5ron:
Would like to discuss hashes, specifically the one that quickest comes to mind is file hashes.

I found some things with [auditbeat](https://github.com/elastic/ecs/blob/54f600d76cebcfcf121e1666a2ea8f2c72c6b892/use-cases/auditbeat.md) that have md5, sha1, etc.
I am thinking that we could use hash as a nested similar to how [Geo fields](https://github.com/elastic/ecs#-geo-fields) are being used.

File hashes would then be something like:
```
file.hash.md5
file.hash.sha1
file.hash.sha256
```
Going on something like x509/SSL/TLS certificates, would then be something like:
```
tls.certificate.hash.md5
tls.certificate.hash.sha1
tls.certificate.hash.sha256
```

Going on something like processes, example [sysmon event id 1](https://docs.microsoft.com/en-us/media/landing/sysinternals/event-screen-optimized.png), would then be something like:
```
process.hash.md5
process.hash.sha1
process.hash.sha256
process.hash.imphash
```

This would then keep searching hashes across disparate log sources like:
`*.hash.md5:60b725f10c9c85c70d97880dfe8191b3`

Especially, if/when, start adding source and destination file when doing more things from Sysmon/endpoint/EDR.

Excuse me if I am missing an open issue, I performed some searches in this repo and didn't find much in the issues.

webmat:
I don't think it's addressed clearly in any issues or PR. Thanks for submitting this.

Yes, adding hash support as a reusable field set makes sense, IMO.

neu5ron:
yep, I will take a stab.
also wil add;
- the other hashes that the auditbeat link had (like sha512 and such)
- ssdeep
- ja3
- ja3s
- whatever else I can find

webmat:
Awesome, thank you!

Don't sweat any missing hashes. We can add them over time if any big ones are missing.

webmat:
@neu5ron Colleague is tackling the hashes in ECS, if you want to take a look & comment: #426

PR #426, opened 2019-04-10
Add hash.* field set

cwurm:
Adds a `hash.*` field set containing various hash algorithms and their values. It is meant to be nested under other entities, notably `file` (file hashes) and `process` (hash of the process executable).

These fields are already used by the Auditbeat FIM (`file_integrity`) module (see [docs](https://www.elastic.co/guide/en/beats/auditbeat/6.7/exported-fields-file_integrity.html#_hash_fields)), and in https://github.com/elastic/beats/pull/11722 I'm proposing to add them to the Auditbeat System process dataset as well.

*Note: The Auditbeat FIM module puts these in a top-level `hash.*` object. I personally think it makes more sense to have it nested under the entity that is hashed (there might even be multiple hashes in one event), which is why I marked it as `reusable.top_level: false`. File hashes should then go under `file.hash.*`.*

Resolves https://github.com/elastic/ecs/issues/388.


schemas/hash.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: hash
+  title: Group
+  group: 2
+  type: group
+  short: Hashes, usually file hashes.
+  description: >
+    The hash fields represent different hash algorithms and their values.
```
andrewkroh: I recommend establishing an expected format of the hashes such as lowercase hex.

andrewkroh:
How would the file hash be different from the process hash? Are they both hashes of files?

webmat:
The subtext of my question about FIM vs network monitoring agent is: do we need to support additional meta-data per hashing result?

```
{ "hash":
  { "md5":
    { "value": "deadbeef",
      "encoding": "latin1" }}}
```

cwurm:
> How would the file hash be different from the process hash? Are they both hashes of files?

My idea was that process hashes would be the hashes of their executables (e.g. `/usr/bin/bash`) - so yes, they'd be file hashes.

To make the expectations clear, it would be nice if we could have separate documentation for `file.hash` and `process.hash`, but I don't think that's currently possible with a reused object like this?

> The subtext of my question about FIM vs network monitoring agent is: do we need to support additional meta-data per hashing result?
> 
> ```
> { "hash":
>   { "md5":
>     { "value": "deadbeef",
>       "encoding": "latin1" }}}
> ```

I would think it's unlikely for hashes to be represented in different encodings. Hexadecimal is the only one I've ever seen, so from my perspective, it would be reasonable to expect that. Happy to add a note to the documentation as @andrewkroh suggested.

webmat:
Ah I should have been a bit clearer by what I meant here. I didn't mean `encoding` for how the hash is displayed, but rather when the file being transferred (e.g. keeping the webserver download), can it change the encoding?

I'm pretty sure it wouldn't for binary transfers, but perhaps this can happen for text files (think code files: js, php, etc).

webmat:
Another thing I'm wondering is whether we already need to make `hash` reusable.

Is there a case where one would have both a `file.hash.xx` and a `process.hash.xx` in the same event?

webmat:
@MikePaquette I'd like your take on a few things here, when you have a moment:

**Official hashes**

We want to trim down the list of hashes in this proposal, and tell people how to add additional hashes. The goal is we don't want to maintain a complete list of hashes (aka add new hashes every week). What's your take on which ones we should include?

The idea would also be to document the hash field set such that if people use a different hash, they just nest it as well under `hash`, with proper instructions on field naming (lowercase, underscores).

**Reusable**

The current proposal is to introduce hash as reusable at `file.hash.*` and `process.hash.*` (which is just a hash of the `process.executable` file). The proposal also specifies that `hash` is not expected at the top level.

I'm questioning this, and wondering whether we actually expect both `file.hash` and `process.hash` to be populated at the same time in one event. If not, I think we could introduce `hash` not as a reusable field set, but just at the top level.

Events would look like:

```JSON
{ "process": { "executable": "/usr/bin/bash" },
  "hash": { "md5": "..." }
}
```

and

```JSON
{ "file": { "path": "/usr/bin/bash" },
  "hash": { "md5": "..." }
}
```

cwurm:
> Is there a case where one would have both a `file.hash.xx` and a `process.hash.xx` in the same event?

I think so. An example would be a process executable hash and a file hash representing either one or more files specified on the command line or a process reading/writing file(s) while it's running. I'm not sure if anything is generating this kind of data at the moment, but it's something I've thought about.

Just in general, there can be multiple files in an event and in fact we already have this situation in Auditbeat with `auditd.paths`.

MikePaquette:
@webmat 
> We want to trim down the list of hashes in this proposal, and tell people how to add additional hashes. The goal is we don't want to maintain a complete list of hashes (aka add new hashes every week). What's your take on which ones we should include?

> The idea would also be to document the hash field set such that if people use a different hash, they just nest it as well under hash, with proper instructions on field naming (lowercase, underscores).

Since these are all "extended" level fields, and beats implementations that use them already exist, I am not opposed to including the entire set in the ECS spec.  I think this will result in more compatibility than if we encourage users to define their own custom `hash.*` fields.

> The current proposal is to introduce hash as reusable at file.hash.* and process.hash.* (which is just a hash of the process.executable file). The proposal also specifies that hash is not expected at the top level.
> 
> I'm questioning this, and wondering whether we actually expect both file.hash and process.hash to be populated at the same time in one event. If not, I think we could introduce hash not as a reusable field set, but just at the top level.

From a search perspective, I can see both positions here.  Allowing `hash.*` to be reusable does allow some granularity when searching, for example `file.hash:"deadf00d"` and you could still use leading wildcard `*.hash:"deadf00d"` to perform a search across all uses of hash.  However, I think you could get similar granularity by including the `event.category` or some other event categorization field in your search, so I'm not sure the re-usability is required.

I think the multiple hashes in one event is the case that could tip the odds one way or the other. How are we doing to handle that case?  Is using the `hash.*` fields as arrays sufficient?  What about `related.hash.*` ?

webmat:
Heads up on the leading wildcard queries. They are purely a Kibana trick, not a core Elasticserch feature. When querying e.g. `*.hash.md5:deaf00d`, Kibana turns this into an `OR` query that explicitly lists all field names that match `*.hash`. But `related.hash` will be the "real" solution to this anyway, so that's no big deal.

With that said, I think I've been convinced that we need the reusability. If an event describes a file interaction by a process, and a hash for each is calculated, it should not be ambiguous which hash matches which entity.

Finally, I think we still need the `hash` field set to document how to add additional hashes correctly, because the current list is missing a bunch more hashes that have already been mentioned in issue https://github.com/elastic/ecs/issues/388. The point I'm making is **not** that we should go add these ones now. Because even with those, the list is still not complete :-)

So with that said, I do think there would be value in trimming down the list somewhat, because the list is quite long, and I wonder how many of these hashes have been added for completeness' sake. Or are all of these shades of SHA hashes actually being used by many threat indicator sources / detection systems?

If they are all legitimately used, I could go with adding them all. But I'd need to be convinced :-)

webmat:
@cwurm When you fix the changelog conflict, please keep your line above the line for PR 439, to reduce chances of further conflicts, if we merge other things before this. It will also keep the entries loosely ordered by PR.

cwurm:
Ok, so my takeaway is three things to be done:

1. Trim list down to most common hashes.
2. Document how to customize with more hash fields.
3. Fix changelog conflict but maintain order in the changelog.

I'll make those changes.

cwurm:
> * Trim list down to most common hashes.

I've trimmed down the list of hashes to MD5, SHA1, SHA256, and SHA512.

> * Document how to customize with more hash fields.

Added `Field names for common hashes (e.g. MD5, SHA1) are predefined. Add fields for other hashes by lowercasing the hash algorithm and using underscore separators as appropriate (snake case).`

@webmat @andrewkroh @MikePaquette Let me know if this looks good now.

schemas/hash.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    Field names for common hashes (e.g. MD5, SHA1) are predefined. Add fields for other hashes
+    by lowercasing the hash algorithm and using underscore separators as appropriate (snake case).
```
webmat: Small clarification: "the hash algorithm name". I'd also give an example at the end of the sentence, something like "for example `sha3_512`"

cwurm:
Thanks @webmat, I fixed the description.

@MikePaquette you requested changes to this, does this look good now?

PR #434, opened 2019-04-17
Backport #430 to 1.0: Docs: Fix a table start to align with the stop (#430)

webmat:
Backport of PR #430 to 1.0 branch. Original message:

Asciidoctor is a bit pickier about start and end delimiter and complains
about this table because the start line has more `=`s than the end line.

This removes `=`s from the table start, so it lines up with the table end.


webmat:
Seeing the email notif I was like "Yay someone else with my sense of humor!" 🤣

Issue #129, opened 2018-09-28
proposal for Event fields


jcallaha:
I would like to propose two new _event_ fields that I think would be useful 
* event.level (keyword)
* event.sequence (long)

_event.level_ is the human readable companion field to _event.severity_, essentially identical in use to _log.level_ (which could be deprecated/removed in favor of _event.level_).

_event.sequence_ is another integrity field (like _event.hash_) for event sources that emit events in sequence, to detect missing events and/or order events that may have time stamps based on receipt time rather than emit time (syslog over UDP comes to mind). 

While I'm on the topic of the Event group, is _event.id_ meant to be a unique id for for an instance of an event (such that two identical events would have different and unique event.id) or is it meant to handle the case where events have numeric or coded identifiers, e.g.:
`%ASA-3-610001: NTP daemon interface interface_name : Packet denied from IP_address`
might look something like the following in ECS 1.0
```
{
  ...
  "message": "000045: %ASA-3-610001: NTP daemon interface interface_name : Packet denied from IP_address",
  "event": {
    "action": "denied",
    "category": "firewall",
    "id": "610001",
    "severity": 3,
    "level": "ERROR",
    "sequence": 45,
    "version": "1.0.0"
  }
  ...
}
```

ruflin:
For `event.level`: Do you see some other `level` types except the common log ones like `ERROR`, `WARN`, etc.?

jcallaha:
@ruflin off the top of my head I can think of the following commonly used "levels" (human readable severity levels):
* syslog/RFC3164/RFC5424: `Emergency`, `Alert`, `Critical`, ` Error`, `Warning`, `Notice`, `Informational`, `Debug`
* log4j: `FATAL`, `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`
* Windows Event Log: `Critical`, `Error`, `Warning`, `Informational`, `Verbose`
* CVSS Scoring: `Critical`, `High`, `Medium`, `Low`

ruflin:
Seems like the first 3 are all log related. Last one is more about scoring. I wonder if `level` would still be the right field here?

One interesting point that shows the above: The naming is very similar but lower / upper case is different which mill make correlation tricky. I think epecially between log4j, syslog there should be correlation. Should we standardise the naming?

willemdh:
Imho there should be an `event.level` too. This should be the 'normalised' / 'standardized' version of `log.level` (which should imho only hold the original level)

[...]

PR #439, opened 2019-04-23
Additions to event

### LINKED ISSUE #321: New field event.provider
I have multiple log types with field named like "source" which is some component/provider of the log. If possible I use non-ECS event.logger field but in fact it is not usable here (logger can be some part of application like class/module itself). I noticed in https://github.com/elastic/beats/pull/10333/files it is named winlog.provider_name but it should be more general and part of ECS because Windows Event log is not the only log type which uses it. So I propose to add "event.provider" field.

webmat:
This PR addresses/improves a few things in the the "event" field set.

If any of these turns out controversial / complicated, I'll be happy to extract them to a separate PR. The goal here is to get the simple things in.

- Added `event.code` (See https://github.com/elastic/beats/pull/10333)
- Added `event.sequence` (See https://github.com/elastic/ecs/issues/129, https://github.com/elastic/beats/pull/10760)
- Added `event.provider` (See https://github.com/elastic/ecs/issues/321)
  - Note: Beats modules currently put the Syslog "programname" in `process.name` which is sometimes accurate, sometimes not (e.g. "kernel"). event.provider would be a better field for this.
- Explain event.module and event.dataset without mentioning Beats

Closes #321
Addresses part of #129

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -13,6 +13,7 @@
 * Generator for the asciidoc rendering of field definitions. #347
 * Generator for the Beats fields.ecs.yml file. #379
 * Added field formats to all `.bytes` fields and `event.duration`. #385
+* Added `event.code`, `event.sequence` and `event.provider`. #439
 
 ### Improvements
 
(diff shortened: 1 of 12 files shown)


webmat:
@ph Proposal to add `event.sequence` to ECS is in here :-)

webmat:
Thanks guys! Waiting for Mike's input as well on this one, since it's on `event`. ETA Wednesday :-)

PR #441, opened 2019-04-24
Additions to "file"

### LINKED ISSUE #322: add file.name
ECS has a lot of great file fields but one that's very important but seems to be missing is file.name.

Thanks.

webmat:
This PR addresses/improves a few things in the the "file" field set.

If any of these turns out controversial / complicated, I'll be happy to extract them to a separate PR. The goal here is to get the simple things in.

- Added `file.name` (See #322)
- Added `file.directory`
- Added a lot of examples
- Adjusted a few field definitions

Closes #322


neu5ron:
BOOM!

schemas/file.yml
```diff
@@ -25,67 +38,92 @@
...
-    description: File size in bytes (field is only added when `type` is
-        `file`).
+    short: File size in bytes.
+    description: >
+      File size in bytes.
+
+      Only relevant when `file.type` is "file".
+    example: 16384
+
+  - name: crtime
```
ruflin: This is pretty cryptic for ECS standard I would say? What about `created`?

schemas/file.yml
```diff
@@ -25,67 +38,92 @@
...
     description: File type (file, dir, or symlink).
+    example: file
 
   - name: device
     level: extended
     type: keyword
     description: Device that is the source of the file.
+    example: sda
+
+  - name: system
```
ruflin: I'm wondering if this belongs here. `file` is information about a specific file. I see how it could fit it in but stumbled over it.

The other part I would not call it `system` as I expect us to have a system object in the future related to system metrics. `system` already exists today in Metricbeat. Alternative names?

schemas/file.yml
```diff
@@ -25,67 +38,92 @@
...
 
   - name: ctime
     level: extended
     type: date
-    description: Last time file metadata changed.
+    short: Last time the file attributes changed.
+    description: >
+      Last time the file attributes or metadata changed.
+
+  - name: atime
```
ruflin: I guess these are the "standard" names that is why we use them here?

schemas/file.yml
cwurm: `file.filesystem`?

schemas/file.yml
cwurm: It could also be `creationtime` to keep in line with the `*time` naming schema we have going on.

The `stat` output calls it `Birth` (though always empty on Linux since it does not expose the creation time) - but I don't think that's a good name.

It's maybe unfortunate we already have `ctime` and `mtime` which are UNIX names that Windows does not have. `ctime` could be confused as the creation time. If we ever want to make a change we could go for more descriptive names, e.g. `created|modified|changed|accessed`.

webmat: Yes, I was specifically pondering this, as I was working on this PR.

The `file` field set didn't get much attention before 1.0.0, unfortunately. But I'm not a fan of the existing field names `ctime` and `mtime`, because they go against one of the principles we're aiming for with ECS. We try to avoid contractions and acronyms, unless the convention is overwhelmingly in favour of them (e.g. `ip`, `pid`).

Seeing that we already had both of these fields, I went with the other 2 standard field names. I went for internal consistency in `file`. But this is getting on my nerves even more, as we discuss this 😂 

As @cwurm points out, `ctime` vs `crtime` are confusing. Any article about Linux filesystem timestamps (debugfs, stat) mention this common confusion.

So I'm very open to revisiting these field names. It's too late for now to change `ctime` and `mtime`, although we could create 4 new fields with less ambiguous names and mark the first two as deprecated for the life of ECS 1.x, and remove them in 2.0.

Proposed names:

- `file.created`: creation time
- `file.accessed`: last access time
- `file.modified` (deprecates `file.mtime`): last modification time for the file content
- `file.metadata_modified` (deprecates `file.ctime`): last modification time for the file metadata.

I'm not a big fan of this last one. Ideas welcome.

For now we can push this naming decision to a bit later, focus on the two new fields, and make a note to deprecate mtime/ctime.

schemas/file.yml
webmat: I'm pretty strongly in favour of adding this field even though it's not information directly about a given file. It's still useful metadata about the place where a file lives. I wanted to point to `host` as very general metadata about an event, but `host` is a field set, not a field inside one specific field set.

Perhaps we could do the same with filesystem information, and introduce a `volume` field set?

Volume fields:

- name
- filesystem
- size (bytes)
- used (bytes)
- letter (Windows drive letter)
- kind (ssd, spinning rust)
- type (platform specific types, such as AWS' gp2, io1 etc)
- id
- serial_number
...

webmat:
For now I'll take out `file.system` and the timestamps, so we can move forward with this PR.

I'll open a PR right away for the timestamps with clearer names.

I'll make a note for the `volume` field set. We haven't had a pressing need for it yet, so it may take a bit longer before I get around to adding it as a field set. Although the more I think about it, the more I think it makes sense.

webmat:
Ok, I've taken out the new timestamps as well as `file.system`. Only non-controversial changes are left here, IMO.

Please have a quick look, to see if this PR is ready.

New, more readable timestamps proposed in #445, and #444 opens the discussion for `volume` (including `volume.filesystem`).

schemas/file.yml
```diff
@@ -25,67 +38,74 @@
...
     level: extended
     type: date
-    description: Last time file content was modified.
+    description: Last time the file content was modified.
 
   - name: ctime
     level: extended
     type: date
-    description: Last time file metadata changed.
+    description: Last time the file attributes or metadata changed.
```
cwurm: On Linux and macOS, `ctime` will also be updated when the file contents change (together with `mtime`).

webmat: I've pushed a documentation tweak about this. I find it a bit clumsy, and I'm not 100% convinced this belongs in ECS.

In a sense this helps explain the relationship between two ECS fields.

On the other hand, I feel like we're getting too close to documenting OS quirks & the like, which can be different from one platform to the next.

But in this case I'm ok with this addition. I'll merge tomorrow, unless there's disagreement.

webmat: @cwurm Any other objections about this PR? Does [this documentation](https://github.com/elastic/ecs/pull/441/files?file-filters%5B%5D=.yml#diff-776162149979f3e81d54d5bf9b843027R115) update work for you?

It's not the best prose, but we can always tweak improve it later if needed :-)

cwurm: > @cwurm Any other objections about this PR? Does [this documentation](https://github.com/elastic/ecs/pull/441/files?file-filters%5B%5D=.yml#diff-776162149979f3e81d54d5bf9b843027R115) update work for you?

Yes, LGTM

> On the other hand, I feel like we're getting too close to documenting OS quirks & the like, which can be different from one platform to the next.

True, though I'm not sure we can avoid it entirely. Only UNIX has atime/mtime/ctime/crtime, [Windows has only](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfiletime) lpCreationTime/lpLastAccessTime/lpLastWriteTime (and lpLastAccessTime is not tracked by default).

So for Windows then, is `lpLastWriteTime == mtime` (`modified` in a future version?), with `ctime` empty (later `changed`?)? Or is `lpLastWriteTime == mtime == ctime`? The Auditbeat FIM module must be implementing this in some way, but I don't know how off the top of my head.

webmat:
@ruflin Agreed. I was just trying to streamline my time, before the month of May, where I have 3/4 weeks **not** on ECS :-)

PR #450, opened 2019-05-07
Create new cloud field to hold account alias

ansoni:
As I was upgrading our proprietary schema to ECS, I could not find a proper place for an AWS Account Alias - https://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html#AboutAccountAlias

This is an important field for us because AWS Account IDs are not great for humans.  Since this is user-supplied and volatile, I kept the AWS "alias" jargon instead of "name".

### DIFF
--- a/schemas/cloud.yml
+++ b/schemas/cloud.yml
@@ -68,3 +68,13 @@
 
         Examples: AWS account id, Google Cloud ORG Id, or other unique
         identifier.
+    - name: account.alias
+      level: extended
+      type: keyword
+      example: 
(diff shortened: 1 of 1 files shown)


ruflin:
@ansoni Do you know if something similar also exists in GCP, Azure or other providers?

SHolzhauer:
> @ansoni Do you know if something similar also exists in GCP, Azure or other providers?
I believe GCP has "projects", don't known about Azure though.

ebeahan:
`cloud.account.name` introduced in https://github.com/elastic/ecs/pull/816

PR #454, opened 2019-05-09
Format port numbers and numeric IDs as strings

cwurm:
It has been bugging me for some time that in Kibana things like port numbers and PIDs are displayed using a thousand separator, e.g. `process.pid: 20,123`.

This PR changes the display format to `string` where appropriate (I went through all numeric fields).

Changed fields are:
1. client.port
2. destination.port
3. event.severity
4. event.sequence
5. http.response.status_code
6. process.pid
7. process.ppid
8. process.pgid
9. process.thread.id
10. server.port
11. source.port
12. url.port


webmat:
This has been bugging me for a long time as well! Thanks for submitting this.

I'll try to review and merge next week :-)

cwurm:
> Could you add a changelog entry.

Done.

@webmat @ruflin Let me know if this looks good.

webmat:
I'm all for rendering more straightforward numbers for things that should be easy to copy / paste around, such as PIDs, and the other ones you're improving here. This was much needed 👍

But I have one worry: does formatting numbers as a string have an impact on what we can do with the field in visualizations? Because we can also use a numeral.js format for `number` fields, such as `'0[.]000'` to render integers with no noise, and optional decimals.

I'm not a huge fan of having Kibana implementation details in ECS, for obvious reasons 😆 But https://github.com/elastic/ecs/pull/425 is how I ended up mitigating an issue with `event.duration`, during the Beats migration. So it's totally possible to dictate a bit more of the format all the way from here.

The fact that it's a really Kibana-specific implementation detail I think is acceptable for now. If it becomes a problem, we can always add a layer of indirection in the future 😉

cwurm:
> But I have one worry: does formatting numbers as a string have an impact on what we can do with the field in visualizations? Because we can also use a numeral.js format for `number` fields, such as `'0[.]000'` to render integers with no noise, and optional decimals. See below.

Formats do not affect how fields can be queried or aggregated on. It's still possible to create a histogram on any of these fields, or to filter for low port numbers in Discover (e.g. `destination.port < 1000`).

Randy-312:
Have you talked to the SIEM folks?
Being able to do some of the aggregations and queries on port numbers seems extremely useful in that realm.  I'd hate to see you limit what can be done with ECS in the Security realm.

cwurm:
@Randy-312 I'm on the SIEM team :) Changing the display format does not affect the type of queries or aggregations you can run on these fields.

PR #458, opened 2019-05-17
Add object_type to fields

codebrain:
`base.yml` contains object_type, but this is missing from the documentation.
https://github.com/elastic/ecs/blob/652f7250f9c3acfec9538b00ad60053b7042ba20/schemas/base.yml#L39

### DIFF
--- a/schemas/README.md
+++ b/schemas/README.md
@@ -35,6 +35,7 @@ Supported keys to describe fields
 - multi\_fields (optional):
 - reusable (optional):
 - index (optional): If `False`, means field is not indexed (overrides type)
+- object_type
 
 ### Multi\_fields
 


schemas/README.md
```diff
@@ -35,6 +35,7 @@ Supported keys to describe fields
 - multi\_fields (optional):
 - reusable (optional):
 - index (optional): If `False`, means field is not indexed (overrides type)
+- object_type
```
ruflin: Do you want also to add a line on what is used for? In summary it defines a dynamic template for all the fields inside this object with the above field type.

webmat: Thanks for this, @codebrain!

Yeah I wrote up this file as I was playing around, to help my faulty memory. It's by no means exhaustive. So good catch :-)

webmat:
This is a pull request to branch 1.0. Closing.

PR #459, opened 2019-05-17
Update log.yml

codebrain:
YAML file is missing the `group` parameter as defined as required by the documentation, default value set to 2.


webmat:
Thanks @codebrain!

PR #461, opened 2019-05-20
Fix exmple type to example for fields.yml and process.yml

codebrain:
Fix typo, should read as `example`

### DIFF
--- a/fields.yml
+++ b/fields.yml
@@ -1363,7 +1363,7 @@
           type: long
           description: >
             Process id.
-          exmple: ssh
+          example: ssh
     
         - name: name
           level: extended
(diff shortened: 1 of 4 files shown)


codebrain:
@ruflin - its proving to be a little difficult to run this in Windows :)

codebrain:
Looks like I might be able to workaround this by editing the generated files directly?

ruflin:
@codebrain Sorry about that Windows issue.We should probably fix that. If you give me permission on your branch, I can run it for you.

webmat:
@codebrain I'd love your help on making this more easy to run on Windows at some point. What are the failures you're seeing?

Oh, is it just because of `make`? I see Beats partially solves this with mage...

Any suggestions welcome on replacing make :-) Could be mage, could be another task runner in Python.

webmat:
I just noticed this is a PR to the 1.0 branch. I'll reapply your contributions on top of master, and then backport to 1.0.

Need to keep the workflow uniform :-)

PR #462, opened 2019-05-20
Update url.yml

codebrain:
YAML file is missing the group parameter as defined as required by the documentation, default value set to 2.

### DIFF
--- a/schemas/url.yml
+++ b/schemas/url.yml
@@ -1,6 +1,7 @@
 ---
 - name: url
   title: URL
+  group: 2
   short: Fields that let you store URLs in various forms.
   description: >
     URL fields provide support for complete or partial URLs, and supports the


codebrain:
@webmat - any chance you could have a look at this and merge if OK? Its the only element throwing off my YAML validation...

webmat:
@codebrain Done! Sorry about the delay :-)

codebrain:
@webmat - can this be ported to `1.1` branch too?

webmat:
The files in this directory should not be used directly by anything other than ECS scripts.

If you want a fully fleshed out equivalent of the schema files (e.g. with all the defaults filled in, etc) you should instead use [generated/ecs/ecs_nested.yml](https://github.com/elastic/ecs/blob/v1.1.0/generated/ecs/ecs_nested.yml), which is the same nested YAML format.

Unrelated but worth mentioning: if you need the fully fleshed out ECS fields but would rather not walk a nested tree, you have an equivalent file called [generated/ecs/ecs_flat.yml](https://github.com/elastic/ecs/blob/v1.1.0/generated/ecs/ecs_flat.yml), which is keyed on the full dotted field names, so you can directly loop over everything.

If you're under time pressure and would like me to just backport anyway once, just this time, I'm more than happy to do it :-) But make sure to switch over to one of these other two files ASAP after that.

Do you still need the backport, or does one of the other two files work?

PR #464, opened 2019-05-22
Fix exmple typo in process.yml

webmat:
This closes #461, which was submitted by @codebrain, who use Windows --
ECS tooling doesn't currently run on Windows. Thanks Stuart!


codebrain:
Thanks for your help!

PR #472, opened 2019-05-24
Finalize 1.0.1 changelog

webmat:


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file based on the
...
 ## [1.0.1](https://github.com/elastic/ecs/compare/v1.0.0...v1.0.1)
 
+### Bugfixes
+
+* Fix typo in "process" fieldset definition that prevented the example value for
+  `process.pid` from being picked displayed. #464, #470
+
 ### Added
 
 * Add generated source code for Go. #249
```
ruflin: I wonder if in the future most of these entries should go into something like "developer changelog" because it does not affect the schema. Would be good to have a place where only changes about the schema are noted.

webmat: Well the example now shows up in the doc. I agree it's not a change to the schema per se, though.

Actually, perhaps the changelog entry should be more straightforward, however:

> Add example for `process.pid` and `process.ppid`

ruflin: My comment was more general and not only for this specific line.

PR #477, opened 2019-05-31
(#475) Add support for server and process uptime fields

### LINKED ISSUE #475: Support for server.uptime and process.uptime
Is is worth it adding the milliseconds a server has been up? It's a value coming directly from the OS (both *NIX and Windows) using the `uptime` command.

For process it already exists a `process.start` which could define how to calculate the uptime.

I.e., as described here: (https://superuser.com/a/464413) in linux/mac, it's possible to gather this info:
```
$ ps -eo pid,comm,lstart,etime,time,args
  PID COMM             STARTED                          ELAPSED      TIME ARGS
    1 /sbin/launchd    Sun May 19 10:10:53 2019     09-04:18:37  38:47.10 /sbin/launchd
   39 /usr/sbin/syslog Sun May 19 10:11:01 2019     09-04:18:29   0:18.02 /usr/sbin/syslogd
   40 /usr/libexec/Use Sun May 19 10:11:01 2019     09-04:18:29   0:34.19 /usr/libexec/UserEventAgent (System)
   43 /System/Library/ Sun May 19 10:11:01 2019     09-04:18:29   0:14.74 /System/Library/PrivateFrameworks/Uninstall.framework/Resources/
   44 /usr/libexec/kex Sun May 19 10:11:01 2019     09-04:18:29   0:22.23 /usr/libexec/kextd
```

But again, please close this issue if you do not consider it :)

Thanks!

mdelapenya:
Closes #475

## What does this PR do?
This PR adds fields for `process.uptime` and `server.uptime`. As suggested by @webmat, the uptime field will be measured in **seconds**.

### DIFF
--- a/schemas/process.yml
+++ b/schemas/process.yml
@@ -87,6 +87,13 @@
       description: >
         The time the process started.
 
+    - name: uptime
+      level: extended
+      type: long
+      example: 1325
+      description: >
+        Seconds the process has been up.
+
     - name: working_directory
       level: extended
       type: keyword
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -6,6 +6,7 @@
 
 ### Added
 
+* Added `process.uptime` and `host.uptime` fields. #475
 * Added field formats to all `.bytes` fields and `event.duration`. #385, #425
 * Added `hash.*` field set. #426
 * Added `event.code`, `event.sequence` and `event.provider`. #439
--- a/code/go/ecs/host.go
+++ b/code/go/ecs/host.go
@@ -53,6 +53,9 @@ type Host struct {
 	// meaningful in your environment.
 	Type string `ecs:"type"`
 
+	// Seconds the host has been up.
+	Uptime int64 `ecs:"uptime"`
+
 	// Operating system architecture.
 	Architecture string `ecs:"architecture"`
 }
--- a/code/go/ecs/process.go
+++ b/code/go/ecs/process.go
@@ -60,6 +60,9 @@ type Process struct {
 	// The time the process started.
 	Start time.Time `ecs:"start"`
 
+	// Seconds the process has been up.
+	Uptime int64 `ecs:"uptime"`
+
 	// The working directory of the process.
 	WorkingDirectory string `ecs:"working_directory"`
 }
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -1531,6 +1531,17 @@ type: keyword
 
 // ===============================================================
 
+| host.uptime
+| Seconds the host has been up.
+
+type: long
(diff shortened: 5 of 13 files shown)


elasticcla:
Hi @mdelapenya, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecs/pull/477.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

mdelapenya:
Hey @webmat, I updated the PR to include unit suffix as first-class citizen, so the new fields are
`process.uptime.seconds` and `server.uptime.seconds`

cc/ @axw

webmat:
@mdelapenya Thanks for updating the PR. My suggestion was to make it seconds, but I should have been more explicit: the unit shouldn't be part of the field name. We should have the description state that it's seconds.

In the context of the ecs-metrics experimental branch, we're even considering making this directive official, see https://github.com/elastic/ecs/pull/481

mdelapenya:
Oh my! I understood totally the opposite from the thread :) Also reading [how prometheus folks define metrics](https://prometheus.io/docs/practices/naming/#metric-names) made me think that way.

Anyway, I'll revert back to the original, where units were explicitly described in fields comments/descriptions

Thanks!

mdelapenya:
PR updated with your suggestions, thanks!

schemas/server.yml
```diff
@@ -74,3 +74,10 @@
       example: 12
       description: >
         Packets sent from the server to the client.
+
+    - name: uptime
```
webmat: Sorry I hadn't noticed this. The field set `server` is meant to descrive the server side of a network connection. We have two pairs: client/server and source/destination. As such, this isn't the right place for this.

`server.uptime` should be moved to `host.uptime` :-)

Let's also make `host.uptime` extended for now.

schemas/process.yml
```diff
@@ -87,6 +87,13 @@
       description: >
         The time the process started.
 
+    - name: uptime
+      level: core
```
webmat: Please make this `extended` for now.

schemas/server.yml
mdelapenya: Is `extended` a matter of adding a `level: extended`, or is there any other consideration?

mdelapenya:
Changes ready!

mdelapenya:
Thanks @webmat!! What about the docs? Is this something I have to handle too?

ruflin:
@mdelapenya Docs were generated automatically.

webmat:
Yes, generated every half hour. But note that master != current on the docs. So you can see your contribution if you go to the docs and select "master" in the version selector.

PR #486, opened 2019-06-21
Added domain field to user

thomaspatzke:


schemas/user.yml
```diff
@@ -56,3 +56,10 @@
 
         Useful if `user.id` or `user.name` contain confidential information and
         cannot be used.
+
+    - name: domain
+      level: extended
+      type: keyword
+      example: domainname
```
webmat: I'm not a fan of this example, as it doesn't add any information, or doesn't jog the user's memory in giving a domain name that people may recognize.

I would either remove the example, or perhaps replace with something else that's commonly seen as an LDAP or AD domain name. Not sure what that would be, though. "corp", "intranet", "acme"?

schemas/user.yml
```diff
@@ -56,3 +56,10 @@
...
+      description: >
+        Name of user domain, e.g. Active Directory or LDAP domain name.
```
webmat: Can you adjust the description to the following, please? Note that the two paragraphs will give us a "short" description automatically, here.

"Name of the directory the user is a member of.

For example, an LDAP or Active Directory domain name."

schemas/user.yml
thomaspatzke: Removed this, I think it should be obvious for users of this field.

schemas/user.yml
thomaspatzke: Changed it. Update PR in few seconds.

thomaspatzke:
Implemented changes as proposed.

webmat:
The field description in all the files is based on what's in `schemas/*.yml`. Please run `make generate` to re-generate all other files. That's why the build is breaking :-)

thomaspatzke:
Sorry, forgot this! Just pushed the commited build.

webmat:
@thomaspatzke Thanks for your contribution :-)

webmat:
And of course I thought "oh wait, what about the Changelog" one millisecond after merging 😂

PR #487, opened 2019-06-21
Optional dockerized build

thomaspatzke:


webmat:
Love the idea as well! Thanks for submitting this, @thomaspatzke.

I don't have time to dig into this, but the docker run command fails for me. Here's the end of the output:

```
...
go: downloading go.uber.org/multierr v1.1.0
build/ve/bin/python scripts/use-cases.py --stdout=true >> /dev/null
/bin/sh: 1: build/ve/bin/python: Exec format error
Makefile:79: recipe for target 'legacy_use_cases' failed
make: *** [legacy_use_cases] Error 2
```

I assume the Docker image is trying to use my workstation's venv and its binaries (and I'm on a Mac)? The docker image should instead have its own Python (2.7 for now).

Also note that in building this, I think it may make more sense to start from a Python base image, then add Go on top, rather than the reverse. I'd like to get rid of Go in this repo, eventually.

thomaspatzke:
> I assume the Docker image is trying to use my workstation's venv and its binaries (and I'm on a Mac)? 

The container uses the Python provided with the Golang image. Likely, the Golang container image lacks portability for the contained Python or it's something completely different. Wouldn't be the first such issue I've observed with MacOS.

> Also  note that in building this, I think it may make more sense to start  from a Python base image, then add Go on top, rather than the reverse.  I'd like to get rid of Go in this repo, eventually.

My main reason was the issue to get the appropriate Go version on my system. It would be more complicated to get the appropriate Go version into a Python container than the other way around, especially because the required Python version is already there. Dropping the Go dependency would likely eliminate my requirement for the Docker container.

djptek:
Closing as at this point we are not planning to add docker to our build toolset support schedule, thanks for your contribution

Issue #446, opened 2019-04-29
Documentation: list of all available fields.


Forbo:
It would be nice to have an index or table of all currently available fields in the documentation.

webmat:
If you mean a single documentation page with all fields, yes I agree with you. This is on the short list. It will not replace the current structure of the documentation (a page per field set), but will just be an additional page.

If you mean something else than the above, perhaps you can look in this repo for other generated files. Particularly https://github.com/elastic/ecs/blob/master/generated/csv/fields.csv

Forbo:
That's exactly what I meant, glad to hear I'm not the only one wanting it. Is there an active issue or branch I could follow?

webmat:
There's no PR or branch yet

ruflin:
+1 on having all fields on one page back ;-)

PR #488, opened 2019-06-25
Implement field list all on one page

ruflin:
I personally prefer to have all field definitions on one page to have it Cmd-F friendly and make browsing between fields efficient. This implements this mostly.

Screenshot below on how it looks at the moment:

Screenshot 2019-06-25 at 16 08 02

### DIFF
--- a/scripts/generators/asciidoc_fields.py
+++ b/scripts/generators/asciidoc_fields.py
@@ -183,6 +183,9 @@ def index_row():
 
 def index_footer():
     return '''
+
+=== Fields
+
 include::field-details.asciidoc[]
 '''
 
@@ -194,6 +197,7 @@ include::field-details.asciidoc[]
 
 def field_details_table_header():
     return '''
+[float]
 [[ecs-{fieldset_name}]]
 === {fieldset_title} Fields
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -1,4 +1,5 @@
 
+[float]
 [[ecs-base]]
 === Base Fields
 
@@ -72,6 +73,7 @@ example: `["production", "env2"]`
 
 |=====
 
+[float]
 [[ecs-agent]]
 === Agent Fields
 
@@ -154,6 +156,7 @@ example: `6.0.0-rc2`
 
 |=====
 
+[float]
 [[ecs-client]]
 === Client Fields
 
@@ -283,6 +286,7 @@ type: long
 
 |=====
 
+[float]
 [[ecs-cloud]]
 === Cloud Fields
 
@@ -377,6 +381,7 @@ example: `us-east-1`
 
 |=====
 
+[float]
 [[ecs-container]]
 === Container Fields
 
@@ -460,6 +465,7 @@ example: `docker`
 
 |=====
 
+[float]
 [[ecs-destination]]
 === Destination Fields
 
@@ -587,6 +593,7 @@ type: long
 
 |=====
 
+[float]
 [[ecs-ecs]]
 === ECS Fields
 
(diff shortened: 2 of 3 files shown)


scripts/generators/asciidoc_fields.py
```diff
@@ -183,6 +183,9 @@ def index_row():
 
 def index_footer():
     return '''
+
```
ruflin: @lcawl I wanted to add a `[float]` here to but get errors about unexpected title levels if I do this. My goal was to have the overview table also on the same page as all the fields.

webmat:
This is great! Works well on a single page already.

I think we'll need to adjust the heading size. Right now a field set and all headings inside it are all at h3, which makes it hard to understand the hierarchy.

The way I initially envisioned it was always to keep the "page per field set" layout, however. And we'd have an "all in one" page listed at the bottom of the list.

I think the page per field set layout is important in helping people focus one section at a time, initially. Then the "all in one" page comes in when you just need to move around quickly, presumably later on.

What do you think?

scripts/generators/asciidoc_fields.py
webmat: > My goal was to have the overview table also on the same page as all the fields.

Great idea!

ruflin:
I don't think we should duplicate data as it will be confusing to have the same in 2 places. As normally 1 field set fills one page well, I don't worry that users get confused having it all in one page. I even think it's beneficial that users accidentially can discover other fields. A typical one I jump for and back in between is source and destination. 

+1 on adjusting the titles.

webmat:
Only that one big list 🤔 TBH up to now, I hadn't considered **only** having one single page. I always assumed the page per field set remained, and the all in one page was for advanced users.

Note that I see the field details table becoming more complex over time, with support for enums, deprecations, additional type or data formatting details (e.g. array of X vs single X). If each of these more complex table has their own page, I think the complexity won't be overwhelming; then having that second "all in one" page isn't that big an issue, even if it's extremely busy.

But if the only representation of the fields is this one massive page, that grows in complexity on top of that, I think this will make for a poor user experience, especially initially.

@karenzone What do you think about having only one page for all the fields, and no longer having one page per field set?

webmat:
I'm curious to see how the sidebar will look after adjusting the heading hierarchy. Hopefully of we downgrade the subsections to h4, they'll disappear from there?

Screenshot 2019-06-26 09 20 45

karenzone:
@webmat The Details level heading should go away entirely (exp: Agent Field Details). The table is nested under "Agent Fields" and one assumes that you get more details as you read deeper into a topic.  That additional heading is not adding value. 

Cmd-F is awesome when you know what you're looking for.  The page-per-field-set (ppfs) approach is better for navigation and for showing field sets in relation to other field sets.

Asciidoctor gives us new capabilities with tagged regions.  What about keeping ppfs as the default presentation, and offering a "Wanna see everything on one page?" alternate presentation?

ruflin:
I think what we have today is not a good user experience. I may be an expert user, but when I look at ECS I never think of one field set alone. I always jump forth and back between definitions and it's really painful right now, not even talking about being on mobile yet.

I think the part I'm missing is on what the benefit of having one field set per page is. What is our target user for this and what task does he try to do? Not sure I got the part about "showing field sets in relation to other field sets".

I would keep everything we have with the top level navigation and navigation on the right side but instead of links having ankers.

ruflin:
One more note on this: I would go so far, that our fields should almost be the landing page of the ECS docs. That is why we had it on the top in our initial README. Because users come once to learn about ECS and then next 25 times to search for a field or learn about one. So I would definitively optimise for the second use case.

MikePaquette:
Adding link to issue:  https://github.com/elastic/ecs/issues/446

ruflin:
Closing this as since last year, I assume many things have changed and currently don't get to update this.

PR #490, opened 2019-06-27
Add missing changelog for #486

webmat:
Add missing changelog for #486

cc @thomas ;-)


thomaspatzke:
The commit history is the changelog ;-)

PR #491, opened 2019-06-28
added nat.ip and nat.port to source and destination fields

dainperkins:
To better support Port/Address translation in flow/fw type network events added source.nat.ip, source.nat.port, destination.nat.ip, destination.nat.port.

If, in the case of e.g. 1:1 nat, the observer doesn't report nat ports, the original port should be copied to the nat.port field for ease of searching all fields of the connection (e.g. searching nat ips/ports for threat hunting etc)

To - do: 
If there is telemetry provided across devices implementing NAT (e.g. internally load balanced source nat sessions being reported across multiple devices before and after S/NAT) we may want to also include:
  - nat.session_id
     - copy native session id at NAT gateway to stitch together pre/post NAT sessions

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -13,6 +13,8 @@
 * Added `file.name` and `file.directory`. #441
 * Added `file.created`, and `file.accessed`. #445
 * Added `domain` field to user. #486
+* Added `.nat.ip` and `.nat.port` to `source`, `destination`, `client` and `server`. #491 
+
 
 ### Improvements
 
--- a/code/go/ecs/client.go
+++ b/code/go/ecs/client.go
@@ -58,4 +58,14 @@ type Client struct {
 
 	// Packets sent from the client to the server.
 	Packets int64 `ecs:"packets"`
+
+	// Translated IP of source based NAT sessions (e.g. internal client to
+	// internet).
+	// Typically connections traversing load balancers, firewalls, or routers.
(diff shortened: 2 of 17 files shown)


webmat:
@dainperkins Thanks for the PR.

At this time, the network pairs are source/destination as well as client/server. We keep them in sync. Could you add the fields to client/server too, please?

Now let's clarify the descriptions a tiny bit. The `nat.ip` and `nat.port` are meant to be the private values, correct? Could you state this more explicitly in the description?

After these adjustments, please run `make` to generate all the other files, and commit all of the generated files as well. The build will fail if generated files are not updated and committed according to your changes.

dainperkins:
@webmat 
I specifically avoided specifying public/private as theres plenty of cases where all IPs will be private or public (nat and natural), but will better scope the field definitions and copy to client/server

webmat:
Ah you're right. Perhaps use wording such as "the internal side of..."

webmat:
@dainperkins One last little thing and then we merge.

Could you add the following entry to the file `CHANGELOG.next.md`, please?

```
* Added `.nat.ip` and `.nat.port` to `source`, `destination`, `client` and `server`. #491 
```

dainperkins:
added, & made (and pushed to my nat-ips fork - do I have to pull again?)

webmat:
Great! Thanks @dainperkins

neu5ron:
awesome work guys!

PR #494, opened 2019-07-02
Add ability to include external directory of schemas when running generator

mcpate:
This adds the ability to pass a path to `make` (which is used by the `generator` target) to a directory where users might have their own schemas defined. 

The use case this is solving for is when a user has custom schemas that are required due to being unable to map fields directly to ECS Core/Extended fields. This change allows them to run the following to have their schemas included in the output of `make` or `make generate`:
```
$> make INCLUDE=/path/to/dir/of/yml/files
or 
$> make generator INCLUDE=/path/to/dir/of/yml/files
```

If `INCLUDE` does not exist, everything works as per usual. If `INCLUDE` does not contain any yml files everything works as per usual. If `INCLUDE` contains yml files the keys/values are merged with those found in the `schemas` directory and the output is for both sets of values. If the user tries to include a yml file that will overwrite a key from the `schemas` directory, an exception is raised.

As mentioned above, the `generator` target was the focus here. Should this PR add this functionality to any other targets/scripts?

### DIFF
--- a/scripts/tests/test_ecs_helpers.py
+++ b/scripts/tests/test_ecs_helpers.py
@@ -66,11 +66,11 @@ class TestECSHelpers(unittest.TestCase):
         result = ecs_helpers.safe_merge_dicts(a, b)
 
         self.assertEquals(result,
-            {
-                'cloud': {'group': 2, 'name': 'cloud'},
-                'agent': {'group': 2, 'name': 'agent'},
-                'base': {'group': 1, 'name': 'base'}
-            })
+                          {
+                              'cloud': {'group': 2, 'name': 'cloud'},
+                              'agent': {'group': 2, 'name': 'agent'},
+                              'base': {'group': 1, 'name': 'base'}
+                          })
 
     def test_merge_dicts_raises_if_duplicate_key_added(self):
         a = {'cloud': {'group': 2, 'name': 'cloud'}}
--- a/scripts/tests/test_ecs_helpers.py
+++ b/scripts/tests/test_ecs_helpers.py
@@ -56,6 +56,29 @@ class TestECSHelpers(unittest.TestCase):
         result = ecs_helpers.dict_sorted_by_keys(dict, ['group', 'name'])
         self.assertEqual(result, expected)
 
+    def test_merge_dicts(self):
+        a = {
+            'cloud': {'group': 2, 'name': 'cloud'},
+            'agent': {'group': 2, 'name': 'agent'},
+        }
+        b = {'base': {'group': 1, 'name': 'base'}}
+
+        result = ecs_helpers.safe_merge_dicts(a, b)
+
+        self.assertEquals(result,
+            {
+                'cloud': {'group': 2, 'name': 'cloud'},
+                'agent': {'group': 2, 'name': 'agent'},
+                'base': {'group': 1, 'name': 'base'}
+            })
+
+    def test_merge_dicts_raises_if_duplicate_key_added(self):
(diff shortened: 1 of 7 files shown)


webmat:
Thanks a lot for submitting this, @mcpate! This will be very useful. I can't review in depth at this time, but I can still answer your question:

The `make generator` target is for the "new" generator, the other targets generating files are to be considered legacy, mostly. So you don't need to worry about adding support elsewhere for now. If we determine we need to add support for this elsewhere, it'll be addressed as separate PRs.

scripts/tests/test_ecs_spec.py
```diff
@@ -1,13 +1,14 @@
...
+import glob
 import unittest
 
 sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
 
 from scripts import schema_reader
 
 
-(nested, flat) = schema_reader.load_ecs()
+(nested, flat) = schema_reader.load_ecs(sorted(glob.glob('../../schemas/*.yml')))
```
webmat: Hey @mcpate, the tests are still failing because the glob path is relative to where the script is invoked from, not from the script file itself.

What I'll propose here is not perfect, but I can clean up the code around this later.

For now, if you just apply the following change, the tests will pass and I'll merge.

```suggestion
(nested, flat) = schema_reader.load_ecs(sorted(glob.glob('schemas/*.yml')))
```

scripts/tests/test_ecs_helpers.py
```diff
@@ -56,6 +56,29 @@ def test_sorted_by_multiple_keys(self):
...
+        b = {'base': {'group': 1, 'name': 'base'}}
+
+        result = ecs_helpers.safe_merge_dicts(a, b)
+
+        self.assertEquals(result,
+            {
+                'cloud': {'group': 2, 'name': 'cloud'},
+                'agent': {'group': 2, 'name': 'agent'},
+                'base': {'group': 1, 'name': 'base'}
+            })
```
webmat: Ah sorry! Seems like Python's autopep8 (code formatting) isn't happy with the nesting here.

If you run `make check` locally, autopep8 will modify the file in place, to be formatted correctly. Looks like it's aligning the opening curly with the function's opening parens.

mcpate: Nice catch! Sorry about that! I think I've got it fixed.

nick-george:
Hi @mcpate, 

Thanks very much for writing this and apologies for commenting on an already merged PR. Happy to take this discussion somewhere else if you like. 

In my opinion, this feature would be more useful if it allowed for adding fields to schemas that already exist in the ECS. For example, add a new field to the 'source' object. 

I have taken code from https://stackoverflow.com/questions/20656135/python-deep-merge-dictionary-data to make your feature do what I need. See below. Obviously this could be shortened somewhat, but I wanted to leave the stackoverflow answer unmodified. 

I believe merging two ECS dictionaries is safe if they contain the same key, and I also believe that the user should be able to override any field within the ECS too. 

Thanks for your time!
Nick

```
def deep_merge(source, destination):
    for key, value in source.items():
        if isinstance(value, dict):
            # get node or create one
            node = destination.setdefault(key, {})
            deep_merge(value, node)
        else:
            destination[key] = value

    return destination

def safe_merge_dicts(a, b):
    """Merges two dictionaries into one."""
    c = deep_merge(a,b)
    return c
```

webmat:
@nick-george Sorry, forgot to respond. A PR that would make it possible to do that is always welcome.

If you look at https://github.com/elastic/ecs/issues/497 you'll see it's actually something we'd like to flesh out, as time permits :-)

PR #496, opened 2019-07-02
Update schema.json on 'make'

andrewkroh:
The schema.json was not being rebuilt when `make` was executed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,11 @@ reload_docs: generator docs
 schema.json: ve
 	$(PYTHON) scripts/schemas.py
 
+# Download and setup tooling dependencies.
+.PHONY: setup
+setup: ve
+	cd scripts && $(FORCE_GO_MODULES) go mod download
+
 # Build an Elasticsearch index template.
 .PHONY: template
 template:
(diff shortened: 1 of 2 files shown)


webmat:
@andrew-goldstein @XavierM Hey guys, did you say this file was no longer in use?

If that's the case, I'd rather drop it completely. This is still based on the old file generation code, which doesn't include the nested fields.

Makefile
```diff
@@ -94,10 +94,10 @@ misspell:
 .PHONY: reload_docs
 reload_docs: generator docs
 
-# Download and setup tooling dependencies.
-.PHONY: setup
-setup: ve
-	cd scripts && $(FORCE_GO_MODULES) go mod download
```
webmat: Why are you deleting the `setup` target?

andrewkroh: I think cut instead of copied. Fixed now.

webmat:
Thanks for the fix!

Will probably delete it soon, as it's no longer being used by the SIEM team.

But I'll merge anyway for now. Perhaps we'll move it to legacy before deleting. I'm debating whether we should keep this legacy stuff. As Nic pointed out in the past, legacy is in the Github history too.

Issue #497, opened 2019-07-05
Make ECS tooling friendly to generating custom templates based on ECS


webmat:
A frequent request about ECS is: "How can I generate Elasticsearch templates that include ECS fields and my custom fields?"

There's many ways we could provide tooling for that: a Kibana app would probably be the most user-friendly. However the tooling in the ECS repository is very close to being able to offer this already. As a short term solution, we will be tweaking the generator in this repo to make it easier for users to manage their templates with it. As a bonus, the generator will continue generating all of the artifacts that includes people's fields (asciidoc, csv), not only the ES templates. Some users may find that interesting.

Here's a list of tasks that will gradually make it easier to achieve this goal:

- [x] #494 Add the ability to include external directory of schemas
- [x] #748 Add the ability to specify an output directory
- [x] #746 Document how to use the ECS tooling to generate custom templates
- [ ] Let users select which artifacts they want to generate
- [x] #737 Let users specify which ECS schemas to include, to skip the fields they will not use
- [ ] Forcefully set `level: custom` for any field loaded from an external directory
- [x] Make the index template settings configurable
- [x] Allow for adding fields within existing ECS field sets from external files

These will help address #95 and #324

Randy-312:
We were going to do this with ordering for elasticsearch templates.
Have our base fields in a 001, and all of the various version specific fields in 002.
I've not thought of all the problems in that approach.. what are they?  _i.e, if we change OUR fields and dont' want to use those everywhere._

webmat:
Please move your question to discuss.elastic.co, and I'll answer it there. This issue is meant to discuss the plan I've outlined above.

pmuellr:
I'm starting to look into adopting ECS into my life, and this sort of tooling is of interest to me.  Just starting, haven't written any code for this yet, so ... can't speak with any real credence.  We'd either be taking base ECS and adding more bits to it, or starting with our own ECS-compatible-ish base, and adding relevant ECS bits on top of that.

My initial thoughts were to write some of my own tooling in node, that would just read one of the generated schema files, and then operate on that and my custom definitions.  I'm actually quite comfortable doing that, vs using an existing tool - it's just JSON algebra in the end, and it wouldn't surprise me to find I need a bit more than some generic tooling might provide.  And I guess I've done this so many other times over the years :-)

[...]

PR #499, opened 2019-07-09
A few tweaks on top of recent work to include external dirs

webmat:
- Keep the ECS file glob encapsulated, to avoid relative path mistakes
- Rename variables to better represent that they're no longer necessarily ECS per se
- tweak description of the CLI argument


webmat:
Follow-up to #494. Related to #497

PR #502, opened 2019-07-19
Clarify `service.id` description.

webmat:
This removes an example that went contrary to the purpose of a `service.id`.
If someone needs to focus on the events emitted by one host of a
service, they should filter on `host.id` or `host.name`.

`service.id` is meant to identify a service as a whole.


webmat:
I think this clarifies it, WDYT @axw?

axw:
Thanks @webmat, that's 100% clear now.

PR #512, opened 2019-08-01
Fix title and subtitle from Hash Fields section

fabianonunes:


fabianonunes:
Ok. I will send a new PR.

PR #518, opened 2019-08-13
Risk Fields

dainperkins:
Added risk fields for scoring and categorizing atomic users / groups / hosts / cloud / containers etc.


schemas/risk.yml
```diff
@@ -1,40 +1,52 @@
...
       - file
-      - process
+      - threat
+      - vulnerbility
+
 
   fields:
 
-    - name: md5
+    - name: label
```
MikePaquette: @dainperkins Is there another name we could use instead of "label" ?  As you know, one ECS base field is "labels" (plural), but I am worried about creating confusion, so I wonder if we could use something like `.name` to  be consistent with other field sets?

schemas/risk.yml
```diff
@@ -1,40 +1,52 @@
...
       level: extended
       type: keyword
-      description: MD5 hash.
+      description: locally relevant label to describe the risk type of asset or data identified either
```
MikePaquette: Can we enhance this definition to make it clearer to those transforming data what they should put here?  What is the relationship of this field value to the `risk.score`?  Do we expect this to be an array, holding multiple values?

schemas/risk.yml
```diff
@@ -1,40 +1,52 @@
...
-      description: SHA256 hash.
+      type: long
+      description: comparative risk score quantitavely scoring the level of risk of the asset or data, 
+        or risk represented by a given threat (typically provided by )
 
-    - name: sha512
-      level: extended
-      type: keyword
-      description: SHA512 hash.
+      example: 1-10
```
MikePaquette: Is this really just an example, or should we define this must contain a 0-10 value?   At the least, we should recommend a 0-10 scoring system.

dainperkins:
I think my gitFoo has failed me...

the hashes shouldn't be there...  will verify and update the label field to
something else

dainperkins:
>
> Question: With this change (adding risk.* fields), are you suggesting
> that we deprecate current ECS fields event.risk_score and/or
> event.risk_score_norm ?
>

While i hadn't previously considered it, I think I am suggesting
deprecating risk_score / risk_score norm in favor of a single reusable risk
field set that can represent both atomic entity risk factors (user, group,
asset, threat, ioc lookups, etc.) as well as the normalized risk of an
event (based on atomic risk scores of available entities , or signal (based
on normalized atomic *event* risk scores)

That being said, it does seem like a non-normalized representation (e.g. PA
uses low, medium, high, very high, critical or something iirc, firepower
something similar) and we should have a spot for risk scores as reported

add risk.original or risk.reported to the risk fields?

or

leave event.risk_score as the original risk label (quantitative or
qualitative) for searching on risks from a specific device standpoint (e.g.
show me everything PA labeled a high risk)



/d

schemas/risk.yml
dainperkins: updated to risk.tag for clarity

schemas/risk.yml
dainperkins: It might help to consider how we would apply asset / resource risk levels to a normalized scoring system to prioritize signals by business risk - tho I suppose we could allow for multiples.

dainperkins:
@MikePaquette  How about risk.scope (thinking of PCI in scope / out of scope from an Audit perspective instead of tag/label

I think pulling risk fields out of nesting with event makes more sense - that way theres a specific place for event risk (e.g. palo says this is a high risk alert) as well as normalized_risk for business level sorting of signals (which combines signal risk with some d&d style modifier based on assets, users, zones, etc...

ebeahan:
@dainperkins We're going to close this out for now as it's been opened for a while without activity. If you want to move forward, let's open a stage 0 RFC proposal to capture these ideas.

PR #520, opened 2019-08-13
Fix title and subtitle from Hash Fields section

fabianonunes:


webmat:

Issue #281, opened 2018-12-17
Logging libraries for ECS


ruflin:
Most ECS data is fetched from log files. To make generation of ECS based logs easier, ECS could provide logging libraries or tools on top of the existing logging libraries for the different programming languages. This should make generation of ECS based logs much easier. 

This is partially related to https://github.com/elastic/ecs/issues/280 as it could be used as a foundation for the logging libraries.

tsg:
+1. Somewhat related, I think we should also standardise on some sort of Elastic or ECS protocol, which can be as simple as JSON over TCP/UDP, with the twist that the JSON respects ECS. This means we could have a Filebeat/Logstash input for this protocol, and then any 3rd parties that want to integrate with our stack can implement this.

ruflin:
@tsg I like the idea, should we also open a separate issue for it to not mix discussions?

webmat:
I see some merit in looking into this. I'm curious what form this would take concretely, in the usage of such a library. When doing structured logging, there's usually still a free form message (key `message`), then there's the meta-data, which falls in two categories. First is the metadata collected automatically by the library, which could benefit from being named according to ECS (e.g. PID). But I think in general, most of the metadata is specified upon emission of the log event, and is therefore named by the developer themselves.

How did you see the featureset offered by these logging libraries? Especially wrt the ad hoc key naming.

If we decide to implement this, I call dibs on implementing the Ruby library :trollface:

The one thing we need to steer away from is actually shipping the logs. An application should never have to manage (or suffer from) problems in the logging pipeline. It shouldn't do the queueing, the retries, etc. Hopefully we're just talking about logging libraries that format a text logfile or a syslog payload.

urso:
This library combines a many concepts and separates between ECS and custom user defined fields. 
It does not define the output or shipping part, but provides building blocks for serializing logs to json, ubjson, cbor, plain text, `map[string]interface{}` (yes, one can capture the log in an object and post-process). Capturing meta-data is optional for all logging backends.

It is based on a many experiments with a many other go libraries and some [Serilog](https://serilog.net/). The type safe part is based on generated code from the fields.yml files. Users can define their own type safe namespaces and even generate a template.json file if required. Everything is meant to be compatible to ECS.

I found the key difference between structured loggers is how the context is handled in all go libraries. I was hoping to do a meta-logger that can operate with other structured loggers, but unfortunately this didn't work out well. Yet I didn't find a library doing exactly as I wanted, so I did roll my own.

[...]

Issue #501, opened 2019-07-18
What field to use for Java / Jee logs


disaster37:
Hi,

you think that I use the right field for Java / Jee logs ?

Sample of logs
```
2019-07-18 09:43:41,689 ERROR [main] org.springframework.boot.SpringApplication (SpringApplication.java:837) - Application run failed
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'directoryInboundGatewayFlow' defined in class path resource [fr/hm/fwk/batch/directory/config/DefaultDirectoryBatchIntegrationConfiguration.class]: Initialization of bean failed; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'directoryInboundGatewayFlow.org.springframework.integration.config.SourcePollingChannelAdapterFactoryBean#0.source': Invocation of init method failed; nested exception is java.lang.IllegalArgumentException: Source directory [/mnt/CMN0B00_ACDC-DEMAT] does not exist.
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:581)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:495)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:317)
	at org.springframework.beans.factory.support.AbstractBeanFactory.dt_access$1246(AbstractBeanFactory.java)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:222)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:315)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:759)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:867)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:548)
	at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:140)
[...]

PR #521, opened 2019-08-14
Add log.logger

felixbarny:
As discussed in https://github.com/elastic/ecs/issues/281#issuecomment-520567252, `event.dataset` is not a good place to store the name of the logger. Let's create a more suitable field for that.

### DIFF
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -35,3 +35,11 @@
 
         This field is not indexed and doc_values are disabled so it can't be
         queried but the value can be retrieved from `_source`.
+
+    - name: logger
+      level: core
+      type: keyword
+      example: org.elasticsearch.bootstrap.Bootstrap
+      short: Name of the logger.
+      description: >
+        The name of the logger, usually the name of the class which initialized the logger.
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -42,4 +42,4 @@
       example: org.elasticsearch.bootstrap.Bootstrap
       short: Name of the logger.
       description: >
-        The name of the logger, usually the name of the class which initialized the logger.
+        The name of the logger inside an application. This is usually the name of the class which initialized the logger, or can be a custom name. 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -13,8 +13,9 @@
 * Added `file.name` and `file.directory`. #441
 * Added `file.created`, and `file.accessed`. #445
 * Added `domain` field to user. #486
-* Added `.nat.ip` and `.nat.port` to `source`, `destination`, `client` and `server`. #491 
+* Added `.nat.ip` and `.nat.port` to `source`, `destination`, `client` and `server`. #491
 * Added `as` fields for Autonomous System information (i.e. ASN). #341
+* Added `process.thread.name` field. #517
 * Added `log.logger` field. #521
 
 ### Improvements
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -15,6 +15,7 @@
 * Added `domain` field to user. #486
 * Added `.nat.ip` and `.nat.port` to `source`, `destination`, `client` and `server`. #491 
 * Added `as` fields for Autonomous System information (i.e. ASN). #341
+* Added `log.logger` field. #521
 
 ### Improvements
 
--- a/code/go/ecs/event.go
+++ b/code/go/ecs/event.go
@@ -90,12 +90,6 @@ type Event struct {
 	// module name, followed by a dot, then the dataset name.
 	Dataset string `ecs:"dataset"`
 
-	// Name of the component.
-	// Similar to dataset but more fine-grained. It is used, for example, to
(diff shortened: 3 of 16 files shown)


ruflin:
This reminds me off https://github.com/elastic/beats/blob/master/filebeat/module/elasticsearch/_meta/fields.yml#L10 which we use in the Elasticsearch logs. Do you see this as specific to Java or more general? Perhaps `java.component`?

felixbarny:
I'm not entirely sure whether logging implementations in other languages tend to have name for a logger which is usually scoped to one class. @elastic/apm-agent-devs do you know?

But in general, it's not specific to Java. Even if it turns out only Java loggers tend to use it I think it makes sense to have this as a general field.

ruflin:
If we tie it to Logging but not Java, what about `log.component`?

felixbarny:
that would also work for me

ruflin:
Currently +1 on `log.component` then, but probably miss something ;-)

SergeyKleyman:
[Logger-per-class pattern](https://stackoverflow.com/questions/3143929/why-do-loggers-recommend-using-a-logger-per-class) is definitely not something specific to Java. But I don't know what are the prevalent logging approaches in languages without class concept.

ruflin:
`logger` also exists in the Beats (Golang) code. And we also call it `logger`. Should we perhaps even call it `log.logger`?

felixbarny:
simple, clean and straight to the point. +1
I'll change the PR accordingly. But also curious what the other reviewers have to say 🙂

beniwohli:
Another +1 for `log.logger`, that's what it is called in Python logging.

webmat:
I can also confirm that the logging class/component can be a thing in Ruby as well.

I wonder if we shouldn't take a step back and think about this more holistically, though. Even in normal logs, you can have the function/method name provided by some loggers (e.g. Rails' `MyController#my_method`). And once we're at it, it's also possible to have line numbers and file paths in logs (exceptions and otherwise). [This blog post](https://www.scalyr.com/blog/rails-logger/) shows default Rails log samples, which contain all of these.

However, the more we dig into these details, the less it makes sense to cram all this in `event.*` or `log.*`, IMO.

Should we look into putting together a more complete application logging / debugging field set? This field set could encompass the source code details as I'm discussing here, as well as optionally exception details, if the event is about an exception.

Let's use `debug.*` as an example field set name. Could be something else, but I think this could work well for these two cases:

```
- debug.component
- debug.method (or function?)
- debug.file.path (optionally a place where someone can nest file.*, if they care?)
- debug.file.line (or directly debug.line)
- debug.exception.stack_trace
- debug.exception.class
- debug.exception.fields (for exceptions that take params)
- debug.exception.exception... (nested)
```

I'm excluding `message` in here because I think the exception message should be in the base field `message`.

I'm sure there's a few more attributes we could add, here. This is just a quick first draft.

These ideas are related to #154 and #501.

felixbarny:
Good point but I think linking to source code is a separate beast and we already have an issue for that: #154.

The `log.logger` is in addition to, but not the same as the class name. Although they are often the same it's important to not use them interchangeably. It's often much more expensive to compute the actual class/file name (by traversing the stack). The `log.logger` can be a user-supplied string value, like `audit_logs` or `foobarbaz`.

schemas/log.yml
```diff
@@ -35,3 +35,11 @@
...
         This field is not indexed and doc_values are disabled so it can't be
         queried but the value can be retrieved from `_source`.
+
+    - name: logger
+      level: core
+      type: keyword
+      example: org.elasticsearch.bootstrap.Bootstrap
+      short: Name of the logger.
+      description: >
+        The name of the logger, usually the name of the class which initialized the logger.
```
webmat: I would flesh the description out a little more, to try to clarify what scenarios this is meant to be used in. Could be something like this:

```suggestion
        The name of the logger inside an application. This is usually the name of the class which initialized the logger, or can be a custom name. 
```

webmat:
Note that you'll have to pull these changes, run `make` and push the re-generated files as well :-)

felixbarny:
> Did you want to wait for feedback from more people from the APM side?

I just wanted to confirm that `log.logger` is not just a Java thing which there is agreement that it's not. So I don't think we need to wait for more APM feedback (although still appreciated ofc).

Maybe wait another week to see if there are any objections and merge if there are? How do you normally handle that in the ECS repo? I just added a bunch of people as reviewers who might have a stake in this.

webmat:
I'm fine with waiting a little more.

For merging, my heuristic is: I merge community PRs. Elasticians are free to merge, or ask me to merge, whichever they prefer. I'll be more proactive in taking over if there's backports needed (e.g. for small clarifications to fields that were in 1.0), as I don't want the burden of understanding the intricacies of the repo's branching & management to be pushed to all Elastic contributors.

ruflin:
Let's get it in by end of day if no objects. We always have time up to the next release to change our opinion, but I prefer not to have pending PR's.

PR #526, opened 2019-08-22
Move 1.1 release branch to 1.1.0

webmat:
This removes the `-dev` label from the 1.1.0 version


webmat:
This version number will only live in the 1.1 branch, and will not be forward-ported to master.

PR #531, opened 2019-08-24
Added parent domain field to source, destination and url.

mbudge:
The parent domain field is the domain without any sub-domain. The domain field is a exact-match keyword field, which means it is not possible to search for all connections to a domain when a sub-domain is involved. The parent domain field will allow users to store normalized domains using the public suffix list.

For example, the registered domain for "foo.malware.com" is "malware.com".

This value can be determined precisely with a list like the public suffix list (http://publicsuffix.org). Trying to approximate this by simply taking the last two labels will not work well for TLDs such as "co.uk". If the parent domain normalization process fails, users should store with original domain with the sub-domain in the parent-domain field. Punny-code domains are more likely to fail when using common TLD extract libraries which using the public suffix list to get the parent domain, hence using a best-effort approach means users can still search one field to more accurately find network connections. 

This is important in SIEM and log management functions, as users need to be able to find all logs when they are searching for a known bad IOC domain. Users could index domains into an extra text field in their schema, but this is slow and expensive when searching many TB's of data in Elasticsearch.

### DIFF
--- a/code/go/ecs/destination.go
+++ b/code/go/ecs/destination.go
@@ -42,6 +42,14 @@ type Destination struct {
 	// Destination domain.
 	Domain string `ecs:"domain"`
 
+	// Destination parent domain domain stripped of any subdomain.
+	// For example, the registered domain for "foo.google.com" is "google.com".
+	// This value can be determined precisely with a list like the public
+	// suffix list (http://publicsuffix.org). Trying to approximate this by
+	// simply taking the last two labels will not work well for TLDs such as
+	// "co.uk".
+	DomainParent string `ecs:"domain.parent"`
+
 	// Bytes sent from the destination to the source.
 	Bytes int64 `ecs:"bytes"`
 
--- a/code/go/ecs/source.go
+++ b/code/go/ecs/source.go
@@ -42,6 +42,14 @@ type Source struct {
 	// Source domain.
 	Domain string `ecs:"domain"`
 
+	// Source parent domain domain stripped of any subdomain.
+	// For example, the registered domain for "foo.google.com" is "google.com".
+	// This value can be determined precisely with a list like the public
+	// suffix list (http://publicsuffix.org). Trying to approximate this by
+	// simply taking the last two labels will not work well for TLDs such as
+	// "co.uk".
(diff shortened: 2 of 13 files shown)


mbudge:
I made a feature request to add a TLD extract filter to Logstash. 

https://github.com/elastic/logstash/issues/11079

MikePaquette:
@mbudge Thanks for creating this PR.  I think there is wide agreement in the community for adding a field to capture the parent/higher level/registered domain.  However, there is not agreement about what to call it :-).  Please see Issue #84 for a further discussion.  There is some support (https://github.com/elastic/ecs/issues/84#issuecomment-432346994) for creating a field `*.registered_domain` to capture what you are looking for. 

I have three suggestions:

- Let's finish the conversation in #84 about what to name this field.
- Let's not use the naming you suggested `domain.parent`, as this would cause `domain` to become an Elasticsearch object, something ECS tries to avoid for fields that are already defined, such as `*.domain`.  (i.e. let's not use the "." character after "domain", regardless of what name we agree upon.)
- If/when we do add this field, let's add it in all the places that `*.domain` is used today in ECS 1.1, which includes: 

1. `client.domain` 
2. `client.user.domain`
3. `destination.domain`
4. `destination.user.domain`
5. `host.user.domain`
6. `server.domain`
7. `server.user.domain`
8. `source.domain`
9. `source.user.domain`
10. `url.domain`
11. `user.domain`

MikePaquette:
@mbudge A related question please:  Would you also propose to add related ECS fields for the subdomain portion of the domain?  (i.e. the part left over after extracting the parent/higher level/registered part)

Previous sentiment on adding such fields has been mixed, with some votes against it (https://github.com/elastic/ecs/issues/84#issuecomment-434334603) but would love your thoughts.

mbudge:
It's true, parent child traverses the DNS hierarchy. 

Submitted a new pull request to add registered_domain. 

https://github.com/elastic/ecs/pull/533

mbudge:
This can be rejected in favor of registered_domain.

https://github.com/elastic/ecs/pull/533

webmat:
Indeed, closing in favor of #533. Thanks @mbudge for making the change!

Issue #515, opened 2019-08-08
Define ECS set of fields for (runtime) dependencies


simitt:
Defined fields for software dependencies in ECS. Auditbeat already collects information around installed packages and can be used as starting point:

[auditbeat system/package](https://www.elastic.co/guide/en/beats/auditbeat/current/exported-fields-system.html#_package):
 `system.audit.package.entity_id`
`system.audit.package.name`
`system.audit.package.version`
`system.audit.package.release`
`system.audit.package.arch`
`system.audit.package.license`
`system.audit.package.installtime`
`system.audit.package.size`
`system.audit.package.summary`
`system.audit.package.url`

---
Update: PRs implementing `package` specs:
* https://github.com/elastic/ecs/pull/532 main PR covering agreed on `package` fields
* https://github.com/elastic/ecs/pull/585 `package.url`
* https://github.com/elastic/ecs/pull/586 `package.extended_version`
* https://github.com/elastic/ecs/pull/587 `package.type`

Qard:
In the case of the Node.js APM agent:
- `name`, `version`, `license` , `summary` should all be easy to collect
- _Sometimes_ we'll have something useful for `url`
- `size` would be rather difficult/expensive to get

Everything else is not really relevant to Node.js, in my opinion.

webmat:
Good idea, thanks @simitt!

If the APM team needs this, a pull request would be really appreciated. Packages aren't on the top of the priority list over here, but we're definitely able to help hammer out a PR and get this merged in, if someone is willing to do the legwork :-)

@Qard That's all good. The idea is to define fields that make sense when looking at the broad picture. Then we fill as many as reasonable, depending on package type. If the size isn't readily available, it would be skipped.

Speaking of which, we may want to have a field `package.type` to track which kind of package this is: `system`, `node`, `ruby`, etc...

PR #532, opened 2019-08-27
ECS software packages and runtime dependencies

simitt:
Initial draft for getting the process started to define ECS fields for installed software packages / runtime dependencies. 

adressing #515


schemas/user.yml
```diff
@@ -18,6 +18,7 @@
       - host
       - server
       - source
+      - package
```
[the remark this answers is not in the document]
simitt: Yes

schemas/package.yml
```diff
@@ -0,0 +1,83 @@
...
+  group: 2
+  short: These fields contain information about an installed software package.
+  description: >
+    These fields contain information about an installed software package. It contains general information about a
+    package, such as name, version or size. It also contains installation details, such as time, URL or
+    user who installed the package.
+  type: group
+  fields:
+
+    - name: artifact_id
```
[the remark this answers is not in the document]
simitt: Agree to use `package.id`. 

It is meant to be usable for [entity_id](https://www.elastic.co/guide/en/beats/auditbeat/current/exported-fields-system.html#_package). IMO it should be uniquer per package + host it is installed. So reinstalling the same package on the same host would result in the same ID. Which information exactly is used for that depends on the collector. 
I suggest to at least use `package.name` + `package.versions` + `host.name`. We should probably align in how to collect it between beats and APM agents (or server if it is computed by the server).

schemas/package.yml
```diff
@@ -0,0 +1,83 @@
...
+      example: Open source programming language to build simple/reliable/efficient software.
+
+    - name: size
+      format: string
+      level: core
+      type: long
+      description: Package size in bytes.
+      example: 62.231
+
+    - name: installed
```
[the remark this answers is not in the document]
simitt: Actually I searched for other `date` types before I settled on a name, and found `event.created` as one date field not defining a start or end date and followed along with that naming pattern. 
No strong feelings though, auditbeat currently uses `installtime`.

ruflin:
Great idea to get this into ECS. @andrewkroh you might be able to chime in here as I think auditbeat also collects some of this info?

schemas/package.yml
```diff
@@ -0,0 +1,83 @@
...
+      type: keyword
+      description: Build version of the installed package, e.g. commit hash.
+      example: 9b52cc0d83f794f6fc6190f603fa2f62dd9aa2017165d19ff1d62742955a7a05
+
+    - name: build_time
+      level: extended
+      type: date
+      description: Time when package was built.
+
+    - name: checksum
```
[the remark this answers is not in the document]
ruflin: ++ on a simple checksum field.

schemas/package.yml
```diff
@@ -0,0 +1,88 @@
...
+  group: 2
+  short: These fields contain information about an installed software package.
+  description: >
+    These fields contain information about an installed software package. It contains general information about a
+    package, such as name, version or size. It also contains installation details, such as time, URL or
+    user who installed the package.
+  type: group
+  fields:
+
+    - name: id
```
cwurm: How would this ID typically be generated? Is APM already doing this?

schemas/user.yml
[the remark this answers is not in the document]
cwurm: I think this could be misleading - it could also be the user who "owns" a package or who it was installed for (e.g. for applications on Windows or macOS). I suspect in many cases we would not have the information who installed a package (i.e. Auditbeat today does not have it, and could not get it from dpkg/rpm/homebrew).

There is no great harm in putting it under `expected`, but I would not "expect" it in most cases.

schemas/package.yml
```diff
@@ -0,0 +1,88 @@
...
+      type: long
+      description: Package size in bytes.
+      example: 62231
+
+    - name: installed
+      level: core
+      type: date
+      description: Time when package was installed.
+
+    - name: remote_repository
```
cwurm: What would this be filled with for host packages, installed from DPKG or RPM? Most of them would probably share the same remote repository (e.g. `packages.debian.org`).

Auditbeat fills a `url` field with a URL specific to the package when available, e.g. for Homebrew it's `https://docs.docker.com/compose/` for `docker-compose`.

schemas/package.yml
simitt: No APM is not collecting any `package` related data yet. It was mainly inspired by `entity_id` from the [auditbeat system package](https://www.elastic.co/guide/en/beats/auditbeat/current/exported-fields-system.html#_package)

simitt: Discussion around ID was started in https://github.com/elastic/ecs/pull/532#discussion_r318101822

schemas/package.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: checksum
+      level: extended
+      type: keyword
+      description: Checksum of the installed package
+      example: 9b52cc0d83f794f6fc6190f603fa2f62dd9aa2017165d19ff1d62742955a7a05
+
+    - name: install_type
+      level: extended
+      type: keyword
+      description: Indicating when the package was installed, e.g. at runtime or at buildtime.
```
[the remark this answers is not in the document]
simitt: You're right, the field would make sense to use for distinguishing the install scope. Should we also rename to `install_scope` (e.g. user-local, global)?

schemas/package.yml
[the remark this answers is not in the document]
simitt: >  I'm not sure I would nest the hash field set here.

@webmat I not clear on what you mean with that. Could you please explain.

schemas/package.yml
cwurm: Ah sorry, didn't see that. Let's keep the discussion there.

schemas/package.yml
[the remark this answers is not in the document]
cwurm: On `entity_id`: I introduced it when creating dashboards for the Auditbeat system module (https://github.com/elastic/beats/issues/10463). Without it, there was no way to display a total count of installed packages across all hosts.

To be honest, today I'm not sure it was a good idea. It makes it hard to combine data from Beats (which could theoretically all have it, while in practice so far only the Auditbeat system module has it), with data from third parties that will almost certainly never have it.

And I'm not sure it's needed - the total number of packages across hosts might not be terribly important to most people. Events such as package installs, uninstalls, and upgrades are, but those don't need the entity Id.

So in short: I'm not sure we should be specifying this in ECS.

simitt:
>     1. Package type or origin. Values: RPM, DPKG, Homebrew, NPM, Rubygem, etc.

@cwurm how about adding `package.manager` for this kind of information?

>     2. Path (where the package is located): e.g. `/usr/local/Cellar/go/1.10.3/`

I like the idea, will add it.

schemas/package.yml
simitt: @webmat when using `url` rather than `remote_repository` would you expect to embed a full `url` component, as defined in [ECS Url](https://github.com/elastic/ecs/blob/master/docs/field-details.asciidoc#url-fields) or would it be ok from an ECS perspective to define a url as a keyword?

cwurm:
> > ```
> > 1. Package type or origin. Values: RPM, DPKG, Homebrew, NPM, Rubygem, etc.
> > ```
> 
> 
> @cwurm how about adding `package.manager` for this kind of information?

I'm not sure I have a strong opinion, but I think `package.type` would be more general. Not all packages are managed by a package manager - e.g. macOS applications under `/Applications` or Windows programs under `C:\ Program Files` both of which can be just a download.

schemas/package.yml
simitt: @cwurm I agree that we should also add `url`, most probably as `url.origin`. But I see it a bit different from `remote_repository`, which you are right it would be the same information for most packages, but repositories can also be manually configured, so information might be interesting.

cwurm: I see the difference - but do we have a way of collecting it (the remote repository) at the moment? I don't think Auditbeat does.

simitt: @SergeyKleyman can you chim in here, do you see it feasible to collect the information from the agent?

schemas/package.yml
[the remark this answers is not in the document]
simitt: My thoughts were that this might be interesting information to query for unique package installations in an easy way, e.g. to update an existing document rather than creating a new one, or skipping to store a doc if it is already there. This idea is not mature enough yet from my side. With your comment to rather remove it from ECS based on your experience with auditbeat, I suggest to remove for now, but revisit if it makes sense for APM use case before the package field definition goes GA.

schemas/user.yml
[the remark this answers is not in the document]
simitt: Not sure the APM agents can collect the information, but in my opinion the information about the user installing a package could be interesting for SIEM, that's why I added it. 

Would it make sense to add it as `package.installed_by`? Or would you rather not define information we are not sure our shippers can collect yet?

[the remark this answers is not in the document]
cwurm: Yeah, I'm hesitant to define fields that we don't fill at the moment. We can always do it later when we have more information and concrete examples.

schemas/package.yml
SergeyKleyman: I am not sure I understood from conversation what is the difference between `url` and `remote_repository` but if ECS has both we can include both as optional for agents to set.

cwurm: Correct me if I'm wrong @simitt, but I think it's like this:
1. `url` - Most packages come with URLs pointing to their project websites, e.g. the Homebrew `go` package has `https://golang.org`. DPKG and RPM have this, too.
2. `remote_repository` - also a URL, but the repository URL where the package was installed from. I imagine this would be something like `packages.debian.org`. However, we don't collect anything like that in Auditbeat and I don't think we can, because this information is not available where we collect the data from (at least not for Homebrew or DPKG which I checked).

If we don't have implementations of collecting `remote_repository` I would prefer not to specify it in ECS. We can always add it later if we need it.

CHANGELOG.next.md
```diff
@@ -10,6 +10,7 @@ Thanks, you're awesome :-) -->
 ### Bugfixes
 
 ### Added
+* Add support for installed software packages. #TODO(simi) add PR number
```
andrewkroh:
```suggestion
* Add support for installed software packages. #532
```

schemas/package.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: Checksum of the installed package for verification.
+      example: 9b52cc0d83f794f6fc6190f603fa2f62dd9aa2017165d19ff1d62742955a7a05
+
+    - name: install_scope
+      level: extended
+      type: keyword
+      description: Indicating how the package was installed, e.g. user-local, global.
+      example: global
+
+    - name: license
```
webmat: I like this.

I would add a suggestion in the definition, to try to use the license identifier from https://spdx.org/licenses/, if possible (and if it's open source). Not a hard requirement, though. Plus there's no listing that I know of for popular non-OS licenses anyway. Just meant as a nudge to try to keep this clean. We want the short name, if possible.

schemas/package.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: package
+  title: Package
+  group: 2
+  short: These fields contain information about an installed software package.
+  description: >
+    These fields contain information about an installed software package. It contains general information about a
+    package, such as name, version or size. It also contains installation details, such as time, URL or
+    user who installed the package.
```
webmat: Do you have a way to determine which user installed the package, after the fact? If so, it may be worth making `user` reusable under package (add `- package` [in this list](https://github.com/elastic/ecs/blob/v1.1.0/schemas/user.yml#L15-L20)) :-)

If you're unsure, that's also fine. We can take out the mention in the description for now, and add it later if needed. It's much easier to continue adding than to take out or change after it's in ;-)

schemas/package.yml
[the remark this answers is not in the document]
webmat: I see you've renamed to `install_scope` already, but I'll keep the discussion here for a bit ;-)

So `install_scope` makes tons of sense to me. I could see the following values in there:

* global
* user-local
* vendor (do you agree with this one?)

And I'd rather have the definition list them all up front (e.g. "The expected values are..."), so the field content is consistent across sources. 

WDYT? And do you agree with my current list of expected values, or do you see more?

webmat:
@andrewkroh Can you give an example of such a release string?

Is this meant to support their extra long version strings, where they mix the original package version + their backporting patches on top?

schemas/package.yml
```diff
@@ -0,0 +1,96 @@
...
+      type: keyword
+      description: Information about package type. When installed from a package manager, this would contain the
+        package manager name.
+      example: npm
+
+    - name: url.original
+      level: core
+      type: keyword
+      description: URL from where the package was installed.
+      example: https://docs.docker.com/compose/
```
webmat: Another thing about the URL, I would give an example of a full path to the package archive in full.

If places like Docker Hub don't provide that, perhaps that's acceptable. But I expect this URL to be complete as often as possible. So let's lead by `example:`. Pun intended ;-)

schemas/package.yml
webmat: Never mind. Just realized this was already in.

It's Friday night, time to log off 😂

schemas/package.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: Time when package was installed.
+
+    - name: type
+      level: core
+      type: keyword
+      description: Information about package type. When installed from a package manager, this would contain the
+        package manager name.
+      example: npm
+
+    - name: url.original
```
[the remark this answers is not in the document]
webmat: I also realize I'm the one who suggested `url.original` here, sorry for the back & forth

andrewkroh:
> @andrewkroh Can you give an example of such a release string?

2.6.32-754.el6

In this example `2.3.6` is the package version. and `754.el6` is the release. The version is generally the upstream software version (like the linux kernel version) and the release is the distribution specific value that gets rev'ed when a patch is backported.

schemas/package.yml
```diff
@@ -0,0 +1,96 @@
...
+  short: These fields contain information about an installed software package.
+  description: >
+    These fields contain information about an installed software package. It contains general information about a
+    package, such as name, version or size. It also contains installation details, such as time, URL or
+    user who installed the package.
+  type: group
+  fields:
+
+    - name: name
+      level: core
```
ruflin: General note: I would put all these fields into extended for now.

schemas/package.yml
cwurm: After the thread [here](https://github.com/elastic/ecs/pull/532#discussion_r318958329), this might now refer to the "generic URL" of the package and not the URL where it's installed from (what was `remote_repository` when this PR started). If so, the description should be updated to say something neutral like "URL of this package" / "URL with more information about this package".

On where the package was installed from - I don't know any data source (DPKG, RPM, Homebrew, Chocolatey) that provides this? So I would rather keep that out of ECS, and we can always add it later when we have a concrete example and need to add it.

cwurm:
> Do you think there should be a `release` field for release number used in Redhat packaging? Or would that be too specific to one packaging type.

I personally think it would be too specific. Only RPM has it. DPKG has a similar one though, called `revision` (not parsed out by Auditbeat, but libpdkg can do it).

My preference would be to keep ECS simple and not include it at all, with users always welcome to add extra fields that are useful to them (but are not required or even expected). But if we want to have it, I'd rather have something that covers RPM, DPKG, and anything else (and maybe that's `release`, `revision`, or something else generic).

webmat:
On the distro package versions, do rpm and dpkg give easy access to the original/upstream version number as well? In other words, to get the `2.6.32` in [Andrew's example](https://github.com/elastic/ecs/pull/532#issuecomment-526788805), do we need to parse it out of the release version? Or is there another attribute that contains `2.6.32`?

If extra work is required to parse out the "actual" version number, it will get nasty real quick. E.g. what if some original packages contain `-` in their version numbers? I wouldn't be surprised at all 😂

So I'm wondering if we need to specify this at all, at this time. Perhaps just having `package.version` without specifics is enough for now? Platforms that have full release versions will get that in `package.version`, and platforms that only have "simple" versions will get that in `package.version`...

If on the other hand we can get the upstream version easily from rpm/dpkg, I'd be open to discuss the addition of a field for the distro's release version as well.

schemas/package.yml
webmat: Agreed

schemas/package.yml
webmat: Ok, if the exact package source URL isn't available, I'll defer to you here.

But I agree we could perhaps flesh out the description of the field a little more.

schemas/package.yml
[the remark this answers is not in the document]
simitt: As long as we are open for _more_ values in the future, I think defining a starting list would be fine. I am not sure about `vendor`, as depending on the vendor mechanism this could also be system-wide for a specific version or local to a project.

schemas/package.yml
simitt: I am fine with removing `remote_repository` for now, and reconsider adding it later when collecting it somewhere.

simitt:
> Do you think there should be a release field for release number used in Redhat packaging? Or would that be too specific to one packaging type.

@andrewkroh have you seen the suggested `detailed_version`? I originally suggested it with the intention of having fine granular information for unreleased versions, but maybe this could serve both purposes.

andrewkroh:
I think `detailed_version` could work for RHEL packaging assuming we document it should be formatted as `%{VERSION}-%{RELEASE}`. Like the value of `rpm -qp --queryformat '%{VERSION}-%{RELEASE}' example.rpm`. Then since we have a separate field for architecture I think this would cover the RPM details.

schemas/package.yml
```diff
@@ -0,0 +1,90 @@
...
+      level: extended
+      type: date
+      description: Time when package was installed.
+
+    - name: type
+      level: extended
+      type: keyword
+      description: Information about package type. When installed from a package manager, this would contain the
+        package manager name.
+      example: npm
```
webmat: In the spirit of keeping the examples consistent from one another, please put an example here that works with the "golang" example used throughout the field set. So perhaps put example "rpm" or "homebrew".

schemas/package.yml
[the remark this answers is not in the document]
simitt: @webmat I think we need to clear this out - should I add the suggested list?

simitt:
I incorporated all the feedback, from my perspective we need to come to a conclusion in the following points before merging:

@webmat I don't have strong opinions about https://github.com/elastic/ecs/pull/532#discussion_r330393197, please let me know how you prefer to move forward.

@cwurm I gave the [`package type | origin | manager` discussion](https://github.com/elastic/ecs/pull/532#issuecomment-525664722) another thought. Reading `package.type` I think of `.deb` , `.rpm`, `.tgz` and not of the _way_ the package was installed. Maybe using `package.origin` or `package.install_type` (we already use `package.install_scope`, so that would fit) would be preferable here. WDYT?

schemas/package.yml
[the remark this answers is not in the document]
webmat: The definition currently gives "user-local" and "global" as examples. I think that's enough for now.

If there's uncertainty about vendoring, we can skip for now, and we'll hash it out when it comes up with a specific use case. WDYT?

cwurm:
> Reading `package.type` I think of `.deb` , `.rpm`, `.tgz` and not of the _way_ the package was installed.

I struggle with that - I'm not sure we could always fill it with a sensible value. What would the value for a Homebrew package be? When installed, it's just a folder in `/usr/local/Cellar`. Or a Windows program? It's just an `.exe` in `C:\Program Files` - but that was not how it was installed and it's probably impossible to know what the installation method was (could have been an `.msi`, or just copying the `.exe`).

schemas/package.yml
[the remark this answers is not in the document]
cwurm: I'm +1 on skipping it for now, we have no real-world usage examples. We can always add it later.

webmat:
@simitt Thanks for all of the adjustments!

It looks like `package.type` needs more thought as well. Please take it out of the PR and submit a new one. We'll take the time to hash this one out stress-free ;-)

webmat:
Note: Auditbeat will benefit from the package field set, and currently doesn't seem to have the package type as an attribute.

simitt:
@webmat I removed `package.type`, the PR is ready to merge if you approve.

schemas/package.yml
```diff
@@ -0,0 +1,74 @@
...
+      example: global
+
+    - name: license
+      level: extended
+      type: keyword
+      short: Package license
+      description: >
+        License under which the package was released.
+
+        Use a short name, e.g. the license identifier from [SPDX License List](https://spdx.org/licenses/) where possible.
```
webmat: Sorry I just noticed this link. The descriptions in here are used many places, with no, or varying support for links. 

Please put a plain link, instead of a Markdown link:

> Use a short name, e.g. the license identifier from SPDX License List where possible (https://spdx.org/licenses/).

schemas/package.yml
```diff
@@ -0,0 +1,74 @@
...
+      level: extended
+      type: keyword
+      description: Path where the package is installed.
+      example: /usr/local/Cellar/go/1.12.9/
+
+    - name: architecture
+      level: extended
+      type: keyword
+      description: Package architecture.
+      example: runtime
```
webmat: > x86_64

webmat:
Thanks for all the last minute adjustments 🙂

cwurm:
I think without the package type (deb, rpm, homebrew, etc.) it's hard to distinguish between different packages? For example, there is an `elasticsearch` package in pretty much every package manager, but we cannot distinguish between them without the package type. 

Also, pretty much every package manager has a URL/homepage for the package (e.g. `https://www.elastic.co/products/elasticsearch`) and that's very useful for finding out what the package is (ok, `elasticsearch` is pretty obvious, but there are many obscure packages, e.g. `libewf`). In my opinion, this field would be a lot more useful than let's say the package architecture (almost always amd64 these days I assume).

Do we have a need to merge this now (i.e. is something waiting and ready to implement these fields), or can we wait and get these useful fields in?

webmat:
@cwurm As far as I can tell, PR as it stands covers all of what Auditbeat captures about packages, except for entity_id. Is the absence of a place for the URL a blocker for getting this in 1.2?

Auditbeat 7.4 | ECS
-- | --
system.audit.package.arch | package.architecture
system.audit.package.entity_id | -
system.audit.package.name | package.name
system.audit.package.size | package.size
system.audit.package.summary | package.description
system.audit.package.version | package.version

cwurm:
@webmat That the package dataset does not have a `type` field is an oversight and my hope is that when we touch it again we can add it. We can get this into ECS 1.2, then get `package.type` into 1.3. For the development, it doesn't matter all that much if something is released in ECS, just that it's agreed upon and merged.

PR #525, opened 2019-08-21
A new take on adding missing Syslog fields to ECS.

webmat:
This PR is meant to replace and close #301.

Unfortunately #301 is too old to be brought up to date with the current state
of the ECS repo.

While bringing the discussion we had in the original PR to present day ECS,
I'm proposing a few changes to what we had there. Comments welcome, if you
agree/disagree.

In this PR:

- Modify description of `log.level` and associate it with Syslog's severity label.
- Modify description of `event.severity` and associate it with Syslog's severity numeric code.
- Add field `log.facility` as the text-based representation.
- Add field `log.facility_code` as the numeric representation.
- Add field `log.priority` (numeric only: 8 * facility + severity)

Differences with #301:

- I'm nesting under `log.*` instead of `event.*`, because I feel like these very
  Syslog-specific fields work better there than in the already crowded `event.*`.
- I'm not adding a text field for the priority label. There's no such thing as a priority label in Syslog.
- I'm specifically calling out the Syslog mapping in the existing field `log.level` and `event.severity`.

Issue #516, opened 2019-08-09
Is ECS confusing to you?


anhlqn:
I've been using Elastic stack for Log Analytics use case for a few years since v1.4.0 with more than 30 types of logs. I think most of us know the challenge of not having common fields across log types, so we came up with our internal core and standard fields to enable log correlation. When I heard about the release of ECS, I thought I can just look at the standard document and easily map our current fields to ECS fields.

That's not really the case, at least for me. I guess some new Elastic users may have the same frustration when looking to adopt ECS.

> The goal of ECS is to enable and encourage users of Elasticsearch to normalize their event data, so that they can better analyze, visualize, and correlate the data represented in their events.

> ECS fields follow a series of guidelines, to ensure a consistent and predictable feel, across various use cases.

If Elastic expects the community to adopt ECS, that means ECS should be as simple as possible, not confusing. I've read and re-read the ECS document multiple times and end up having to search through issues in the Github repo to know how to properly map a field to ECS. Almost have to pull a few people in a meeting when we have a new field.

I like the idea of grouping related fields into fieldsets, but I think there is room for improvement.

**source/destination vs. client/server**
source/client and server/destination have the same nested fields, so why don't we just choose one pair and get rid of the other? source/destination pair is more generic and will fit most use cases.

>Client / server representations can add semantic context to an exchange, which is helpful to visualize the data in certain situations. If your context falls in that category, you should still ensure that source and destination are filled appropriately.

I think this approach is unnecessarily complicated. It confuses users and can significantly increase the storage size at the same time if we are looking at billions of events. Netflow, firewall, web access, or IPS/IDS logs can all use source/destination pair. With two pairs, the questions users face are:

- When to populate source/destination?
- When to populate client/server?
- When to populate two pairs?
- Which pair should I use when searching by field:value syntax?
- Should correlation be done on one pair or the other?

These questions aren't answered in the current ECS document.

**log vs. event**
I have the same confusion with these two field sets. I believe ECS is designed for Log Analytics use case, which aims to support the new Kibana SIEM app. Usually, any log message that enters the SIEM is treated as an event, so **event** field set is a great one. Why also **log** to confuse users? If I want to map attributes of a message to ECS, do I have to jump between log and event field sets?

log.level vs. event.severity? and someone has proposed a event.level here https://github.com/elastic/ecs/issues/129

[...]

PR #540, opened 2019-09-04
Beef up the description of the `log` field set.

webmat:
This is meant to help clarify the difference between `log` and `event`.

Relates to #525 and address some concerns voiced in #516.

### DIFF
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -2,14 +2,16 @@
 - name: log
   title: Log
   group: 2
-  short: Low level details specific to log events.
+  short: Details about the event's logging mechanism.
   description: >
     Low level details specific to log events.
 
-    `log.*` typically contains lower level details about the logging mechanism
-    than `event.*`. For example, The Syslog transport details belongs under `log.*`,
-    whereas the message payloads transmitted via Syslog will likely be parsed out
-    into `event.*` or elsewhere in ECS.
+    The log.* fields are typically populated with details about the logging
+    mechanism used to create and/or transport the event.
+    For example, syslog details belong under `log.syslog.*`.
+
+    The details specific to your event source are typically not logged under `log.*`,
+    but rather in `event.*` or in other ECS fields.
   type: group
   fields:
 
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -2,8 +2,14 @@
 - name: log
   title: Log
   group: 2
+  short: Low level details specific to log events.
   description: >
-    Fields which are specific to log events.
+    Low level details specific to log events.
+
+    `log.*` typically contains lower level details about the logging mechanism
+    than `event.*`. For example, The Syslog transport details belongs under `log.*`,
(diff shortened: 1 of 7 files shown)


schemas/log.yml
```diff
@@ -2,8 +2,14 @@
 - name: log
   title: Log
   group: 2
+  short: Low level details specific to log events.
   description: >
-    Fields which are specific to log events.
+    Low level details specific to log events.
+
+    `log.*` typically contains lower level details about the logging mechanism
```
MikePaquette: Suggest: "Details about the event's logging mechanism.  The `log.*` fields are typically populated with details about the logging mechanism used to create and/or transport the event.  For example, syslog details are typically mapped here.  Note that details about the event that is transported over syslog are not typically stored here, but are instead stored in the `event.*` or other ECS fields.

Optional:  <why not invest some text in making this super clear, by adding a more complete and correct sentence, like:>   "When mapping syslog details, the following mappings are recommended:  facility->`log.facility`, severity name->`log.level`, `severity number`->`log.severity`, <add them all>)"

schemas/log.yml
```diff
@@ -2,8 +2,14 @@
 - name: log
   title: Log
   group: 2
+  short: Low level details specific to log events.
```
MikePaquette: Suggest:  "Details about the event's logging mechanism."   <I prefer "logging mechanism" vs. "log events" since metrics can be carried over syslog as well, as mentioned in  https://github.com/elastic/ecs/pull/301#issuecomment-455617131)>

schemas/log.yml
webmat: I'm putting in place the first suggestion. With the latest developments in #525, I think we won't need the optional idea. :-)

webmat:
Ok, just updated this PR with Mike's comments.

I think this clarification about the `log` field set definition will really help users understand the difference between the what goes in `log` and what goes in `event` & elsewhere in ECS.

Let's review whether this new proposed definition fits well with the fields under `log.*`.

**What works well with the new definition**

* `log.original`: There is also a lot of confusion between the fields `log.original` and `event.original`. I think we should deprecate one of the two, as I cannot imagine anyone deciding to store two full, but slightly different original copies of the original log event. After this PR, I think it would make most sense to keep `log.original` as the one to use. `event.*` is not for the low level details, this is now the clearly stated purpose of `log.*`.
* `log.logger`: I think this makes sense. This field is for application level logging, not transport details per se. However this is a low level detail about the source of a log event, in the context of application logs. 👍
* `log.origin.file.*` and `log.origin.function`: I was initially on the fence here. I've been pushing back on these in #563, but I've come around. This is a **really** low level detail about the source of an application log. :-)

**What does not work as well with the new definition**

* `log.level`: after #525, this field will be clearly mapped to the source's severity, not the transport's severity (when they differ). However "log level" is an expression so widespread that I think it's ok to have it under here.

I'll merge this tomorrow, unless @ruflin & @MikePaquette chime back in with additional comments / suggestions.

PR #541, opened 2019-09-06
Added top_level_domain field

mbudge:
Added top_level_domain field so Elasticsearch users with security use cases, can identify logs where the domain has a particular suffix. For example, if a company is experiencing a spear-phishing campaign, they may want to identify all connections to .xyz or .ru domains if the attackers using these domains in their links. 

Having a top_level_domain suffix will allow users to find these connections without having to index domains in text fields, so they won't need to do expensive wildcard queries - "*.ru". Instead they can just do a fast keyword search in their analytics.

Users will be able to create a unique list of top_level_domains by using if statements in Logstash pipelines/filters.

### DIFF
--- a/code/go/ecs/client.go
+++ b/code/go/ecs/client.go
@@ -53,6 +53,17 @@ type Client struct {
 	// Client domain.
 	Domain string `ecs:"domain"`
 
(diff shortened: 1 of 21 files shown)


mbudge:
Forgot to pull from the master branch.

PR #542, opened 2019-09-06
Added top_level_domain field to identify connections to a domain suffix in log data.

mbudge:
Added top_level_domain field so Elasticsearch users with security use cases, can identify logs where the domain has a particular suffix. For example, if a company is experiencing a spear-phishing campaign, they may want to identify all connections to .xyz or .ru domains if the attackers using these domains in their links.

Having a top_level_domain suffix will allow users to find these connections without having to index domains in text fields, so they won't need to do expensive wildcard queries - "*.ru". Instead they can just do a fast keyword search in their analytics.

Users will be able to create a unique list of top_level_domains by using if statements in Logstash pipelines/filters.

### DIFF
--- a/schemas/url.yml
+++ b/schemas/url.yml
@@ -68,6 +68,24 @@
         simply taking the last two labels will not work well for TLDs such as "co.uk".
       example: google.com
 
+    - name: top_level_domain
+      level: extended
+      type: keyword
+      short: The top level domain is the last part of the domain (com, net, org).
+      description: >
+        The top level domain (TLD) also known as the domain suffix is the last part of the domain name.
+        For example, the top level domain for google.com is "com".
+
+        The following groups of top level domain are maintained by the Internet Assigned Numbers Authority (IANA).
+
+        Infrastructure top-level domain (ARPA)
+        Generic top-level domains (gTLD)
+        Restricted generic top-level domains (grTLD)
+        Sponsored top-level domains (sTLD)
+        Country code top-level domains (ccTLD)
+        Internationalized country code top-level domains (IDN ccTLDs)
+        Test top-level domains (tTLD)
+
     - name: port
       format: string
       level: extended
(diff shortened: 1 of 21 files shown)


MikePaquette:
@mbudge I think this is a good addition to ECS!  I see that you've also added it consistently to everywhere that the `*.registered_domain` field is used.  Just wondering if you think you'd use it in the dns field set as `dns.question.top_level_domain` ?

mbudge:
Hi Mike, 

We'll mainly use it in the other schema's so maybe DNS is not necessary. 

I'm happy for it to be left out. 

Thanks.

webmat:
Note: only actual requested change for now is for the changelog.

We can decide whether we re-add under DNS once the discussion has concluded :-)

MikePaquette:
@webmat I was just questioning whether it was necessary.  If it is useful, then I agree with your suggestion to add it back.  In a spec like ECS consistency is goodness :-).

webmat:

schemas/url.yml
```diff
@@ -68,6 +68,24 @@
...
 
+    - name: top_level_domain
+      level: extended
+      type: keyword
+      short: The top level domain is the last part of the domain (com, net, org).
+      description: >
+        The top level domain (TLD) also known as the domain suffix is the last part of the domain name.
+        For example, the top level domain for google.com is "com".
+
+        The following groups of top level domain are maintained by the Internet Assigned Numbers Authority (IANA).
```
webmat: I think this sentence and the subsequent list should be replaced by a mention of Mozilla's public suffix list. You can take inspiration from the `registered_domain` definition for wording.

Of course both approaches can be used to figure out the eTLD, but Mozilla's list is the more direct method. One file, and you're able to determine the eTLD. Pointing users to a slew of lists like this will discourage them from even trying, IMO.

webmat:
Here's a tip to preview the rendered docs, btw. New CI feature.

You can preview how the docs will look after this PR here: http://ecs_542.docs-preview.app.elstc.co/diff. Clicking on any link in there will take you to the relevant field set, including your changes :-)

Example for this PR's changes to URL: http://ecs_542.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-url.html

PR #550, opened 2019-09-11
Docs: Update an include file

nik9000:
This file has moved. It *is* interesting to think about *which* include
file we should use. There are lots of different "versions" of the stack
that we might we use here, but this uses the same version that we used
in the file that has moved.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -2,7 +2,7 @@
 :doctype: book
 :ecs: ECS
 
-include::{asciidoc-dir}/../../shared/versions.asciidoc[]
+include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
 [[ecs-reference]]
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -2,7 +2,7 @@
 :doctype: book
 :ecs: ECS
 
-include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
+include::{asciidoc-dir}/../../shared/versions/stack/7.x.asciidoc[]
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
 [[ecs-reference]]


webmat:
Awesome, thanks Nik! Is this ready to be merged, or should I wait for something to be merged in the docs repo?

Also, we may backport some small ECS tweaks to branches 1.1 and 1.0. This should be backported these branches as well, correct?

nik9000:
> Awesome, thanks Nik! Is this ready to be merged, or should I wait for something to be merged in the docs repo?

It is ready any time!

> Also, we may backport some small ECS tweaks to branches 1.1 and 1.0. This should be backported these branches as well, correct?

It *must* be backported or we'll have to stop building the docs for it. Thought on backport we may want to use a different branch than `master` so we link to a released version of the stack.

karenzone:
Interesting point, @nik9000. 

>  It is interesting to think about which include
file we should use. There are lots of different "versions" of the stack
that we might we use here, but this uses the same version that we used
in the file that has moved.

Because ECS doesn't version with the stack, does it make sense to handle the versions we point to in the ECS index.asciidoc file?

nik9000:
> Because ECS doesn't version with the stack, does it make sense to handle the versions we point to in the ECS index.asciidoc file?

We can do that by including a file like `shared/versions/stack/7.4.asciidoc` or something like that. We make a file for every stack version.

webmat:
@karenzone Right, so I would backport this file and instead of `master.asciidoc` I would adjust so it looks like the mapping table we discussed a few days ago?

> * master = 1.1
> * 7.x = 1.1
> * 7.4 = 1.1
> * 7.3 = 1.0
> * 7.2 = 1.0
> * 7.1 = 1.0
> * 7.0 = 1.0

Although from the ECS POV, it's one to many. ECS doesn't always rev as often as the stack. So perhaps I should point to the latest version of the stack that used a given version of ECS?

* 1.0 => 7.3
* 1.1 => 7.4 (an actual release, not ship them to master)
* master => master

karenzone:
Pointing to `master` doesn't seem right.  That'll link us to `8.0.0-alpha` versions for everything.

We *could* pick a version, like 7.3 or 7.4.  Since it wouldn't impact any other books, I like the control and precision of handling in the file.  Whether we hardcode version filename or hardcode the versions, it'll have to be updated with every stack release anyway.

nik9000:
> Pointing to `master` doesn't seem right. That'll link us to `8.0.0-alpha` versions for everything.

We were doing that before which is why I pointed to master. Would you prefer to point to 7.x?

nik9000:
> Whether we hardcode version filename or hardcode the versions, it'll have to be updated with every stack release anyway.

We were talking about making a `current` file for this which might be nice. I think that'd be a followup change though because we don't have it in the docs repo now.

webmat:
Right now ECS doesn't have master = ECS 2.0, as opposed to most of the stack. We're keeping things simple by reducing the amount of backports.

So yeah our master should likely point to 7.x indeed, good point

karenzone:
I can't think of any downsides of handling versioning for ecs in the ecs index.asciidoc file.  (Like we do for Beats, but for different reasons.). And, we don't get the upsides that we get in stack docs.

If we determine that linking to a particular version file is the way to go, then @webmat 's suggestion makes the most sense:

> Although from the ECS POV, it's one to many. ECS doesn't always rev as often as the stack. So perhaps I should point to the latest version of the stack that used a given version of ECS?

    1.0 => 7.3
    1.1 => 7.4 (an actual release, not ship them to master)
    master => master

karenzone:
I'm not sure why this breaks the doc build only intermittently. (Note to self: Save for another discussion.)

We could set to 7.x and merge while we continue the discussion.

webmat:
Yes let's link to 7.x for now. I'll merge when the PR is updated.

Then I'll try to find time backport. Or either of you can do it if you like :-)

@karenzone was asking if I thought it made most sense to do here or over in the docs repo. I'm not 100% sure of the implications. But my gut feeling is that we should do this here. I'll add this to my release notes, when new branches are cut & so on.

We can decide otherwise later, if I mess things up too often 😂 

However I think the upside of doing it here is that I will be the one most knowledgeable of the mapping, as we continue moving forward.

nik9000:
> I'm not sure why this breaks the doc build only intermittently. (Note to self: Save for another discussion.)

It'll only break it if we try to rebuild the ecs docs. So only if we try to rebuild `--all` right now. It'll fail all the time if we push a patch to the ECS repo that contains docs though.

> However I think the upside of doing it here is that I will be the one most knowledgeable of the mapping, as we continue moving forward.

If you want to own the mapping then you picking the version file to include is perfect.

I'll update this PR to 7.x.

nik9000:
> I'll update this PR to 7.x.

Done.

nik9000:
Thanks!

PR #552, opened 2019-09-11
Docs: Update an include file (backports #550)

nik9000:
This file has moved. Also, we'd prefer to link to the last released
version.


nik9000:
Thanks for reviewing @karenzone!

PR #556, opened 2019-09-11
Added appliance schema for network devices like web proxies, firewalls, WiFi AP's

mbudge:
Added a schema for physical network devices such as web proxies, firewalls, routers and wifi access points. These devices aren't software agents, but they can report on observed events on the network.
Added appliance schema

Added a schema for physical network devices such as web proxies, firewalls, routers and wifi access points. These devices aren't software agents, but they can report on observed events on the network.

Users might not want to identify these devices using the ECS fields like host.name and platform, as the devices report on events observed on the network, and not the device itself (the device isn't the source of the network connection..).


mbudge:
Closed, didn't see observer.

MikePaquette:
@mbudge If I'm reading your PR correctly, this use case should be satisfied by the ECS `observer.*` field set.

I think it looks like this:
image

From the ECS Docs:

> An observer is defined as a special network, security, or application device used to detect, observe, or create network, security, or application-related events and metrics.
> 
> This could be a custom hardware appliance or a server that has been configured to run special network, security, or application software. Examples include firewalls, intrusion detection/prevention systems, network monitoring sensors, web application firewalls, data loss prevention systems, and APM servers. The observer.* fields shall be populated with details of the system, if any, that detects, observes and/or creates a network, security, or application event or metric. Message queues and ETL components used in processing events or metrics are not considered observers in ECS.

PR #557, opened 2019-09-11
Added location and name field to observer.

mbudge:
Web proxies can block traffic in the same way as an Intrusion Prevention System which is listed as a type of Observer. Added a location and name field so observer can also store fields use to describe web proxy appliances found in web proxy logs.

### DIFF
--- a/schemas/observer.yml
+++ b/schemas/observer.yml
@@ -36,11 +36,13 @@
       type: keyword
       description: >
         Name of the observer.
-    - name: location
+      example: 1_proxySG
+    - name: product
       level: core
       type: keyword
       description: >
-        Location of the observer.
+        The observers product name.
+      example: s200
     - name: vendor
       level: core
       type: keyword
--- a/schemas/observer.yml
+++ b/schemas/observer.yml
@@ -9,7 +9,7 @@
 
     This could be a custom hardware appliance or a server that has been configured
     to run special network, security, or application software.
-    Examples include firewalls, intrusion detection/prevention systems,
+    Examples include firewalls, web proxies, intrusion detection/prevention systems,
     network monitoring sensors, web application firewalls, data loss prevention systems, and APM servers.
     The observer.* fields shall be populated with details of the system, if any,
     that detects, observes and/or creates a network, security, or application event or metric.
@@ -31,6 +31,16 @@
       type: keyword
       description: >
         Hostname of the observer.
+    - name: name
+      level: core
+      type: keyword
+      description: >
+        Name of the observer.
+    - name: location
+      level: core
+      type: keyword
+      description: >
+        Location of the observer.
     - name: vendor
       level: core
       type: keyword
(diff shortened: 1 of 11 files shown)


MikePaquette:
@mbudge Indeed a proxy would be considered an observer in ECS.  In fact, it is listed as one of the examples for the `observer.type` field in the ECS [observer docs](https://www.elastic.co/guide/en/ecs/current/ecs-observer.html).

Please note that the geo.* fields maybe nested below `observer.*` providing you with multiple fields to store the location-related information regarding an observer.  For example: 
`observer.geo.city_name`, `observer.geo.continent_name`, `observer.geo.country_iso_code`, `observer.geo.country_name` `observer.geo.location`, `observer.geo.name`, `observer.geo.region_iso_code`, and `observer.geo.region_name`.

In addition, ECS specifies `observer.vendor` and `observer.version` and observer.os.name` and `observer.os.version`.  

For example:  you might have:  `observer.vendor:"symantec"`, `observer.version`:"s200"`, `observer.os.name:"proxysg", and `observer.geo.name:"3rd_floor_mdf"`.

So I think we already have enough observer fields to capture this information, so I don't think we should proceed with your proposed `observer.location` and `observer.name`.

mbudge:
Hi, 

Didn't see the observer.os.location so we'll use that. 

Not sure about observer.os.name for the appliance name, as we want to store the appliances custom name used to identify the appliance. This is similar to the agent.name field but for appliances instead of an agent. 

There could be several if there are multiple web proxies on the network.
1_proxysg
2_proxysg
3_proxysg

Thanks!

schemas/observer.yml
```diff
@@ -31,6 +31,18 @@
       type: keyword
       description: >
         Hostname of the observer.
+    - name: name
+      level: core
+      type: keyword
+      description: >
+        Name of the observer.
```
webmat: Please flesh out this description a little more, to make it clear that this is meant to be user-configured.

For inspiration, here's two other places in ECS where we have `.name` meant to be a user-customizable value: [agent.name](https://github.com/elastic/ecs/blob/v1.1.0/schemas/agent.yml#L30-L38) and [host.name](https://github.com/elastic/ecs/blob/v1.1.0/schemas/host.yml#L29-L33)

schemas/observer.yml
```diff
@@ -31,6 +31,18 @@
...
+      example: 1_proxySG
+    - name: product
+      level: core
+      type: keyword
+      description: >
+        The observers product name.
```
webmat: For consistency's sake with the other field definitions, please use

```suggestion
        The product name of the observer.
```

webmat: Another request, since we're going with @MikePaquette's s200 example, could you also add a corresponding example to the "vendor" field, with value "Symantec", please?

schemas/observer.yml
```diff
@@ -31,6 +31,18 @@
       type: keyword
       description: >
         Hostname of the observer.
+    - name: name
+      level: core
```
ruflin: can we make these fields extended for now? Easy to move it to core later but not sure about the other way around.

MikePaquette: 👍  for `level: extended` for these new additions

PR #558, opened 2019-09-11
Added event.detection to identify suspicious activity or files

mbudge:
Added a field to store a detection which triggered the event or was observed during processing of the event.

An example of a detection which triggered an event is an anti-virus alert caused by malware.
An example of a detection observed during processing of an event, is a web-proxy identifying a suspicious file while processing http requests between an internal host and external web-server.

Detection's could come from the following sources
Anti-virus logs
Intrusion Detection Systems
Intrusion Prevention Systems
Endpoint Agent logs
Web-Proxy logs 
Email solution providers 
Cloud platforms

### DIFF
--- a/code/go/ecs/event.go
+++ b/code/go/ecs/event.go
@@ -76,6 +76,15 @@ type Event struct {
 	// Please avoid using this field for user data.
 	Type string `ecs:"type"`
 
+	// The detection which triggered the event or was observed during
+	// processing of the event.
+	// An example of a detection which triggered an event is an anti-virus
+	// alert caused by malware. An example of a detection observed during
+	// processing of an event, is a web-proxy identifying a suspicious file
+	// while processing http requests between an internal host and external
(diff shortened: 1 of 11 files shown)


MikePaquette:
@mbudge ECS definitely needs a place to store information about the "detection" associated with an event.  

In fact, there are a number of attributes of such a detection that we think will be interesting to a security analyst.  We've been moving forward with a `threat.*` field set for just this purpose.   

Because of the multiple attributes (e.g., name, id, tactic, technique, etc.) required to accommodate threat information, I think it will be better for ECS to keep these in a separate field set, and not move the security-specific "detection" into the `event.*` field set.   

Would you mind taking a look at https://github.com/elastic/ecs/pull/505 ?  I think this will provide at the very least a `threat.name` field that you would use in place of your proposed `event.detection`.

mbudge:
The threat fields look good.

Issue #290, opened 2018-12-18
Clarification for type vs event.type


ceeeekay:
Hi there, 

We're currently trying to align as closely as possible with ECS as part of a major version upgrade.

I notice from https://github.com/elastic/ecs/pull/242 that `event.type` has recently been set to reserved when previously it seemed the intent was to replace any root `type` field, (e.g., for Logstash conditionals, etc) with `event.type`.

We are already some way along with converting all our `type` fields to `event.type` and need clarification on what we should be doing at this point.  What is the best place to move any `type` fields to?

Thanks.

ruflin:
Can you share what kind of values you have in the `type` field today?

ceeeekay:
@ruflin sure thing: anything like: apache, apache-[app_name]-error, auditd, postfix, etc.

Just the regular stuff you would expect to find in a root `type` field for a long-existing elastic stack.

We're fine with moving it (if we do it soon), but we really need to know where to put it to match ECS.  The original ECS docs gave the impression the `event.type` was the right place for this.

MikePaquette:
@ceeeekay great question that pulls on two related threads 1) event categorization and 2)source types.  

1) The ECS fields: `event.kind`, `event.category`, `event.type`, and `event.outcome` are reserved so that we can define a canonical ECS event categorization scheme that can be used across all data sources.

2) There are two ECS fields, that although they may have some overlap with the above, should be used for mapping your existing data sources (fka types) to ECS.  These are `event.module` and `event.dataset`.  We expect that `event.module` would be used as a higher-level field, while `event.dataset` would be used to provide more details.  Also, it may be possible that there could be multiple datasets in a given event.  The Elastic team expects to use these fields similarly, populating the `event.module` field with the Beats or Logstash module name, and the `event.dataset` field with the name(s) of the dataset, metricset, or fileset associated with the module.

For example, an apache error log event could populate these fields like:
`event.module: apache2`
`event.dataset: error`

Another example, an apache access log event could populate these fields like:
`event.module: apache2`
`event.dataset: apache_access`  <-- you may want the dataset name to include the module name

Another example, auditd file events could populate these fields like:
`event.module: auditd`
`event.dataset: file_integrity`

Another example, auditd system call events could populate these fields like:
`event.module: auditd`
`event.dataset: system_call`

Another example, Palo Alto _pan_traffic_ events could populate these fields like:
`event.module: traffic`
`event.dataset: scan`

[...]

PR #563, opened 2019-09-16
Add log.origin fields

felixbarny:


schemas/log.yml
```diff
@@ -43,3 +43,28 @@
       short: Name of the logger.
       description: >
         The name of the logger inside an application. This is usually the name of the class which initialized the logger, or can be a custom name. 
+
+    - name: source.file
```
ruflin: I worry if what we call source here will conflict with what we have in source under https://github.com/elastic/ecs/blob/master/schemas/source.yml At the same time I think some argument could be it is similar?

felixbarny: > At the same time I think some argument could be it is similar?

I'd say it's actually not really similar to `source`. We could also rename to `log.origin.*` to avoid ambiguities.

ruflin: +1 on origin.

schemas/log.yml
```diff
@@ -43,3 +43,28 @@
...
+      description: >
+        The name of the source file which originated the log event.
+
+    - name: origin.function
+      level: core
+      type: keyword
+      example: init 
+      short: The function which originated the log event.
+      description: >
+        The name of the function or method which originated the log event.
```
ruflin: Out of curiosity: If it's a method, would we expect it to also indicate what class it belongs to? Something like `Foo.bar`?

schemas/log.yml
```diff
@@ -43,3 +43,28 @@
       short: Name of the logger.
       description: >
         The name of the logger inside an application. This is usually the name of the class which initialized the logger, or can be a custom name. 
+
+    - name: origin.file
```
ruflin: Let's reuse the `file` object here: https://github.com/elastic/ecs/blob/master/schemas/file.yml#L15

So this becomes `log.origin.file.path` I guess?

schemas/log.yml
felixbarny: No, it would just be the `bar` in this case. We could, however, add an optional `origin.class` for languages which support that concept.

schemas/log.yml
felixbarny: Good call! It would be `log.origin.file.name` though. Should we then also do `log.origin.file.line` instead of `log.origin.line`?

schemas/log.yml
ruflin: Got it, SGTM. Lets only add it if we already have a use case for it. But

schemas/log.yml
ruflin: That would make sense I guess. I wonder if we "must" extend `file` for this or we just use it here? I'm good with just adding it here but I'm sure at @webmat will have an opinion on this.

schemas/log.yml
felixbarny: We actually do have a use case for logging. For java logging libraries, we usually get both the file name and the class name of the origin. See also https://docs.oracle.com/javase/7/docs/api/java/lang/StackTraceElement.html#getClassName().

I didn't add it at first because not all languages have the concept of a class. But when it's optional, why not?

ruflin: Yeah, I could name quite a few that have a similar concept ;-) In the case of Golang I would also use `.class` even though that is not 100% correct. @axw Hope you agree :-)

+1 on adding class directly.

axw: :) I'm fine with class as a name, but I would have made it a qualified name like `Foo.bar` as you described. @felixbarny what was the reason behind excluding the class name? Is it useful to search on them independently?

felixbarny: I don't see a reason why the class name should be included `log.origin.function`, am I missing something?

We can definitely have an additional field for the class (`log.origin.class`) which contains the fully qualified class name like `org.elasticsearch.bootstrap.Bootstrap`. But we currently don't have that field in the APM schema where we have module instead, which would translate to a Java package.

I'd like to defer these potentially controversial discussions and get the lowest common denominator in which, I think, is `log.origin.file.name`, `log.origin.file.line` and `log.origin.function`.

ruflin: In the description it mentions, it can be a function or method. So if it is a method, would it be wrong to put here `Bootstrap.foo`? What I'm getting at, it would be nice if the shortcut to this discussion is that both options are fine. Note: I left out the package name on purpose, was not even thinking of it before @felixbarny brought up the example.

felixbarny: That would seem a bit weird to me as the method name is foo, not Bootstrap.foo. https://docs.oracle.com/javase/7/docs/api/java/lang/StackTraceElement.html#getMethodName() would also just return the method name. 

What would be the benefit of concatenating the simple class name and the method name?

axw: @felixbarny I'm not too fussed if this is deferred. I don't like to sound like a broken record, but the use case I have in mind is once again profiling, and I was hoping we could find a common definition of fields.

I'd say the benefit of having them combined is having a simple display name, so that UIs don't need to know how to recombine the class and method name. e.g. it's "Class::Method" in C++.

CHANGELOG.next.md
```diff
@@ -10,6 +10,7 @@ Thanks, you're awesome :-) -->
 ### Bugfixes
 
 ### Added
+* Added `log.origin.file`, `log.origin.function` and `log.origin.line` fields. #563
```
ruflin: Needs updating

ruflin:
@felixbarny @axw Suggestion for getting this PR in quickly: Remove `function` from this PR and open a separate one with it like this we can merge the other two fields quickly.

axw:
@ruflin @felixbarny after further consideration, I'm personally happy for this to go in as-is. What I have in mind is still a bit future-looking, and could just as easily be implemented as a different field, such as "displayname".

schemas/log.yml
```diff
@@ -43,3 +43,28 @@
       short: Name of the logger.
       description: >
         The name of the logger inside an application. This is usually the name of the class which initialized the logger, or can be a custom name. 
+
+    - name: origin.file.name
+      level: core
```
ruflin: Could we make all these extended? Sorry, should have spotted this earlier.

felixbarny: sure, np 🙂

schemas/log.yml
webmat: I'm fine with directly adding 2-3 specific fields for now 👍 

We're going for something like this for `package.*` as well

webmat:
Sorry for not responding to that PR last week, when this was happening.

I'm not 100% sold on the addition of `origin` under `log.*`. In another context, we had been discussing people's confusion between `log.*` and `event.*`, and therefore wanted to introduce [this clarification](https://github.com/elastic/ecs/pull/540/files?file-filters%5B%5D=.yml#diff-be3ee609762d18a7b12bc471fd6db8f2) to the description of the `log` field set. So the thinking was: `log.*` is for the archaic or low level details, whereas `event.*` is for the more interesting or relevant details.

I'm not saying we need to move this into `event.origin.*` necessarily, but simply that I think we need to move this out from under `log.*` and find a better place for it.

schemas/log.yml
MikePaquette: Note: Using the `file.*` object here would require a change to ECS, since it is not currently defined as a reusable (nestable) object.  I'm not saying this is wrong, as the idea has come up before to capture events like file-copy events, but we should consider this as we work through this PR, addressing where else `file.*` could be nested.  For now, I would suggest we don't re-use the `file.*` object here.

MikePaquette:
Likewise, sorry for the late response.  I have concerns similar to what @webmat has raised. To help with ECS spec consistency and to avoid ECS user confusion, we've been providing high-level guidance for the understanding of certain fields and field sets (e.g.,  https://github.com/elastic/ecs/issues/516#issuecomment-52281749, https://github.com/elastic/ecs/issues/290#issuecomment-449384800) as follows:

1. **Where the event comes from:**
`event.module`, `event.dataset`
2. **How the event should be analyzed/visualized/correlated:**
`event.kind`, `event.category`, `event.action`, `event.type`, `event.outcome`
3. **How the event was created or transported (e.g., syslog details)**
`log.*`
4. **Everything else about the event**
all other ECS field sets

I am not sure I fully understand the proposed use of `log.origin`, but it seems to me that it fits better into No. 1 and/or No. 4 above rather than into No. 3.

felixbarny:
> 3. How the event was created or transported (e.g., syslog details)
log.*

Sounds like `log.origin.*` makes sense in this case then 🙂 

It captures in which source file (incl. name of the function and line number) the log event was created. For example:

https://github.com/elastic/ecs/blob/c6e5685451cf6e90aef1626abadf7d5a015da437/scripts/generator.py#L27

`log.origin.file.name: generator.py`
`log.origin.file.line: 27`
`log.origin.funtion: main`

schemas/log.yml
felixbarny: It's not using the full `file` object here but re-uses the same structure. Coming up with slightly different names just not to re-use the `file` object structure doesn't feel quite right. But I think I don't quite understand what you have in mind here. Which names would you suggest?

webmat:
Filebeat currently populates `log.file.name` with the name of the log file an event was read from. This is not currently an ECS field, but could eventually become one.

We're now introducing `log.origin.file.name`, and I get the feeling people will start using this to track the log file name / line where a log originated from, and be confused why Filebeat uses a different field.

I think creating a code-specific field set would add a lot of semantic meaning to these events. What about this?

```YAML
code.file.name: generator.py
code.file.line: 27
code.function: main
```

The `code` field set could be used for the source of errors/messages, and also for other things eventually. I'm sure there's events in the upcoming Code Kibana app that would benefit from having a `code` field set.

felixbarny:
> I get the feeling people will start using this to track the log file name / line where a log originated from (e.g. /var/log/system.log)

That would be a misuse of `log.origin` then. Something we can't really avoid with any field. But we could improve the documentation to be more clear about that.

I'm generally fine with `code` too but down the line, we probably need a structure which represents a stack trace element which is reusable in different contexts. For example, to represent profiling data, as @axw mentioned, to represent parsed stack traces, which contain an array of stack trace elements and also as a single stack trace element for the origin of a log event. I'm not sure `code` is the best namespace for the logging use case. Yes, it's about a stack trace element which basically represents a code reference but it's in the context of logging. Hence my vote would still go for `log.origin`.

Let's include @mw-ding and @zfy0701 from the code team to this discussion who are working on APM/Code integration.

mw-ding:
Thanks @felixbarny for cc. It's great to know that we can have the origin of the logs. This will definitely enable some use cases for Code integrations.

Questions. This change is for elastic common schema, so when will this be available in the actually data? Also, even with this schema, the data is not guaranteed to have them, right? Any ideas on the coverage?

ruflin:
Also +1 to continue with `log.origin` and improve documentation and tackle `code` afterwards.

felixbarny:
> when will this be available in the actually data? Also, even with this schema, the data is not guaranteed to have them, right? Any ideas on the coverage?

This data should always be considered optional. It's basically just a recommendation for people to use these fields. Additionally, I have created https://github.com/elastic/java-ecs-logging which is a plugin for the most commonly used Java loggers which format log events to an ECS compatible structure. By default, they don't include the `log.origin.*` fields but it can be enabled with a flag. The reason it's disabled by default is that this information is quite expensive to gather. The loggers basically have to capture a stack trace for each log event.

webmat:
@mw-ding The coverage will be dependent on adding support in the sources of these events, so it will come over time. You should sync up with folks implementing the sources you have in mind to get their take on implementing this -- I assume it's mostly APM clients? But in short, your solutions needs to support this being present or absent.

These fields being in ECS also means that users will be able to populate these fields in their custom events as well. This means more opportunities for the Code app to hook up with arbitrary sources for integration :-)

webmat:
So the consensus is to use `log.origin.*` for now, and potentially do a breaking change in 8.0 / ECS 2.0?

Seems like an unnecessary pain, since this is still new and unpublished (master is just a dev branch).

ruflin:
I think I will be happy with `log.origin.*` for logs also in the long run, so no breaking changes needed ;-)

mw-ding:
Yes, this is new and super helpful for Code. Will keep this in mind for future integrations. @felixbarny @webmat

webmat:
@felixbarny @ruflin I've been finishing up the work on #540 -- meant to clarify the purpose of `log.*`.

As you can see from https://github.com/elastic/ecs/pull/540#issuecomment-535194080, I've come around: I agree with `log.origin.*`, after all. I retract my hesitations with the new fields.

I think I was conflating the `log.origin.*` details with the work on stack traces :-)

beawolf:
I have a follow-up question if I may ask: Is it the correct approach to concatenate full namespace with method name for `log.origin.function` field? Or is it more related to `event.dataset` field? Suppose that I have an application named _TestWebApp_ and a code block like this:

```csharp
namespace MyCompany.TestWebApp
{
  public class AccountsController
  {
    public List<Account> GetAccounts()
    {
      logger.Info("Accounts have been retrieved successfully.");
      
      // Do other things.
      // ...
    }
  }
}
```

What is the correct approach?
1. log.origin.function: GetAccounts and event.dataset: MyCompany.TestWebApp.AccountsController
2. log.origin.function: MyCompany.TestWebApp.AccountsController.GetAccounts
3. Or something else?

felixbarny:
Here's what I would recommend:

`log.origin.file.name`: `AccountsController.cs`
`log.origin.file.line`: `42`
`log.origin.function`: `GetAccounts`
`log.logger`: `MyCompany.TestWebApp.AccountsController`

`event.dataset` is not recommended to use to store the name of the logger. See https://github.com/elastic/ecs/issues/281#issuecomment-520567252 for more details.

beawolf:
Thank you @felixbarny, that explanation is very useful for me.

One more question about `log.origin.file.name`: Is this field intended just for file name or full path? I mean by full path something like this: `/home/beawolf/MyProjects/TestWebApp/AccountsController.cs`

felixbarny:
That would be just the name. Do you need to store the full path? A good place for that would be `log.origin.file.path`, although that field is not part of ECS yet.

Another potentially useful addition would be `log.origin.class` which would be the fully qualified class name of the class which originated the log event - `MyCompany.TestWebApp.AccountsController` in your case. 

Not sure if that's possible in .NET but in Java, the logger name (`log.logger`) may be set to something other than the class name.

beawolf:
In .NET side, log4net provides full location information in a single field, hence my question was based on that actually. We have just started our implementation, so I haven't got any solid argument for usefulness of storing file path right now.

log4net is the counterpart of log4j in .NET world, and you can set logger name other than the class name as you suspected.

felixbarny:
Could you link at the docs for which properties are available for the location information?

beawolf:
Here is the related doc page: [log4net LocationInfo](https://logging.apache.org/log4net/release/sdk/?topic=html/Properties_T_log4net_Core_LocationInfo.htm) (Not so informative unfortunately.)

And here are some actual example values for it in my application:

```
ClassName: TestWebApp.AccountsController
FileName: /home/beawolf/MyProjects/TestWebApp/AccountsController.cs
FullInfo: TestWebApp.AccountsController.GetAccounts(/home/beawolf/MyProjects/TestWebApp/AccountsController.cs:46)
LineNumber: 46
MethodName: GetAccounts
```

PR #565, opened 2019-09-18
Adding service.node.name

### LINKED ISSUE #538: Add service node name
Whenever a `service` is composed of multiple instances, there may be a need to distinguish between them for filtering and aggregation purposes.

The context for this issue is https://github.com/elastic/kibana/issues/43765 - a service (eg a web application) may be composed from multiple instances. In APM, we recently introduced dedicated UI for showing JVM specific metrics. Currently, in order to drill down to see specific JVMs, you would need to use the query bar with other fields, that do not always guarantee uniqueness of JVMs (ie service instances). We now want to further enhance the Metrics feature and provide better options for drilling down to individual JVMs. For this purpose, we want to store a unique service instance name.

We are already using the `agent` fields and in terms of data-correctness we could use `agent.id`, but it doesn't really reflect the meaning of that field, which is naming the service instance, and not the agent, so we are not keen on using that. Besides, service instance fields may be valid for agent-less scenarios as well.

Would adding `service.instance.name` be proper? If so, I assume it would make sense to add `service.instance.id` as well, but we currently only need the name (we **are** going to require uniqueness, but it should be a human readable name, therefore - a name rather than an ID).

eyalkoren:
Closes #538

### DIFF
--- a/schemas/service.yml
+++ b/schemas/service.yml
@@ -36,16 +36,31 @@
       description: >
         Name of the service data is collected from.
 
-        The name of the service is normally user given. This allows if two
-        instances of the same service are running on the same machine
-        they can be differentiated by the `service.name`.
+        The name of the service is normally user given. This allows for
+        distributed services that run on multiple hosts to correlate the
+        related instances based on the name.
 
-        Also it allows for distributed services that run on multiple hosts to
-        correlate the related instances based on the name.
-
-        In the case of Elasticsearch the service.name could contain the cluster
-        name. For Beats the service.name is by default a copy of the `service.type`
+        In the case of Elasticsearch the `service.name` could contain the cluster
+        name. For Beats the `service.name` is by default a copy of the `service.type`
         field if no name is specified.
+
+    - name: node.name
+      level: extended
+      type: keyword
+      example: "node1"
+      short: Name of the service node.
+      description: >
+        Name of a service node.
+
+        This allows for two nodes of the same service running on the same
+        host to be differentiated. Therefore, `service.node.name` should
+        typically be unique across nodes of a given service.
+
+        In APM, this field is populated by default with the container ID
+        (if the node is running within a container) or the host name
+        (automatically discovered, or manually configured). When the default
+        is not unique for each node within a service, it can be manually
+        configured through the APM agents.
 
     - name: type
       level: core
--- a/schemas/service.yml
+++ b/schemas/service.yml
@@ -47,7 +47,7 @@
     - name: node.name
       level: extended
       type: keyword
-      example: "node1"
+      example: "master"
       short: Name of the service node.
       description: >
         Name of a service node.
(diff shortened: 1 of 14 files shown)


schemas/service.yml
```diff
@@ -36,17 +36,32 @@
...
+        name. For Beats the `service.name` is by default a copy of the `service.type`
         field if no name is specified.
 
+    - name: node.name
+      level: extended
+      type: keyword
+      example: "node1"
+      short: Name of the service node.
+      description: >
+        Name of a service node.
```
webmat: I have a few things I'd like to be adjusted in this description, but I agree with the intent.

1. Please mention the use case for Elasticsearch nodes again. It's useful to have the same example reused across a field set. It will help cement the distinction between `service.name` and `service.node.name`. 
2. I like the thinking described in the last paragraph as well, as a general approach for multi-host applications that aren't specifically clustered (e.g. web apps in general). However right now it's phrased as APM documentation. Please phrase it more generally instead, here's an example:

> In cases where the service is not a clustered or doesn't have the concept of a node name, the host name or container name can be used to distinguish running instances that make up this service.

eyalkoren: I actually tried to follow the `service.name` description - tell the general use (first paragraph) and then provide examples (second paragraph). In the case of `service.name` the examples are Elasticsearch and Beats. 
I will add the Elasticsearch example and minimize the APM example description.

schemas/service.yml
```diff
@@ -36,17 +36,31 @@
...
+
+        This allows for two nodes of the same service running on the same
+        host to be differentiated. Therefore, `service.node.name` should
+        typically be unique across nodes of a given service.
+
+        In the case of Elasticsearch, the `service.node.name` could contain
+        the node name, e.g. "ingest" or "master".
```
webmat: I would remove the "ingest" or "master" examples here. This will confuse people vs the Elasticsearch roles. This field is rather meant to hold something like "instance-0000000016", if someone were to use Elastic Cloud. But we don't need to have an explicit string example for the ES use case. I think establishing that an ES node name maps to `service.node.name` is sufficient.

The last paragraph still reads like APM documentation. Documentation of our products doesn't belong in ECS, it belongs in each product's documentation. However, as I've already mentioned, I like the thinking you were describing earlier. I think it makes sense for other general cases, such as parsing web access logs. The quotation under point 2 above was specifically meant to replace this paragraph. This description would actually map well to how APM behaves, and users will be able to follow this to implement compatible mappings in similar logs.

eyalkoren: > The quotation under point 2 above was specifically meant to replace the last paragraph. This description would actually map well to how APM behaves, and users will be able to follow this to implement compatible mappings in similar logs.

This quotation doesn't really captures what I mean with this addition:

> In cases where the service is not a clustered or doesn't have the concept of a node name

This is irrelevant- even if there is a concept of a node name, whatever sends the data for it (filebeat, metricbeat, APM agent) still needs to discover it and put it in a common place, where this field will come into play and be used for filtering/aggregations.

> the host name or container name can be used to distinguish running instances that make up this service.

This maps well to how APM behaves, but, even if true, it specifically describes places where we didn't need this field 🙂. We DO use the container name for containerized services and the host name for others. We needed something in addition for cases where two service nodes are not containerized and are colocated on the same host (eg for redundancy purposes).

I'll try again, without any APM specifics.

PR #566, opened 2019-09-19
Add error.type field

felixbarny:
At first I tried to map the exception class name to error.code.
But I don't think that's the best place for that.
In APM we also have both a code and a type for an exception.

### DIFF
--- a/schemas/error.yml
+++ b/schemas/error.yml
@@ -29,6 +29,12 @@
       description: >
         Error code describing the error.
 
+    - name: type
+      level: core
+      type: keyword
+      description: >
+        The type of the error, for example the class name of the exception.
+
     - name: stack_trace
       level: extended
       type: keyword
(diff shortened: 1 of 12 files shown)


schemas/error.yml
```diff
@@ -29,6 +29,12 @@
       description: >
         Error code describing the error.
 
+    - name: type
+      level: core
```
ruflin: I kind of agree that this should be core but in general I would try to have as little field as possible in core as we will not get them "out there" anymore :-) Could we make it extended for now?

schemas/error.yml
```diff
@@ -29,6 +29,12 @@
...
+      type: keyword
+      description: >
```
ruflin: Could you add an example under `example` key? Would also be great to have at least 2 examples in the description. The reason I'm insisting on this is that `type` fields are always these "well we could fit everything in here" and its always hard to explain what is in. Having example should help.

felixbarny: Added an `example` field but could not come up with a good second example for the description

PR #567, opened 2019-09-23
add JSON + omitempty tags to generated Go code

mark-ignacio:
Specifying JSON tags with explicit lowercase names and the `omitempty` tag allows for easier (and slimmer, in terms of JSON payload size) integration of data processed by Go.

A lot of libraries (for me, https://github.com/segmentio/kafka-go and https://github.com/olivere/elastic) rely on the built-in `json.Marshal()` function to serialize things before they get shipped off, so this ensures that field names are canonical.

### DIFF
--- a/code/go/ecs/as.go
+++ b/code/go/ecs/as.go
@@ -26,8 +26,8 @@ package ecs
 type AS struct {
 	// Unique number allocated to the autonomous system. The autonomous system
 	// number (ASN) uniquely identifies each network on the Internet.
-	Number int64 `ecs:"number"`
+	Number int64 `ecs:"number" json:"number,omitempty"`
 
 	// Organization name.
-	OrganizationName string `ecs:"organization.name"`
+	OrganizationName string `ecs:"organization.name" json:"organization.name,omitempty"`
 }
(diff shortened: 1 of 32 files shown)


code/go/ecs/as.go
```diff
@@ -26,8 +26,8 @@ package ecs
 type AS struct {
 	// Unique number allocated to the autonomous system. The autonomous system
 	// number (ASN) uniquely identifies each network on the Internet.
-	Number int64 `ecs:"number"`
+	Number int64 `ecs:"number" json:"number,omitempty"`
 
 	// Organization name.
-	OrganizationName string `ecs:"organization.name"`
+	OrganizationName string `ecs:"organization.name" json:"organization.name,omitempty"`
```
andrewkroh: The `encoding/json` package is going to produce key names containing dots and I don't think this is what we want.

`{"organization.name": "Google LLC"}` vs `{"organization": {"name": "Google LLC"}}`

webmat: Agreed, everything must be nested. No dots in key names.

mark-ignacio:
Thanks for the feedback! I'll reopen with something that uses `go/ast` or similar if I find out that amending the new code would suck too much.

PR #575, opened 2019-10-02
Add user_agent.original.text and fix multi_fields support

webmat:
This PR has a dual purpose:

- It's meant to finalize the work on #555. Thanks for getting this started, @mbudge!
- Since this is the first multi-fields in ECS since 1.0.0, we needed to fix
  support for them in the generator. After this PR, the multi-fields appear:
  - as a separate line in the csv.
  - as a simplified mention in the main docs, we will likely want to improve this.
  - as a proper definition in the Elasticsearch template samples.
    Caveat: there isn't full support for setting defaults per type on multi-fields.
    The only default being set is setting `norms:false` for `text`.


webmat:
Note: I don't think this will make 1.2.0, so no rush.

scripts/generators/asciidoc_fields.py
```diff
@@ -76,8 +77,15 @@ def render_field_details_row(field):
     example = ''
     if 'example' in field:
         example = "example: `{}`".format(str(field['example']))
+
+    field_name_with_mf = field['flat_name']
```
ruflin: I think we will need to add some more docs to this code or extract it to documented functions. Not required now).

scripts/helper.py
```diff
@@ -60,18 +60,18 @@ def clean_fields(fields, prefix, group):
...
-                clean_string_field(f, "example")
-                clean_string_field(f, "type")
-
-                # multi fields always have a prefix
-                f["name"] = field["name"] + "." + f["name"]
-
-                if 'group' not in f.keys():
-                    # If no group set, set parent group
-                    f["group"] = group
+        # if "multi_fields" in field:
```
ruflin: Instead of commenting out, can this code be removed?

scripts/generators/asciidoc_fields.py
webmat: @ruflin Yes, I would like to have the field set pages be expanded this way eventually:

- Title & description of field set at the top (like now)
- More compact table with short definitions only, and field names are links to a more complete section lower in the page
- Perhaps a "field reuse" section like now, or integrate field reuse directly in this summary table described above
- One linkable section per field name, with more complete details. Using headings for those will allow for the linking, and will also allow for additional information that's currently difficult to capture in the table we have not (should it be an array or a single value?, multi-fields details, deprecation, is the field index:false, etc.)

scripts/helper.py
webmat: Yes, thanks for pointing out.

webmat:
I'd like to merge this soon and follow up with another PR adding the `text` analyzer to other fields.

I've made minor adjustments since your review, @ruflin:

- moved the mention of multi-fields from the left column to the center column (below type)
- changed user_agent.original description to remove the word "keyword"
- adjusted CSV line for multi-fields to the new CSV format
- added changelog entries

It's important to review how it looks in the docs preview. Prior to today's adjustments it looked like this (no longer the case):

Screenshot 2019-12-05 16 47 56

The new format ([see preview](http://ecs_575.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-user_agent.html)) is more complete and clearer, but I'm not a big fan. A page redesign is out of the question for this PR, but I'm very much open to suggestions on how we can make this look acceptable in the meantime.

cc @rw-access @randomuserid and 🎁 @neu5ron ;-)

webmat:
Ok, merging this now, as I need to make progress. Please chime in here if this causes issues, and I'll address those in a follow-up PR

PR #581, opened 2019-10-07
Created the Vulnerability Schema

peasead:
The thought is that this would be used for vulnerability scanners or associated projects (Qualys, Nexpose, Nessus, OpenVas, VulnWisperer, etc.) so they can send their data to Elasticsearch with ECS.

Some suggested fields were from #113 

Fields created for the `vulnerability-*` schema:
```
cvss.score.base
cvss.score.temporal
cvss.score.environmental
category
cve.description
cve.id
severity
status
detected.first
detected.last
scanned.first
scanned.last
```


schemas/vulnerability.yml
```diff
@@ -0,0 +1,144 @@
...
+    top_level: true
+    expected:
+      - client
+      - host
+      - os
+      - observer
+  type: group
+  fields:
+
+    - name: cvss.score.base
```
andrewstucki: It might be nice to version this since CVSS scores can vary dramatically depending on if v2 v. v3 calculations are used.

dainperkins: thats probably a good idea (at least for the base fields - iirc theres a bunch of differences)

peasead:
I had _over_ filtered Github notifications. I'm reviewing these comments now. Sorry for the delay.

schemas/vulnerability.yml
peasead: There are differences in the scoring between versions, but both include a Base, Temporal, and Environmental calculation.

I have added `cvss.score.version` and will update the PR.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,144 @@
...
+
+    Fields can have one entry or multiple entries.
+
+  reusable:
+    top_level: true
+    expected:
+      - client
+      - host
+      - os
+      - observer
```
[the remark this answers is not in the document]
peasead: Okay, thanks for that explanation @webmat. I was assuming that "reused" meant from other schemas. Based on the criteria you mentioned, I agree those shouldn't be reusable. 

I'll remove the `reusable`.

joshbressers:
It could make sense to have an arbitrary ID field. While many of the findings have CVE IDs, not all do. There are other vulnerability identifier that show up every now and then. Here's an example I pulled from Snyk
https://snyk.io/vuln/SNYK-JS-PM2-474345

On that note as well, being able to link to one or more reference URLs would be helpful. In the case of CVE IDs you could easily just link to MITRE or NVD. Other identifiers may be less obvious and it's possible some identifier won't even have a central location.

dainperkins:
> It could make sense to have an arbitrary ID field. While many of the findings have CVE IDs, not all do. There are other vulnerability identifier that show up every now and then. Here's an example I pulled from Snyk
> https://snyk.io/vuln/SNYK-JS-PM2-474345
> 
> On that note as well, being able to link to one or more reference URLs would be helpful. In the case of CVE IDs you could easily just link to MITRE or NVD. Other identifiers may be less obvious and it's possible some identifier won't even have a central location.

dainperkins:
good idea

vulnerability.scanner.id
vulnerability.scanner.name

or we could use the observer fields to detail scan vendor, etc.

webmat:
Yeah I like the idea of supporting an additional arbitrary ID. We need to find a good name for that, though. I don't think `vulnerability.arbitrary_id` works well ;-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,144 @@
...
+        For example (ex. https://qualysguard.qualys.com/qwebhelp/fo_portal/scans/vulnerability_status.htm)
+
+      example: New, Active, Fixed, Re-Opened
+
+    - name: detected.first
+      level: extended
+      type: date
+      short: Vulnerability first observed.
+      description: >
+        The first time that the vulnerability was observed on the asset.
```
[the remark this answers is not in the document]
peasead: I like this, great idea.

I updated.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,144 @@
+---
+- name: vulnerability
+  title: Vulnerability
+  group: 2
+  short: Fields to describe the vulnerabilty relevant to an event.
+  description: >
+    The vulnerability fields describe information about a vulnerabilty that is relevant to an event.
+
+    Fields can have one entry or multiple entries.
```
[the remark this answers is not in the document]
peasead: I think I took this from a template description, so that may be an artifact of that.

I'll remove.

peasead:
> good idea
> 
> vulnerability.scanner.id
> vulnerability.scanner.name
> 
> or we could use the observer fields to detail scan vendor, etc.

Awesome ideas @dainperkins and @stiltz 

I took those changes and updated the PR.

dainperkins:
Am I reading this correctly that the vulnerability fields end up as items nested under host/source/destination/observer, etc?

peasead:
> Am I reading this correctly that the vulnerability fields end up as items nested under host/source/destination/observer, etc?

@dainperkins The top level is `vulnerability` or am I misunderstanding what you're asking?

dainperkins:
generated/csv/fields.csv showed vulnerability as both top level and nested e.g. client.vulnerability...  Wasn't sure if that was on purpose, an option, or mistake.

vulnerability.name: Flash buffer overflow
vulnerability.client: [array of affected hosts]

vs.
vulnerability.name: Flash buffer overflow
host.name: [array of affected hosts]

could just be I was thinking about it differently but for reporting vunerabilities I was thinking the vulnerability info would all be top level, with teh addition of a [host] array or similar to define the vulnerability to vulnerable system mappings

webmat:
@dainperkins No, just a matter of re-running `make`. Was puzzled by that too, initially :-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      example: AIX, Firewall, Malware, Office Application, Solaris, Windows
+
+    - name: description
+      level: extended
+      type: keyword
+      short: Description of the vulnerability.
+      description: >
+        The description of the vulnerability that provides additional context of the vulnerability. For example (ex. https://cve.mitre.org/about/faqs.html#cve_entry_descriptions_created)
+
+      example: In Keybase before 2.12.6 on macOS, the move RPC to the Helper was susceptible to time-to-check-time-to-use bugs and would also allow one user of the system (who didn't have root access) to tamper with another's installs.
```
webmat: Since this example is pretty long, I'd prefer if we removed it, or cut it off early with an ellipsis...

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: scanner.id
+      level: extended
+      type: keyword
+      short: ID of the vulnerability.
+      description: >
+        The identification (ID) is the number portion of a vulnerability entry. It includes a unique identification number for the vulnerability. For example (ex. https://cve.mitre.org/about/faqs.html#what_is_cve_id)
+
+      example: CVE-2019-00001, CVE-2019-00002, QID: 38739
```
webmat: Just giving a single example here would be best.

When a list of examples is needed, we put the list of examples in the description. Our rendering of examples expects a single value.

Here's where you can preview it, btw: http://ecs_581.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-vulnerability.html  :-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: scanner.vendor
+      level: extended
+      type: keyword
+      short: Name of the scanner vendor.
+      description: >
+        The name of the vulnerability scanner vendor.
+
+      example: Tenable, Qualys
```
webmat: Single example as well, here. Pick your fav scanner ;-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: severity
+      level: extended
+      type: keyword
+      short: Severity of the vulnerability.
+      description: >
+        The severity of the vulnerability can help with metrics and internal prioritization regarding remediation. For example (ex. https://nvd.nist.gov/vuln-metrics/cvss)
+
+      example: Critical, High, Medium, Low, None
```
webmat: Please move the full list of examples to the description, and update to a single example in the `example:` attribute

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: status
```
webmat: Do vulnerability scanners keep track of past runs, and can they report a status of "re-opened"? Or is this more of a concept that you'd expect to see in a UI that knows about that past history of scans?

I'm wondering if it's a situation like the first/last seen timestamps we discussed, last time around :-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      level: extended
+      type: keyword
+      short: Status of the vulnerability.
+      description: >
+        The current status of the vulerability on the asset:
+
+        New - The first time a vulnerability is detected by a scan the status is set to New.
+        Active - A vulnerability detected by two or more scans is set to Active.
+        Fixed - A vulnerability was verified by the most recent scan as fixed, and this vulnerability was detected by the previous scan.
+        Re-Opened - A vulnerability was reopened by the most recent scan, and this vulnerability was verified as fixed by the previous scan.
```
webmat: Good stuff, love this list.

They're tricky to get right in the ECS YAMLs. Another place we have them is for [network.direction](https://github.com/elastic/ecs/blob/v1.2.0/schemas/network.yml#L84). Please start each line with `*`, see if it renders better in the docs preview http://ecs_581.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-vulnerability.html

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+        For example (ex. https://qualysguard.qualys.com/qwebhelp/fo_portal/scans/vulnerability_status.htm)
+
+      example: New, Active, Fixed, Re-Opened
```
webmat: Single example here as well

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: id
+      level: extended
+      type: keyword
+      short: Scan identification number.
+      description: >
+        The report or scan identification number.
+
+      example: 20191018.0001, SCAN0001
```
webmat: Single example here as well

schemas/vulnerability.yml
webmat: Two more things about this example.

1) I think we should use a CVE as the example, because it's the most widely known kind of vulnerability IDs. So it'll be best at explaining the purpose of the field.

2) But the QID is interesting, because is a string that includes spaces. I wonder if we need to have this reformatted without spaces (either "QID:38739" or "QID-38739"). 

It's not a requirement from the Elasticsearch or Kibana POV, a keyword field can totally contain spaces and it'll still be interpreted as a single keyword.

I just wonder if it will confuse the humans :-)

Any thoughts on this @ruflin, @MikePaquette ?

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+        The classification of the vulnerability scoring system. For example (ex. https://www.first.org/cvss/)
+
+      example: CVSS
+
+    - name: enumeration
+      level: extended
+      type: keyword
+      short: Enumeration of the vulnerability.
+      description: >
+        The enumeration of the common identifiers for publicaly known vulnerabilities. For example (ex. https://cve.mitre.org/about/)
```
webmat: I think it's not obvious enough that this is meant to capture the "type of vulnerability ID" (CVE), not the actual ID (CVE-2019-0000).

Perhaps update the description with more explicit wording like the following:

> The type of identifier used for this vulnerability, for example "CVE-2019-0000" would have `vulnerability.enumeration:CVE`

Perhaps the word "enumeration" has this clear meaning for security specialists, but it's not obvious to a mere security fan like me :-)

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+    - name: score.temporal
+      level: extended
+      type: long
+      short: Vulnerability Temporal score.
+      description: >
+        Scores can range from 0.0 to 10.0, with 10.0 being the most severe.
+
+        Temporal scores cover an assessment for code maturity, remediation level, and confidence. For example (ex. https://www.first.org/cvss/specification-document)
+
+    - name: score.evironmental
```
webmat: Typo :-)

```suggestion
    - name: score.environmental
```

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+        New - The first time a vulnerability is detected by a scan the status is set to New.
+        Active - A vulnerability detected by two or more scans is set to Active.
+        Fixed - A vulnerability was verified by the most recent scan as fixed, and this vulnerability was detected by the previous scan.
+        Re-Opened - A vulnerability was reopened by the most recent scan, and this vulnerability was verified as fixed by the previous scan.
+
+        For example (ex. https://qualysguard.qualys.com/qwebhelp/fo_portal/scans/vulnerability_status.htm)
+
+      example: New, Active, Fixed, Re-Opened
+
+    - name: id
```
webmat: I think we should rename this. If it's meant to capture the ID of the report, I would name this `vulnerability.report_id`.

To me `vulnerability.id` sounds like the identifier of one of the findings in the report.

webmat:
How do we expect people to use the `vulnerability` field set? My current understanding is that this is meant to import report findings, not to enrich events with vulnerability details. Correct?

If the idea is indeed to import scan findings, should we explore support for two related kinds of documents?

- document for the report itself (not explicitly addressed right now)
- documents for each finding of the report (what this PR currently describes)

Or are we good with just documents per findings + the report ID?

cc @joshbressers @stiltz

schemas/vulnerability.yml
webmat: Yet another comment about this example 😂 The `QID: 38739` here breaks Python's YAML parser. Moving to a single example with one of the CVEs will fix that.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+    - name: enumeration
+      level: extended
+      type: keyword
+      short: Enumeration of the vulnerability.
+      description: >
+        The enumeration of the common identifiers for publicaly known vulnerabilities. For example (ex. https://cve.mitre.org/about/)
+
+      example: CVE
+
+    - name: url
```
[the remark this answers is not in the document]
peasead: Done.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      type: keyword
+      short: URL describing the vulnerability.
+      description: >
+        A URL to a resource that provides additional information, context, and mitigations for the identified vulnerabiltiy. For example (ex. https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=ssh)
+
+      example: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-6111
+
+    - name: score.base
+      level: extended
+      type: long
```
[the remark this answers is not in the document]
peasead: Done.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      description: >
+        Scores can range from 0.0 to 10.0, with 10.0 being the most severe.
+
+        Base scores cover an assessment for exploitability metrics (attack vector, complexity, privileges, and user interaction), impact metrics (confidentialy, integrity, and availability), and scope. For example (ex. https://www.first.org/cvss/specification-document)
+
+      example: 5.5, 6.6
+
+    - name: score.temporal
+      level: extended
+      type: long
```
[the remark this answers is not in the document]
peasead: Done.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      description: >
+        Scores can range from 0.0 to 10.0, with 10.0 being the most severe.
+
+        Environmental scores cover an assessment for any modified Base metrics, confidentiality, integrity, and availability requirements. For example (ex. https://www.first.org/cvss/specification-document)
+
+      example: 5.5, 6.6
+
+    - name: score.version
+      level: extended
+      type: long
```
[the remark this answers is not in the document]
peasead: Done.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,149 @@
...
+      short: Vulnerability Temporal score.
+      description: >
+        Scores can range from 0.0 to 10.0, with 10.0 being the most severe.
+
+        Temporal scores cover an assessment for code maturity, remediation level, and confidence. For example (ex. https://www.first.org/cvss/specification-document)
+
+    - name: score.evironmental
+      level: extended
+      type: long
```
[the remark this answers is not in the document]
peasead: Done.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: This is now formatted the same way as `network.direction`.

elasticcla:
Hi @peasead, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecs/pull/581.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

schemas/vulnerability.yml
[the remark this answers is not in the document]
webmat: 🧐 you renamed the field above 🙂 

Here's what I meant:

* `vulnerability.url` should become `vulnerability.reference`
* The field you have renamed used to be named `vulnerability.enumeration`. I'm ok with keeping that original field name, as long as the description is fleshed out a little, like described [here](https://github.com/elastic/ecs/pull/581#discussion_r346985294).

schemas/vulnerability.yml
```diff
@@ -0,0 +1,176 @@
...
+    - name: severity
+      level: extended
+      type: keyword
+      short: Severity of the vulnerability.
+      description: >
+        The severity of the vulnerability can help with metrics and internal
+        prioritization regarding remediation.
+        For example (https://nvd.nist.gov/vuln-metrics/cvss)
+
+      example: Critical
```
webmat: @randomuserid @dainperkins @MikePaquette @FrankHassanabad 

I'd like us to discuss `vulnerability.severity` a bit.

ECS already has a few fields close to this. I don't think we should use them, but perhaps we need to call out to them, one way or another.

- `event.severity` is numeric, so not a good fit for this.
- `log.level` and `event.severity` are meant to be equivalent, where `log.level` is the textual representation of `event.severity`. But I also wouldn't expect importing vulnerabilities and having their severity in `log.level`. So also a no-go.

But should we give instructions to e.g. copy `vulnerability.score.base` to `event.severity`? Any other opinions on this?

schemas/vulnerability.yml
```diff
@@ -0,0 +1,176 @@
...
+      type: keyword
+      short: Description of the vulnerability.
+      description: >
+        The description of the vulnerability that provides additional context of the
+        vulnerability.
+        For example (https://cve.mitre.org/about/faqs.html#cve_entry_descriptions_created)
+
+      example: In macOS before 2.12.6, there is a vulnerability in the RPC...
+
+    - name: scanner.id
```
webmat: This field name bugs me. It sounds like an internal ID for a single finding, when actually the example shows the public ID of a vulnerability.

If this is meant to be the public ID, I'd have this straight at `vulnerability.id`

CHANGELOG.next.md
```diff
@@ -3,18 +3,9 @@
 - Please insert your changelog line ordered by PR ID.
 Thanks, you're awesome :-) -->
 
-## Unreleased
-
-### Breaking changes
-
-### Bugfixes
-
```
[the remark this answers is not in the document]
peasead: Done.

CHANGELOG.next.md
```diff
@@ -3,18 +3,9 @@
...
 ### Added
 
-### Improvements
-
-### Deprecated
-
```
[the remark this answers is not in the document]
peasead: Done.

CHANGELOG.next.md
```diff
@@ -24,8 +15,6 @@ Thanks, you're awesome :-) -->
 
 ### Bugfixes
 
-### Added
-
```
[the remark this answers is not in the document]
peasead: Done.

peasead:
> A few more things to address, see review comments below.
> 
> I would also like clarifications on a few questions:
> 
> * @peasead do you think we need to be able to ingest a document representing the report itself, in addition to each finding? Or are we good with the report details just being referenced via `vulnerability.report_id`?

I think that the `vulnerability.report_id` is sufficient.

> * Do we need `vulnerability.status`? See longer version of the question [here](https://github.com/elastic/ecs/pull/581/files?file-filters%5B%5D=.csv&file-filters%5B%5D=.md&file-filters%5B%5D=.yml#r346969591)

I think that this is probably scanner/UI dependent and if we removed it, we'd be okay. To whit, I'm not sure other scanners beyond Qualys track this information, so it may be very scanner dependent. I'll remove this.

> * There's a few fields that I would like to know whether we should expect a single value or an array of values, please let me know your thinking about each of them:
>   
>   * `vulnerability.category`
>   * `vulnerability.scanner.id`

I think `category` and `scanner_id` would generally be a single value. In the event that they cover more than one category/ID, I believe that there'd be an additional entry, which would remove the need for an array consideration.

> Additional note on the array question above. Even if some scanners make a decision to pick the "best match" (or highest severity) when assigning a category or vuln ID to a finding, that's not the standard we have to match. What we have to cater to is if _any_ scanner can associate multiple categories or return multiple CVE IDs for a finding, then we should make these field arrays.
> 
> Note that there's no direct support for arrays in ECS right now, but it's coming. So any of those should be arrays, we would simply mention it in the description, and give an example that's actually an array (`example: '[CVE-2019-0001]'`).

While I agree that we don't want to be in the business of assigning a category, we're collecting the data that is being presented by the scanner. So if the scanner says "Firewall", we should take the observer at its word vs. trying to maintain a matrix.

schemas/vulnerability.yml
peasead: Severities are generally calculated by the scanner through an internal algorithm. I can't say definitively if the "high" severity by Qualys is the same as the "high" from Tenable, but like the `vulnerability.category`, I think we should let the scanner decide on this vs. us maintaining some matrix.

schemas/vulnerability.yml
peasead: Agreed. Makes sense.

schemas/vulnerability.yml
webmat: Totally agree on that point, I don't want to introduce a mapping of these values.

And likewise, if we were to instruct people to copy e.g. the base score out to a more central field like `event.severity`, the idea is still that the value is source-specific, not mapped.

But perhaps we shouldn't worry about this for now, and only address this if/when we have a concrete need for that. I looked around SIEM, and `event.severity` is not a default column anywhere that I could see.

webmat:
> I think category and scanner_id would generally be a single value. In the event that they cover more than one category/ID, I believe that there'd be an additional entry, which would remove the need for an array consideration.

By additional entry here, do you mean a distinct field? In other words, would they have a "main category" and a list of other applicable categories?

>> Additional note on the array question above. Even if some scanners make a decision to pick the "best match" (or highest severity) when assigning a category or vuln ID to a finding, that's not the standard we have to match. What we have to cater to is if any scanner can associate multiple categories or return multiple CVE IDs for a finding, then we should make these field arrays.
> While I agree that we don't want to be in the business of assigning a category, we're collecting the data that is being presented by the scanner. So if the scanner says "Firewall", we should take the observer at its word vs. trying to maintain a matrix.

I think my comment was misunderstood there. I don't want to override or map anything there. What I'm saying is much more simple: if any scanner is going to give us an array of categories or multiple CVEs for a finding, I want our field to be an array that can take it all in as is. This would also mean that in most cases when we have a single category/CVE id, we'd have arrays of 1 item, but that's fine.

webmat:
I'll take care of the CHANGELOG.next.md conflict at merge time, btw. No need to worry about that. If you fix now, it may complain again after the next PR gets merged 🙂

schemas/vulnerability.yml
```diff
@@ -0,0 +1,159 @@
...
+      example: CVE
+
+    - name: reference
+      level: extended
+      type: keyword
+      short: Reference of the vulnerability.
+      description: >
+        A resource that provides additional information, context, and
+        mitigations for the identified vulnerabiltiy.
+        For example (https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=ssh)
```
webmat: I would remove this line. I think the `example:` attribute below is sufficient.

schemas/vulnerability.yml
```diff
@@ -0,0 +1,159 @@
+---
+- name: vulnerability
+  title: Vulnerability
+  group: 2
+  short: Fields to describe the vulnerabilty relevant to an event.
+  description: >
+    The vulnerability fields describe information about a vulnerabilty that is
+    relevant to an event.
```
webmat: Typo first "vulnerabilty" => "vulnerability", then a more philosophical point:

The description of the field set makes it sound like users are expected to enrich "normal" events with vulnerability information. Is that our expectation?

My expectation would be more along the lines of users *directly importing scan findings*. In other words, this isn't another kind of event being enriched with this info. The event/document coming in is literally one finding from the report, which is getting imported for correlation in the SIEM.

If you agree with this, we should address this directly. Perhaps:

```YAML
  short: Fields to describe a vulnerability finding.
  description: >
    The vulnerability fields describe information about a vulnerability that was
    reported by a scan.
```

schemas/vulnerability.yml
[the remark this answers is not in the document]
peasead: Sorry about that. I've made the adjustments.

peasead:
> How do we expect people to use the `vulnerability` field set? My current understanding is that this is meant to import report findings, not to enrich events with vulnerability details. Correct?
> 
> If the idea is indeed to import scan findings, should we explore support for two related kinds of documents?
> 
> * document for the report itself (not explicitly addressed right now)
> * documents for each finding of the report (what this PR currently describes)
> 
> Or are we good with just documents per findings + the report ID?
> 
> cc @joshbressers @stiltz

To address your question on how the `vulnerability.*` field-set could be used, I think both importing vulnerability information as well as enrichment. Importing scan findings is obvious, but this information can also be used to enrich if you're tracing an event back to "patient zero", knowing what vulnerabilities could have been leveraged is important to understand adversary:
* resourcing
* sophistication
* reconnaissance
* capabilities

In the second part of your question regarding documents per findings and report ID, I think that having the report ID is sufficient to guide the analyst to a place to get additional information i.e. the scanner vs. importing the entire scan report. I see a lot of products in the intelligence space that attempt to replicate "all the things" instead of providing a launch point for analysts. 

Maybe we could adjust `vulnerability.reference` to be a link to the scanner report instead of an external reference? <-- I like this idea.

Just my $0.02 from some time the field.

schemas/vulnerability.yml
peasead: Done.

schemas/vulnerability.yml
peasead: Fixed typo.

I think adjusting this description makes sense. That said, just as some context: https://github.com/elastic/ecs/pull/581#issuecomment-555550726

schemas/vulnerability.yml
```diff
@@ -0,0 +1,158 @@
...
+      level: extended
+      type: keyword
+      short: Category of a vulnerability.
+      description: >
+        The type of system or architecture that the vulnerability affects. These may be
+        platform-specific (for example, Debian or SUSE) or general (for example, Database
+        or Firewall).
+        For example (https://qualysguard.qualys.com/qwebhelp/fo_portal/knowledgebase/vulnerability_categories.htm)
+
+      example: Firewall
```
webmat: To make this an array, here's what I have in mind. Feel free to adjust:

```
      description: >
        The type of system or architecture that the vulnerability affects. These may be
        platform-specific (for example, Debian or SUSE) or general (for example, Database
        or Firewall).
        For example (https://qualysguard.qualys.com/qwebhelp/fo_portal/knowledgebase/vulnerability_categories.htm)

        This field must be an array.

      example: '["Firewall"]'
```

I don't know if we need to expand on main category being first + other categories coming next. We can add that later, if there's confusion.

peasead: > * I'm also good on only capturing the report ID and not capturing "all the things". I just wanted to make sure this was also how you saw it. Looks like we're on the same page. 👍

Roger roger.

> Maybe we could adjust vulnerability.reference to be a link to the scanner report instead of an external reference? <-- I like this idea.
> 
> * I think we have a need for both URLs.
>   
>   * For now I would say `.reference` should remain a URL that points to general documentation about a thing, in this case the CVE. This is the semantics we've been establishing, so far for "reference URL" fields.
>   * I also really love the idea of capturing the URL an analyst would want to visit, in order to view the actual finding in their vulnerability scanner. It's another type of URL we will need in a few places (e.g. endpoint alerts), but we don't have a good name for that field yet. So IMO we leave this out of this PR, and add it as a follow-up PR.

Agreed, this makes sense.

> On category and ID as arrays:
> 
> My thinking there isn't whether a published vulnerability could have more than one ID or category necessarily. It's rather "can one scanner finding correspond to more than one CVE", or could it have more than one category (orthogonal ones: "Windows" & "Firewall", or a wide problem: "SUSE" + "Debian" + "RHEL" ).
> 
> I do think it would be better to go on the safe side here, and the effort is pretty trivial. Elasticsearch actually doesn't care if docs contain a single value vs an array. The reason in ECS we try to flesh this out more explicitly is that the pipelines handling the events or generated language libs like [ecs-dotnet](https://github.com/elastic/ecs-dotnet) have to know, in order to do the right thing.
> 
> We could even promote lightweight semantics, where if a scanner has such a thing as "main category" and "other categories", the main category should be first in the array.
> 
> To close on this, I'm 99% sure it's useful for category; for ID I'm not sure yet, I'm starting to think not. Perhaps the way scanners are structured means each CVE has its own detection, and a given finding will always only have one CVE? In other words the same piece of software would simply have multiple findings (one per CVE) if applicable...
> 
> See PR comment, for a concrete suggestion on what we can do to make category into an array.

I like this and will make the adjustment declaring that this must be an array and formatting the example properly for `.category`.

PR #585, opened 2019-10-16
Add `url` to package definition

simitt:
Add information from where the package was installed.

related to #515

Follow up from https://github.com/elastic/ecs/pull/532 to focus on adding `URL` from where a software package was installed.

----

Follow up from previous discussions https://github.com/elastic/ecs/pull/532#discussion_r318443653 and https://github.com/elastic/ecs/pull/532#discussion_r319697166 

Last comment from https://github.com/elastic/ecs/pull/532#discussion_r329182793:

>  webmat 16 days ago Member
>
>I was initially misunderstanding the purpose of this field. Its description is not clear enough, and the example is misleading (Docker also has a registry of packages in Docker Hub).
>
>Please remove the field from this PR, and reopen a PR with just this field.
>
>When reopening the new PR, please adjust the following things to remove the ambiguity:
>
>    clarify the description, perhaps something like "Main URL of the software included in the package"
>    change the example so it's aligned with the other examples in the package field set. In other >words, this should be Golang's URL, which according to Homebrew, is simply "https://golang.org".
>
>Another discussion point that I'd like to hash out: instead of partially nesting the URL field set here, I think we could use another pattern we've started using elsewhere, and name the field .reference. WDYT?

@webmat I didn't find a lot of context around using `url.reference`, but from the examples I saw it really looked like it was pointing to a page with a threat reference. Not sure this would translate to the package url here, so for now I went with adding `url.original`. Let's discuss this.


ruflin:
I think this will need a changelog entry.

simitt:
Giving this some more thoughts I am in favor of changing to `package.url.reference`, and making it clear in the description that this URL is e.g. the URL the package manager is pointing to or the URL of the packages project site, included in the package itself (as pointed out by @cwurm in https://github.com/elastic/ecs/pull/532#discussion_r318958329)

SergeyKleyman:
@simitt It might be helpful to understand the differences between these fields if you provide a couple of examples where fields are different. I would recommend examples based on packages installed by sysadmin (such as Debian packages) and packages coming with an application (such as the packages from Java's Maven repository).

simitt:
I changed the name and description. @cwurm and @webmat could you have another look please.

webmat:
I'll address the changelog conflict when merging

Issue #589, opened 2019-10-17
Flesh out support for multiple users in an event


webmat:
We need to flesh out support for multiple users in an event to ECS, when there is a semantic relationship between them. Here are some use cases (please comment if you see additional use cases):

1. User A manages User B (create, modify, delete)
2. Directory User A logs into a machine as User B
3. User A escalates privileges / assumes identity of User B

This has been discussed a bit in the past in #234 , but no concrete plans were made at the time.

ECS currently supports `source.user` and `destination.user`, which may be enough to satisfy the requirements for no 2 above. However since `source` and `destination` are both designed to represent each side of a network connection, the semantics may be problematic when trying to use `source.user` and `destination.user` for no 1 and 3, as these can be purely local events.

*Questions*

Has anyone used `source.user` and `destination.user` in ECS to model these use cases? If so, did this work well, and are there edge cases where it doesn't work as well?

Should we try to model all cases with one naming recommendation for all use cases? As an example, should we always use `user` & `affected.user`. Or should we adjust the naming depending on semantics? E.g. 1) `user` & `affected.user`, 2) `source.user` & `destination.user`, and a different way for 3).

*Points to consider*

* Some schemas use the word "target" to describe the affected user. However the word also has the connotation of sounding like a victim, or the goal of an investigation. It may be a good idea to avoid using this word if we can.
* It may be a good idea to define the non-nested `user` as the user initiating the action in an event, rather than nesting both sides. This way searching for `user.name:bob` would return any events where "bob" was doing something. Searching for events affecting "bob" would require searching for the more qualified field name, like e.g. `affected.user.name:bob`. Pivoting to find anything related to "bob" can somewhat be done in Kibana by searching for `*.name:bob` (not supported via the ES API), or could be done by populating `related.user.name` systematically in all sources, then searching for `related.user.name:bob`.

*Naming ideas*

Please comment with additional naming ideas.

* `user` & `affected.user`
* `user` & `target.user`
* `subject.user` & `target.user` (like Windows)
* `user` & `assumed.user` (privilege escalation)
* ... (your suggestions)

webmat:
cc @neu5ron, @eternalyperplxed, @vbohata, @willemdh 

Your thoughts on this would be appreciated, as always :-)

webmat:
cc @rw-access if you can comment, or cc the right Endgamer to chime in for this discussion :-)

willemdh:
Personally I think for clarity, it would be good to nest all user related data under `user.*`

[...]

PR #591, opened 2019-10-22
Add host.domain field

janniten:
I'm working with windows events 4741,4742 and 4743.
When mapping into ECS I realize that I need a field host.domain 
I found also that host.domain is necessary when mapping certain Fortigate Logs

### DIFF
--- a/schemas/host.yml
+++ b/schemas/host.yml
@@ -88,5 +88,5 @@
         Name of the directory the group is a member of.
 
         For example, an LDAP or Active Directory domain name.
-
+      example: CONTOSO 
 
--- a/schemas/host.yml
+++ b/schemas/host.yml
@@ -85,8 +85,9 @@
       type: keyword
       short: Name of the directory the group is a member of.
       description: >
-        Name of the directory the group is a member of.
+        Name of the domain of which the host is a member. 
 
-        For example, an LDAP or Active Directory domain name.
+        For example, on Windows this could be the host's Active Directory domain or NetBIOS domain name. 
+        For Linux this could be the domain of the host's LDAP provider.
       example: CONTOSO
 
--- a/schemas/host.yml
+++ b/schemas/host.yml
@@ -79,3 +79,14 @@
       example: "x86_64"
       description: >
         Operating system architecture.
+
+    - name: domain
+      level: extended
+      type: keyword
+      short: Name of the directory the group is a member of.
+      description: >
+        Name of the directory the group is a member of.
+
+        For example, an LDAP or Active Directory domain name.
+
+
--- a/schemas/host.yml
+++ b/schemas/host.yml
@@ -88,5 +88,5 @@
         Name of the directory the group is a member of.
 
         For example, an LDAP or Active Directory domain name.
-      example: CONTOSO 
+      example: CONTOSO
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -10,6 +10,7 @@ Thanks, you're awesome :-) -->
 ### Bugfixes
 
 ### Added
+* Added `host.domain` field. #591
 
 ### Improvements
 
--- a/code/go/ecs/host.go
+++ b/code/go/ecs/host.go
@@ -59,7 +59,9 @@ type Host struct {
 	// Operating system architecture.
 	Architecture string `ecs:"architecture"`
 
-	// Name of the directory the group is a member of.
-	// For example, an LDAP or Active Directory domain name.
+	// Name of the domain of which the host is a member.
+	// For example, on Windows this could be the host's Active Directory domain
+	// or NetBIOS domain name.  For Linux this could be the domain of the
+	// host's LDAP provider.
 	Domain string `ecs:"domain"`
 }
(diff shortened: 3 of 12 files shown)


webmat:

neu5ron:
@janniten what specific windows/beats field name are referring to mapping to `host.domain`?  
I want to make sure at the same time if we should add `host.fqdn`

@webmat this LGTM otherwise, definitely necessary.

janniten:
Hi @neu5ron 
Some examples of specific fields are in windows events 4741,4742 and 4743.
Those events are related to computer account management, i.e, computers as objects in an Active Directory structure.
The field winlog.data.TargetDomainName in these events represent the domain in which the computer account was created. This is an Active Directory Domain name diferent to a fqdn which is a DNS domain name.

As an analogy there exists the user.domain and group.domain which are populated in the same way when  user or group account events are mapped.

janniten:
> I want to make sure at the same time if we should add `host.fqdn`
@neu5ron I think this is neccesary too

webmat:
Yes now that we're fleshing out support for AD/LDAP domains (user, group, now host), perhaps this will create confusion vs DNS domain names (source, dest, client, server, url). We'll see if it becomes a problem, and we can adjust if that's the case.

But for now, let's confidently move forward ;-)

- [ ] Please add an example for the field
- I'd say let's not add fqdn just yet

Once we have the example in place, I think we can merge this. Thanks for your contribution!

Finally, this is unrelated to this PR specifically, but related to your work here and on https://github.com/elastic/beats/pull/14299, I'd love to get your thoughts on https://github.com/elastic/ecs/issues/589, @janniten.

janniten:
Hi @webmat 
I have added the example in the host.yml (same domain Microsoft uses in their examples)
I'll take a look into the discussion in https://github.com/elastic/ecs/issues/589 and leave my comments
Regards!

schemas/host.yml
```diff
@@ -79,3 +79,14 @@
...
+
+    - name: domain
+      level: extended
+      type: keyword
+      short: Name of the directory the group is a member of.
+      description: >
+        Name of the directory the group is a member of.
+
+        For example, an LDAP or Active Directory domain name.
+      example: CONTOSO
```
webmat: 🤔 I expected this would look more like a domain name, actually.

I suppose it's something optional? Companies that run their domain 100% internal could name their domain CONTOSO, and if the company wanted to make this publicly accessible, it may be "contoso.net"?

Or was the `.fqdn` discussion meant to be about the FQDN of the domain controller? When I read it, I understood `host.fqdn` to be the fqdn of the host, which could in this case be "bob-laptop.CONTOSO"?

webmat: > same domain Microsoft uses in their examples

Love the approach in any case :-)

janniten: Hi @webmat it is only a name (can be anything) that refers to an Active Directory Domain or Windows Domain.
It is not optional, always exists. If you do not have an AD the domain name in the event is the name of the local computer. 
Users, Groups and Computers (and in general, objects within the domain) are referenced as domain_name\object-name.
In the case of bob's laptop the reference will be CONTOSO\bob.laptop 
It is not related to domain name in the context of DNS nor with the fact that the domain permit the object will be publicly accesible.
Real domain name examples and user names

An old company I worked for (and doesn't exist anymore :) )
domain: NEXTEL_AR

My home local domain
TargetUserName	Anna
TargetDomainName	DESKTOP-4564S0K

Either is an Active Directory or an stand alone computer in a private network , Windows always asigns a domain name :)

schemas/host.yml
```diff
@@ -79,3 +79,14 @@
       example: "x86_64"
       description: >
         Operating system architecture.
+
+    - name: domain
+      level: extended
+      type: keyword
+      short: Name of the directory the group is a member of.
+      description: >
+        Name of the directory the group is a member of.
```
andrewkroh: I would think this would read like:

"Name of the domain of which the host is a member. For example, on Windows this could be the host's Active Directory domain or NetBIOS domain name. For Linux this could be the domain of the host's LDAP provider."

janniten: Fixed!

janniten:
Working today with events 4741-2-3 came up an observation about the use of the fields host.* for this events.
@andrewkroh @webmat What do you think?

https://discuss.elastic.co/t/winlogbeat-evt-4741-4742-4743-computer-management-events/206069

janniten:
@andrewkroh @webmat  
I found that mapping event log's Computer Management Events  to hosts.* will cause some inconsistencies.
For all events, the host.name field is filled with the host that is the source of the event. I think I cannot use the host.name in order to put the name of the computer name being created/deleted/modified.
Instead of that maybe we can introduce some winlog especifc fields in order to map those fields
I propose these ones: 
winlog.event_data.TargetSid                 -> winlog.computer.id
winlog.event_data.TargetUserName     -> winlog.computer.name
winlog.event_data.TargetDomainName -> winlog.computer.domain

What do you think?
Thank You

andrewkroh:
@janniten I think the issue you describe relates to #589. While that issue is about multiple users (source and target) I think the concept of a "target" (or whatever we call it) should expand to include a target machine.

I think this PR is good to merge. @webmat agree?

PR #592, opened 2019-10-22
Add link to ecs-logging project, pointing users to log formatters that support ECS.

webmat:


webmat:
Should be merged only when the readme actually lists the logging libraries :-)

webmat:
Actually I'll tone down the "many languages" for now, since there's only two present at the destination :-)

@ruflin You asked (forget where) why this wasn't added directly to the main ECS repo. For now I think the separation makes sense, as the other repo may end up containing implementation details that are common to the logging libraries and formatters, and not directly related to the schema itself. Also this is a greenfield project, so moving fast over there makes sense, until the initiative finds its stride.

If the repo ends up not being needed, agreed we could perhaps eliminate it later.

PR #599, opened 2019-10-29
Add process.command_line

webmat:


schemas/process.yml
```diff
@@ -45,15 +45,26 @@
...
+    - name: command_line
+      level: extended
+      type: keyword
+      short: Full command line that started the process.
+      description: >
+        Full command line that started the process, including the absolute path
+        to the executable, and all arguments.
+
+        Some arguments may be filtered to protect sensitive information.
+      example: "/usr/bin/ssh -l user 10.0.0.16"
```
andrewkroh: When do we use `process.command_line` vs `process.args`? They seem a little redundant.

webmat: I totally agree they are redundant.

Having the args as an array is exceptionally powerful (more than I initially realized) when detections are looking for an exact term in a command.

However especially in the context of endpoint detections, many of them will still have to rely on fuzzier searches e.g. with wildcards, to detect some attack patterns such as obfuscated commands, base64 decoding, webshells & so on.

So both fields are actually needed, in order to perform each class of detections.

Also unrelated to detections, displaying `process.command_line` in Kibana will be much nicer :-)

cc @rw-access  @randomuserid

rw-access: Yeah exactly.
I think **process.args** will have more utility for search than **process.command_line**, but **process.command_line** gives us full recoverability of the original command line as presented to the OS. For one odd edge case, cmd.exe essentially puts 2 spaces between arg[0] and arg[1], but arg parsing is lossy, so we lose that artifact.

PR #606, opened 2019-11-12
Add TLS schema to ECS

dcode:
I compared the output of Zeek 3.x, Suricata 5.x, and Packetbeat 7.5 and selected these fields as a common set that would cover the TLS protocol. You can see my [scratchpad](https://gist.github.com/dcode/df8670ef49413aa8b5b7908c3491ff27#file-tls_schema-yaml) where I compared field names and came up with an example data record (YAML formatted).

While selecting fields, I aimed to keep only the data that centers around the SSL/TLS negotiation and avoid digging too deep into the x.509 certificates themselves. The certificates contain much more data than is extracted here, but this is similar to how Zeek and Suricata handle SSL/TLS protocol metadata by default, where only the high-level data is extracted. 

With that in mind, there should be sufficient information recorded that will allow an analyst to pivot from a TLS event to a first class file object document (not currently defined by ECS) that extracts and analyzes all the x.509 fields. Zeek provides this as a separate x.509 event, for example. This data could also be provided by recursive file scanners like [Strelka](https://github.com/target/strelka/) or potentially endpoint agents.

**Formatting Question**: I broke all the "objects" into their own group. It might be better to just use a dotted field notation, understanding that the tooling takes care of the rest. I'm not sure on this one.


schemas/tls.yml
```diff
@@ -0,0 +1,189 @@
...
+    Fields related to TLS activity.
+  type: group
+  fields:
+
+    - name: version
+      level: extended
+      type: group
+      description: Fields related to the SSL/TLS version used.
+      fields:
+        - name: original
```
andrewkroh: I wonder how useful this one will be since we have a normalized protocol name and version number. When information is coming from a log we will still have the original value in the `message` (albeit unparsed). When the info is coming directly from network traffic there isn't really a value for this unless we write some bytes (e.g. `0x303`). Perhaps we don't need this field.

schemas/tls.yml
```diff
@@ -0,0 +1,189 @@
...
+          description: Original, unparsed version string provided by generating application.
+
+        - name: protocol
+          level: extended
+          type: keyword
+          description: Normalized protocol name parsed from original string (e.g. ssl, tls).
+
+        - name: number
+          level: extended
+          type: number
```
andrewkroh: This should probably be treated as a keyword. Otherwise it would need to be indexed as a float.

schemas/tls.yml
```diff
@@ -0,0 +1,189 @@
...
+  description: >
+    Fields related to TLS activity.
+  type: group
+  fields:
+
+    - name: version
+      level: extended
+      type: group
+      description: Fields related to the SSL/TLS version used.
+      fields:
```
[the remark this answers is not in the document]
dcode: There was some formatting issues, but I have since flattened this and I think fixed this issue.

schemas/tls.yml
dcode: That's a fair point. I just noticed that Suricata, Zeek, and Packetbeat all provide a differently formatted value for this field, but if that's being preserved in `message`, that's probably good enough.

schemas/tls.yml
dcode: Thanks. Fixed. I just went through this all blindly, and @webmat told me how to run the other tools this morning. 👍

webmat:
> Thanks for providing the output from the three different tools.

Where is this? :-) I'm curious as a 🐱

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    - name: version.protocol
+      level: extended
+      type: keyword
+      description: Normalized protocol name parsed from original string (e.g. ssl, tls).
+
+    - name: version.number
+      level: extended
+      type: keyword
+      description: Numeric part of the version parsed from the original string (e.g. 1.2, 3).
```
webmat: These two fields cannot be nested under `tls.version`, as this is a hard break vs Packetbeat's current definition for `tls.version`, where the field is type `keyword`. This would lead to a mapping conflict within the 7.x product line, which we can't do.

What would work well, I think is the following:

- `tls.version` (`keyword`)
  - Field content would need to move from Packetbeat's content of e.g. "TLS 1.3" to "1.3", however
  - perfectly aligned with all other ECS `*.version` fields.
- `tls.version_protocol` (`keyword`)
  - This is where we would put either "tls" or "ssl"
  - The underscore avoids the mapping conflict
  - The definition of `tls.version_protocol` should mention that the protocol must be lowercased, in order to be aligned with network fields such as `network.protocol`, `network.application`, `network.type` and `network.transport`.

What I'm laying out above however still leads to a minor break within the 7.x product line, where Packetbeat would have to start putting "pure" version values such as "1.2", "1.3" instead of its current "TLS 1.2", "TLS 1.3".

I'd like to get people's take on that kind of break for the `tls.version` field's content for Packetbeat. @andrewkroh @tsg @ruflin WDYT?

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+      type: keyword
+      level: extended
+      description: String indicating the curve used for the given cipher, when applicable.
+
+    - name: resumed
+      type: boolean
+      level: extended
+      description: Boolean flag indicating if this TLS connection was resumed from an existing TLS negotiation.
+
+    - name: next_protocol
```
webmat: This field should also mention that the value must be lowercased, just like the slew of other protocol fields I mentioned in the comment above :-)

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    - name: client.not_after
+      type: date
+      level: extended
+      description: Timestamp indicating when client certificate is no longer considered valid.
+
+    - name: client.certificate_chain
+      type: keyword
+      level: extended
+      description: List of PEM-encoded certificates that make up the certificate chain offered by the client.
```
webmat: Like we discussed, it would be good to mention in both `*.certificate` and `*.certificate_chain` that they are more or less mutually exclusive. If the source has the cert chain fill that, and cert can be omitted. If only cert is available, then fill that one instead.

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+      type: boolean
+      level: extended
+      description: Boolean flag indicating if this TLS connection was resumed from an existing TLS negotiation.
+
+    - name: next_protocol
+      type: keyword
+      level: extended
+      description: String indicating the protocol being tunneled (e.g. http/1.1, spdy/3, imap, webrtc)
+
+    - name: client.ja3
```
webmat: I'm not an expert on the subject, but is it intentional that the 2 JA3 fields aren't exactly named the same? Currently we have:

`tls.client.ja3`
`tls.server.ja3s`

Just wondering about the trailing "s"

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+      type: keyword
+      level: extended
+      description: A hash that identifies clients based on how they perform an SSL/TLS handshake.
+
+    - name: client.server_name
+      type: keyword
+      level: extended
+      description: >
+        Also called an SNI, this tells the server which hostname to which the client is attempting to connect.
```
webmat: We should likely call out explicitly that `tls.client.server_name`, when available, should be copied to `destination.domain`.

dcode:
> > Thanks for providing the output from the three different tools.
> 
> Where is this? :-) I'm curious as a 🐱

@webmat I think he's referring to [my gist](https://gist.github.com/dcode/df8670ef49413aa8b5b7908c3491ff27#file-tls_schema-yaml). I can provide some actual output from the three tools. It'll take me a little bit, but it could probably make this process easier to digest.

webmat:
@dcode Ah right. No the gist is fine. I do have Suricata, Zeek and Packetbeat all monitoring my own workstation. If you want to see me spying on myself, I'll happily give you access to the ECS cluster :-)

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
+---
+- name: tls
+  title: TLS
+  group: 2
+  short: Fields describing a TLS connection.
+  description: >
+    Fields related to TLS activity.
```
webmat: I prefer the wording of the "short" description, using the word "connection" rather than "activity", so let's align the main description to use this word as well.

Note that "short" is used in places like [the listing of all field sets](http://ecs_606.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-field-reference.html), so the current curt sentence is perfect there. However "description" is meant to populate [the paragraph(s) used to introduce the field set](http://ecs_606.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-tls.html#ecs-tls) at length.

Feel free to flesh out "description" some more, if needed. If you feel nothing else is needed at this time, then perhaps we can simply align on usage of the word "connection" and call it done. I guess Packetbeat's "main" [description for the field set](https://www.elastic.co/guide/en/beats/packetbeat/current/exported-fields-tls.html) is just as short :-)

schemas/tls.yml
webmat: I like that the example shows the chain as an array, good clarification.

Could you adjust the description so it's mentioned explicitly there as well? Could be as simple as replacing the word "List" with "Array".

schemas/tls.yml
webmat: Another question around consistency: isn't JA3 a hash as well? Shouldn't we add it at `*.hash.ja3`?

On the other hand I do get the sense JA3 is especially powerful/relevant for TLS, and if we want to make it more of a first class field this way, perhaps that's ok as well

schemas/tls.yml
```diff
@@ -0,0 +1,174 @@
...
+      level: extended
+      description: PEM-encoded stand-alone certificate offered by the client.
+
+    - name: client.hash.md5
+      type: keyword
+      level: extended
+      description: >
+        Certificate fingerprint using the MD5 digest of DER-encoded version of certificate offered by the client.
+        For consistency with other hash values, this value should be formatted as an uppercase hash
+        (e.g. `0F76C7F2C55BFD7D8E8B8F4BFBF0C9EC`).
```
webmat: I'd take out the parens and literally make it an example that gets picked up by the ECS scripts:

```YAML
      description: >
        Certificate fingerprint using the MD5 digest of DER-encoded version of certificate offered by the client.
        For consistency with other hash values, this value should be formatted as an uppercase hash.
      example: 0F76C7F2C55BFD7D8E8B8F4BFBF0C9EC
```

This will ensure it shows up as an example in the docs, but also in various other places such as the CSV

schemas/tls.yml
```diff
@@ -0,0 +1,196 @@
...
+  description: >
+    Fields related to TLS activity.
+  type: group
+  fields:
+
+    - name: version.protocol
+      level: extended
+      type: keyword
+      description: Normalized lowercase protocol name parsed from original string.
+      example: "tls` or `ssl"
```
webmat: ECS can only handle single examples in here. Please use either "tls" or "ssl", but not both, it'll render weird :-)

scripts/go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+go 1.13
```
webmat: Is this true of Beats as well? Not sure if this will cause issues

schemas/tls.yml
webmat: My review had been in progress while you moved things around for this part of the file, but this comment is still applicable after your recent push.

We can't nest under `tls.version.*` as this is a hard breaking change vs Packetbeat.

dcode:
I added the raw output for the same session to the gist from zeek, suricata, and packetbeat

schemas/tls.yml
dcode: JA3 is specific to TLS/SSL. There are some similar hash schemes for things like SSH and others. Both fingerprints are ultimately MD5s, but they are hashing a specific thing.

For `ja3`, the following fields are hashed:

```SSLVersion,Cipher,SSLExtension,EllipticCurve,EllipticCurvePointFormat```

For `ja3s`, the following fields are hashed:

```SSLVersion,Cipher,SSLExtension```

File: schemas/tls.yml (hunk 4 above)
Comment (dcode) [reply schemas/tls.yml]: I don't think that really _breaks_ the version string. The only thing that is broken is doing analysis across agent versions. Even then, analysis will generally still work, but be degraded, which seems like a reasonable progression.

Comment (webmat) [reply schemas/tls.yml]: That's also my take. I hope we can take this step forward. But I also feel it's important we get alignment from folks on this :-)

File: schemas/tls.yml (hunk 6 above)
Comment (webmat) [reply schemas/tls.yml]: You know what, I'm wondering if we should really make them mutually exclusive.

If the source only gives access to the cert itself, of course populate that one.

But if the source gives access to the whole chain, I think it's worth populating both. This would ensure that the clean single value `*.certificate` is almost always populated, and therefore consumers of the data can reliably depend on its presence.

Then if the chain is also available, consumers can go dig deeper there as well. The schema can push to fill both. If users of the solution want to trim down index size, they're free to remove the chain, at the cost of losing this information.

File: schemas/tls.yml (hunk 8 above)
Comment (andrewstucki) [reply schemas/tls.yml]: Does it make sense to add a field for SAN specification? As the SNI stuff kind of implies, there are going to be plenty of times where the CN (`server.subject` here I believe) doesn't match a SAN specified in the cert.

File: scripts/go.mod (hunk 12 above)
Comment (dcode) [reply scripts/go.mod]: The `make` process did this, but I can revert that particular file if will cause issues.

File: schemas/tls.yml (hunk 11 above)
Comment (dcode) [reply schemas/tls.yml]: Oh, okay. I checked the rendered asciidoc output and it actually looked good, but I kinda hacked it with the backticks

**Comment by webmat** [comment]
Also don't forget to add a changelog entry to `CHANGELOG.next.md` :-)

**Comment by dcode** [comment]
I'm not sure about the SAN thing. That's digging into the certificate itself. Packetbeat is the only application I know that provides that out of the box as a list of all SANs. I suspect that Packetbeat will continue to ship x509 data in TLS events, but we don't _have_ to define that here. I think a natural progression of this PR is defining a `x509` schema to capture this information for systems the provide file analysis.

Open to discussion though.

**Comment by webmat** [comment]
I'll let other folks chime in on @andrewstucki's point. I'm not familiar enough with that

**Comment by dcode** [comment]
Subject to any additional input or dissenting opinions I'm done hacking on this for now.

**Comment by dcode** [comment]
Thinking through the `SAN` list point from @andrewstucki, I see the analytic need for sure, but I want to minimize data duplication. It might make sense to put the things that packetbeat currently offers in `tls.server_certificate.*` into a structure located at `tls.server.x509.*`. The reason being that x509 can exist in other places. I previously mentioned file analysis (as found in stelka and zeek), but also in things like S/MIME digitally signed email, digital signatures in office/pdf documents, IPSEC authentication, and others.

My goal is merely to think through a uniform way of representing that if we're defining it at all. It goes beyond the scope of this PR, but I think `x509` as a nested field would allow a wildcard field search such as `*.x509.issuer: CN=Let's Encrypt` and you would be able to find all the certificates regardless of use.

Once we get that hashed out, we could do a fun ML job to find anomalous alt names for a given server_name or even destination IP address. 🤷‍♂

File: schemas/tls.yml (hunk 7 above)
Comment (webmat) [reply schemas/tls.yml]: Got it so ja3 and ja3s it is, thanks for the clarification :-)

**Comment by andrewstucki** [comment]
I too like the idea of having a reusable x509 namespace that can be either embedded under TLS or as a standalone top-level thing for the SAN data, reasoning makes sense to me--also, totally cool addressing it in a subsequent PR

File: use-cases/tls.md [hunk 13]
```diff
@@ -11,7 +11,7 @@ You can store TLS-related metadata under `tls.`, when appropriate.
 | [source.ip](../README.md#source.ip)  | IP address of the source.<br/>Can be one or multiple IPv4 or IPv6 addresses. | core | ip | `10.1.1.10` |
 | [destination.ip](../README.md#destination.ip)  | IP address of the destination.<br/>Can be one or multiple IPv4 or IPv6 addresses. | core | ip | `5.5.5.5` |
 | [destination.port](../README.md#destination.port)  | Port of the destination. | core | long | `443` |
-| <a name="tls.version"></a>*tls.version* | *TLS version.* | (use case) | keyword | `TLSv1.2` |
+| [tls.version](../README.md#tls.version)  | TLS version. | extended | keyword | `TLSv1.2` |
```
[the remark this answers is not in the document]
dcode: I don't know what this means, but I'll ignore it as directed ;-)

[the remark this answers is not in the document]
webmat: Yes, nobody knows what the "use cases" are in the repo, anymore.

The folks that find them get confused by them, because we haven't been able to maintain them for a long time.

PR #611, opened 2019-11-15
Add 'file.attributes'

webmat:


webmat:
@rw-access @andrewstucki @paulewing  This field would remove the need for wildcards on file attributes.

webmat:
Currently `file.mode` is a field that only makes sense in posix environments. With a bit more work (not currently reflected in the def), we could map that posix bit field to a predictable set of values for `file.attributes`. E.g. world-read, world-write, world-executable, group-read, etc. 

I think `file.attributes` would be a better way to capture the attributes of a file, independent of platform.

generated/beats/fields.ecs.yml
```diff
@@ -1129,6 +1129,16 @@
       description: 'Last time the file was accessed.
 
         Note that not all filesystems keep track of access time.'
+    - name: attributes
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: 'Array of file attributes.
+
+        Attributes names will vary by platform. Here''s a non-exhaustive list of values
```
rw-access:
```suggestion
        Attributes names will vary by platform. Here's a non-exhaustive list of values
```

webmat:
Ok, I don't want to over-complicate this, so I think I'll merge as is.

Let's leave out the Posix mapping of the octal attributes for now, as this could lead to more questions, such as exact naming, and additional selinux/apparmor attributes. But this is definitely something we should dig into later.

I'll open the follow-up issue and merge this one.

webmat:
Issue opened https://github.com/elastic/ecs/issues/685

PR #612, opened 2019-11-15
Add process.parent by duplicating all fields explicitly

### LINKED ISSUE #597: Consider adding an equivalent field to endgame.parent_process_name in ECS
In the current version of ECS (`1.2` at the time of this writing), the [ECS Process fields](https://www.elastic.co/guide/en/ecs/current/ecs-process.html) have a field to store parent process id:

```
process.ppid
```

, however I'm not seeing a field to store the parent process name, which would be equivalent to:

```
endgame.parent_process_name
```

Please consider adding an equivalent field to `endgame.parent_process_name` in ECS.

_Originally posted by @andrew-goldstein in https://github.com/elastic/ecs/issues/589#issuecomment-543907968_

webmat:
Doing this via normal field reuse is pretty tricky at this time. This is the
first time a field set gets reused under a different name ("process" nested as "parent").

So for now I suggest simply duplicating all fields.

In order to avoid issues with this duplication:

- I call this out and explain it at length in a comment at the top of the `process.yml` file.
- Each duplicate entry is directly below the entry it copies in this file
  (e.g. `process.pid` is directly followed by `process.parent.pid`)
  - Note that the generated docs are sorted by the flattened key name. So all `parent.*`
    fields will actually be grouped in the generated files.

Closes #597.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -15,6 +15,7 @@ Thanks, you're awesome :-) -->
 * Added `process.exit_code`. #600
 
 * Added fields in `tls.*` to support analysis of TLS protocol events. #606
+* Added `process.parent.*`. #612
 
 ### Improvements
 
--- a/code/go/ecs/process.go
+++ b/code/go/ecs/process.go
@@ -31,22 +31,42 @@ type Process struct {
 	// Process id.
 	PID int64 `ecs:"pid"`
 
+	// Process id.
+	ParentPID int64 `ecs:"parent.pid"`
+
 	// Process name.
 	// Sometimes called program name or similar.
 	Name string `ecs:"name"`
 
+	// Process name.
+	// Sometimes called program name or similar.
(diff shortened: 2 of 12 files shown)


webmat:
@ruflin @andrewkroh I'd like to get feedback on the approach I'm taking here. I think it's the pragmatic approach to get this done sooner than later, but it may be controversial.

Also cc @andrew-goldstein, as this solves your issue #597 :-)

ruflin:
I'm ok with getting this in as a hack. We must make sure to get a proper implementation long term as otherwise it is guaranteed that the two get out of sync.

webmat:
Totally agreed. I will have to adjust this PR because we've recently been adding a few other fields to `process.*`. So I'm already feeling the pain ;-)

Work has started on adjusting the scripts [in this branch](https://github.com/webmat/ecs/tree/process-parent), but still very much a WIP.

PR #620, opened 2019-11-20
Added `file.drive_letter`

webmat:


code/go/ecs/file.go
```diff
@@ -34,9 +34,19 @@ type File struct {
...
 
 	// Directory where the file is located.
+	// On platforms with drive letters, the drive letter should not be part of
+	// the directory field. For example, "C:\\Program Files\\Foo" should be
+	// parsed as `file.directory` "Program Files\\Foo" and `drive.letter` "C".
 	Directory string `ecs:"directory"`
 
-	// Full path to the file.
+	// Drive letter where the file is located.
+	Drive string `ecs:"drive"`
```
rw-access: Should this bee `drive_letter`? I think there's a difference, like for . mounted vs unmounted in Windows

code/go/ecs/file.go
```diff
@@ -34,9 +34,19 @@ type File struct {
...
+
+	// Full path to the file, including the file name.
+	// On platforms with drive letters, the drive letter should not be part of
+	// the `file.path` field. For example, "C:\\Program Files\\Foo\\Bar.exe"
```
rw-access: I'm not sure if this is a good idea.
This could get really messy, and the format should match `process.path` and other fields that might be worth joining on.

We could add `relative_path` if we wanted to, but that ends up meaning we index even more data.

code/go/ecs/file.go
webmat: Oh jeez yeah, good point 🤦‍♂ 😂

code/go/ecs/file.go
webmat: I approached it this way in order to address detections that need to be drive-letter independent.

But I think you're right about pivoting between process.executable, and frankly, user expectations.

I'll ponder this a bit (aka let other people chime in), but I'm thinking I'll go with your suggestion here, and leave the drive letter in the path.

schemas/file.yml
```diff
@@ -21,13 +21,33 @@
...
+      the directory field. For example, "C:\\Program Files\\Foo"
+      should be parsed as `file.directory` "Program Files\\Foo" and
+      `drive.letter` "C".
     example: '/home/alice'
 
+  - name: drive
+    level: extended
+    type: keyword
+    description: Drive letter where the file is located.
+    example: C
```
webmat: @rw-access @andrewkroh Any preference on whether this should be `C:` or `C`?

I'm currently proposing stripping off the colon because it doesn't carry any information. But if people are used to seeing the `:` as part of their "drive letter", that's fine too. It's been so long since I used Windows seriously that I have no recollection what the convention is, for that 😬

dcode:
My only hesitation on this is that `drive_letter` is very OS specific. Something like `filesystem_root` might be more agnostic, but you can't tell that directly from a path. From a POSIX `stat` call on a given file, you would know what device that file existed on (and its inode). You can essentially get almost this same information with a Windows `stat` call (see [_stat structure](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2019#remarks)), but by convention this isn't as common, I think.

For a given Linux file, for counter example, it might be nice to know that a binary existed in a filesystem mounted at `/home`, which existed on `/dev/sda5`. You can't tell that from directly looking at the path, but if I'm populating data from an agent on a system, that information would be available via the `mount` table and/or `stat` call.

At the end of the day, we're trying to facilitate analysis of the data though, not to be pedantically correct. So, I think this is a reasonable approach to facilitate analysis, but I wanted to share food for thought to see if others had a better abstraction.

NeilADesai:
There are also times where someone on a Windows host might access the drives without using the drive letter: https://support.microsoft.com/en-us/help/100027/info-direct-drive-access-under-win32. I have seen this used in forensics and by malicious users/software.

webmat:
@NeilADesai Actually we already have `file.device`, whose definition right now is pretty Linux specific. Do you think it would make sense to expand its definition for capturing drives such as `\\.\PhysicalDrive1` in there? Is it the canonical ID of the drive? The drive numbering sounds ephemeral to me (haven't read the whole thing).

webmat:
I'm satisfied with this one. I'm ready to merge.

Just waiting for official reviews :-)

schemas/file.yml
```diff
@@ -33,13 +33,30 @@
...
+      when appropriate.
     example: '/home/alice'
 
+  - name: drive_letter
+    level: extended
+    type: keyword
+    ignore_above: 1
+    short: Drive letter where the file is located.
+    description: >
+      Drive letter where the file is located.
```
andrewkroh: I think the docs should mention this is a Windows concept to alleviate any confusion on the part of implementers working on non-Windows file info.

webmat: @andrewkroh Done.

Do you see any issue with stripping off the `:`?

andrewkroh: No issue. It doesn't provide any additional information.

PR #621, opened 2019-11-21
Flesh out the csv a little

webmat:
In this PR:

- Add the field set column
- Add the short description
- Add whether the field should be indexed
- Reorder the columns a bit, so the stable length values are first

Preview GitHub's rendering of [generated/csv/fields.csv](https://github.com/webmat/ecs/blob/flesh-out-csv/generated/csv/fields.csv) (scroll right for definitions)

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -43,7 +43,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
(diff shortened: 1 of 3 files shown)


webmat:
cc @MikePaquette :-)

MikePaquette:
Looks great @webmat !

Issue #619, opened 2019-11-20
Add Elastic SIEM and Elastic Endpoint Security to list of Products and Solutions


MikePaquette:
Documentation page https://www.elastic.co/guide/en/ecs/current/ecs-products-solutions.html should be updated to list Elastic SIEM App and Elastic Endpoint Security Server.

![image](https://user-images.githubusercontent.com/22619280/69254782-c6602600-0b84-11ea-812b-bb361316939e.png)

webmat:
Makes sense. I'd like these new entries to link to somewhere.

* For SIEM I would link to the SIEM documentation section
* What should the link be for Endpoint?

MikePaquette:
> What should the link be for Endpoint? [This product page](https://www.elastic.co/products/endpoint-security)?

Yes, for now that seems best, thanks.

webmat:
Would be nice to link up the other bullet points as well.

@benskelker would you have some free cycles to tackle this issue?

benskelker:
@MikePaquette @webmat 
yes

webmat:
Hey no arm twisting ;-)

I want to close on #569 after the release next week. Since it's docs, I'll be able to backport to a few past versions (so it's not time sensitive vs the actual release)

ebeahan:
This has been added to https://www.elastic.co/guide/en/ecs/current/ecs-products-solutions.html

PR #622, opened 2019-11-21
Adds links to the SIEM app and Endpoint server

benskelker:
Adds links to SIEM and Endpoint server on the Products and Solutions that Support ECS page, as requested in #619.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -2,7 +2,7 @@
 :doctype: book
 :ecs: ECS
 
-include::{asciidoc-dir}/../../shared/versions/stack/7.x.asciidoc[]
+include::{asciidoc-dir}/../../shared/versions/stack/current.asciidoc[]
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
 [[ecs-reference]]
--- a/docs/products-solutions.asciidoc
+++ b/docs/products-solutions.asciidoc
@@ -5,6 +5,9 @@ The following Elastic products support ECS out of the box, as of version 7.0:
 
 * {beats}
 * APM
+* {siem-guide}[Elastic SIEM app]
+* https://www.elastic.co/products/endpoint-security[Elastic Endpoint Security 
+Server]
 * Infrastructure UI and Logs UI
 * Log formatters that support ECS out of the box for various languages can be found
   https://github.com/elastic/ecs-logging/blob/master/README.md[here].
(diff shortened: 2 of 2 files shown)


benskelker:
Updated index file so links always point to the latest released version

docs/products-solutions.asciidoc
```diff
@@ -3,8 +3,11 @@
 
 The following Elastic products support ECS out of the box, as of version 7.0:
 
-* {beats}
-* APM
+* {beats-ref}/beats-reference.html[{beats}]
+* {apm-get-started-ref}/overview.html[APM]
+* {siem-guide}[Elastic SIEM app]
```
webmat: Can we link to the overview here as well?

docs/products-solutions.asciidoc
```diff
@@ -3,8 +3,11 @@
...
+* https://www.elastic.co/products/endpoint-security[Elastic Endpoint Security 
+Server]
 * Infrastructure UI and Logs UI
```
webmat: Let's take this opportunity to rephrase this bullet point to as well to "Logs monitoring UI", and link it to the docs: http://ecs_622.docs-preview.app.elstc.co/guide/en/logs/guide/current/logs-app-overview.html

docs/index.asciidoc
```diff
@@ -2,7 +2,7 @@
 :doctype: book
 :ecs: ECS
 
-include::{asciidoc-dir}/../../shared/versions/stack/7.x.asciidoc[]
+include::{asciidoc-dir}/../../shared/versions/stack/current.asciidoc[]
```
webmat: Not 100% sure what the effects of this are. Could you expand on this?

webmat: Saw your comment on this, but I'm just wondering if this will mean past versions of ECS will all link to the latest version of the stack

benskelker: If we back port, yes. But does it matter?

docs/products-solutions.asciidoc
benskelker: Updated, but I'll change the the URL when 7.5 is released and becomes the current branch .

webmat: Oh I didn't even realize this was a brand new page :-)

docs/index.asciidoc
webmat: Ok I'll trust you on this. I think it'll be fine.

By the way, looking at the other files in this directory, it looks like the variable `ecs_version` is incorrect. It should be 1.2 in `7.4.asciidoc` and in `7.x.asciidoc`. I guess next week after the ECS release we'll increase 7.x to 1.3.

PR #624, opened 2019-11-21
Set the correct version in the 1.3 branch

webmat:


webmat:
Oops, should have been opened against 1.3 branch

Issue #603, opened 2019-11-05
Usage of generated Go types


hamishforbes:
Are the generated types in this repo actually intended to be usable or are they a WIP / experiment that's not ready yet?

I'm working on a tool to filter and re-structure our logging data at the edge (before shipping from hosts to save network bandwidth and also Logstash resource usage), being able to use these generated types to validate outgoing messages are correctly formed would be very useful.

However the current generated code doesn't seem quite ready. It's got these custom `ecs` tags and dot-format field names which the default Go JSON marshaller doesn't understand.

Is there intended to be a custom marshaller to go along with these?

Related to #343 but the refactor PR mentioned in that issue is already merged without fixing any of these issue that I can see?

mytototo:
I'm also trying to use generated Go types for ECS. I tried to find how it works but didn't find anything relevant. For now I just have a parent struct containing all other struct, but it seems a bit silly and doesn't support JSON.

Where can I find docs or examples about it ?

webmat:
I'm not sure this library is in use nor how well it works. I've been wanting to remove this from the ECS repo for a while.

You're welcome to check out this repo https://github.com/elastic/ecs-logging, which links to concerted efforts to building ECS-compatible logging libraries.

Golang isn't there yet, but it's one of the languages we want to cover eventually as well.

mytototo:
Thank you for the details @webmat.

So in your opinion, what is the best way to respect the ECS in a Go application-- in a very short term ? Do you think replicating the schema as Go types with JSON support is the best way to handle this ?

hamishforbes:
In the end I hacked up a python script (loosely based on some of the scripts in this repo) to generate proper nested and JSON annotated Go types from the yaml schemas in https://github.com/elastic/ecs/tree/master/generated/ecs

webmat:
I think if the goal is solely to emit correct events, you can get away with generating JSON documents that would be ingested properly by Elasticsearch in an ECS index. By this I mean:

* numerics and booleans would be unquoted
* everything else (including IP and date/times) would be quoted
* respecting which fields should be arrays (to be formalized soon via #727 

But I'm a Rubyist, and that's just my preference :-)  

[...]

Issue #343, opened 2019-02-25
Dots in field names for generated Go code


praseodym:
The [ECS guidelines](
https://github.com/elastic/ecs/blob/166358d5610127146537c3108a75be55bd9bfe88/docs/guidelines.asciidoc) mention that “The document structure should be nested JSON objects.”. However, some ECS fields, such as `cloud.instance.id` are [defined as dotted fields under a group](
https://github.com/elastic/ecs/blob/825ee39a3dc6f8b827df534858130d709e33f46d/schemas/cloud.yml#L40).

While this is handled just fine in the [Elasticsearch template](
https://github.com/elastic/ecs/blob/825ee39a3dc6f8b827df534858130d709e33f46d/template.json#L94) using a nested properties object, the [generated Go code](
https://github.com/elastic/ecs/blob/825ee39a3dc6f8b827df534858130d709e33f46d/code/go/ecs/cloud.go#L35) contains the dotted field name just like any regular field name. I don’t think a document using a nested JSON object for such a field will properly deserialize using the generated code (aside from the Go code not having any `json:` keys in the struct field tags at all).

andrewkroh:
There's a [plan](https://github.com/elastic/ecs/pull/336) in the works to refactor all the generators used in this repo. It should be possible to modify the generator to produce code that is compatible out-of-the-box with the encoding/json package.

I added the existing code because I found myself implementing these types in multiple places and keeping them in sync with the spec was hard. I figured I'd come back to add a means of marshaling to JSON, but haven't found time. For my use case I have custom marshaler that reads the `ecs` tags.

sporkmonger:
Also none of the generated Go code uses `omitempty` in tags, which I suspect will cause issues for most real-world marshaling usage.

willie:
I came here to basically say the same thing @sporkmonger. Not having `json` or `omitempty` (for non required fields) really makes the generated Go code difficult to use in real world code where you want ecs compatible JSON. I tried to add `omitempty` in `gocodegen.go`, but the loops and tricks to find that field after the YAML is marshaled is extraordinarily obscure (`libbeat` is weird place, once you manage to find the right version). I don't mean to come off angry, I just really wanted to fix it and submit a PR for it, but the what is using what to determine the fields stopped me cold. I see the plan for the new generators and that's interesting. I subscribe to that PR. In the meantime, if you have suggestions for how to find the status of `required` in the YAML so I can alter the generator, I'd be happy to fix it.

PR #628, opened 2019-11-22
Remove the Go library and the code generating it.

webmat:
Elastic is in the process of releasing ECS logging formatters and libraries for
multiple languages. Each of them will consume ECS artifacts as needed, to generate
the code & so on. None of them will be hosted directly in the ECS repo itself,
however.

This PR removes the Golang library, as recently discussed, @andrewkroh. Happy
to hold off a bit, until a new repo for it has been created.

Issues to carry over to that new repo: https://github.com/elastic/ecs/issues/603, https://github.com/elastic/ecs/issues/343


ruflin:
@webmat I think the logging library and this go code are two different things. It might be used for a go library but can be used for other things too. Be aware, this is a breaking change in case someone depends on it.

Could you share some background on why these don't belong into the ECS repo and where they should land in the future? I would prefer to keep ECS "things" together in one repo and only scale out when we really have to.

webmat:
Yes, it's true that this isn't one of the upcoming logging formatters per se. Part of the logging formatter initiative is also focusing on creating actual libraries like this one. Case in point: the new [ecs-dotnet](https://github.com/elastic/ecs-dotnet).

Having ECS libraries specific to the languages people use is a thing that makes sense for generating and consuming ECS events. However for obvious reasons, these language libraries can't all be built in this repo. I can't deal with all the languages at the same time here, especially if I don't maintain the libraries themselves.

Since none of these libraries have anything to do with publishing the schema per se, I think none of them actually belong here. It was fine to start with the Golang one in this repo, because it was the simplest way to accomplish that at the time, and we had other Golang things in here. But now it's time to make the separation IMO.

I'm happy to wait a bit on this one if there's a need, but I don't think moving this library is going to be a big deal. The two issues linked in the body seem to indicate that this library isn't actually ready for prime time: it looks like it's still at the experimental stage, and issues aren't being addressed.

Note that for CI, I've started talking to the infra team about setting up multiple related projects, so one build here could trigger builds of other dependent projects.

webmat:
- [ ] Add a changelog entry after https://github.com/elastic/ecs/pull/633

ruflin:
I think it wouldn't be a technical problem to have all the libs also in the spec folder. For me the strongest argument to have it separate is the release cycles. As far as I understand, the logging library have their own release versions? 

For the Golang files they are tied to the schema as far as I can tell. Updating it in a single PR instead of having to go to a second repo sounds beneficial to me.

webmat:
But if we push that to the extreme, are you saying the .Net library should move here as well? If not, why would we have the Golang library here and not the .Net one?

There's no material difference between the two, IMO.

webmat:
To be clear, I'm conscious of the fact that this will require a few more manual steps to build the code libraries by being outside this repo. But we can automate this as needed.

ruflin:
I think so, yes. It would for example solve this "version" directory: https://github.com/elastic/ecs-dotnet/tree/master/src/Specification

webmat:
And what about when there's a Java library? Should this repo require Python, Golang, .Net and Java just to run make? 😬

I would much rather have this repo's CI reuse a Docker artifact from these other repositories, to ensure things still work on every PR. But there's no need to have up to date artifacts for every commit to master for all of these libraries. These can be triggered once, when an ECS release is out.

webmat:
On the "version" directory in ecs-dotnet: actually having all versions at the same time in this repo is a benefit that's enabled by this separation. I don't see it as a problem.

ruflin:
We consolidated repos in the past, for example getting all Beats into one repo to move faster which I think was very beneficial. But one or many repos, both is going to work. Not going to further push back on this.

I don't think this PR should be merged before the new code exists somewhere in a go repo.

gen0cide:
Not asking for an endorsement, but just so you can have peace of mind, I published a Go code generator for ECS:

https://github.com/gen0cide/ecsgen

webmat:
@gen0cide Does it use the Go code from the ECS repo?

gen0cide:
Nope! Its goal is to follow convention and generate code based on the generated ECS YML. (In my case, Flat YML). 

It parses the YML into an AST, then generators walk/enumerate the AST to generate code. Right now it supports Go, but working on an Avro schema generator as well.

ebeahan:
Separating this code from elastic/ecs is still something in the works (#630), but I'm closing this PR as it's been inactive for some time.

PR #631, opened 2019-11-22
Run misspell on the schema definitions too

webmat:
This used to be covered by running spell-checker on the readme. But the schema's no longer there.


webmat:
I'd like to run this on all of the asciidoc files as well, but misspell seems to completely ignore them.

webmat:
@ruflin Thanks for your reviews on all of the controversial PRs.

This one shouldn't be controversial, though. WDYT? ;-)

Issue #601, opened 2019-10-30
Shared knowledge base on mapping log events to Elastic Common Schema


Duchadian:
Issue / roadmap suggestion posted here at the suggestion of some of the devs.

Elastic Common Schema has provided a unified format for logs ingested using Beats in those cases where a module for parsing and mapping the events in the log is available. However, it can be difficult to parse and map events to ECS for the log sources for which a module is not available. 

It would be nice to have a shared knowledge base for previous log sources that have been mapped to ECS. This would make converting new log sources to ECS significantly easier, as there would be a central location with examples of previous log sources that have been converted to ECS. Right now, that is scattered across various repos where it is relatively difficult to find.  

The knowledge base could take a similar form to the [grok-patterns repo](https://github.com/logstash-plugins/logstash-patterns-core/tree/master/patterns).

webmat:
Hi folks, the closest thing we have to this right now is opening an issue on the ECS repo, with your mapping, and tagging it with [mapping](https://github.com/elastic/ecs/issues?q=is%3Aissue+is%3Aopen+label%3Amapping). Ideally the mapping can be shared via a public web spreadsheet, but it can be shared another way too.

Opening such issues will allow us to provide feedback and thoughts on what needs adjustments, explain edge cases that apply to your situation & so on. These issues and discussions are then available for the community to look at, see the discussion, the thought process & so on.

Does that help for what you had in mind?

Duchadian:
Hi @webmat, that's sort of what I had in mind. 

From my perspective, what ECS lacks is a centralised repo / website / something where the community (in collaboration with Elastic) can maintain parsers and mappings to ECS for all kinds of log sources across various versions of the log source. This would ensure that ECS can be more easily adopted for all kinds of log sources by users, vendors, and Elastic itself. 

What most of the issues with the mapping label are missing is the _original_ event. That way, it is difficult to see which part of the original log maps to which field in ECS. A side-by-side of the original event, the parsed event, and the event mapped to ECS would be a way to make this clearer in my opinion. Otherwise, it is more difficult to get the semantics of the message right in the mapping from any new log source to ECS. 

I hope it is at least sort of clear what I mean.

Randy-312:
I second this ask!  
It would also be good to highlight where each mapping is in the cycle, and IF a module is on the way for it to be used as a plugin.

Wikipedia could handle it, but I'd rather have it on GitHub where we can find it easily, and contribute to it.

webmat:
[...]

PR #632, opened 2019-11-22
Remove use cases directory and code

webmat:
They have not been maintained for a long time. With the recent addition
of the TLS field set (#606), the information in there was outright incorrect.


ruflin:
I agree it needs cleanup but not sure removal is the right step here as it was a place to "share" mappings which go beyond ECS. For example the kubernetes ones. Do you have a replacement in mind?

webmat:
Yes, there is something superior coming up soon. A description is available in the first part of this comment https://github.com/elastic/ecs/issues/601#issuecomment-553711655.

This will allow us and the community to share ECS mappings in a way that can also be leveraged programmatically.

webmat:
Happy to wait until after the above project is available, if you think the use cases are still this useful.

If you think we should wait, please let me know. I'll have to fix the TLS stuff in the use cases, as it's now proposing something that's wrong.

webmat:
- [ ] Add a changelog entry after https://github.com/elastic/ecs/pull/633

ruflin:
@webmat I'm ok moving forward with removing it first before we have a replacement. Sometime I pointed users to the use cases for inspiration, but having them outdated is also not good. For the replacement: Where you expect the "example csv" to end up? In this repo?

webmat:
I'm not 100% what the end result will look like. So during the experimentation phase, I will be using a separate repo. When the POC is ready, we can look at it and decide if we should move it here or not.

ebeahan:
We'll revisit this change when ready, but since this PR is quite out-of-date let's close.

New issue for tracking: https://github.com/elastic/ecs/issues/1391

PR #636, opened 2019-11-25
Remove TLS use case.

webmat:
With TLS in ECS (starting with 1.3), this use case is obsolete and misleading.


webmat:
@ruflin I opened this PR in case you think we should wait on #632.

#632 can wait, but the TLS use case needs to be removed, with the advent of TLS in ECS.

PR #640, opened 2019-11-26
Backport #621 to 1.3: Flesh out the csv a little

webmat:
Backport of PR #621 and #638 to 1.3 branch. Original message:

In this PR:

* Add the field set column
* Add the short description

Preview GitHub's rendering of [generated/csv/fields.csv](https://github.com/webmat/ecs/blob/backport_622_1.3/generated/csv/fields.csv) (scroll right for definitions)

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -37,9 +37,13 @@ Thanks, you're awesome :-) -->
 
 #### Breaking changes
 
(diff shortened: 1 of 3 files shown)


webmat:
Yeah I agree. In this case though I want to backport this new csv format all the way back, though (after the release). This single PR will make further backports a little easier.

PR #650, opened 2019-11-26
Backport #640 to 1.2: Flesh out the csv a little

webmat:
Backport of PR #640 to 1.2 branch. Original message:

In this PR:

* Add the field set column
* Add the short description

Preview GitHub's rendering of [generated/csv/fields.csv](https://github.com/webmat/ecs/blob/backport_640_1.2/generated/csv/fields.csv) (scroll right for definitions)

### DIFF
--- a/generated/csv/fields.csv
+++ b/generated/csv/fields.csv
@@ -1,397 +1,397 @@
-Field,Type,Level,Example,ECS version
-@timestamp,date,core,2016-05-23T08:05:34.853Z,1.2.0
-labels,object,core,"{'application': 'foo-bar', 'env': 'production'}",1.2.0
-message,text,core,Hello World,1.2.0
-tags,keyword,core,"[""production"", ""env2""]",1.2.0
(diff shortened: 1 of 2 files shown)


webmat:
@ruflin Unsure what to do about messaging this backport. I wouldn't rev the ECS version, as the schema hasn't changed, only the csv file has.

Also, in 1.3 we're introducing the separation in the changelog for schema changes vs tooling/artifact changes. For 1.2.0 and before, we rarely messaged anything about tooling/artifact changes, so the changelog entries no longer have a home in older versions. I see two paths forward:

1. Just backport without any changelog entry, consistent with earlier versions' use of the changelog.
2. Retroactively introduce the changelog separation for all past releases, and reintroduce the changelog entries there.

I would tend to favor option 1, for simplicity's sake. WDYT?

ruflin:
Why would you backport this in the first place?

webmat:
To have the same format available for all versions.

Do you see a problem with this?

PR #651, opened 2019-11-26
Backport #650 to 1.1: Flesh out the csv a little

webmat:
Backport of PR #650 to 1.1 branch. Original message:

In this PR:

* Add the field set column
* Add the short description

Preview GitHub's rendering of [generated/csv/fields.csv](https://github.com/webmat/ecs/blob/backport_650_1.1/generated/csv/fields.csv) (scroll right for definitions)


ruflin:
Why is this backported?

PR #655, opened 2019-11-28
Fix typo in DNS fields documentation

marcocova:


elasticcla:
Hi @marcocova, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecs/pull/655.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

webmat:

PR #661, opened 2019-12-02
Marks items as arrays if required.

codebrain:
This is to aid code generation for languages that support array constructs.

Note: This field is optional, the default value is false. Lack of field presence implies a value of false.

### DIFF
--- a/schemas/base.yml
+++ b/schemas/base.yml
@@ -37,6 +37,7 @@
       level: core
       type: object
       object_type: keyword
+      is_array: true
       example: {env: production, application: foo-bar}
       short: Custom key/value pairs.
       description: >
--- a/schemas/dns.yml
+++ b/schemas/dns.yml
@@ -48,6 +48,7 @@
     - name: header_flags
       level: extended
       type: keyword
+      is_array: true
       short: Array of DNS header flags.
       description: >
         Array of 2 letter DNS header flags.
@@ -129,6 +130,7 @@
     - name: answers
       level: extended
       type: object
+      is_array: true
       short: Array of DNS answers.
       description: >
         An array containing an object for each answer section returned by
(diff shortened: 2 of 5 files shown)


webmat:
Thanks for getting this started, @codebrain. This will be useful.  :-)

webmat:
@codebrain If you don't mind I'll take over the PR from here. That was the plan, right? 🙂 

I'll change the way we're specifying what's an array to something like:

```YAML
format:
- array
```

For now I will not add more under `format`, but eventually I'd like to formalize other kinds of normalization (e.g. lower/upper casing, perhaps date formats) with this...

The purpose of this PR is still 100% focused on clarifying what's an array, however. So:

* marking all the fields that should be arrays
* tweaking the code to display the info in the docs
* having the information show up in the generated/ecs yaml files

webmat:
Not quite ready to update this PR yet, but I've started work on adapting it here https://github.com/webmat/ecs/commits/is-array-markers, if you want to check it out, @codebrain

codebrain:
https://github.com/webmat/ecs/commit/843e521873c1692bfb53a0cdcc4c8605f68b6027 - looks good.

I didn't realise, but it looks as if there are more arrays. I will wait for this to be merged in and then tweak our generator for 1.3.

Thanks Mat, yes, more than happy for you to pick this up from here on in :)

jeskew:
@webmat do you know if there's a target release for https://github.com/webmat/ecs/commit/843e521873c1692bfb53a0cdcc4c8605f68b6027 to be incorporated into master? I use ECS for a system that tees logs into Elasticsearch and parquet files and have been maintaining a separate manifest of which fields should be treated as arrays in the parquet schema.

webmat:
@jeskew We can't make any promises on the release of this branch. But it's towards the top of the list of priorities, as many folks need this, like you do. I'm hoping it will make it into the next ECS release.

Note also that merging something to master doesn't make it official. The master branch is a development branch. Only the [releases](https://github.com/elastic/ecs/releases) make new developments official 😉

PR #662, opened 2019-12-02
Remove tls.server.supported_ciphers field

adriansr:
Unlike the client, the server doesn't provide a list of supported ciphers in its hello message.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -14,7 +14,7 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
-* Removed unnecessary field `tls.server.supported_ciphers`.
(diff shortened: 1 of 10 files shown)


ruflin:
@webmat Was this already released?

webmat:
@ruflin Yes it was.

Since it's a field that doesn't make sense, I propose we release ECS 1.3.1 with only this removal in it.

@dcode or @dainperkins you guys agree this field doesn't make sense, right?

webmat:
@ruflin To clarify.

During the exchange, the client says which ciphers it supports -- captured at `tls.client.supported_ciphers`.

The server picks which one to chose, and the selected cipher should be captured in `tls.cipher`.

At no point are there server side "supported ciphers" we can capture. I think this was a copy/pasta mistake ;-)

PR #665, opened 2019-12-02
added rule fields

### LINKED ISSUE #598: Add Rule Fields for Observers and Agents
ECS needs a place to capture information about rules used to create detections, for example, web proxy rules, firewall rules, IDS/IPS signatures, EDR rules, WAF rules, SIEM rules, etc.  

**Proposed Fields** 

- [ ] `rule.id` - A rule ID that is unique within the scope of the entity using the rule for detection of this event
- [ ] `rule.type` - a categorization value keyword used by the entity using the rule for detection of this event
- [ ] `rule.name` - A keyword name for the rule used to create this event
- [ ] `rule.description` - a text description of the rule used to create this event
- [ ] `rule.reference`- Text URL pointing to additional information about the rule used to create this event
- [ ] `rule.version` -  Rule version of the rule used to created this event. Used by systems that allow multiple rule versions to exist.
- [ ] `rule.ruleset` - name of the ruleset, policy, or group in which the rule used to create this event is a member.

**Usage Example** 

- `observer.type:"proxy"`
- `observer.name:"1_proxysg"`
- `observer.vendor:"symantec"`
- `observer.product:"s200"`
- `observer.os.name:"sgos"`
- `observer.os.version:"7.1"`
- `observer.geo.name:"3rd_floor_mdf"`
- `observer.rule.id:"00012345"` 
- `observer.rule.type:"url"`
- `observer.rule.name: "no gambling at work"` 
- `observer.rule.description:"blocks access to known gambling web sites` 
- `observer.rule.reference:"http://sitereview.bluecoat.com/#/lookup-result/betthefarm.xyz"`
- `observer.rule.version:"1"` 
- `observer.rule.ruleset:"Standard Employee Browsing Policy"` 

**Related Fields** 
In security use cases, there may be other identifiers used for classifying  rules, such as the MITRE ATT&CK framework.  If this is the case, such information should be populated in the ECS `threat.*` fields.

**Field Reuse** 
Other field sets must not be nested under the `rule.*` field set
The `rule.*` fields must not be used at the top level.

_Nested usage_ | Description
-- | --
observer.rule.* | Fields describing rules running on an observer used to detect this event
agent.rule.* | Fields describing rules running on an agent used to detect this event

dainperkins:
Added rule fields for agent / observer rule based events/actions

Closes #598


schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used by observers and agents to trigger events and/or actions.
```
MikePaquette: can we change "trigger events and/or actions." to "generate alerts or other events."

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that are associated with a given action or event.
```
MikePaquette: can we change "are associated with a given action or event." to "generate alerts or other events."

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+   Typical sources of rule data might include any of the following: acl enabled network access equipment, network admission control
```
MikePaquette: request change "Typical sources of rule data might include any of the following:" to "Examples of data sources that would populate the rule fields include:"

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used by observers and agents to trigger events and/or actions.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that are associated with a given action or event.
+   Typical sources of rule data might include any of the following: acl enabled network access equipment, network admission control 
+   platforms, network or host IDS/IPS, network firewalls, web access firewalls, url filters, etc.
```
MikePaquette: consider changing: "platforms, network or host IDS/IPS, network firewalls, web access firewalls, url filters, etc." to "platforms, network or host IDS/IPS, network firewalls, web access firewalls, url filters, endpoint detection and response (EDR) systems, etc."

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+         A categorization value keyword used by the entity using the rule for detection of this event
+
+      example: DNS
+      
+    - name: ruleset
+      level: extended
+      type: keyword
+      short: Rule Ruleset
+      description: >
+         The parent category of the rule used to create this event.
```
MikePaquette: can we change "The parent category of the rule used to create this event." to "name of the ruleset, policy, group, or parent category in which the rule used to generate this event is a member."

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      example: Standard_Protocol_Filters
+
+    - name: reference
+      level: extended
+      type: keyword
+      short: Rule Reference
+      description: >
+        Text URL pointing to additional information about the rule used to create this event.
```
MikePaquette: change "create" to "generate"

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: done

dainperkins:
also added "a link to the vendors documentation regarding the rule" to the reference field (e.g. https://www.snort.org/rule_docs/1-50032)

Should we have multiple examples?

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: done

dainperkins:
> Where do we add text about field re-use?

unsure @webmat?

Also @MikePaquette & @webmat 
  - should we provide some rule.type examples?
      * acl (base ip / port decision)
      * url  (e.g. blocked url)
      * protocol (e.g. malformed or tunneled protocol)
      * intrusion (detection/prevention)

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+        A rule ID that is unique within the scope of an agent, observer, or other entity using the rule for detection of this event
+
+      example: 101 
+
+    - name: uuid
+      level: extended
+      type: keyword
+      short: Rule UUID
+      description: >
+        A rule ID that is unique within the scope of a set or group of agents, observers, or other entities using the rule for detection of this event.
```
webmat: The distinction is not clear enough vs `rule.id` IMO. @enotspe or @MikePaquette any ideas here?

Are there actually situations where one would populate both `rule.id` and `rule.uuid`? If you have both, is `rule.id` less relevant? I'm asking because perhaps we can simplify this and specifically say `rule.id` may be used to capture the rule UUID when available, and otherwise capture whatever rule ID makes sense from the device.

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      example: 1100110011
+
+    - name: version
+      level: extended
+      type: keyword
+      short: Rule Version
```
webmat: Please use sentence capitalization, not title cap. 
```suggestion
      short: Rule version
```

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      description: The version / revision of the rule being used for analysis.
+      example: 1.1
+
+    - name: name
+      level: extended
+      type: keyword
+      short: Rule Name
```
webmat:
```suggestion
      short: Rule name
```

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      description: The name of the rule generating the event.
+      example: Filter_DOH
+
+
+    - name: description
+      level: extended
+      type: keyword
+      short: Rule Description
```
webmat:
```suggestion
      short: Rule description
```

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      example: Block requests to public DNS over HTTPS / TLS protocols
+
+    - name: type
+      level: extended
+      type: keyword
+      short: Rule Type
+      description: >
+         A categorization value keyword used by the entity using the rule for detection of this event
+
+      example: DNS
```
webmat: If it's a category, should we call this field `rule.category`?

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      type: keyword
+      short: Rule Version
+      description: The version / revision of the rule being used for analysis.
+      example: 1.1
+
+    - name: name
+      level: extended
+      type: keyword
+      short: Rule Name
+      description: The name of the rule generating the event.
```
webmat: I would add wording here about "name or signature". For example Suricata uses "signature" for the rule's name and "category" for categorization: https://github.com/elastic/beats/blob/master/x-pack/filebeat/module/suricata/eve/test/eve-small.log-expected.json#L62-L67

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      short: Rule Ruleset
+      description: >
+         Name of the ruleset, policy, group, or parent category in which the rule used to generate this event is a member.
+
+      example: Standard_Protocol_Filters
+
+    - name: reference
+      level: extended
+      type: keyword
+      short: Rule Reference
```
webmat:
```suggestion
      short: Rule's reference URL.
```

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: type
+      level: extended
+      type: keyword
+      short: Rule Type
+      description: >
+         A categorization value keyword used by the entity using the rule for detection of this event
+
+      example: DNS
+      
+    - name: ruleset
```
webmat: What's the relationship between ruleset and type? Is there a hierarchy there?

schemas/vlan.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: vlan
```
webmat: Please remove `schemas/vlan.yml` from this PR and open a different PR for it ;-)

dainperkins: This is NOT supposed to be here... github insanity

webmat:
> Where do we add text about field re-use?

@MikePaquette What do you mean by that? We agreed that rule would be at the top level, right? Nesting it under agent + observer would mean we can't analyze both sets of rules together.

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used by observers and agents to generate alerts or other events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts or other events..
+   Examples of data sources that would populate the rule fields include: platforms, network or host IDS/IPS, network 
+   firewalls, web access firewalls, url filters, endpoint detection and response (EDR) systems, etc.
```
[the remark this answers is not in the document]
dainperkins: network admission control...copy/paste debaucle

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used by observers and agents to generate alerts or other events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts or other events..
```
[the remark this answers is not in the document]
dainperkins: since you asked so nicely... done

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+  
+  type: group
+  fields:
+
+    - name: id
+      level: extended
+      type: keyword
+      short: Rule ID
+      description: >
+        A rule ID that is unique within the scope of an agent, observer, or other entity using the rule for detection of this event
```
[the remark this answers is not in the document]
dainperkins: done

schemas/rule.yml
dainperkins: I'm wondering if an observer event (e.g. a single PA firewall log) will have UUID for e.g. a Panorama rule?

schemas/rule.yml
dainperkins: changed to "rule or signature"

schemas/rule.yml
dainperkins: how about "Rule Reference URL"

schemas/rule.yml
dainperkins: Ruleset is a logical grouping (e.g. web filter rule group) type is what kind of action does the rule concern itself with, I'm not sure type is necessary across the board.

schemas/vlan.yml
dainperkins: done

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used by observers and agents to generate alerts or other events.
```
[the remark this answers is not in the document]
dainperkins: ok by me

schemas/rule.yml
dainperkins: Was thinking of this as more a categorization of what type of action/information is being analyzed by the alert - e.g. url for url filtering, dns for nxdomain results, acl for basic access control, etc.

schemas/rule.yml
webmat: Yes, perfect

schemas/rule.yml
dainperkins: per ecs meeting - I'm thinking this can probably be dropped for now....

webmat: Ok, I'm good with dropping `.type` for now.

Just to be sure we're on the same page, `.ruleset` is meant to capture a value according to the source of these events?

For example, using this to map Suricata alerts [like this one](https://github.com/elastic/beats/blob/6f14a606f1518a96c4378d3a16a8fd06266a20e7/x-pack/filebeat/module/suricata/eve/test/eve-alerts.log-expected.json#L45-L49), we would have this?

`rule.category="Attempted Information Leak"` (from `suricata.eve.alert.category`)

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+      example: Standard_Protocol_Filters
+
+    - name: reference
+      level: extended
+      type: keyword
+      short: Rule Reference URL
+      description: >
+        Text URL pointing to additional information about the rule used to generate this event.  
+        For example a link to the platform used to manage the observers, a link to the vendors documentation regarding the rule, 
+        or a link to more information on the specific action that triggered the rule.
```
webmat: I'd like to rephrase this section a bit, and split in two paragraphs. Here's my take.

```
      description: >
        Reference URL to additional information about the rule used to generate this event.

        The URL can point to the vendor's documentation about the rule.
        If that's not available, it can also be a link to a more general page describing this type of alert.
```

You'll note that I'm dropping the part about linking directly to another platform. If I understand correctly, this was meant to link directly in a system, to the page about this specific instance the alert.

While this is a very relevant piece of information, I'd like to capture such specific links in a different field. I'm trying to establish a pattern where the `.reference` URL fields are meant to be documentation-style references about a thing.

I'm still looking for a good name for the field meant to capture deep links to another system. Suggestions welcome :-)

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: done

schemas/rule.yml
dainperkins: Done

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used to generate alerts or other events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts or other events.
+   Examples of data sources that would populate the rule fields include: network admission control platforms, network or
```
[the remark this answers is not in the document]
dainperkins: done

schemas/rule.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used to generate alerts or other events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts or other events.
+   Examples of data sources that would populate the rule fields include: network admission control platforms, network or 
+   host IDS/IPS, network firewalls, web access firewalls, url filters, endpoint detection and response (EDR) systems, etc.
```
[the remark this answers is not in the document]
dainperkins: done

schemas/rule.yml
dainperkins: sounds good to me

schemas/rule.yml
dainperkins: Dropped, tho now I am reading the above and thinking maybe we need discrete rulesets (groups of rules/signatures that may apply to e.g. specific networks - e.g. GUEST NET IPS SIGNATURE <rule group> AND rule.type or rule.category - e.g. "Attempted Information Leak" as they are definitely different things logistically

I think category is more descriptive so I will add that in.

schemas/rule.yml
dainperkins: brought this back re: ruleset discussion below, as rule.category

schemas/rule.yml
webmat: Yeah I'd like the two descriptions to be a little more differentiated. But at least it's clear that uuid is needed. (thanks for the links, @enotspe 🙂)

Let's leave it as is, and improve as needed later.

schemas/rule.yml
```diff
@@ -0,0 +1,84 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used to generate alerts or other events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts or other events.
```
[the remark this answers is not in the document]
dainperkins: I'm good with that, will update today

on the other question - rule.category is, typically, the category of action the rule is alerting on (e.g. "Web Attacks" rulesets would include xss, directory traversal, etc) and you could then apply the "WEB ATTACKS" rule wherever it needed to be applied, instead of enabling each individual rule.  

ruleset is a logical configuration which (often times across multiple managed implementations of a solution - e.g. Panorama managing multiple Palo Altos) combines rules into a single logical set (e.g. on the GUEST NETWORK apply the GUEST rule.ruleset which includes [blocks DNS-over_TLS, blocks nsfw/guns/and competitors web sites,  limits bandwidth, etc.)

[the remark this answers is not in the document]
webmat: So ruleset is a wider grouping than category? Or are these orthogonal?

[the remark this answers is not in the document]
dainperkins: orthogonal...

Ruleset is localized configuration/implementation specific (as in apply ruleset MY_RULES to this subnet / vlan / group) where MY_RULES have been defined to include e.g. 10 specific locally relevant individual rules.   

Category is a grouping of rules by type (think Mitre technique or vulnerable system type)

schemas/rule.yml
```diff
@@ -0,0 +1,84 @@
+---
+- name: rule
+  title: Rule
+  group: 2
+  short: Fields to capture details about rules used to generate alerts or other notable events.
+  description: >
+   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts oor other notable events.
```
webmat:
```suggestion
   Rule fields are used to capture the specifics of any observer or agent rules that generate alerts ooor other notable events.
```

What about this? ;-)

webmat: Kidding, I prefer "or" :-)

dainperkins: DOH! fixed

schemas/rule.yml
```diff
@@ -0,0 +1,84 @@
...
+      short: Rule name
+      description: The name of the rule or signature generating the event.
+      example: BLOCK_DNS_over_TLS
+
+    - name: description
+      level: extended
+      type: keyword
+      short: Rule description
+      description: The description of the rule generating the event.
+      example: Block requests to public DNS over HTTPS / TLS protocols
```
andrewstucki: I'm assuming the plan was to do a subsequent PR with the corresponding #680 changes that likely ought to be added?

webmat: Yeah the default text indexer would be useful here, but we should do another round later.

Issue #630, opened 2019-11-22
Remove Golang dependencies


webmat:
I'd like to get rid of the dependencies around the Golang ecosystem in this repo.

For this to happen, the following dependencies have to be removed or replaced:

- [x] https://github.com/elastic/ecs/pull/629 Remove legacy template
- [x] https://github.com/elastic/ecs/pull/1567 Remove Go library, it should move to its own repo
- [ ] https://github.com/elastic/ecs/pull/1585 misspell (either use compiled `misspell` or use another tool)
- [ ] https://github.com/elastic/ecs/pull/1585 license check

Reasons why I'd like to remove dependencies on Go:

- Simplify setup for all contributors
- Make it possible for Windows users to run the ECS tooling
- Reduce distractions from supporting 2 languages & related dependencies

webmat:
@andrewkroh Does "elastic/go-licenser" support looking at more than Go files?

webmat:
cc @codebrain thinking about you on this one ;-)

ruflin:
@webmat Do you plan to make all python based?

Can you elaborate how "Make it possible for Windows users to run the ECS tooling" is related to Golang?

webmat:
Stuart was unable to run the tooling on Windows, and I think it had to do with Golang. Since I don't know the Go ecosystem and tools very much, I couldn't help him.

I'm pretty sure it was related to Go. Perhaps I'm mis-remembering.

webmat:
And yes, the goal is to have only Python code in this repo. For simplicity's sake.

I don't mind pulling in compiled binaries like misspell another way, but I want to cut out the dependency on Golang.

ruflin:
I would assume the Windows issue has more to do with bash scripts than Golang? +1 on removing dependencies but also be aware that it will probably need rewriting of some tooling. And Beats is moving in the opposite direction, getting rid of python and move to Golang. One reason is the cross platform support ;-)

ebeahan:
Packetbeat is the main consumer of this generated library, and the Beat's module dependencies [lists](https://github.com/elastic/beats/blob/23e4403ae093fcc8f7905345cad2c7ad256976d8/go.mod#L62) an old version of ECS: `v1.8.0`.

@andrewstucki @andrewkroh Do you see an eventual need to support a more recent ECS version in this library?

[...]

PR #674, opened 2019-12-04
Convert to Python 3

rw-access:
Getting ahead of python2's EOL. There were a few minor changes, and I believe that whitespace changes in `json.dump`, and [dictionary default ordering](https://portingguide.readthedocs.io/en/latest/dicts.html#changed-key-order) may have changed some of the autogenerated code.

Unit tests are still passing, although there's one deprecation warning that's easy to address
```consolemake test
build/ve/bin/python -m unittest discover --start-directory scripts/tests
..
ecs/scripts/tests/test_ecs_helpers.py:72: DeprecationWarning: Please use assertEqual instead.
  'base': {'group': 1, 'name': 'base'}
........................
----------------------------------------------------------------------
```

If you [disable whitespace changes](https://github.com/elastic/ecs/pull/674/files?utf8=%E2%9C%93&diff=unified&w=1), it'll be easier to view the diff.


generated/elasticsearch/6/template.json
```diff
@@ -1,926 +1,912 @@
...
               }
             }
           }
-        }, 
+        },
         "dns": {
           "properties": {
             "answers": {
               "properties": {
-                "class": {
```
rw-access: Although, I'm not sure why this disappeared

scripts/generators/ecs_helpers.py
```diff
@@ -6,11 +6,11 @@
 # Dictionary helpers
 
 
-def dict_copy_keys_ordered(dict, copied_keys):
+def dict_copy_keys_ordered(dct, copied_keys):
```
rw-access: I renamed this to avoid variable shadowing, since `dict` already means something.
It wasn't strictly necessary, but it makes the code easier to read, and makes PyCharm more pleased with me.

scripts/generators/es_template.py
```diff
@@ -20,29 +20,32 @@ def generate(ecs_flat, ecs_version):
...
 
 
-def dict_add_nested(dict, nestings, value):
+def dict_add_nested(dct, nestings, value):
     current_nesting = nestings[0]
     rest_nestings = nestings[1:]
     if len(rest_nestings) > 0:
-        if current_nesting not in dict:
-            dict[current_nesting] = {'properties': {}}
-        elif 'type' in dict[current_nesting] and 'object' == dict[current_nesting]['type']:
```
rw-access: We got lucky before, because with Python 2, the [ordering was deterministic](https://portingguide.readthedocs.io/en/latest/dicts.html#changed-key-order)

rw-access: The previous version was passing unit tests but generated code that was inconsistent, because `"type": "object"` was missing. I can make it so this is always present, but I don't exactly know the intendend behavior.
https://github.com/elastic/ecs/blob/94e7376ed1e618bec7128e4b14f579a5c0111343/generated/elasticsearch/7/template.json#L419-L496

Replacing these lines
https://github.com/elastic/ecs/blob/94e7376ed1e618bec7128e4b14f579a5c0111343/scripts/generators/es_template.py#L27-L30
with this should do what I _think_ is our intended behavior, but it's very clear to me what that is.
```python
        dct[current_nesting].setdefault("properties", {})
```

webmat:
Huge thanks for this! I'll review shortly :-)

I'm specifically using [OrderedDict](https://github.com/elastic/ecs/blob/48f0cb887f70944e656800804ca8c5865b9e4f87/scripts/generators/ecs_helpers.py#L3) in a few strategic places. But the default dict becoming ordered does show up in a few more places, like the `labels` example :-)

rud:
Tiny suggestion: update the python version in `CONTRIBUTING.md`.

scripts/generators/es_template.py
webmat: Yeah iirc it's a nasty bit that I added when adding `dns.answers`. Adding `type: object` everywhere we have further nesting would be fine and a noop, as far as I understand.

I'd prefer to keep it as it was, and only handle the explicit `type: object` in place when it's already there, for now.

webmat:
The build is now failing because the Golang base image for 1.13.x comes with 5 year old Python 3.5.2. Not sure why the Travis config for python 3.7 isn't kicking in 🤦‍♂

webmat:
Related: https://github.com/elastic/ecs/issues/630

webmat:
The only oddity left is around the two timestamp field's examples. I had never noticed that it was getting interpreted before. It was unquoted in the YAML source files, and was being changed slightly by Python 2.7. Upgrading to 3.6+ was changing it slightly, but in a different manner.

event.created:

2.7

- source value was: `2016-05-23T08:05:34.857Z`
- output value (e.g. in asciidoc) was: `2016-05-23 08:05:35.101000`

3.6/3.7

- source value was: `2016-05-23T08:05:34.857Z`
- output value (e.g. in asciidoc) was: `2016-05-23 08:05:34.857000+00:00`

I've quoted the example in the source YAML, to avoid the timestamp being interpreted. In the end, ECS is not actually mandating a precise timestamp format. So this is different than before, but this makes the timestamp examples consistent with what's in the source YAML everywhere.

Now:

- source value is: `'2016-05-23T08:05:34.857Z'`
- output value (e.g. in asciidoc) is: `2016-05-23T08:05:34.857Z`

webmat:
@rw-access @ruflin  This PR is ready for final review.

rw-access:
thanks for the updates @webmat. LGTM

Issue #570, opened 2019-10-01
further .text and keyword discussion


neu5ron:
for security use cases, large error messages, etc.. `.text` is an important aspect in searching.

**pro(s):**
 - faster searches
 - severely reduces impact on the backend (and ultimately customer/user experience)  

**con(s):**
 - increases storage

temporary proposal to meet in the middle (for now) instead of adding `.text` into the dynamic template for all strings fields VS having no `.text` fields at all
ECS and thus corresponding beats index templates have `.text` field for high value/impacting fields. List of fields off the top of my head would be:
- `process.executable.text`
- `process.args.text`
 - `url.original`
- `user.email`
- `user_agent.original`
- `error.stack_trace`
- `file.path`
- `host.name`
- any `.user.name` fields
- any `.domain` fields
- any `.as.organization.name` fields
- `http.request.body.content`
- `http.response.body.content`
- `os.name.full`

existing ECS reference:
https://github.com/elastic/ecs/issues/340
https://github.com/elastic/ecs/issues/104

webmat:
Thanks @neu5ron! Yes, the subtext here is that when only `keyword` fields are available, people will do searches with wildcards, potentially multiple wildcards. Prefix searches (aka trailing wildcard: "myword*") are very fast on `keyword` fields, but a search such as "*myword*" is essentially equivalent to a full table scan.

Another subtext is that users are free to add their own `.text` multi-fields (as custom fields) wherever they need it, even if it's not in ECS. On the other hand, I don't think people do that a lot in real life. They either don't know they can do that, or it's way too late and they are threat hunting right now. A project to go back & identify which fields should have a `.text` multi-fields, modifying the template, then reindexing their data is obviously out of the question when that happens. So "multiple wildcard search" it is... And the cluster 💥

So I think this is worth addressing in ECS, and making sure that a baseline of fields where full text search is useful are defined from the get go.

Other suggestions and thoughts about this are welcome :-)

webmat:
I would add `os.name.full` to the list. See why in https://github.com/elastic/ecs/issues/576#issuecomment-540748059

randomuserid:
So I reviewed all of our current rule sets and there are a few more ECS fields that are used by hunting searches. This list is not ranked but the process fields are clearly the most heavily used and process.args  / process.command_line may be the most frequently used. These are the fields not listed above: 

process.command_line
process.name
process.parent.args
process.parent.executable
process.pe.original_file_name
process.pe.signer_name
process.working_directory

[...]

PR #680, opened 2019-12-06
Add the default text analyzer to some fields

webmat:
This implements many fields mentioned in https://github.com/elastic/ecs/issues/570.

Note: I'm not adding `host.name`, mentioned in 570 because the default analyzer doesn't split on `-`. So I'm not sure it's worth adding. Please let me know if I'm missing something.

In going over the fields, I've identified a few more that I think may be interesting to add to this PR. Please voice your opinions. I'm happy to hold off or add now:

  * dns.question.name
  * threat.technique.name
  * vulnerability.description
  * tls.client.issuer, tls.client.subject, tls.server.issuer, tls.server.subject

@neu5ron @randomuserid @rw-access @MikePaquette @peasead

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,9 +16,10 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Add default `text` analyzer to `user_agent.original`. #575
+* Add default `text` analyzer as a multi-field to `user_agent.original`. #575
 * Added `file.attributes`. #611
 * Add `file.drive_letter`. #620
+* Added default `text` analyzer as a multi-field to around 25 more fields. #680
 
 #### Improvements
 
--- a/code/go/ecs/threat.go
+++ b/code/go/ecs/threat.go
@@ -22,14 +22,14 @@ package ecs
 // Fields to classify events and alerts according to a threat taxonomy such as
 // the Mitre ATT&CK framework.
 // These fields are for users to classify alerts from all of their sources
-// (e.g. IDS, NGFW, etc.) within a  common taxonomy. The threat.tactic.* are
-// meant to capture the high level category of the threat  (e.g. "impact"). The
+// (e.g. IDS, NGFW, etc.) within a common taxonomy. The threat.tactic.* are
+// meant to capture the high level category of the threat (e.g. "impact"). The
 // threat.technique.* fields are meant to capture which kind of approach is
-// used by  this detected threat, to accomplish the goal (e.g. "endpoint denial
+// used by this detected threat, to accomplish the goal (e.g. "endpoint denial
 // of service").
 type Threat struct {
 	// Name of the threat framework used to further categorize and classify the
-	// tactic and technique of the reported threat.   Framework classification
+	// tactic and technique of the reported threat. Framework classification
 	// can be provided by detecting systems, evaluated at ingest time, or
 	// retrospectively tagged to events.
 	Framework string `ecs:"framework"`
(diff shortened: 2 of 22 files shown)


peasead:
@webmat I think that `vulnerability.description` is a fantastic use case for `.text` as it would allow someone to search for `microsoft` instead of `microsoft has a vulnerability in...` (if I'm understanding the difference between keyword and text properly).

mbudge:
Would ngram be better for filepath and process path?

These tend to be longer strings.

Wildcard search against a text field when searching TB's of data might be slow, if a company is collecting logs from a medium to large enterprise network.

`
"analyzer": {
        "ngram_analyzer": {
          "type": "custom",
           "tokenizer": "ngram_tokenizer",
           "filter": [
              "lowercase"
           ]
        }
      },
      "tokenizer": {
        "ngram_tokenizer": {
          "type": "ngram",
          "min_gram": 3,
          "max_gram": 3
        }
      }
  	},
`

webmat:
@peasead Yes, that's exactly the reason I'm adding it there :-)

@mbudge Agreed, there are better ways to index the path-like fields like path, url etc. Still, I think it's good to add the default `.text` analyzer anyway for a few reasons:

* it works surprisingly well nonetheless. Words are tokenized on `/` and the stemming doesn't seem to be causing issues.
* We will keep working on crafting a great analyzer for these more structured fields, but this will take more time. When that's ready, we will introduce this as a differently named multi-field (not `.text`), since it will work differently. @neu5ron already provided a great example of a better analyzer for these fields in HELK in #570.

So overall the thinking for the path fields should be interpreted as "progress over perfection". But we'll still deliver the perfection in time ;-) If the fields specific to paths are superior enough, we could even deprecate their `.text` and take them out in ECS 2.0 / Elastic Stack 8.0, and keep only the specialized analyzers. We'll see

webmat:
@mbudge And you also bring another good point on performance. Right now we're adding these in order to enable efficient detections, mostly for the SIEM alerting engine. If users want to remove these analyzers and lose the ability to do these detections, they're free to do that.

webmat:
@dainperkins What would you think about having the default analyzer (full text search) on `threat.technique.name`? I'm not sure it's worth adding to `threat.tactic.name` as these are very high level and 1-2 words. But techniques are usually more detailed, and one could very well want to search for `threat.technique.name:denial` or something. WDYT?

dainperkins:
I think thats an excellent idea - I'll make a PR if you show me what needs to be done :)

peasead:
Whoops, thanks for the reminder @dainperkins 

@webmat do you need me to make a fresh PR with the changes to `vulnerability.description`?

webmat:
@peasead @dainperkins Meant to respond earlier, sorry I forgot to hit "Comment" 😂  I added them both to this PR directly.

peasead:
> @peasead @dainperkins Meant to respond earlier, sorry I forgot to hit "Comment" 😂 I added them both to this PR directly.

Roger roger. I'll drop the PR. 👍

webmat:
I consider this PR ready for final review, please voice opinions (esp. disagreement) soon, I'd like to merge tomorrow.

If you think additional fields would benefit from this, please voice your opinion. But this shouldn't be considered a blocker for the PR, they will be addressed in follow-up PRs :-)

dainperkins:
looks good to me

peasead:
LGTM

webmat:
Agree with some of those. I'd like to do subsequent PRs (in or after 1.4) for them, however.

Quick note on values that are dot separated (domains, hostnames) or even dash-separated (hostnames): the default analyzer doesn't deal well with them. Those would require a specially crafted analyzer that breaks them up correctly :-)

PR #687, opened 2019-12-10
Workaround for Beats issue with default_field growing too big

adriansr:
This updates the Beats generator to undestand the `default_field` option and add it to the generated yaml, as well as update the fields added by 1.3.0 to include the `default_field` tag.

This change was discussed in https://github.com/elastic/beats/pull/14844


CHANGELOG.next.md
```diff
@@ -18,6 +18,7 @@ Thanks, you're awesome :-) -->
 
 * Add default `text` analyzer to `user_agent.original`. #575
 * Added `file.attributes`. #611
+* Added the `default_field` option for Beats. #NNN
```
webmat:
```suggestion
* Added the `default_field` option for Beats. #687
```

:-)

webmat:
Note that I'm happy to adjust the generators a bit more around this as well. For example, generating 1 output file per Beat instead of a global one.

adriansr:
@webmat I think I've addressed all your comments. Now it's using `beats.default_field` and the documentation can be found in `schemas/README.md`.

Your higher level suggestions make sense but those would require a bigger refactor in how Beats integrates fields from ECS.

webmat:
Yeah I understand why this is easier for now.

I'm not clear on the semantics, however. This PR adds `beats.default_fields: false` on 29 fields. Is the expectation that all further fields added to ECS, after 1.2 always have `beats.default_fields: false`?

andrewkroh:
With the assumption that this is more or a temporary fix, I would like do this is a way that the files under schema/ never need to have changes. This way ECS contributors don't need to be burdened with understanding the effect of setting default_field or even what Beats are.

This could be achieved by have a snapshot of the current fields we allow for Beats to go into default_field. Then in the generator for the Beat's fields.yml we "whitelist" those fields and for any other fields not in that list we give them `default_field: false`.

webmat:
Thanks @andrewkroh, picking this up now.

Are you ok with me wording this (e.g. changelog, schema/readme, etc) specifically as a temporary workaround for Beats 7.6?

webmat:
Actually I'll put that wording in only for changelog. Because I agree with your assessment that this shouldn't burden ECS contributors, so I'll remove the readme mention.

PR #690, opened 2019-12-10
updated description to text datatype

peasead:
Referencing elastic/ecs#680  to change the datatype of `vulnerability.description` from `keyword` to `text`.


webmat:
Sorry Andrew, I meant to respond much earlier to #680. I added the `text` analyzer over on the other PR instead.

Also, the way we're doing it is by adding a multi-field, not by changing the type on the field itself. End result is the following:

* `vulnerability.description` remains `keyword` for what kw is good at (aggregations, exact matches, etc)
* `vulnerability.description.text` is added as `text`, and enables full text search (e.g. `vulnerability.description.text:microsoft`)

Changing the datatype itself would be a breaking change 🔕 😉 

I'll close this one, as it's implemented in 680. Sorry for the delay in my response :-)

peasead:
Already fix't by @webmat

PR #692, opened 2019-12-12
Introduce the initial values for ECS categorization

webmat:
This pull request introduces the first set of official values for the 4 ECS categorization fields, which were reserved up to this point (event.kind, event.category, event.type, event.outcome).

Note that more values are currently being considered for some of these fields. Users and implementers can preview the other values and offer feedback on them in this public document: https://ela.st/ecs-categories-draft

**TODO**

- [x] Changelog entry :-)
- [x] Make values themselves linkable.
  - [x] Add table of content at top of "Accepted Values" pages
- [x] Fix the horizontal list of values in the event field page
- [x] Resolve/remove inconsistencies and flesh out the new documentation pages & descriptions

[Preview](http://ecs_692.docs-preview.app.elstc.co/diff)


webmat:
Note: The purpose of this PR at this point is simply to preview the rendering with a more realistic set of values. Not all of these values are expected make it for the initial release of the categorization values. We're still hammering out the details of some of these values.

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
+            Indicates that this event describes only an attempt for which the result
+            is unknown. For example, if the event contains information only about a
+            request in an entity transaction that usually results in a response,
+            populating `event.outcome:unknown` is appropriate.
 
     - name: type
       level: core
       type: keyword
-      short: Reserved for future usage.
+      short: Event type (Categorization Field)
```
webmat:
```suggestion
      short: Event type. The third categorization field in the hierarchy.
```

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            is often populated for events coming from firewalls, intrusion detection
+            systems, endpoint detection and response systems, etc.
+            It is used to indicate that an alert was triggered.
+        - name: event
+          description: >
+            `event.kind:event` is the most general and most common value of this
+            field. It is used to represent events that indicate that something happened.
+        - name: metric
+          description: >
+            Used to indicate that this event is a measurement taken at given point in time.
```
andrewstucki: So, I'm not too huge of a fan of the inconsistency in the sentence syntax here--`alert` and `event` values both start off with `event.kind:*` while the others just start with a predicate or a reference back to the field value (i.e. 'This value')--we should be consistent--visually I'm not a fan either.

If we want to follow the format of the top-level docs for the fields, it makes sense to me to start everything with either "This value" or with `event.kind:*`.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+          description: >
+            The signal value is used by Elastic Kibana apps, such as SIEM, for app-specific purposes.
+            `event.kind:signal` is thus reserved and should not be used for the ingestion
+            of events into Elasticsearch.
 
     - name: category
       level: core
       type: keyword
-      short: Event category.
+      short: Event category. (Categorization Field)
```
webmat:
```suggestion
      short: Event category. The second categorization field in the hierarchy.
```

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
       example: 4648
 
     - name: kind
-      level: extended
+      level: core
       type: keyword
-      short: The kind of the event.
+      short: The kind of the event. (Categorization Field)
```
webmat:
```suggestion
      short: The kind of the event. The highest categorization field in the hierarchy.
```

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
       example: user-password-change
 
     - name: outcome
-      level: extended
+      level: core
       type: keyword
-      short: The outcome of the event.
+      short: The outcome of the event. (Categorization Field)
```
webmat:
```suggestion
      short: The outcome of the event. The lowest categorization field in the hierarchy.
```

Wording question: "lowest" or "last"? "lowest" sounds less important, which is not the case here. I really do want to know if the event of category "authentication" had outcome "success" or "failure" :-)

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            Elasticsearch, MongoDB, etc. Use this category to visualize and analyze
+            database activity such as accesses and changes.
+          expected_event_types:
+            - access
+            - change
+            - info
+            - error
+        - name: driver
+          description: >
+            Having to do operating system device drivers and similar software entities
```
andrewstucki: Should we make these into full sentences too? Something like:

> Events in the driver category have to do with ...

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+          description: >
+            Relating to intrusion detections from IDS/IPS systems and functions,
+            both network and host-based. Use this category to visualize and analyze
+            intrusion detection alerts from systems such as Snort, Suricata,
+            and Palo Alto threat detections.
+          expected_event_types:
+            - info
+        - name: malware
+          description: >
+            Malware detection events and alerts. Use this category to visualize and analyze
```
andrewstucki: Besides all of the "Relating to"--these are also partial sentences that I'm wondering if we want to fix.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+        - name: pipeline_error
+          description: >
+            This value indicates that an error occurred during the ingestion of this event,
+            and that event data may be missing, inconsistent, or incorrect.
+            `event.kind:pipeline_error` is often associated with parsing errors.
+        - name: signal
+          description: >
+            The signal value is used by Elastic Kibana apps, such as SIEM, for app-specific purposes.
+            `event.kind:signal` is thus reserved and should not be used for the ingestion
+            of events into Elasticsearch.
```
webmat: Even if users should not be populating this value in their pipelines, I think we should explain what this value actually means. This description doesn't do that.

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
 
     - name: outcome
-      level: extended
+      level: core
       type: keyword
-      short: The outcome of the event.
+      short: The outcome of the event. (Categorization Field)
       description: >
-        The outcome of the event.
+        This is one of four ECS Categorization Fields, and indicates the
```
andrewstucki: Not that big of a deal, but wondering if this fully makes sense to everyone contextually? "This is one of..." in the context of "Allowed Values"? Wondering if we want to change the heading docs in a subsequent PR, maybe make the page header be something like `ECS Categorization Field: event.outcome` and then have "Table of Contents" say something like "Allowed Values"

Just raising this here since the "description" of the field makes sense to me, but it doesn't quite seem to fit the context of where it is on the particular page, since I'd assume at that point we'd already be talking about the "Allowed Values" rather than the field itself.

Screen Shot 2019-12-16 at 11 14 16 AM

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
+          description: >
+            The change event type is used for the subset of events within a category
+            that indicate that something has changed. If semantics best describe an
+            event as modified, then include them in this subcategory.
+            Common examples include `event.category:process AND event.type:change`,
+            and `event.category:file AND event.type:change`.
+            You can further distinguish change operations using the ECS `event.action` field.
+        - name: creation
+          description: >
+            The "creation" event type is used for the subset of events within a category
```
andrewstucki: should we quote the other values like this? (i.e. "access", "change", "deletion", etc.) Also do we need the sentence structure to be consistent with whatever we do for `event.category` values?

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+        This field is closely related to `event.type`, which is used as a subcategory.
+      example: authentication
+      allowed_values:
+        - name: authentication
+          description: >
+            Events in this category are related to the challenge and response process
+            in which credentials are supplied and verified to allow the creation of a session.
+            Common sources for these logs are Windows event logs, ssh logs, etc.
+            Visualize and analyze events in this category to look for unusual login activity,
+            failed logins, etc.
```
andrewstucki: Can we change this to say

> Visualize and analyze events in this category to look for authentication-related activity such as unusual or failed login attempts.

? Ending two sentences in a row in `etc.` makes me 😢

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            - creation
+            - deletion
+            - info
+        - name: host
+          description: >
+            Events and metrics about hosts. Usually higher-level information about host
+            activity from an external perspective. Different than operating system
+            in the sense that host events are usually externally visible and independent from the OS.
+            Note that `event.category:host` is not meant to indicate events that are simply
+            "happening on a host". Use this category to visualize and analyze inventories
```
andrewstucki: Maybe this?

> Use this category to visualize and analyze information such as host inventories or host lifecycles.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            intrusion detection alerts from systems such as Snort, Suricata,
+            and Palo Alto threat detections.
+          expected_event_types:
+            - info
+        - name: malware
+          description: >
+            Malware detection events and alerts. Use this category to visualize and analyze
+            malware detections from EDR/EPP systems such as Elastic Endpoint Security,
+            Symantec Endpoint Protection, Crowdstrike, and network IDS/IPS systems and
+            functions such as Palo Alto Networks threat and Wildfire logs.
```
andrewstucki: Does this mean the following?

> ... such as Palo Alto Networks threat **_logs or_** Wildfire logs.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            - change
+            - deletion
+            - info
+            - installation
+            - start
+        - name: process
+          description: >
+            Relating to the operation of software processes executing within operating
+            systems on hosts. Use this category of events to visualize and analyze process
+            starts, process parents, process relationships, etc.
```
andrewstucki: Maybe?

> Use this category of events to visualize and analyze process-specific information such as lifecycle events or process ancestry.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+          description: >
+            Used to indicate that this event is a measurement taken at given point in time.
+            Examples include CPU utilization, memory usage, or a vulnerability scan result.
+            Events with `event.kind:metric` indicate that a measurement was taken.
+        - name: state
+          description: >
+            This value is similar to metric, except that the entity being measured does not
+            provide a numeric metric value, but rather one of a fixed set of conditions or states.
+            For example a periodic event reporting a "fin_wait" state of a TCP connection
+            on a host might use `event.type:state`.
```
webmat: I wonder if we should hold off on releasing this one until we've confirmed the semantics a little more. Here are two reservations I still have, at this time:

The state that can be observed at one moment in time (the cluster is green). The state transitions (e.g. cluster goes from yellow to green). This definition should address whether both kinds of events belong in event.kind:state. My understanding is that only "point in time" observations, and not the state transitions should belong here. I'm not sure if we have agreement on that yet.

Another worry I have is that in many cases, checking on the state of a system can actually report back metrics. So in these cases, these state checks would be `event.kind:metric` because they contain metrics. If that's the case, how useful/reliable will `event.kind:state` be? Perhaps it's better to have all state checks in event kind metric?

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
+        - name: failure
+          description: >
+            Indicates that this event describes a failed result. A common example is
+            `event.category:file AND event.type:access AND event.outcome:failure`
+            to indicate that a file access was attempted, but was not successful.
+        - name: success
+          description: >
+            Indicates that this event describes a successful result.  A common example is
+            `event.category:file AND event.type:create AND event.outcome:success`
+            to indicate that a file was successfully created.
```
andrewstucki: 👍 on the specific examples of how to use these values.

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
-
-      - name: end
-        description: >
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
-          incididunt ut labore et dolore magna aliqua.
+        This is one of four ECS Categorization Fields, and indicates the
+        third level in the ECS category hierarchy.
+
+        `event.type` represents a categorization "sub-bucket" that, when used along with
+        the `event.category` field values, enables filtering events down to a level
```
andrewstucki: Maybe?

> when used along with the `event.category` field values _**it**_ enables

schemas/event.yml
andrewstucki: Can we make this consistent with the `kind` and `outcome` fields, see sentence structure:

> The outcome of an event.

v.

> Event category.

schemas/event.yml
andrewstucki: same as `category` wonder if we can make "Event type" into a full sentence.

schemas/event.yml
webmat: I agree there's an inconsistency there. I think starting each definition with "This value" reads a little more easily.

However whenever need to expand a little more or give additional examples, going with the more explicit `fieldname:value` is much easier.

Perhaps we can always start with "This value" and always use `fieldname:value` at any other time where we need to mention a field and a value later?

I don't have a strong opinion on this either.

schemas/event.yml
webmat: Yeah I think these all need more fleshing out. If we mark all of this as beta and subject to change, I think we can keep refining and backport to 1.4 aggressively, so the docs website gets the improvements even after the release.

Otherwise we typically only backport fixes like obvious typos or inconsistencies and so on.

schemas/event.yml
webmat: Yeah, I think they work well for the feedback document, but in here we definitely want to aim at complete sentences. Totally agree with that.

schemas/event.yml
andrewstucki: If there's not clarity in what this category is used for, I'd tend to agree. **But** isn't there agreement on what this field is used for? If the answer is "no" we should strip it out, but if others have been thinking of this the same way I don't see a need to remove it from the initial release.

As you state, the language seems to indicate any sort of data state falls under data involving "one of a fixed set of conditions or states." I'm not sure that there's really a meaningful distinction between "heartbeat" style periodic events or events that have the same exact data, but are only fired on state transitions--so I would imagine that this would cover either use-case.

Just for illustration purposes, what I've been thinking of using this category for is for any sort of Endpoint internal state transitions--i.e. policy applications, upgrades, lifecycle, etc.--so besides the surrounding metadata of things like versions, categorization fields, and host-level information, I envision that implementations that would use this category similarly to how I envisioned would contain a lot of hybrid data, with much "application-specific state" in custom fields. That's how I see this differ from the idea of "metrics."

andrewstucki:
@webmat I like the idea of having a `beta` marker since fleshing this fields out fully are still a pretty large WIP. Until we've released a fully fleshed out list where we only have a couple of potentials that we're occasionally working with (v. introducing like 13+ values at once), I'm cool with this staying in "beta" mode.

schemas/event.yml
rw-access: Agreed, and we should communicate all reserved values similarly.

schemas/event.yml
rw-access: I'm for consistency, but I think we should also strive to describe fields without using the name of a field.
Neither "The outcome of an event", nor "the event's outcome" tell me what an _outcome_ actually is.

Similarly, telling its position in the hierarchy doesn't help much. I think need to link to a separate ECS hierarchy page/doc from here, where we can explain more about what the hierarchy is and how it works.

schemas/event.yml
rw-access: I would avoid enumerating these examples. Mentioning Elastic Endpoint Security is fine, but anymore can look like bias or omission, and I think we can safely communicate `malware` without needing to venture into that territory.

schemas/event.yml
rw-access: I think we need a visualization to communicate these better. For instance,
```
event.kind:                   event                     alert
                          /     |    \                /      \
event.category:      process  file  network       malware    intrusion_detection
                       ...   /     \
event.type:              access     change
                           |      /    |    \
event.action:            read  create modify rename
                               /   \
event.outcome:            success   failure
```

schemas/event.yml
```diff
@@ -99,45 +220,108 @@
...
+        appropriate for single visualization.
```
rw-access: Yes, it enables visualization, but we should communicate also the value of this field beyond organization.
What does event type _mean_? How should I approach this field?

schemas/event.yml
MikePaquette: +1 for Mat's suggestion - start each with with "This value..."   Use `fieldname:value` for subsequent discussion.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            For example a periodic event reporting a "fin_wait" state of a TCP connection
+            on a host might use `event.type:state`.
+        - name: pipeline_error
+          description: >
+            This value indicates that an error occurred during the ingestion of this event,
+            and that event data may be missing, inconsistent, or incorrect.
+            `event.kind:pipeline_error` is often associated with parsing errors.
+        - name: signal
+          description: >
+            The signal value is used by Elastic Kibana apps, such as SIEM, for app-specific purposes.
```
MikePaquette: This value is used by the Elastic SIEM app to denote an Elasticsearch document that was created by a SIEM detection engine rule.

schemas/event.yml
MikePaquette: OK, How's this text?

schemas/event.yml
MikePaquette: The shortness of the short definition make it challenging to keep good style practices, but I agree we can do better. I vote for keeping the hierarchy comments for now, but I love the idea of a diagram to be added later.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+        `event.category` represents the "big buckets" of ECS categories. For example,
+        filtering on `event.category:process` yields all events relating to process activity.
+        This field is closely related to `event.type`, which is used as a subcategory.
+      example: authentication
+      allowed_values:
+        - name: authentication
+          description: >
+            Events in this category are related to the challenge and response process
+            in which credentials are supplied and verified to allow the creation of a session.
+            Common sources for these logs are Windows event logs, ssh logs, etc.
```
MikePaquette: Common sources for these logs include Windows event logs and ssh logs.

schemas/event.yml
MikePaquette: failed logins, and other authentication-related activity.

schemas/event.yml
MikePaquette: Agreed that full sentences will read better in the documentation.

schemas/event.yml
```diff
@@ -38,53 +38,174 @@
...
+            both network and host-based. Use this category to visualize and analyze
+            intrusion detection alerts from systems such as Snort, Suricata,
+            and Palo Alto threat detections.
+          expected_event_types:
+            - info
+        - name: malware
+          description: >
+            Malware detection events and alerts. Use this category to visualize and analyze
+            malware detections from EDR/EPP systems such as Elastic Endpoint Security,
+            Symantec Endpoint Protection, Crowdstrike, and network IDS/IPS systems and
```
MikePaquette: network IDS/IPS systems such as Suricata, and

schemas/event.yml
MikePaquette: I think it was Nate who suggested explicitly including IDS type events from systems that are firewalls or UTM devices, so I'd like to include these. I've suggested a compromise approach here.

MikePaquette: other sources of malware-related events such as Palo Alto Networks threat logs and Wildfire logs.

schemas/event.yml
MikePaquette: I love the diagram.  I think that the use of "high" to "low"  (and similarly top to bottom) when talking about hierarchical structures is well known, and I don't think readers will associated lowest with "less important."

schemas/event.yml
MikePaquette: Could re-structure, but I think it reads correctly as is.

schemas/event.yml
MikePaquette: The subcategory of an event.

schemas/event.yml
MikePaquette: the `event.category` field values, enables filtering events down to a logical set that is appropriate for common analysis and

schemas/event.yml
MikePaquette: visualizations.

schemas/event.yml
MikePaquette: I vote for consistency.  In this case, let's just remove the quotes.

MikePaquette: The creation event type is used for the subset of events within a category

scripts/generators/asciidoc_fields.py
```diff
@@ -295,91 +294,106 @@ def nestings_row():
...
     for cat_field in category_fields:
         section_text += render_field_values_page(ecs_flat[cat_field])
     return section_text
 
 
-def values_header():
+def values_section_header():
     return '''
 [[ecs-category-field-values-reference]]
 == {ecs} Category Field Values
```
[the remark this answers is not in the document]
webmat: Thanks, I had missed this rename

schemas/event.yml
webmat: Now that we've agreed on adding a beta warning to the section, I'm less worried here.

Let's leave it in for now and see if we take it out later.

schemas/event.yml
webmat: The heading of that documentation section is indeed hidden far away from these 4 pages (it's in the breadcrumbs and the sidebar only).

I'll experiment with bringing this front and center on each of the 4 pages. Good point 👍

webmat: I think it looks a little funny in the sidebar, but it works better on the page itself:

ECS Categorization Field in each page heading

andrewstucki: That's fine, still think we should circle back at some point and figure out something that works a bit better contextually--but yeah, the navigation looks nasty with that convention.

webmat: I just realized I hadn't renamed the heading before the list... Adjusting now.

schemas/event.yml
rw-access: do we also want to communicate other conventions for the values (nouns, etc.)?

webmat:
Since we have the beta warning on the section, we'll be able to keep clarifying and improving the definitions and the wording over the next little while, and backport to the 1.4 docs (although additional allowed values would be in subsequent releases).

I'd like to request a final review, make sure there's no blocker left.

webmat:
I'll reword the warnings on each of the 4 allowed values pages. Right now it's contradictory with the warning at the top of the whole section.

Currently, the gist is

**Top of section**: "these values may change in the future"

**On each page**: "warning if you don't use these values bad things will happen"

I'll reword the warning on each page to the effect of

"warning, once we remove the beta label, if you don't use these values bad things will happen"

So not a big change, just a nod to the effect that this is still being worked on and may still change.

webmat:
Yeah I'll start collecting this feedback today, and perhaps even start a follow-up PR

Issue #678, opened 2019-12-05
Source/target and diff modeling


andrewstucki:
This is a more generic application of modeling either "source/target" style events or "diff" events that I'm spinning off from https://github.com/elastic/ecs/issues/589

As I initially mentioned (https://github.com/elastic/ecs/issues/589#issuecomment-555203355). There are a number of things that ECS should support modeling including things like:

- setuid/setgid operations
- file modification events (i.e. renames, permissions)
- IPC calls
- network requests/flows (current use of `source` and `destination`)
- user modifications
- process execs
- registry modifications
- windows source/target audit log info

Overall these more or less fall into two categories:

1. Modelling communication between two like things (network connections, IPCs, windows audit log)
2. Modification events

Currently the way ECS has started to approach this is to make fields that are specific to each of these domains, i.e. `source`/`destination` are currently for network modeling only, and then there's also `client`/`server`

I'm advocating for adopting a more generic field set that allows you to do generic source/target or diff modeling which would essentially allow you to embed any other field set in it.

For example--something like `origin` and `target` (sad that `source`/`destination` is already taken):

**For file modifications:**

```
origin.file.name = "foo"
target.file.name = "bar"
```

**For process execs:**

```
origin.process.path = "foo"
target.process.path = "bar"
```

**For network requests (slightly difficult because of the lack of `port` info):**

```
origin.host.ip = "foo"
target.host.ip = "bar"
```

**For user modification (maybe by another user `baz` who did the modification?):**

```
user.name = "baz"
origin.user.name = "foo"
target.user.name = "bar"
```

Thoughts?

janniten:
> **For user modification (maybe by another user `baz` who did the modification?):**
> 
> ```
> user.name = "baz"
> origin.user.name = "foo"
> target.user.name = "bar"
> ```

Hi @andrewstucki,
baz is the user performing the modification?
Thank you

andrewstucki:
@janniten that was the idea in that particular example, but the desire was just to highlight the need to model any sort of diffing or source/destination style events in a sizeable portion of the existing field sets out there.

I'm fairly open to ideas on naming/implementation (would this be at the top-level as I suggested or embedded inside each top-level field set, etc.), but just calling out the need to solve this issue for a number of use-cases so that we don't just solve the same problem a myriad of different ways as they arise. So, any suggestions/opinions on how we should go about this are more than welcome!

rw-access:
[...]

PR #694, opened 2019-12-16
Added related.user Field

janniten:
From discussions https://github.com/elastic/ecs/issues/589 and https://github.com/elastic/ecs/issues/678  and from working with windows events https://github.com/elastic/beats/pull/13530  in the winlogbeat security module many events has related users.
In spite of what naming convention/schema will be chosen to represent n-ary user relationships it would be useful to have the related.user field in order to facilitate pivoting.


CHANGELOG.next.md
```diff
@@ -1,63 +1,34 @@
 <!-- When adding an entry to the Changelog:
-
```
webmat: @janniten You're accidentally overwriting the current changelog with an older format.

To fix this, you can run `git checkout CHANGELOG.next.md` and then introducing your changelog entry after that ;-)

janniten: Fixed (I think so : ) )

PR #696, opened 2019-12-17
Set master version to 1.5.0-dev

webmat:


webmat:
This will be merged after 1.4 is released

PR #698, opened 2019-12-17
Add missing hash fields to process.parent

marshallmain:
Recently process.parent fields that are identical to the process fields were added to ECS, but the hash fields that are nested under "process" did not get added to process.parent.  This PR adds "process.parent" to the list of nestings in "hash" and also support for nesting fields at levels other than the top level of the destination schema using the '.' notation (e.g. process.parent).  Previously nested fields could only be inserted at the top level.

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -1967,7 +1967,7 @@ type: keyword
 
 ==== Field Reuse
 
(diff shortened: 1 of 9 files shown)


webmat:
Huge thanks for submitting this @marshallmain 💯

PR #702, opened 2019-12-19
Use an example IP instead of letters for clarity

rud:


webmat:
Hello! I think that was actually a good idea, thanks for point that out!

I'll make that adjustment in the next few days, as a Christmas present 🎁  ;-)

Thanks for dropping by!

webmat:
I've been meaning to add this for a while https://github.com/elastic/ecs/pull/711, so I improved the IP example at the same time :-)

rud:
I was trying to fix this in the right location, but got sniped installing python 2 and ran out of time. I'm glad you could use the idea 👍

PR #706, opened 2019-12-20
Move to --direct_html doc build option:

webmat:
- Better asciidoctor compatibility
- Moar fast builds

Relates to https://github.com/elastic/docs/pull/1559

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ docs:
 	if [ ! -d $(PWD)/build/docs ]; then \
 		git clone --depth=1 https://github.com/elastic/docs.git ./build/docs ; \
 	fi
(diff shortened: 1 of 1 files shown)


nik9000:
I don't think you need to do this now that I've merged https://github.com/elastic/docs/pull/1659

nik9000:
With that PR adding the option will just log a happy message. Nice and all, but not needed.

webmat:
I always build the docs locally though, when working on docs. So that helps speed there too, no?

nik9000:
> I always build the docs locally though, when working on docs. So that helps speed there too, no?

Nah. Once you pull the most recent version of the docs that flag doesn't do anything. It was important for every other repo to add it because it wasn't the default. But because you were last I switched it to the default as soon as I merged you.

webmat:
Ah, thanks for the clarification!

PR #708, opened 2019-12-20
Add session id for process and process.parent

andrewstucki:
So, as the docs state, this field would be representative of a session identifier for processes. Currently in our endpoint pre-elastic schema we ship two fields for process events:

1. a Windows specific "authentication_id" which corresponds to the `AuthenticationId` LUID [here](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control)
2. a POSIX specific "session_id" which corresponds to the value returned from a call to [`setsid`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html)

These fields are somewhat complex in the Windows and POSIX parlance because:

1. A Windows session LUID corresponds directly to a user's logon session, but is also passed to every process spawned during that logon session
2. A POSIX `sid` value corresponds often to a user's terminal session, but, in the case of daemonized processes, is decoupled from an actual "user session"
3. Windows LUIDs need to be represented in a string format
4. POSIX `sid` values are just a `pid_t` structure, so they're numbers

That said, despite not necessarily being coupled to a "user session" -- both values are directly tied to a process executing and represent some sort of session information. Therefore, I don't think it's a stretch to unify on a platform-agnostic definition that can be added to the ECS spec under `process` rather than `user`.

Additionally, since this value could either be a string or an integer, I made them type `keyword` so we could leverage either.

Thoughts?

### DIFF
--- a/schemas/process.yml
+++ b/schemas/process.yml
@@ -102,6 +102,35 @@
         Identifier of the group of processes the process belongs to.
 
 
+    - name: session_id
+      level: extended
+      type: keyword
+      short: The Windows or POSIX session identifier for a process.
+      description: >
+        The Windows or POSIX session identifier for a process.
+
+        On Windows systems, this corresponds to the LUID of the session itself
+        rather than and individual process access token. For more information
+        see `AuthenticationId` at https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control.
+
+        On POSIX systems, this corresponds to the session id returned from a
+        call to `setsid` (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html).
+
+    - name: parent.session_id
+      level: extended
+      type: keyword
+      short: The POSIX or Windows session identifier for a process.
+      description: >
+        The Windows or POSIX session identifier for a process.
+
+        On Windows systems, this corresponds to the LUID of the session itself
+        rather than and individual process access token. For more information
+        see `AuthenticationId` at https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control.
+
+        On POSIX systems, this corresponds to the session id returned from a
+        call to `setsid` (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html).
+
+
     - name: command_line
       level: extended
       type: keyword
--- a/code/go/ecs/process.go
+++ b/code/go/ecs/process.go
@@ -54,6 +54,26 @@ type Process struct {
 	// Identifier of the group of processes the process belongs to.
 	ParentPGID int64 `ecs:"parent.pgid"`
 
+	// The Windows or POSIX session identifier for a process.
+	// On Windows systems, this corresponds to the LUID of the session itself
+	// rather than and individual process access token. For more information
+	// see `AuthenticationId` at
+	// https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control.
+	// On POSIX systems, this corresponds to the session id returned from a
+	// call to `setsid`
+	// (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html).
+	SessionID string `ecs:"session_id"`
+
+	// The Windows or POSIX session identifier for a process.
+	// On Windows systems, this corresponds to the LUID of the session itself
+	// rather than and individual process access token. For more information
+	// see `AuthenticationId` at
+	// https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control.
+	// On POSIX systems, this corresponds to the session id returned from a
+	// call to `setsid`
+	// (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html).
+	ParentSessionID string `ecs:"parent.session_id"`
+
 	// Full command line that started the process, including the absolute path
 	// to the executable, and all arguments.
 	// Some arguments may be filtered to protect sensitive information.
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -3311,6 +3311,21 @@ example: `4241`
 
 // ===============================================================
 
+| process.parent.session_id
+| The Windows or POSIX session identifier for a process.
+
+On Windows systems, this corresponds to the LUID of the session itself rather than and individual process access token. For more information see `AuthenticationId` at https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_control.
+
+On POSIX systems, this corresponds to the session id returned from a call to `setsid` (https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsid.html).
+
+type: keyword
+
+
+
+| extended
+
+// ===============================================================
+
 | process.parent.start
 | The time the process started.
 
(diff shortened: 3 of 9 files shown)


rw-access:
I think token information may need to be its own field set that can be nested under process. Windows allows for impersonation tokens and thread specific tokens. Each process will have a primary token as well. For some of the points

> 1. A Windows session LUID corresponds directly to a user's logon session, but is also passed to every process spawned during that logon session

It corresponds to a token. Every user's logon session will have a token, but there are more tokens than there are sessions, from what I understand.

> 3. Windows LUIDs need to be represented in a string format

I don't think so. These are always unsigned 64 bit integers. Sysmon and Windows security events represent this as hex, but Endgame represents this as uint64. Sysmon also generates a UUID for this, because these values can get reused after reboot and across hosts.

> 4. POSIX sid values are just a pid_t structure, so they're numbers

I think session_id and authentication_id have basically nothing to do with each other.

schemas/process.yml
```diff
@@ -102,6 +102,35 @@
         Identifier of the group of processes the process belongs to.
 
 
+    - name: session_id
+      level: extended
+      type: keyword
+      short: The Windows or POSIX session identifier for a process.
```
rw-access: Windows has a concept of session ids, but this actually has nothing or little to do with LUIDs.
I think it would be a mistake to conflate the two. If you rip out all mentions of LUIDs, and refer to what Windows calls "sessions", then you're fine.

https://techcommunity.microsoft.com/t5/Ask-The-Performance-Team/Sessions-Desktops-and-Windows-Stations/ba-p/372473

schemas/process.yml
```diff
@@ -102,6 +102,35 @@
...
+      description: >
```
rw-access: Can you add examples please?

schemas/process.yml
rw-access: Also for reference, here's an example 4624 for a Windows logon event. I believe elevation and calls to APIs like CreateProcessAsUser will cause this (just a different logon type). However, the child processes should be in the same session.
```
Subject:
    Security ID: SYSTEM
    Account Name: DESKTOP-LLHJ389$
    Account Domain: WORKGROUP
    Logon ID: 0x3E7

Logon Information:
    Logon Type: 7
    Restricted Admin Mode: -
    Virtual Account: No
    Elevated Token: No

Impersonation Level: Impersonation

New Logon:
    Security ID: AzureAD\RandyFranklinSmith
    Account Name: rsmith@montereytechgroup.com
    Account Domain: AzureAD
    Logon ID: 0xFD5113F
    Linked Logon ID: 0xFD5112A
    Network Account Name: -
    Network Account Domain: -
    Logon GUID: {00000000-0000-0000-0000-000000000000}

Process Information:
    Process ID: 0x30c
    Process Name: C:\Windows\System32\lsass.exe

Network Information:
    Workstation Name: DESKTOP-LLHJ389
    Source Network Address: -
    Source Port: -

Detailed Authentication Information:
    Logon Process: Negotiat
    Authentication Package: Negotiate
    Transited Services: -
    Package Name (NTLM only): -
    Key Length: 0
```

rw-access: Finally (maybe), here's what Sysmon reports
https://github.com/hunters-forge/OSSEM/blob/master/data_dictionaries/windows/sysmon/event-1.md

Note that it has
- **TerminalSessionId**: which should be session_id
- **LogonId**: which is authentication_id, I believe
- **LogonGuid**: a UUID generated for the LogonID, which avoids the same reuse problem we have with PIDs

Processes can only have a single session_id, and a single primary token/authentication_id. However, because threads can use impersonation tokens, this means that we will need to somehow track the active/impersonated tokens for I/O.

For example, if svchost.exe uses an impersonated token while modifying a file, we might expect:
- **process.session_id** to indicate the session_id for svchost.exe. This will match the corresponding process creation event
- **process.logon_id** to match the authentication id for the process' primary token
- **token.logon_id** to indicate the impersonated token
- **token.logon_uuid** for the UUID to match token.logon_id

marshallmain:
@rw-access one snag with representing the windows LUID is that elasticsearch does not support unsigned 64 bit ints.  In the past we've gotten around this by representing uint64s as keyword fields instead once they get to elasticsearch.

schemas/process.yml
andrewstucki: So, from what I can tell based off the MSDN docs, it almost looks like `AuthenticationId` is a LUID that corresponds to an otherwise "user friendly" session id (the low number in task manager) and is unique across OS restarts--so it really is a session identifier, but one that's not as "user friendly" as the thing you see in task manager, but more useful.

From what I can tell looking at the structs from the docs, I might be wrong, but I don't believe you can actually impersonate an `AuthenticationId`--my naive reading of the docs is that this is where tokens split between authentication and authorization--where the `AuthenticationId` is tied explicitly to the session of "who logged in" v. impersonation tokens which allow you to change the ACLs associated with the process.

But I would gladly be corrected if this is a misunderstanding of the (somewhat confusing) MSDN docs.

So, long story short (from my perspective):

1. Don't think we necessarily need to handle modeling token impersonation here, and
2. The `session_id` v. `logon_id` distinction makes sense, but if they're literally 1-1 (in the sense that in a single boot of the OS you'll always have the same LUIDs correspond to session ids 0, 1, 2, etc.) then, apart from matching what's seen in the task manager and hence being more "user friendly" is it really useful to have this distinction?

webmat:
Quick note, haven't reviewed everything. Performance wise, `keyword` is faster for exact match filtering and aggregations than integers anyway. So I think it would make sense for this field to remain `keyword`.

marshallmain:
With that in mind we may want to turn a lot of our integers into multi fields then so we can get the advantages of both numeric and keyword types.

webmat:
If you need the numeric abilities like numeric sorting and arithmetic (and aren't dealing with uint64 or bigger), then I think you should stick to numeric fields, though.

schemas/process.yml
gabriellandau: @rw-access is correct.  Even though they both have the word 'session' in their name, they are different entities.

A session is a logical separation created by Windows to isolate users and services sharing the same server, and to enforce different types of quotas (memory, etc).  It was later extended in Vista add some security properties via [Session Isolation](https://techcommunity.microsoft.com/t5/ask-the-performance-team/application-compatibility-session-0-isolation/ba-p/372361)

A logon session is a container for security credentials, created when a user or service logs into the system.  A new logon session can be created simply by calling the [LogonUser ](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonuserw) API.

One property of the logon session is the authentication LUID:
```
0:000> dt ntoskrnl!_SEP_LOGON_SESSION_REFERENCES LogonId
   +0x008 LogonId : _LUID
```
> The locally unique identifier (LUID) is a 64-bit value guaranteed to be unique only on the system on which it was generated. The uniqueness of an LUID is guaranteed only until the system is restarted.

https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/igpupvdev/ns-igpupvdev-_luid

I believe including Authentication LUIDs in events lets us write rules to detect certain types of token-based attacks, [like those shown here](https://2017.zeronights.org/wp-content/uploads/materials/ZN17_James_Forshaw_Bypassing_UAC.pdf), however I do not see any such rules acting on the authentication LUID: https://github.com/elastic/eql-rules/search?q=authentication_id&unscoped_q=authentication_id 

CC @wburgess1 @joe-desimone

PR #711, opened 2019-12-23
Add `related.hash`.

webmat:
Also use an actual IP, in an example in the field set description.


code/go/ecs/related.go
```diff
@@ -26,11 +26,16 @@ package ecs
 // A concrete example is IP addresses, which can be under host, observer,
 // source, destination, client, server, and network.forwarded_ip. If you append
 // all IPs to `related.ip`, you can then search for a given IP trivially, no
-// matter where it appeared, by querying `related.ip:a.b.c.d`.
+// matter where it appeared, by querying `related.ip:192.168.1.15`.
```
andrewkroh: Consider using ranges from https://tools.ietf.org/html/rfc5737. (Doesn't really matter but someone reserved ranges specifically for docs so we might as well use them.)

PR #713, opened 2019-12-24
Add a documentation section about custom fields

webmat:
[Preview of the new page](http://ecs_713.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-custom-fields-in-ecs.html)


webmat:
Trying to write a section about "what it means to have a conflict with ECS" is nearly impossible. There's so many ways users can design their pipelines and indices, that we essentially need to explain all possible kinds of problems. Would be more confusing than helpful, IMO.

Perhaps we should introduce this page as is. It's already a step forward.

webmat:

PR #717, opened 2020-01-08
Beats: Set default_field=false in multi_fields

adriansr:
The generator was not adding `default_field: false` to `multi_fields`. This again causes Beats to exceed the number of fields in the default list due to the recent addition of some multi_fields.

This will add the tag to multifields that are not in the whitelist and whose parent field is in the whitelist. If the parent field is not in the whitelist, the tag is already inherited.


adriansr:
Yes, only 1.4. Thanks

PR #719, opened 2020-01-09
Improve array support in Go code generator

adriansr:
ECS has a bunch of fields expected to be arrays. For example, `tls.client.supported_ciphers`:

```yaml
    - name: client.supported_ciphers
      type: keyword
      level: extended
      description: Array of ciphers offered by the client during the client hello.
      example: ["TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "..."]
```

However, the generated Go code is using a type of `string` instead of `[]string`:

https://github.com/elastic/ecs/blob/94e7376ed1e618bec7128e4b14f579a5c0111343/code/go/ecs/tls.go#L65-L66

Looking at the generator's code, it seems that it was updated to output a type of `[]string` for a particular field (`{process|parent}.args`), but there isn't generic support to output arrays:

https://github.com/elastic/ecs/blob/94e7376ed1e618bec7128e4b14f579a5c0111343/scripts/cmd/gocodegen/gocodegen.go#L272-L274

Instead of continuing patching the generator, this PR inspects the description of a field to tell if it should be output as an array.

Fields with a description that contain the term `array` in the first 10 characters of the description will be generated as an array datatype.

This allows for:
```yaml
description: Array of [...]
description: An array [...]
description: The array [...]
description: Sub-array [...]
```
but not:
```
description: Length of the array
```
This is a bit hacky but it seems the best choice given that libbeat doesn't expose the example key that could be used to check for an array neither allows to use custom keys as hints. (i.e. `is_array: true`).

### DIFF
--- a/code/go/ecs/dns.go
+++ b/code/go/ecs/dns.go
@@ -45,7 +45,7 @@ type Dns struct {
 
 	// Array of 2 letter DNS header flags.
 	// Expected values are: AA, TC, RD, RA, AD, CD, DO.
-	HeaderFlags string `ecs:"header_flags"`
+	HeaderFlags []string `ecs:"header_flags"`
 
 	// The DNS response code.
 	ResponseCode string `ecs:"response_code"`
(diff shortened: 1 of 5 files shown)


webmat:
Clearly marking which fields are arrays is something I hope to get done for ECS 1.5 🤞 

You should check out this PR https://github.com/elastic/ecs/pull/661 as well as current follow-up work in this branch https://github.com/webmat/ecs/commits/is-array-markers.

If you base your work on this instead, it'll be much less hacky :-)

As for reviewing the Go code, you should sync up with @andrewkroh

adriansr:
I can't believe I missed that PR 🤦‍♂ 
Closing as there is already an ongoing effort (less hacky than this proposal)

PR #722, opened 2020-01-13
Refactor schema reader to use intermediate schema structure

marshallmain:
This PR improves the ECS ability to nest reusable fields inside other reusable field sets. Currently there is at least one example of multiple reuse: "group" fields are reused in "user", and "user" is then reused by "client", "destination", "host", "server", and "source". However, the implementation in master depends on the "group" fields being copied to "user" before all of the "user" fields are copied to the final locations. This works because the "group" key comes before the "user" key when iterating over the keys in the schema dictionary, but this will not be the case for other places we might want to use multi-level nesting.

This PR adds an intermediate data structure representing the schema which contains references to reused field sets rather than immediately copying them. Using references removes the dependency on the order fieldsets are read in. The intermediate representation is then rendered into the same formats as before by recursively replacing the references with copies.

The output from the refactored version diverges from the original in a few ways. Most notably original_fieldset now refers to the closest reusable fieldset (previously client.user.group.domain would have original_fieldset: user instead of original_fieldset: group). original_fieldset is also populated in every reusable fieldset, even when the fieldset is not embedded. dashed_name now matches flat_name more closely.

Reopened version of #707 reapplying changes on top of master

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -34,6 +34,7 @@ Thanks, you're awesome :-) -->
 #### Improvements
 
 * ECS scripts now use Python 3.6+. #674
+* schema_reader.py now reliably supports chaining reusable fieldsets together. #722
 
 #### Deprecated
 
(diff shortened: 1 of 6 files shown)


webmat:
Once we're done with this PR we should revisit https://github.com/elastic/ecs/pull/698

marshallmain:
Thanks for the review @webmat ! I'm planning to add the tests and changelog tomorrow so we can get this merged.

PR #725, opened 2020-01-15
Endpoint alert schema

marshallmain:
Structure overview:
- Most of the PR diff is autogenerated changes
- Real (hand made) changes to review are in `*.yml `files.  `custom_schemas/custom_*.yml` files **contain fields that I do not think belong in ECS**.  `schemas/*.yml` files contain changes that I think **DO belong in ECS**.  Whether or not a field belongs in ECS is of course open to discussion and one of the purposes of this PR.
- The new fields (both custom and ECS) correspond to fields we use currently in endgame-data-schema.  https://docs.google.com/spreadsheets/d/1upkqdKXqn5D8NlS6HY7UmoovNhB_13ZedV-UxGmOA7E/edit#gid=0 provides a mapping between the field name in endgame-data-schema to the new name in ECS.
- It can be hard to follow the nesting structure throughout the different files.  I find it helpful to look at the final generated elasticsearch mapping at ecs/generated/elasticsearch/7/template.json since it lists all the fields in one file and expands out the nested structure.


schemas/parent.yml
```diff
@@ -0,0 +1,23 @@
...
+  group: 2
+  short: These fields contain information about a parent.
+  description: >
+    These fields contain information about a parent.
+
+    These fields provide more context about the parent process
+    and thread that are related to the data in the document.
+  type: group
+  fields:
+    - name: process
```
rw-access: Is this a breaking change? did `process.parent` -> `parent.process`?

schemas/process.yml
```diff
@@ -23,6 +9,11 @@
     These fields can help you correlate metrics information with a process id/name
     from a log message.  The `process.pid` often stays in the metric itself and is
     copied to the global field for correlation.
+  reusable:
+    top_level: true
+    expected:
+      - parent
```
rw-access: would this mean process.parent.process?

schemas/parent.yml
```diff
@@ -0,0 +1,23 @@
...
+      level: extended
+      type: object
+      description: >
+        Process.
+
+    - name: thread
```
rw-access: I've never seen anything or any use cases for referring to a "parent thread". I think we should be fine sticking with our current process.parent paradigm

schemas/process.yml
marshallmain: No, we would have process and parent.process.  Moving parent to its own top-level field makes it simpler to re-use the process fields with the existing ECS generators.  The way the generators currently work, if we reused the process fields under process.parent then we would get process.parent.process.*

schemas/parent.yml
marshallmain: Yes

rw-access: This doesn't seem like a necessary breaking change. Outside of potentially simplifying some of this code base, this breaks a ton of detections for no added value, so I think it'd make more sense for use to figure out a more flexible way to keep the `process.parent` field set.

custom_schemas/custom_process.yml
```diff
@@ -0,0 +1,523 @@
...
+      - target
+  type: group
+  fields:
+    - name: unique_pid
+      level: custom
+      type: keyword
+      description: >
+        Unique process id.
+
+    - name: is_sensor
```
charlie-pichette: Consider changing to `is_endpoint`.  Sensor is an SMP term not an Elastic Endpoint term.

charlie-pichette: Consider changing to `is_endpoint`. Sensor is an SMP term not an Elastic Endpoint term.

docs/field-details.asciidoc
```diff
@@ -3521,288 +3585,2432 @@ example: `/home/alice`
...
+
+type: keyword
+
+
+
+| custom
+
+// ===============================================================
+
+| process.is_sensor
```
charlie-pichette: Consider changing to `is_endpoint`. Sensor is an SMP term not an Elastic Endpoint term.

generated/beats/fields.ecs.yml
```diff
@@ -2505,877 +3320,4822 @@
...
-    - name: uptime
-      level: extended
-      type: long
-      description: Seconds the process has been up.
-      example: 1325
-    - name: working_directory
-      level: extended
+      description: TODO
+      default_field: false
+    - name: process.is_sensor
```
charlie-pichette: Consider changing to `is_endpoint`. Sensor is an SMP term not an Elastic Endpoint term.

generated/beats/fields.ecs.yml
```diff
@@ -2505,877 +3320,4822 @@
...
+      type: boolean
+      description: TODO
+      default_field: false
+    - name: imphash
+      level: custom
+      type: keyword
+      ignore_above: 1024
+      description: TODO
+      default_field: false
+    - name: is_sensor
```
charlie-pichette: Consider changing to `is_endpoint`. Sensor is an SMP term not an Elastic Endpoint term.

generated/beats/fields.ecs.yml
```diff
@@ -2505,877 +3320,4822 @@
...
+    - name: module_info.mapped_size
+      level: custom
+      type: long
+      description: TODO
+      default_field: false
+    - name: module_info.path
+      level: custom
+      type: keyword
+      ignore_above: 1024
+
```
charlie-pichette: Consider changing to `is_endpoint`. Sensor is an SMP term not an Elastic Endpoint term.

custom_schemas/custom_process.yml
marshallmain: Good point, will do.

custom_schemas/custom_authenticode.yml
```diff
@@ -0,0 +1,80 @@
...
+      - module_info
+      - process.pe_info
+      - process
+  type: group
+  fields:
+    - name: cert_signer.issuer_name
+      level: custom
+      type: keyword
+      description: >
+        TODO
```
james-elastic: @marshallmain there's a lot of TODOs can you add descriptions in too?

marshallmain:
@andrewstucki I'm well aware of this.  If I make a PR to the endpoint-app-team repo then the diff won't highlight changes between the current version of ECS and the proposed version.  I'm not suggesting that we add the custom_schemas to the ECS repo, hence the "in progress" tag.

andrewstucki:
@marshallmain so rather than just dump a host of fields all at once to introduce into core, I'd take and create issues for each field set that you're thinking about introducing/changing. This sort of muddles things up and makes it hard to give constructive feedback and run by the rest of the ECS contributors.

For example: if the idea is to introduce a top-level field set called `parent` that changes the hierarchical ordering so you can do things like `process.parent` to `parent.process` I'd open up a PR specifically for that, or at least a discussion issue.

marshallmain:
@andrewstucki I'm not sure I follow the last paragraph.  The point of the custom pieces is that theyre separate from ECS so they don't introduce any breaking changes but they are inserted alongside the core and extended fields.  I noted in the original comment that the custom_*.yml files don't belong in ECS.  I've designed this PR so that reviewers can, as easily as possible, see the full extent of the fields in an alert and provide feedback on which fields do or do not belong in ECS which is why the custom schemas are included here.

Attempting to split the PR up into little digestible pieces will muddle the high level alert view and hide the full list of fields in an alert.  If we go down that path, reviewers won't see the custom schemas and will only provide input on whether or not the fields I propose to include in ECS are included.  I'd prefer for reviewers to also provide input on the whether or not the rest of the alert fields belong in ECS.

As for the process.parent vs parent.process issue, I'm happy to stick with the current hierarchy and submit a separate PR with enhancements to the generator script to support the existing schema.

schemas/parent.yml
james-elastic: @marshallmain can you explain your thought process for making a breaking change and why you think it's necessary?

james-elastic:
@marshallmain there is a lot of files here with a lot of code 34 files changed with +40,267 −5,816 lines of code. This is a large pr which makes it hard to get a complete picture of what you're trying to do. Can you break it down in separate PRs or make an outline of the files into digestible pieces and attach it here with the individual files?

james-elastic:
Separate commits might make it easier to digest as well

madirey:
Until we get more product guidance, let's just stay consistent with SMP as per searchable fields... i.e. if it was searchable in SMP, then let's make it searchable here. We can always add new fields later. See https://github.com/elastic/endgame-packrat/blob/dev/pkg/elasticsearch/processes.json#L68 for example.

Issue #666, opened 2019-12-03
New top-level object: search


joshdevins:
I'm working on building search event capture and metrics calculation into the Stack (prototype only so far) and want to raise the possibility of adding to the top level of ECS a `search` object. So far I am ingesting two event types into a single index: `query`, `click`. I extended ECS with the following fields and from this, I am able to use [transforms](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/transforms.html) to calculate some basic per-query search metrics: number of clicks, max reciprocal rank, mean reciprocal rank, sum reciprocal rank (used for later aggregations and metrics) time to first click, time to last click. 

I'm not ready to suggest concrete extension and fields but on @MikePaquette 's suggestion, I wanted to get early feedback on the general idea and source input on how to possibly structure the schema.

Additional open topics so far:
 - What should the schema of the transform which adds the metrics be.
   - Should it be ECS as well?
   - How would we support adding new or user-defined metrics then?
 - We can simplify and remove some fields (like rank from click) if we put more responsibility on the transform to calculate fields
 - Need to support pagination fields (which page of the results were clicked)
 - Need to (ideally) support other event types like "business objectives" that resulted from the search results and clicks. Things further down the funnel and value-chain such as, item added to shopping cart, item purchased, track played, user followed, etc.

Prototype schema additions:

- search
  - query
    - id: generated either by the application layer or maybe Elasticsearch
    - value: the query string itself - only included when the event is a `query`
  - results
    - size: size of the result set displayed
    - ids: document IDs returned/displayed
  - click: only included when the event is a `click`
    - result
      - id: document ID of the clicked document
      - rank: rank in the result set of the clicked document
  - ab: A/B testing properties -- these are mostly for demo purposes but could be here as well
    - experiment: name of the experiment
    - variant: name of the variant of the experiment that the user was exposed to

```json
      "search": {
        "properties": {
          "query": {
            "properties": {
              "id": {
                "ignore_above": 1024,
                "type": "keyword"
              },
              "value": {
                "ignore_above": 1024,
                "type": "keyword"
              }
            }
          },
          "results": {
            "properties": {
              "size": {
                "type": "long"
              },
              "ids": {
                "ignore_above": 1024,
                "type": "keyword"
              }
            }
          },
          "click": {
            "properties": {
              "result": {
                "properties": {
                  "id": {
                    "ignore_above": 1024,
                    "type": "keyword"
                  },
                  "rank": {
                    "type": "long"
                  }
                }
[...]

PR #729, opened 2020-01-24
Adds a search field set to represent search events

joshdevins:
Search events (query, page, click) have a common set of fields that are
useful to collect to create metrics from. Search metrics can be
calculated with transforms based on this schema. We expect applications
to emit these events, not Elasticsearch itself (for now).

This change is part of a larger project to collect search events and
calculate metrics in a standard way.

Closes: https://github.com/elastic/ecs/issues/666


joshdevins:
I'm not sure where to make these changes, but I would also like to add `search` as a valid `event.category`. See inline description of the field set for more details.

joshdevins:
Ping @webmat — can you review/comment or suggest another reviewer?

webmat:
Hey @joshdevins, thanks for opening #666 and this PR. I love the idea of adding support for capturing search analytics in ECS.

I'll start with some high level comments, then suggest a few next steps.

## High level comments

Actually, let's start at the highest level possible. How about a bit of philosophy 🙃 When you say "The alternative is to not use ECS at all"; I'd say the design of ECS is specifically to allow you to use the parts of ECS that already make sense, and doing anything else that's not supported yet via custom fields.

As an example, if using `service.*` and/or `host.*` makes sense for your use case, but `search.*` isn't defined in ECS yet, "using ECS" looks like this:
1. populating `host.*` and `service.*` as ECS describes
2. capturing all details of your search analytics in custom fields

Of course when custom fields capture information that are broadly applicable elsewhere than your project, you're more than welcome to propose adding them to ECS. As you're doing here :+1:

When I say "capturing in custom fields", of course there's a risk of conflicting with a future version of ECS that may add the same fields. The next release of ECS lays out a few approaches to adding custom fields while reducing the risk of future conflicts. You can check it out here: [Custom Fields in ECS](https://www.elastic.co/guide/en/ecs/master/ecs-custom-fields-in-ecs.html).

So concretely, you could apply some of the thoughts in there and have your search analytics at `projectname.search.*` for now. This way when we settle on the approach to use in ECS, you'll have to migrate your fields; but you'll be able to do so with "legacy" fields and ECS fields at the same time in your events, which will make the transition smoother.

If your project follows the above recommendations, you're free to proceed right away with implementation. And we're free to take the time to discuss the right way to add this to ECS, without rushing.

## Adding this to ECS

Which takes us to how I would see adding search analytics in ECS :-)

* I like the `search.*` field set. I think it's a fundamental type of activity we can observe and analyze. The field set makes sense to me.
	* For now I wouldn't define an event category/type for it (or for clicks) just yet. Let's wait until we've fleshed this out and used it more.
	* If you need a way to filter for these events, you can use a mix of `event.module`, `event.dataset`, and `event.action` (e.g. "search.query", "search.results"). None of these fields have prescribed values, which frees you up to categorize your events with the content of these fields.
* Clicks are also a fundamental type of activity we may want to analyze. As such, I wouldn't nest it under `search.*`. I'd rather think about it as a full-fledged field set.
	* Doing it this way will let us design a field set that's useable in many kinds of user behaviour analysis. Heat maps (click coordinates), search analysis, email engagement, etc.
* Continuing on behavioural analysis, I would also hold off on adding anything about AB testing. Just like clicks, this is a fundamental concept we may want to define independently.
* Whenever an event has a duration it should be captured at `event.duration`, so I wouldn't add `search.results.took` to ECS
* There's been a desire to add more support for metrics in ECS (#474). Either via many specific fields, or general recommendations on "how to do it right". But we haven't been able to dedicate resources to take this to the finish line. So I'd say metrics that are very specific to a field set are 👍. But we don't have a general recommendation on custom metrics yet.

## Next steps

My recommendations for a "search" field set:
* `search.query.id`
* `search.query.value`
* `search.results.page`: notice I'm suggesting under `.results`, was there a reason for it to be under `.query` initially?
* `search.results.size`
* `search.results.total`
* `search.results.ids`
* Please add a mention of tracking search durations at `event.duration` in the field set description
* Just like for field sets, fields also support a `short` and a `description`. Any time the field description gets a bit long, please add a `short` as well. It's used in places like the CSV, may be used in tooltips & so on.  Check out everything that's supported in field definitions at [schemas/README.md](https://github.com/elastic/ecs/tree/master/schemas/README.md)
* We recently added a way to mark fields that are expected to be arrays. Please add the following to the `.results.ids` field:
```
normalize:
  - array
```

I do see a need for a "click" field set eventually. But I think I'd wait until we have time to look at this kind of analytics in general in the context of ECS, to formulate a clearer plan. So my recommendation for clicks is custom fields all the way. They will inform our thinking, when we get around to that in ECS.

## Conclusion

Please feel free to respond to any of my suggestions above. You have more context on the need than I do.

So clarifications or further discussion are welcome :-)

joshdevins:
> High level comments

> Actually, let's start at the highest level possible. How about a bit of philosophy

Always appreciated! I'm new to the ECS landscape.

> When you say "The alternative is to not use ECS at all"; I'd say the design of ECS is specifically to allow you to use the parts of ECS that already make sense, and doing anything else that's not supported yet via custom fields.

This makes a lot of sense to me and I'm happy to keep things as custom fields until we have a clear way forward to inclusion in ECS (as with clicks).

The part I struggle a bit with is understanding how best to advise or implement the mapping then. Do solutions provide a mapping file that is based on ECS but includes the custom fields they want? Or do they rely on ECS mappings and dynamic mappings for the custom fields?

> Adding this to ECS

> For now I wouldn't define an event category/type for it

That's fair, it was the first place I saw that looked like it made sense. I will probably try using `event.action` as it seems to fit the best (I'll validate this works in my test cases, but it should be fine).

> I would also hold off on adding anything about AB testing.

Agreed. I had included it in the issue description as an example but removed it when I made the PR. I have been putting these values into `labels` for now since they are just key/value `keyword` types.

> Next steps

These sound good — I'll make the recommended changes/additions, including removing `click` fields from this PR 👍🏼

> `search.results.page`: notice I'm suggesting under `.results`, was there a reason for it to be under `.query` initially?

The page is a parameter of the query and not the response, so I thought it made logical sense to group it with those fields. For example, you would "ask for" page 1 or 2 of a search results page by setting the `limit` and `offset` in the Query DSL. The response is then the results for that requested page.

I'm reviewing the metrics I want to provide based on `page` and I'm thinking that I might actually drop this completely for now. We can use a custom field (😉) for now and add it later when/if it becomes more standard.

> I do see a need for a "click" field set eventually. But I think I'd wait until we have time to look at this kind of analytics in general in the context of ECS...

This makes sense to me, and I want to add some context that maybe helps in this PR discussion or for future discussions about clicks/behavioural events. In the case of search metrics, the proposed `search.*` (without `click`) is only useful up to a point. You can calculate some very basic metrics, but nothing that really tells you anything useful about search relevance. This is fine, and we will collect clicks with the help of a custom field set for now, but the trick here is that the two need to work in concert, in order to generate useful metrics. When a click event ocurrs and is captured, it needs to have a `search.query.id` in it, so that we can group the clicks together with the original query event. From this, we calculate search relevance metrics. There's nothing in ECS that prevents us from having clicks using custom fields, but I'm not sure how well this will sit beside other types of behavioural events, if this generalizes in some way. This is the main rationale to put `click` fields also under `search` — they kind of don't mean much on their own 😄.

joshdevins:
I'm just looking at moving `click` fields out and the other thing that strikes me is that all the fields of `click` are specific to search: `result.id`, `result.rank`. So again, it seems to me that it's so specific to search metrics, I don't know if it generalizes and should be it's own top-level field set.

joshdevins:
I also recognize that some of the limits on `keyword` fields will have to go. We don't want to skip storing events if they happen to have long query values, for example. I'll be updating that as well. Happy to hear your thoughts on that change.

joshdevins:
@webmat Have a look again. It's much smaller diff this time since I've removed all the `click` subset for now. Have a look at the recent comments as well if you could.

joshdevins:
@MikePaquette thanks for having a look, hoping to get this in for 1.5.0 release.

Re singular vs plural: I'm pretty indifferent and am happy to comply with the norms of ECS.

> Do you expect to commonly have information about the source of the search? For example, `user.id` does not need to be nested under `source.*`, and could just be populated as top level `user.id`.

I would expect that most apps/sites with logged-in or anonymous users will have that as part of the search context, and it's possibly useful to log. It's really up to the business to decide if it's useful to log user ID's — there is no prerequisite to doing so. I can imagine also just logging a hash so you can do simple metrics like calculating unique search users, for example. There is also no prerequisite to where that information is logged in the schema, so I can also elide it from the comments and let people decide. I liked it in `source` since it sits beside other user context like geo.

> Where would the geo information come from?

Again there is no prerequisite here but it's very common to capture user context information in a search query. The geo information usually comes from geo-IP services that are in a layer above search, or can be from a user profile. Geo information can be important to understand search performance across a range of locations. For example, do my users in North America (and English speaking?) have a better experience than users in CJK language countries? More granular geo information (cities) might be less useful of course. I don't want to put any restrictions on what people log, but being able to view search performance from various angles is what I'm after, and geo is one of those.

> Are we agreed upon the datatype for search.query.value being keyword?

What alternative(s) do you imagine? It's a `keyword` right now because it needs to be logged as-is and un-analyzed so the user/business can decide how to normalize it for metrics (a stage later). There's no need at query logging time to normalize the query strings and they don't need to be searched over at the event-logging stage. In theory, someone could also log it separately as a custom `text` field if they had use-cases for that.

I'm confident that we should include query value but it's also possible that a search solution has no query string or multiple even. I don't know what percentage though, but I'd like to include it to enable things like per-query-string metrics which is quite common.

joshdevins:
@MikePaquette These are seem like pretty small changes that can be made very easily, so I'm happy to jump on Slack/Zoom to just resolve these. I have two teams waiting on this to move forward with other work.

schemas/search.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: search
+  title: Search
+  group: 2
+  short: Fields to describe details of a search event.
+  description: >
+    The Search fields describe information about a search request event: query or pagination. These fields should be used with the corresponding `event.action` set with the corresponding event action: `['search.query', 'search.page']`. Other fields that should be used with this field set include: `event.duration` to describe the duration of a search request, `event.created` to record the event's timestamp and optionally `source` fields to record context information such as `user.id` or `geo`.
```
webmat: In ECS, the content of `event.action` is never mandated. It's meant to be defined by the source of the event. So the guidance you're providing here belongs in your project, not in ECS.

What we need to think about for `event.action`, is that other projects unrelated to yours may want to leverage the `search` field set, but use slightly different event actions that are more tailored to their use case.

schemas/search.yml
```diff
@@ -0,0 +1,53 @@
...
+        An opaque query identifier. This identifier needs to be unique to a user query, and all subsequent events (pagination, clicks) need to have the same query identifier.
+      example: "2dc15175-de0d-44db-86d8-8a99f41b7a11"
+
+    - name: query.value
+      level: extended
+      type: keyword
+      ignore_above: 8191
+      short: The query string being searched on.
+      description: >
+        The query string being search on. This field is not analyzed and should not be pre-processed in any way in the event (e.g. normalization list lowercasing). This is useful for search use-cases that use a one-box style search interface. Other interfaces will have to rely on additional custom fields or labels to represent things like filters applied, extra parameters, user context, etc.
```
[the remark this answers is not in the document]
webmat: I'm curious why you wouldn't want the field to be analyzed. I totally get that the source field in the document should be the unmodified search query. Someone investigating issues of their search solution will want this field untouched. 👍 

But also having an analyzed field in the Elasticsearch mapping would be great as well, to let users investigate a subset of their search events.

So if you'd like to add an analyzed multi-field, you can do so with:

```YAML
      type: keyword
      multi_fields:
      - type: text
        name: text
```

You'll notice that ECS follows the reverse convention vs Elasticsearch' dynamic mappings: the `keyword` field is always the canonical one, and the analyzed field is the nested one.

schemas/search.yml
webmat: Love the guidance of calling out other general ECS fields that should be used in conjunction with the `search` field set 👍

joshdevins:
> One of the ways users can build their templates based on ECS is to leverage the tooling in the repo.

This tooling looks great. Can we advise use of this or not yet?

schemas/search.yml
joshdevins: I'm going to change that wording a bit then to be more suggestive rather than prescriptive.

joshdevins:
> Supporting 8K long values in there for this purpose seems excessive, no?
Yeah probably. I will set it to...4k?

schemas/search.yml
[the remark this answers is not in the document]
joshdevins: This is typically done after an aggregation on something like query string or query ID. You wouldn't tend to investigate search queries in their raw form but really only after some aggregation. The kind of normalization/analysis that you would do it also pretty custom to the business, I would argue. The `standard` analyzer is probably fine for most people of course, but many places will have custom stop words, etc. that they may want to also use to normalize their queries before analysis.

I'm not opposed to adding multi-fields here, but I'm not convinced it's necessary. I'd opt to hold off and add it later if there is a larger observed need. As you say, someone can always index the query in a custom `text` field.

schemas/search.yml
```diff
@@ -0,0 +1,81 @@
+---
+- name: search
+  title: Search
+  group: 2
+  short: Fields to describe details of a search event.
+  description: >
+    The Search fields describe information about a search request event: query
+    or pagination. The fields that should be used with this field set include:
+    `event.action` to describe the search action (`search.query`,
```
webmat: Let's make it a bit clearer that it's not mandatory here, by simply saying that these are examples

```suggestion
    `event.action` to describe the search action (e.g. `search.query`,
```

schemas/search.yml
```diff
@@ -0,0 +1,81 @@
...
+    `search.page`, etc.), `event.duration` to describe the duration of a search
+    request, `event.created` to record the event's timestamp and optionally
```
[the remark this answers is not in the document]
joshdevins: I'll change these references. `@timestamp` is what I want.

vbohata:
I think there is one field missing here: search.query.offset which indicates the "record from" number which is not the same as the page.

vbohata:
One another note about search.query.value. We have a use case where we need to do aggregations per search.query.value during analysing one of our app. I do not know if you plan to make it searchable, but in our case we need it if possible. If not there should be .text alternative if the value is too long.

joshdevins:
@vbohata Unfortunately it was decided to remove these fields from ECS before the 1.5 release (https://github.com/elastic/ecs/pull/812). We're still talking about bringing them back, but nothing concrete.

PR #731, opened 2020-01-24
Add PE field set under file and process

### LINKED ISSUE #676: PE header metadata
Another thing that Sysmon and Endgame both collect for process and DLL (#675) events is PE metadata. 
image

There may be an OS agnostic way to represent this, but I think this has value even if it's Windows only. This field set could be nested within `process` or `dll`/`lib`/`module` and some of the already enumerated fields by Sysmon (see an [example event](https://github.com/hunters-forge/OSSEM/blob/master/data_dictionaries/windows/sysmon/event-1.md#event-xml) in OSSEM]) are below:

- original_file_name: `wuauclt.exe`
- file_version: `10.0.17134.1 (WinBuild.160101.0800)`
- description: `Windows Update`
- product: `Microsoft® Windows® Operating System`
- company: `Microsoft Corporation`

There's also authenticode information for signed binaries. That field set could be nested here, but we first need to solve that issue (#681)

- signer: `Microsoft Corporation`
- status: `trusted` / `untrusted` / `no_signature`, etc. We might want to enumerate some of these values ahead of time

rw-access:
Closes #676 

Adds a PE field set under file and process.


schemas/pe.yml
```diff
@@ -0,0 +1,50 @@
+---
+- name: pe
+  title: PE Header
```
webmat: Please write it in full here. Perhaps "Portable Executable (PE) Header"

schemas/pe.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: original_file_name
+      format: string
+      level: extended
+      type: keyword
+      description: Internal name of the file, provided at compile-time.
+      example: MSPAINT.EXE
+
+
+    - name: file_version
```
webmat: If there's no other version attribute in PE, I think it would make sense to simply call this `pe.version` instead of `pe.file_version`, WDYT?

schemas/pe.yml
```diff
@@ -0,0 +1,50 @@
...
+      level: extended
+      type: keyword
+      short: Process name.
+      description: Internal version of the file, provided at compile-time.
+      example: 6.3.9600.17415
+
+
+    - name: description
+      format: string
```
webmat: This is not needed, please remove all instances of `format:string`.

The places you've seen this are on numeric fields where we don't want thousands separators (e.g. PID). It doesn't do much, it only passes this along to Beats which uses it when it generates Kibana index templates that set this field format attribute. This isn't something that Elasticsearch is even aware of.

webmat:
I think it's time to open a can of worms, here. Let's discuss the question, and see if we think this is worth looking into or not, at this time.

Actually it's two separate cans :-)

1. I've long been thinking that we should have nested `file` under `process`. Perhaps most times we'd just end up populating `process.file.path` -- the equivalent of our current `process.executable`, and that's fine. However, any time we want to capture additional information about the file (owner, size, original_file_name, etc.) we're slowly duplicating fields from `file` to `process`. I think we should instead bite the bullet and nest `file` at `process.file`. Whether or not we deprecate `process.executable` for removal in ECS 2.0 is another matter.
2. Second can of worms is that each one of these fields makes total sense on their own, directly on `file`. Windows could provide it via the PE headers, but they are all fundamental concepts that can totally be gathered in different ways on other platforms. Perhaps I would adjust the names this way, if we add them directly under `file.`:
  * "pe.company" moved to "file.company"
  * "pe.description" moved to "file.description"
  * "pe.file_version" renamed to "file.version"
  * "pe. original_file_name" renamed to "file.original_name"
  * "pe.product" moved to "file.product" - this one I'm not sure whether to rename. Perhaps I would align with "os.platform" in spirit, and call it "file.platform", as the platform this file is destined to run on?

If we decide to tackle these two cans of worms together, then we get a simplification. Add PE headers in `file.` directly, and nest file under `process.` which gives us all of the PE attributes, but also any other we file attributes we may need.

@rw-access WDYT?

rw-access:
1- I think we should hold off on `process.file.*` for now. More on this in my meta comment below.

2 - I don't think it makes sense to put all of these as top-level `file` fields, because most only make sense for a narrow subset of file events. Text files don't have "company" or "description" or "version". These concepts only exist within the file _header_, so they don't necessarily pertain to the metadata surrounding a file event (size, location, acting process, etc). The concept of `file_version` means one thing to PE/COFF, and something entirely different to sharepoint or VCS. A top level field called version is way too subject to interpretation, and I would like to avoid that.

Removing the `pe` fieldset and nestingintroduces significant ambiguities for hypothetical use cases. I would rather stick with `pe` for our well known SIEM uses cases. 

(meta) Honestly, I'm fairly resistant to making the `process.file` major sweeping changes in ECS now. I've been down that road before with https://car.mitre.org/data_model, and you end up recreating CyBox, which was such a perfect model that it was no longer usable. Instead, we opted for flat and took a pragmatic approach and it helped significantly. I think ECS needs to gain more stability first in the eyes of users, and I would rather accumulate the tech debt so that we can move forward with SIEM, and then for ECS 2.0 we can revisit  these things with our lessons learned, and only make breaking changes once.

schemas/pe.yml
rw-access: there are other version attributes within PE headers, but are typically not collected and have less utility. this is a very narrow subset of all of the fields within PE headers, but does include the endgame and sysmon fields that are usable for detection logic. the fields I enumerated more closely resemble the original names, and if we rename we run the risk of causing more confusion

schemas/pe.yml
rw-access: what's the impact on the sort order? I remember running into this problem in another place

webmat: The reason I was asking this isn't for the sorting, but more for clarity. But looking again in both listing in the docs (see page body and sidebar [here](http://ecs_731.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-field-reference.html)), I think I'm good with keeping the title like it is now, and having it spelled out in the description.

So never mind this request.

To answer your question, the sorting is done on the short name (here "pe"). This is done implicitly by Python dicts, since it's used as the key name, in the in memory and intermediate YAML files. We could adjust the asciidoc generator to sort by full name instead, though.

schemas/pe.yml
webmat: Gotcha, let's keep it this way, then.

webmat:
Alright, let's close can of worms no 1 as well, on `process.file`. If we need to track more details about the file behind a process, the guidance could be to populate both field sets at the top level in the same event, instead.

```JSON
{
  "process": { "name": "ruby", "executable": "/usr/bin/ruby" },
  "file": { "path": "/usr/bin/ruby", "owner": "root", "mode": 0755 ... }
}
```

The clarifications on current field names convinced me as well. Let's proceed with the field names as they are. Descriptions look good as well :-)

The only thing we need to do here, is merge master, re-generate the files, and we'll be good to merge. Let's merge this PR first, and adjust https://github.com/elastic/ecs/pull/679 after.

webmat:
> The only thing we need to do here, is merge master, re-generate the files, and we'll be good to merge. Let's merge this PR first, and adjust #679 after.

Sorry, on top of the above, please don't forget to remove all of the unneeded `format:` attributes as well :-)

rw-access:
Thanks for the comments, and good idea for `process.file.attributes`. I hadn't thought about that, but it could be really powerful. I don't think Sysmon or Endpoint collect this yet IIRC, but this could detect a lot of interesting detection use cases.

If DLL merges, first then I'll add the nesting for `dll.pe` here. Otherwise, I'll just flip depends on which one is ready first.

PR #689, opened 2019-12-10
Added interface fields

dainperkins:
Added basic interface fields

PR #732, opened 2020-01-24
Incorporating IPFIX/Netflow fields into existing ECS field sets

jdburche:
Added 2 additional fields to both the destination and source field sets to capture IP network prefixes/masks and text descriptions of common network ports. Added 11 additional fields to the network field set to capture interface, VLAN (Virtual Local Area Network), ICMP, IP, and TCP information as well as bit rates and packet rates.

### DIFF
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -574,6 +574,12 @@
       description: 'IP address of the destination.
 
         Can be one or multiple IPv4 or IPv6 addresses.'
+    - name: ip.network_prefix
+      level: core
+      type: long
+      description: Count of the number of high-order bits that define an IPv4 or IPv6
+        network address in a destination network.
+      default_field: false
     - name: mac
       level: core
       type: keyword
@@ -603,6 +609,17 @@
       type: long
       format: string
       description: Port of the destination.
+    - name: port_name
+      level: core
+      type: keyword
+      ignore_above: 1024
+      description: 'Text description for commonly used ports, e.g. dns, http, https,
+        smtp.
+
+        The field value must be normalized to lowercase for querying. See the documentation
+        section "Implementing ECS".'
+      example: http
+      default_field: false
     - name: registered_domain
       level: extended
       type: keyword
@@ -1915,6 +1932,13 @@
         The field value must be normalized to lowercase for querying. See the documentation
         section "Implementing ECS".'
       example: aim
+    - name: bit_rate
+      level: core
+      type: float
+      description: Calculated by dividing the total number of bits transferred during
+        an arbitrary period of time by that period of time. By convention, this would
+        likely be presented in bps (bits per second).
+      default_field: false
     - name: bytes
       level: core
       type: long
@@ -1924,6 +1948,13 @@
         If `source.bytes` and `destination.bytes` are known, `network.bytes` is their
         sum.'
       example: 368
+    - name: class_of_service
+      level: core
+      type: long
+      description: In the case of IPv4 packets this would correspond to the TOS (Type
+        of Service) field. In the case of IPv6 packets this would correspond to the
+        Traffic Class field.
+      default_field: false
     - name: community_id
       level: extended
       type: keyword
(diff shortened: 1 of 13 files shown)


jdburche:
I have completed the CLA. Is there a way to rerun that check?

dainperkins:
@jdburche 

Thanks for the PR!  I haven't had a chance to go thru the whole thing, but wondering if you had seen the INTERFACE and VLAN fields proposed earlier - we didn't go with the rfc .1q names in the original PR, which I'm thinking are worth putting into the description if not considering renaming the field names. We had a couple of people internally who had run into more than 2 .1q tags in provider networks, etc, which led us towards discrete inner & outer vlan id & names, as well as a related.vlan.id/name for an array of any other additional encapsulations. 

Interface Fields: https://github.com/elastic/ecs/pull/689
VLAN Fields: https://github.com/elastic/ecs/pull/688

We are looking at allowing the interface and vlan fields to be nested to allow for the egress / ingress interface and vlan info for firewalls, routers/switches, load balancers, while maintaining the core set of fields under the same header.

t looked like there were a number of fields allowing for hex representations OR text descriptions, I think separating those out into 2 fields is likely a good thing.

Thanks
/d

jdburche:
@dainperkins Thanks for the comments and feedback!

I glanced over the discussion on VLAN identifiers, but had not noticed the INTERFACE PR, thanks for pointing out those. It looks like the VLAN field set and its fields as proposed in PR #688 could handle the IPFIX/Netflow use case.

Similarly, the Interface field set and its fields as proposed in PR #689 might be able to handle most of the IPFIX/Netflow use cases. In network flow data, the interface information usually maps back to an SNMP ifIndex field, so it would be good to have the option to list both an ID and Name if someone wanted to map more comprehensible names/descriptions to their less comprehensible index values.

What are your thoughts on how to annotate interface information where it is related to an observed network event as entering or exiting a particular interface at that observation point? Do you think "source.interface.id = network.egress_interface" and "destination.interface.id = network.ingress_interface"? I can see where this might confound other source/destination information, at least in the IPFIX/Netflow use case. I am interested in hearing more on your thoughts of how to approach this.

Regarding breaking out fields to allow for numeric and text representations, I agree that would be great. I initially overloaded those in an effort to keep the number of fields I was proposing down to a minimum 😃.

Thanks again for taking the time to review this PR.

dainperkins:
after talking with @webmat I've put in a new PR with the interface / vlan fields as well as updates to observer to include ingress/egress zones. https://github.com/elastic/ecs/pull/752

I'd very much like to get the flow specific stuff here worked thru and in as well, and will try to get some more thoughts down on things like tcp control bits vs syn/fin/rtxmt, naming conventions, etc. so we can push some good updates for ipfix/flow data

dainperkins:
ok, back to it!

- network.fragment_identification. (all set)
Thinking splitting these up would make sense (fairly easy to do on ingest, possibly difficult to handle at query time for analyzing)
- network.icmp.type (int)
- network.icmp.code (int)

TCP flags (I skipped echo & ecn)
- network.tcp.control_bits (integer representation, or just strip the 0x?)
- network.tcp.syn (boolean or an array network.tcp.flags [syn,ack, etc,.])
-   network.tcp.ack (boolean)
-   network.tcp.finish (boolean)
-   network.tcp.reset (boolean)
-   network.tcp.push (boolean)
-   network.tcp.urgent (boolean)
- network.tcp.window_size
- networl.tcp.sequence_number
- networl.tcp.acknowledgement_number
- network.ip.time_to_live (useful for ordering observers in deduplication perhaps?)

For COS, DSCP, & IP Precedence we need to cover ingress & egress in case of a change,  potentially making egress required only when theres a change 
- network.ingress.class_of_service
- network.ingress.ip_precedence
- network.ingress.dscp
- network.egress.class_of_service
- network.egress.ip_precedence
- network.egress.dscp

And then if we get into flexible netflow theres all sorts of really useful N/APM goodies (as well as the metrics I was previously questioning - I'm still not sure as a general "heres a netflow config to use with Elastic" standpoint I would be inclined to include observer aggregates, as opposed to just using visualizations to aggregate, but I'm unsure how relevant that is)
- network.jitter.mean
- network.jitter.max
- network.jitter.min
- network.delay (total network delay)
- network.server.delay 
- network.client.delay
- network.sample.[rate? pct?] (need to check what the typical usage is for sflow / sampled nf)

theres a whole lot there, but its pretty much standard as well, so it should be relatively easy to work thru.

dainperkins:
probably need to add vrfs as well...

generated/beats/fields.ecs.yml
```diff
@@ -574,6 +574,12 @@
       description: 'IP address of the destination.
 
         Can be one or multiple IPv4 or IPv6 addresses.'
+    - name: ip.network_prefix
```
vpiserchia, line 577: just to know if this type of definition is possible:

ip is already a field of type IP not a json object

jdburche: Splitting out the ICMP type and code sounds like a good idea, would make querying a bit easier.

For the TCP flags, an array of flags sounds like a good approach, again to make querying easier.

Perhaps the IP address should become ip.address?

jdburche:
> probably need to add vrfs as well...



> probably need to add vrfs as well...

Splitting out the ICMP type and code sounds like a good idea, would make querying a bit easier.

For the TCP flags, an array of flags sounds like a good approach, again to make querying easier.

dainperkins (commented):
Do you by any chance have e.g. a fnf / ipfix output you could share?  I remember being able to track retransmits, resets, etc. from netflow/ipfix in other products, but i can't for the life of me recall how that was provided (duplicate syns are pretty straightforward, but if they both come in the same netflow interval is the an indicator in the flow record?).

generated/beats/fields.ecs.yml
vpiserchia: honestly, "ip" fields exist all around since the first incarnation of the ECS. I would rather prefer to have a simple "network_prefix" field at the same level as the corresponding ip one

ebeahan:
A [stage 0 ECS RFC proposal](https://github.com/elastic/ecs/blob/master/rfcs/text/0013-network-headers.md) has been submitted to expand the network headers currently defined in ECS. 

It appears there's some overlap between that RFC and the [proposed changes](https://github.com/elastic/ecs/pull/732#issuecomment-608200356) here? This PR been inactive for awhile now, and it's also expanded from the original description quite a bit.

I'd like to centralize the discussion in the network headers RFC and will close this out for now. We can always re-open this PR if needed.

PR #733, opened 2020-01-24
Digital code signatures for process, file and dll events

### LINKED ISSUE #681: Code signature field set
We need a field set to express information about signatures, and in particular code signatures. Already, the word _signature_ has multiple meanings in security, so we'll need to pick a good word that's clear, and doesn't collide with "signature based detections".  Maybe `digsig` or `codesig` or `code_signature`?

For code signatures, I'm most familiar with signing Windows PE files, and lightly familiar with code signing for macOS. I think a new field set could belong in a few places, depending on what feels the most natural.

I'm leaning towards nesting it under `process.signature`, `lib.signature` (#675) and `file.signature` to be more generic. We could also nest it under [`tls`](https://www.elastic.co/guide/en/ecs/master/ecs-tls.html). Even if this happens later, we should at least choose fields to maximize compatibility.

Without diving into the internals of code signatures and trust relationships that I won't pretend to understand, here are some obvious high value fields that are common between different implementations:
- `subject_name`: "Microsoft Corporation" (I believe this was called "signer_name" within Endgame)
- `valid`: bool - maintains integrity, 
- `timestamp`: date - timestamp for when it was signed
- `trusted`: bool
- `expired`: bool
- `exists`: bool - check if the signature exists. useful to disambiguate between if a signature explicitly doesn't exist vs that information is not available

A few things that may be useful for other non-binary use cases:
- certificate authority information?

rw-access:
Closes #681 

I'm not 100% sold on anything yet. I think there may be room to break out this x.509 information into a separate field set. Ideally in a way that's still intuitive enough in naming. This information is pretty sparsely populated by most endpoint solutions. It's typically just signature status -- valid y/n, trusted y/n, etc. as well as the name of the signer (subject name).


webmat:
In order to review this, I'd like to see some data examples from the Mac and Windows platforms. 

I tried viewing some on both, and I'd like to confirm how they map with what you're proposing here.  Perhaps I'm way off track, please correct me if that's the case :-)

## Mac OS

checking the code signature for Safari.app:

```
$ codesign -dv --verbose=4 /Applications/Firefox.app
Executable=/Applications/Safari.app/Contents/MacOS/Safari
Identifier=com.apple.Safari
Format=app bundle with Mach-O thin (x86_64)
CodeDirectory v=20100 size=321 flags=0x2000(library-validation) hashes=3+5 location=embedded
VersionPlatform=1
VersionMin=658944
VersionSDK=658944
Hash type=sha256 size=32
CandidateCDHash sha256=96d97014be33f78e4b7ccec056f34512c21462d3
Hash choices=sha256
Page size=4096
CDHash=96d97014be33f78e4b7ccec056f34512c21462d3
Signature size=4547
Authority=Software Signing
Authority=Apple Code Signing Certification Authority
Authority=Apple Root CA
Info.plist entries=41
TeamIdentifier=not set
Sealed Resources version=2 rules=13 files=2227
Internal requirements count=1 size=64
```

In this case, the best candidate I see for `code_signature.subject_name` is one of the "Authority" entries. `.valid` and `.exists` are straightforward to populate as well.

Not sure about `.trusted` nor `.status`. Do you expect these to be available on the Mac?

## Windows

checking the code signature for the IE executable:

IE signature on Windows

And:

```
PS C:\Users\vagrant> Get-AuthenticodeSignature -FilePath "C:\Program Files\Internet Explorer\iexplore.exe"


    Directory: C:\Program Files\Internet Explorer


SignerCertificate                         Status                                 Path
-----------------                         ------                                 ----
108E2BA23632620C427C570B6D9DB51AC31387FE  Valid                                  iexplore.exe
```

In the Windows case, I don't see a signer name. Unless that would be "Copyright" (doesn't sound right to me)?

* `.valid` and `.exists` seem straightforward to populate just like the Mac
* How is `.trusted` populated?
* And I suppose "Status" from "Get-AuthenticodeSignature" would populate `.status`?
  * I get "UnknownError" if I check the signature on an unsigned file.

rw-access:
re: trusted/status -- those don't exist in the file, but have to be evaluated by the client. for example, the Elastic Endpoint calls several Windows APIs to verify that a signature is cryptographically correct, and then there are additional steps to validate the trust level of the certificate. we get a message as a result "trusted", "noSignature", "untrustedRoot", "badDigest", etc depending on the error message. we typically check for trusted or noSignature and signer. having `valid` and `exists` as their own fields simplifies a lot of the rule logic

rw-access:
For the PowerShell command there should be a lot more fields that aren't being shown.

That screenshot is actually PE data, not signer information. You'd have to click on the "Digital Signatures" tab. That would show this:
image


Sigcheck is the most analogous tool to codesign for windows. That output looks like this, which also mixes in some PE metadata:
image

webmat:
Thanks for the pointers. Pretty awesome, how I skipped right over the "Digital Signatures" tab, there 😂 🤦‍♂ 

@elasticmachine, run elasticsearch-ci/docss

webmat:
I just realized that Python interprets the YAML `true`, then outputs it as "True", which is not actually considered a valid boolean by Elasticsearch 🤦‍♂ 

Can you remove the examples for now? This is the only place in ECS where we give examples for boolean fields. So props for being thorough here, but we'll fix that small issue separately :-)

docs/field-details.asciidoc
```diff
@@ -489,6 +489,89 @@ example: `us-east-1`
...
+// ===============================================================
+
+| code_signature.exists
+| Boolean to capture if a signature is present.
+
+This should only populated if the signature was checked.
+
+type: keyword
+
+example: `True`
```
rw-access: we could also just do `json.dumps()` instead

rw-access: oh, that may break other examples

webmat:

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
...
+  title: Code Signature
+  group: 2
+  description: These fields contain information about binary code signatures.
+  type: group
+  reusable:
+    top_level: false
+    expected:
+      - file
+      - process
+      # - dll
```
webmat: Let's merge this PR before DLL as well, and add this reuse of `code_signature` via the DLL PR (#679) as well.

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: exists
+      format: core
+      level: extended
+      type: keyword
+      short: Boolean to capture if a signature is present.
+      description: >
+        Boolean to capture if a signature is present.
+
+        This should only populated if the signature was checked.
```
webmat: > This should only populated if the signature was checked. 

Why? Isn't this overloading the definition of `code_signature.exists`? Whether a signature exists is orthogonal to whether its validity has been assessed, IMO.

The ternary approach would make more sense on `code_signature.valid`, though:

* haven't checked validity: `code_signature.valid` is absent
* checked validity, good: `code_signature.valid=true`
* checked validity, bad: `code_signature.valid=false`

Or am I missing something?

If you agree, we should adjust the definitions of `valid` and `exists` accordingly.

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
+---
+- name: code_signature
+  title: Code Signature
+  group: 2
+  description: These fields contain information about binary code signatures.
```
webmat: What about a shorter short definition, and a more fleshed out full definition?

```suggestion
  description: These fields contain information about code signatures for various forms of executable code, such as executables, libraries or scripts.
  short: Information about binary code signatures.
```

And the above begs the question: will we be wanting to track signatures other than code signatures? If so, do we want to discuss whether this field set should be named `signature` instead? Happy to close that can of worms immediately, if you think it's an unlikely hypothetical ;-)

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
...
+    top_level: false
+    expected:
+      - file
+      - process
+      # - dll
+      # - driver
+  fields:
+
+    - name: subject_name
+      format: string
```
webmat: remove

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
...
+    - name: valid
+      level: core
+      type: boolean
+      description: Boolean to capture if the digital signature is verified against the binary content.
+      example: "true"
+
+
+    - name: exists
+      format: core
+      level: extended
```
webmat: Remove the `format:` attribute.

Now the question is: was the intent to make this field level core or extended? :-)

schemas/code_signature.yml
```diff
@@ -0,0 +1,59 @@
...
+      type: boolean
+      description: Stores the trust status of the certificate chain.
+      example: "true"
+
+
+    - name: status
+      format: string
+      level: extended
```
webmat: Like for the field `exists`, please remove the `format:` attribute. And let's clarify which fields are core and extended. Here's an approach I think would make sense:

Core: can be determined via direct observation
Extended: need to evaluate the certificate validity to determine

If we go with this heuristic, the result would be:

Field | Level | Notes
--|--|--
`subject_name` | core | 
`valid` | extended | 
`exists` | core | Provided you agree with my comment on the description of the field
`trusted` | extended | 
`status ` | extended |

schemas/code_signature.yml
rw-access: i thought about that, and the big downside I came up with is that the word "signature" is often used like "rule", so I didn't want to potentially clobber up a namespace that may be used.

I definitely want something shorter than `code_signature`, because `process.code_signature.valid` is lengthy

schemas/code_signature.yml
rw-access: i like your heuristic. resolved in 40aec6a4947136339c935855fa29c17107b09247

schemas/code_signature.yml
webmat: I think we should stick with `code_signature`, then. Are you good with that?

schemas/code_signature.yml
```diff
@@ -0,0 +1,55 @@
...
+
+
+    - name: exists
+      level: core
+      type: keyword
+      short: Boolean to capture if a signature is present.
+      description: >
+        Boolean to capture if a signature is present.
+
+        This should only populated if the signature was checked.
```
webmat:
```suggestion
        This should only be populated if the signature was checked.
```

schemas/code_signature.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: valid
+      level: extended
+      type: boolean
+      description: Boolean to capture if the digital signature is verified against the binary content.
+      example: "true"
+
+
+    - name: exists
+      level: core
+      type: keyword
```
webmat:
```suggestion
      type: boolean
```

:-)

webmat:
I spotted two more small details, which should be fixed. Once fixed, and if you're good with sticking with the name `code_signature` for this field set, I think we'll be ready to merge

rw-access:
I thought of another idea... Would it make sense to just include this within `pe` as `pe.signature`?

webmat:
@rw-access Well PE is named really Windows-specific. And code signatures are expected on Mac OS as well.

So I'd rather keep them separate.

PR #737, opened 2020-01-30
Allow generation of a subset of ECS and custom schema fields

marshallmain:
This PR extends prior work on the intermediate data format introduced in https://github.com/elastic/ecs/pull/722 and provides a way to specify a subset of the full set of ECS fields plus any fields defined by a custom schema and render only that subset to the generated files.  This would be particularly useful when multiple types of documents will be stored in the same index: one monolithic set of fields can be defined using the official ECS fields in addition to a custom schema that encompass every possible field for the documents, and the --subset parameter can be used to define which fields exist in each type of document.

An example subset document is
```---
host: "*"
base:
  "@timestamp": "*"
agent:
  id: "*"
  name: "*"
event: "*"
```
This subset includes every field in `host`, `@timestamp` from `base`, `id` and `name` from `agent`, and every field in `event`.  The format supports an arbitrary level of nesting as well.

Finally, multiple subsets can be provided at once and they will be automatically merged together to produce the set of fields which is the union of all the subsets.  This creates the exact mapping that is needed to support the set of documents.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -35,7 +35,7 @@ Thanks, you're awesome :-) -->
 
 * ECS scripts now use Python 3.6+. #674
 * schema_reader.py now reliably supports chaining reusable fieldsets together. #722
-* A subset of fields to render can be defined. #737
+* Allow the artifact generator to consider and output only a subset of fields. #737
 
 #### Deprecated
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -35,6 +35,7 @@ Thanks, you're awesome :-) -->
 
 * ECS scripts now use Python 3.6+. #674
 * schema_reader.py now reliably supports chaining reusable fieldsets together. #722
+* A subset of fields to render can be defined. #737
 
 #### Deprecated
 
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -10,6 +10,7 @@ from generators import beats
 from generators import asciidoc_fields
 from generators import ecs_helpers
 
+
 def fields_subset(subset, fields):
     retained_fields = {}
     for key, val in subset.items():
@@ -19,6 +20,7 @@ def fields_subset(subset, fields):
         elif val == '*':
             retained_fields[key] = fields[key]
     return retained_fields
+
 
 def main():
     args = argument_parser()
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -11,17 +11,6 @@ from generators import asciidoc_fields
 from generators import ecs_helpers
 
 
-def fields_subset(subset, fields):
-    retained_fields = {}
-    for key, val in subset.items():
-        if isinstance(val, dict):
-            retained_fields[key] = fields[key]
-            retained_fields[key]['fields'] = fields_subset(val, fields[key]['fields'])
-        elif val == '*':
-            retained_fields[key] = fields[key]
-    return retained_fields
-
-
 def main():
     args = argument_parser()
 
@@ -48,7 +37,7 @@ def main():
             with open(file) as f:
                 raw = yaml.safe_load(f.read())
                 ecs_helpers.recursive_merge_subset_dicts(subset, raw)
-        intermediate_fields = fields_subset(subset, intermediate_fields)
+        intermediate_fields = ecs_helpers.fields_subset(subset, intermediate_fields)
 
     (nested, flat) = schema_reader.generate_nested_flat(intermediate_fields)
     intermediate_files.generate(nested, flat)
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -18,7 +18,7 @@ def main():
 
     # Load the default schemas
     print('Loading default schemas')
-    (nested, flat) = schema_reader.load_schemas()
+    intermediate_fields = schema_reader.load_schemas()
 
     # Maybe load user specified directory of schemas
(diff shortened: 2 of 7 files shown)


scripts/generator.py
```diff
@@ -2,13 +2,23 @@
...
 import schema_reader
+import yaml
 from generators import intermediate_files
 from generators import csv_generator
 from generators import es_template
 from generators import beats
 from generators import asciidoc_fields
 from generators import ecs_helpers
 
+def fields_subset(subset, fields):
```
marshallmain: this probably belongs in ecs_helpers.py, planning to move before merging

scripts/schema_reader.py
```diff
@@ -92,6 +92,24 @@ def schema_fields_as_dictionary(schema):
         nested_schema[nested_levels[-1]]['field_details'] = field
 
 
+def merge_schema_fields(a, b):
+    for key in b:
+        if key not in a:
```
alexk307: what about duplicate keys?

jonathan-buttner:
I can definitely see the value in this so that you don't have to copy/paste a bunch of fields to describe a specific event (especially an alert since it's huge). I think there are a couple downsides though, the tooling doesn't support multiple nested fields so if you had a schema file like this:

```yaml
- name: endgame
  description: >
    The endgame.
  title: Endgame
  group: 2
  short: short endgame.
  fields:
    - name: test
      type: group
      level: extended
      description: >
        The DNS packet identifier assigned by the program that generated the
        query. The identifier is copied to the response.
      fields:
        - name: final
          type: keyword
          description: >
            blah description.
```
The generated documents won't have the `final` field in them. To get around that you'd have to `test` it's own schema and put it in its own file. We could address that using some of the recursive changes in my PR as a follow up though.

The generated files will have a bunch of fields on them not probably aren't necessary for translation into C++ classes but that's probably not that big of a deal.

marshallmain:
@jonathan-buttner The other way to get around that issue is add a field called `test.final`, similar to https://github.com/elastic/ecs/blob/master/schemas/process.yml#L37.  I would be in favor of adding support for the nested format in your example though since we saw that it's a very small change.

As for the generated files, if we want to generate files in the true nested format I think a good approach would be to implement another generator that renders the fields in the desired format.  Ideally any new renderers would take the intermediate format as an input so we can move towards consolidating the nested and flat formats as well.

marshallmain:
Thinking about this more the subset format should allow fields to be defined as either optional or required to make a better specification for clients that will use the mapping.  This will require modifying the subset format to expect "options" and "fields" keys in each field dictionary.  In the future support for more options could be added.

CHANGELOG.next.md
```diff
@@ -35,6 +35,7 @@ Thanks, you're awesome :-) -->
 
 * ECS scripts now use Python 3.6+. #674
 * schema_reader.py now reliably supports chaining reusable fieldsets together. #722
+* A subset of fields to render can be defined. #737
```
webmat:
```suggestion
* Allow the artifact generator to consider and output only a subset of fields. #737
```

scripts/generator.py
```diff
@@ -48,6 +55,8 @@ def argument_parser():
                         help='generate intermediary files only')
     parser.add_argument('--include', action='store',
                         help='include user specified directory of custom field definitions')
+    parser.add_argument('--subset', action='store',
+                        help='render a subset of the schema')
```
webmat: I noticed I hadn't documented the usage of `--include` yet, so I won't ask you to add documentation for `--subset`, we'll do that as a separate PR. I opened #746 for this.

But just to make sure I understand the usage correctly, is it correct to say that given my custom fields in `my-fields.yml` and my subset selectors at `my-subset.yml`, the command would be the following?

`scripts/generator.py --include my-fields.yml --subset my-subset.yml`

Also, this supports passing multiple subsets as well? I'm not 100% sure about the Python args parser, how would this look at the command line?

Is it a repeatable argument (`--subset my-subset.yml  --subset my-other-subset.yml`) or an argument that takes multiple values (`--subset my-subset.yml my-other-subset.yml`)?

marshallmain: For the combination of --include and --subset that's the right command to use.

Good point, multiple value support is an important part that I missed.  I just pushed another commit up that enables the second style you mentioned (`--subset my-subset.yml my-other-subset.yml`).  It can be combined with wildcards too, i.e. `--subset dir1/*.yml dir2/*.yml` would get all the subsets from both dir1 and dir2.

marshallmain:
@webmat The last thing I want to update before removing the in progress label and merging is the subset yml format - right now I think it's a bit too simple and inflexible.  I'm planning to add one level between each field and the field's children where we can potentially add options or other fields to describe expected usage later.  The use case I have in mind for this is marking whether a particular field is required for a doc type or simply optional.  I'd like to add the extra level now so that we can add features like that later without making breaking changes to the subset yml format.

scripts/generator.py
webmat: Nice. And for the record, the same is true of `--include` already

webmat:
Thanks @marshallmain ❤️ 

For the record, after the latest changes, the "subset" yaml file becomes

```YAML
host:
  fields: "*"
base:
  fields:
    "@timestamp":
      fields: "*"
agent:
  fields:
    id:
      fields: "*"
    name:
      fields: "*"
event:
  fields: "*"
```

This will be documented via #746

PR #740, opened 2020-02-04
Defined observer zones and interfaces

dainok:
Hi,

referring to network flows passing through a firewall, I think we should track ingress and egress security zones and interfaces.
I think we can have two different approaches:
* zones/interfaces are defined on the observer schema, and I think that's more correct
* zones/interfaces are defined on the client/server/source/destination schema
In the latter case, zones/interfaces are still references observer (i.e. firewall) objects.
Any thoughts?

### DIFF
--- a/code/go/ecs/client.go
+++ b/code/go/ecs/client.go
@@ -85,4 +85,7 @@ type Client struct {
 	// internet).
 	// Typically connections traversing load balancers, firewalls, or routers.
 	NatPort int64 `ecs:"nat.port"`
+
+	// Security zone facing the client.
+	Zone string `ecs:"zone"`
 }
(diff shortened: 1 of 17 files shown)


webmat:
Ping @dainperkins to align with #688 and #689

dainperkins:
I was thinking about just this thing over the last couple of days.  I'm going to add some color on the vlan and interface PRs to get the group on board, and then we can look at the observer implementation

webmat:
Also I forgot something important, when I answered earlier: thanks @dainok for opening this. You're welcome to join the discussion on the other PRs as well :-)

webmat:

dainok:
According to https://github.com/elastic/ecs/pull/688 I moved zone under client/source/destination/server.

dainperkins:
after talking with @webmat I've put a new PR with the basic interface & vlan info, as well as observer based zone info for ingress/egress traffic - https://github.com/elastic/ecs/pull/752

dainok:
Agree. Closing because incorporated under https://github.com/elastic/ecs/pull/752

Issue #672, opened 2019-12-04
Unique PID Fields and Generation


andrewstucki:
# Problem

PID re-use is common for both Windows and POSIX systems across:

1. Long periods of time
2. OS restarts
3. Lots of quickly spawned processes

Identifying a process uniquely across time by generating some sort of unique identifier with more specificity than standard system-level PIDs is a fairly common in auditing and security analysis. For example, Sysmon generates unique guids for processes by [roughly incorporating](https://gist.github.com/mattifestation/0102042160c9a60b2b847378c0ef70b4#file-sysmoneventguidparser-ps1-L45):

- the machine ID of the OS installation
- the process start time
- the access token corresponding to the logged in user's session

Clearly some of this is pretty specific to Windows internals and can't necessarily be re-used across POSIX systems.

The thought is to:

1. Introduce the concept of a `unique_pid` into ECS, and
2. Potentially introduce something like network ["community id"](https://github.com/corelight/community-id-spec) except for processes as a well-known standard for generating this value

If we decide that this is a useful construct to introduce into ECS, we have three potential options I can think of:

- Introducing the concept of `unique_pid` and leaving it implementation defined (skipping 2 above).
- Standardizing on the generation mechanism and leaning towards a "globally unique" identifier that incorporates information about the host the process is running on. This would allowsus to uniquely identify a process in a bucket of time series data that includes information from multiple hosts.
- Standardizing on the generation mechanism and leaning towards a "locally unique" identifier based solely off of the process metadata itself. This would allow us to uniquely identify a process in a bucket of time series data corresponding to a single host.

Here are some thoughts on each.

## Implementation Defined

We could punt on the idea of specifying a generation algorithm and let an ECS field be implementation defined. In this case, we'd allow for each data source (beats, endpoint, etc.) to ensure its own generation method is unique enough for its own use cases. Main problem with this is that by punting on making the generation mechanism standard, we lose the ability for applications to correlate events from multiple data sources.

## Globally Unique

While this potentially buys you the ability to easily correlate processes from disparate sources, the difficult issue is being able to incorporate a standard mechanism for truly uniquely identifying a host where a process is running. For example, in the case of the Sysmon guid generation, the host identifying material--the machine ID--is tied explicitly to the installation of the OS. This potentially becomes problematic when you're using any sort of VMs/VDIs/Containerized systems as you could have multiple VMs with the same installation information contained within them.

## Locally Unique

[...]

PR #747, opened 2020-02-13
Add entity_id for process and process.parent

andrewstucki:
This implements the unique process identification field discussed in https://github.com/elastic/ecs/issues/672. We settled on allowing ECS users to use their own implementation-specific values in this field in order to support multiple types of data shippers that may have different requirements around the identifier generation. That said, there's still room for future discussion on some sort of standard generation algorithm that would allow us to correlate processes across multiple sources.

The name corresponds to what beats are already filling in for their unique identifier.

### DIFF
--- a/schemas/process.yml
+++ b/schemas/process.yml
@@ -42,6 +42,31 @@
         Process id.
       example: 4242
 
+    - name: entity_id
+      level: extended
+      type: keyword
+      short: Unique identifier for the process.
+      description: >
+        Unique identifier for the process.
+
+        The implementation of this is specified by the data source, but some
+        examples of what could be used here are a process-generated UUID,
+        Sysmon Process GUIDs, or a hash of some uniquely identifying components
+        of a process.
+      example: c2c455d9f99375d
+
+    - name: parent.entity_id
+      level: extended
+      type: keyword
+      short: Unique identifier for the process.
+      description: >
+        Unique identifier for the process.
+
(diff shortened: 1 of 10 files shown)


schemas/process.yml
```diff
@@ -42,6 +42,31 @@
...
+      level: extended
+      type: keyword
+      short: Unique identifier for the process.
+      description: >
+        Unique identifier for the process.
+
+        The implementation of this is specified by the data source, but some
+        examples of what could be used here are a process-generated UUID,
+        Sysmon Process GUIDs, or a hash of some uniquely identifying components
+        of a process.
```
webmat: Love this. Should we add another sentence about the "why" as well? Example:

```suggestion
        of a process.

        This is a common practice, to mitigate PID reuse, and get a more unique identifier for processes over time, and across monitored hosts.
```

Not absolutely necessary to add: people who know will understand what this is. But it's a good opportunity to give good ideas to people who hadn't thought of that :-)

webmat:
And 👍 on keeping the Beats field name.

webmat:
Oof, second ☕️ hasn't kicked in yet, it seems.

Another request: please add a changelog entry :-)

schemas/process.yml
andrewstucki: Sounds good, added a variation of the suggested text.

webmat:
Let's wait for @rw-access to chime in as well. You can merge after that :-)

PR #748, opened 2020-02-13
Adding --out flag

jonathan-buttner:
This PR adds the `--out` switch to specify where the generated files and docs are to be written. If the directory does not exist it will be created.

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -65,7 +65,7 @@ def argument_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument('--intermediate-only', action='store_true',
                         help='generate intermediary files only')
-    parser.add_argument('--include', action='append',
+    parser.add_argument('--include', nargs='+',
                         help='include user specified directory of custom field definitions')
     parser.add_argument('--subset', nargs='+',
                         help='render a subset of the schema')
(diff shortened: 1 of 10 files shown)


scripts/generator.py
```diff
@@ -40,24 +40,36 @@ def main():
...
+    beats.generate(nested, ecs_version, out_dir)
+    asciidoc_fields.generate(nested, flat, ecs_version, docs_dir)
 
 
 def argument_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument('--intermediate-only', action='store_true',
                         help='generate intermediary files only')
-    parser.add_argument('--include', action='store',
+    parser.add_argument('--include', action='append',
```
jonathan-buttner: With the change for `append` here, if you pass in multiple `--include` flags it will only use the last one. This causes the values to be added to an array.

jonathan-buttner: Ah good point, didn't think of that. I don't think there should be any issue with switching it over to  `nargs`. I'll update it.

jonathan-buttner:
Example run:

```
python scripts/generator.py --include ../test1 ../test2 --out ../test_out
Running generator. ECS version 1.5.0-dev
Loading default schemas
Loading user defined schemas: ['../test1/agent.yml', '../test2/http.yml']
```

webmat:
Awesome, thank!

```
  --include INCLUDE [INCLUDE ...]
                        include user specified directory of custom field
                        definitions
  --subset SUBSET [SUBSET ...]
                        render a subset of the schema
```

❤️

PR #750, opened 2020-02-18
Support nested type at top level of fieldsets

marshallmain:
This PR adds support for specifying the type at the top level of a fieldset.  This is useful for cases where a fieldset needs to have the `nested` type, such as a `call_stack` fieldset that represents an array of stack frames.

### DIFF
--- a/schemas/README.md
+++ b/schemas/README.md
@@ -11,7 +11,7 @@ YAML with a twist: Flattened field names equivalent to nested. E.g. `foo.bar: va
 - group (required for now): TBD. Just set it to 2, for now ;-)
 - description (required): Description of the field set
 - fields (required): YAML array as described below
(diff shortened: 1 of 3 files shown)


schemas/README.md
```diff
@@ -11,7 +11,7 @@ YAML with a twist: Flattened field names equivalent to nested. E.g. `foo.bar: va
 - group (required for now): TBD. Just set it to 2, for now ;-)
 - description (required): Description of the field set
 - fields (required): YAML array as described below
-- type (ignored): at this level, should always be `group`
+- type (optional): at this level, type can be `group`, `object`, or `nested`
```
varshavjm: Can we mention the default value as group if not user-specified? For instance,
- type (optional): at this level, type can be `group`(default), `object`, or `nested`

PR #753, opened 2020-02-19
Fix bug with subsets and reusable fields

marshallmain:
Subset operation needs to be done after reusables are dropped in place, otherwise reusables could be incorrectly removed by the subset operation.  i.e. if `hash` is not included in the subset at the top level, then it won't show up at `process.hash` even if `process.*` is included in the subset.


marshallmain:
This needs to be updated to fix a related bug - now if you apply a subset operation to reusables in one place then the reusable is modified in all places.  I think the right solution is to allow reusables to be subsetted independently in each location where they are reused, which means doing the subset operation after the reusables are converted from references to copies in the intermediate structure.  I'll update the PR and remove the `in progress` label when that's done.

PR #754, opened 2020-02-23
Add rule attribution fields

MikePaquette:
The `rules.*` fieldset allows the inclusion of details about a rule used to perform some kind of detection.  Detection rules are viewed by some as software, and by others as creative content.  Either way, rules will commonly be made available pursuant to a license and/or requirement to provide attribution to the rule author(s.)

This PR adds three new "attribution" fields to the `rule.*` fieldset:

- [x] `rule.author`: Name, organization, or pseudonym of the author or authors who created the rule used to generate this event.
- [x] `rule.license_type`: Name of the license under which the rule used to generate this event is made available.
- [ ] `rule.license_reference`: Reference URL to the license under which the rule used to generate this event is made available.

Related question: since there are two "license" fields being proposed, should we create a `license.*` sub-object?


schemas/rule.yml
```diff
@@ -81,4 +81,31 @@
...
+    - name: license_type
+      level: extended
+      type: keyword
+      short: Rule license
+      description: >
+        Name of the license under which the rule used to generate this event is made available.
+
+      example: Apache 2.0
+
+    - name: license_reference
```
webmat: I'm not sure the reference URL of the license of the rule should be added to every single alert. That would take a whole lot of room for very little benefit, no?

schemas/rule.yml
```diff
@@ -81,4 +81,31 @@
...
       example: https://en.wikipedia.org/wiki/DNS_over_TLS
 
+    - name: author
+      level: extended
+      type: keyword
+      short: Rule author
+      description: >
+        Name, organization, or pseudonym of the author or authors who created the rule used to generate this event.
+
+      example: John R. Doe, Elastic, s3cst1ltsk1n
```
webmat: Two things about the example here.

First, please only put one value in the example. The asciidoc rendering looks funny otherwise ([see here](http://ecs_754.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-rule.html)). To provide multiple different examples, for now we have to do it via the description.

Second, the alphanumeric handle looks strange. I know it's supposed to represent an online handle. Perhaps we could use a known handle like "Citizenfour", or a fictional one like "Starlord"?

Starlord

schemas/rule.yml
```diff
@@ -81,4 +81,31 @@
...
+      normalize:
+        - array
```
webmat: Do we need the `rule.author` to be an array?

schemas/rule.yml
dcode: :+1: for Star-Lord

schemas/rule.yml
dcode: a related discussion for rule development, we discussed an ecosystem where an author was an organization (think Emerging Threats) which publishes many rulesets (maps to their categories, e.g. Malware, BOTCC, Exploit, etc), each of which contains many rules.

And, as an implementation detail, this begs the question regarding rule UUID generation, it may require namespaced UUID generation or a second "author" UUID

schemas/rule.yml
MikePaquette: Good question.  Yes, this does seem like very little benefit to most events/alerts that would be received.  It would be most valuable to a rule document itself, which is not technically an ECS event.

MikePaquette: I've removed this field altogether for now.  If we build a strong case for it later, we can add with another PR.

MikePaquette:
Added Star-Lord as example.  Removed `rule.license_reference` field.

schemas/rule.yml
webmat: For posterity: we agreed offline that this field does actually benefit from being an array.

This allows for clearly capturing different organizational units or people contributing to the rule.

Resolving this discussion.

schemas/rule.yml
```diff
@@ -81,4 +81,22 @@
...
+      type: keyword
+      short: Rule author
+      description: >
+        Name, organization, or pseudonym of the author or authors who created the rule used to generate this event.
+
+      example: Start-Lord
+      normalize:
+        - array
+
+    - name: license_type
```
webmat: I just noticed the field name `license_type` here. I think we should simply name this `license`, in order to be aligned with `package.license`.

schemas/rule.yml
```diff
@@ -81,4 +81,22 @@
...
       example: https://en.wikipedia.org/wiki/DNS_over_TLS
 
+    - name: author
+      level: extended
+      type: keyword
+      short: Rule author
+      description: >
+        Name, organization, or pseudonym of the author or authors who created the rule used to generate this event.
+
+      example: Start-Lord
```
webmat:
```suggestion
      example: Star-Lord
```

MikePaquette:
@webmat Thanks, yes, please make both changes.

PR #755, opened 2020-02-23
Categorization: Add new "iam" category and associated types "admin", "group", "user"

MikePaquette:
This new value of `event.category` is being added to properly categorize Identity and access management events relating to users, groups, and admin.

- [ ] `event.category:"iam"`

It is expected that this category would be used to visualize and analyze iam-related logs and data from sources such as active directory, LDAP, Okta, Duo, and other IAM systems.

There are three new values of `event.type` associated with this new addition.
- [ ] `event.type:"admin"`
- [ ] `event.type:"group"`
- [ ] `event.type:"user"`

### DIFF
--- a/code/go/ecs/rule.go
+++ b/code/go/ecs/rule.go
@@ -57,4 +57,16 @@ type Rule struct {
 	// that's not available, it can also be a link to a more general page
 	// describing this type of alert.
 	Reference string `ecs:"reference"`
+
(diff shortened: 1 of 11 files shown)


MikePaquette:
Oh crumbs, I added in the rule changes to this pull-request by mistake.  Please ignore here.

PR #756, opened 2020-02-24
Categorization: Add new "iam" category and associated types "admin", "group", "user"

MikePaquette:
This new value of `event.category` is being added to properly categorize Identity and access management events relating to users, groups, and admin.

- [x] `event.category:"iam"`

It is expected that this category would be used to visualize and analyze iam-related logs and data from sources such as active directory, LDAP, Okta, Duo, and other IAM systems.

There are three new values of `event.type` associated with this new addition.
- [x] `event.type:"admin"`
- [x] `event.type:"group"`
- [x] `event.type:"user"`

### DIFF
--- a/docs/field-values.asciidoc
+++ b/docs/field-values.asciidoc
@@ -143,6 +143,7 @@ that will require subsequent breaking changes.
 * <<ecs-event-category-driver,driver>>
 * <<ecs-event-category-file,file>>
 * <<ecs-event-category-host,host>>
+* <<ecs-event-category-iam,iam>>
 * <<ecs-event-category-intrusion_detection,intrusion_detection>>
 * <<ecs-event-category-malware,malware>>
 * <<ecs-event-category-package,package>>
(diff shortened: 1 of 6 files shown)


docs/field-values.asciidoc
```diff
@@ -331,6 +349,16 @@ The access event type is used for the subset of events within a category that in
 
 
 
+[float]
+[[ecs-event-type-admin]]
+==== admin
+
+The admin event type is used for the subset of events within a category that are related to admin objects. Common example: `event.category:iam AND event.type:admin`. You can further distinguish admin operations using the ECS `event.action` field.
```
webmat: @MikePaquette Can you add a more specific example of what would map to "admin" to the description? What's an "admin object"?

schemas/event.yml
```diff
@@ -329,6 +349,13 @@
             Note that pipeline errors that occur during the event ingestion process
             should not use this `event.type` value. Instead, they should use
             `event.kind:pipeline_error`.
+        - name: group
+          description: >
+            The group event type is used for the subset of events within a category
+            that are related to group objects.
+            Common example: `event.category:iam AND event.type:group`.
```
webmat: Since we're specifically banking on `event.type` being an array here, perhaps we should make the example more explicit in that regard? Perhaps like this: `event.category:iam AND event.type:creation AND event.type:group`

I'd say the same goes for each of the new `event.type`, WDYT?

MikePaquette: Yes, great idea, will add this.

docs/field-values.asciidoc
dainperkins: Something like administrative changes withingan identity and access management framework that do not specifically affect a user or group (adding new applications to federation solution for instance)?

MikePaquette: Added clarification based on Dain's examples.

schemas/event.yml
MikePaquette: Added array-reinforcing examples for event.type values admin, user, group.

PR #758, opened 2020-02-24
Clarify event.kind definitions re: vulnerabilities

MikePaquette:
During a review of the ECS categorization fields, it was noticed that the examples given for `event.kind:"metric"` were actually a better fit for `event.kind:"state"`

This PR clarifies both definitions.

No ECS fields or values are changed with this PR.

Thanks @dainperkins for noticing this.


schemas/event.yml
```diff
@@ -71,16 +71,18 @@
...
         - name: state
           description: >
             This value is similar to metric, except that the entity being measured does not
-            provide a numeric metric value, but rather one of a fixed set of conditions or states.
+            provide a numeric metric value, but rather one of a fixed set of categorical values
+            that represent conditions or states.
+
             For example a periodic event reporting a "fin_wait" state of a TCP connection
-            on a host might use `event.type:state`.
+            on a host, or the result of a vulnerability scan on a host, could use `event.type:state`.
```
webmat: I don't think we should be using vulnerability scan as an example for `event.kind` state. I think the "state" aspect of a vulnerability scan isn't direct enough.

Clearer examples of state events could be the state of an Elasticsearch cluster (green/yellow/red). Or the state of a TCP socket in general. There's "fin_wait", but there's also all of the other possible states: open, closed, etc.

schemas/event.yml
```diff
@@ -14,7 +14,7 @@
     A metric is defined as an event containing one or more numerical or
     categorical measurements and the time at which the measurement was taken.
     Examples of metric events include memory pressure measured on a host,
-    or vulnerabilities measured on a scanned host.
+    or vulnerabilities measured on a scanned host. See `event.kind` for additional details.
```
webmat: We still have wording of putting vulnerability scans in `event.kind:metric` here

schemas/event.yml
dainperkins: maybe a couple of examples? e.g. hosts state of vulnerability to a given attack, hosts tcp socket state, file state open/locked? Cluster state seems like it would be a periodically measured thing (metric), but I could be basing that on an core understanding of "cluster state" thru the lens of  e.g. cisco ASA HA polling, or MS clustering

schemas/event.yml
MikePaquette: fixed, thanks.

schemas/event.yml
MikePaquette: @webmat I'd really like to keep the vulnerability example, as ECS is otherwise silent on how to categorize events that populate the `vulnerability.*` field set, and I've heard this question asked several times.  I think that `event.kind:state` is the correct kind categorization for a vulnerability check performed on a single host.  i.e., the "state" result will be "vulnerable" or "not vulnerable."

I've added a few more examples from yours and @dainperkins suggestions to help clarify.

> Although I'm not sure I like the approach in the point above. I think all state events should be in event.kind:state, and whether it's a transition or an informational update should be determined via another field.

I disagree. I'd be worried that this could cloud the definition of event.kind"event".   As proposed, we now have an easily summarized anecdote to distinguish the three. 

 _Regarding `event.kind:`, if something happened: use `event`; if something numerical is measured: use `metric`; if something categorical is measured: use `state`._

Accordingly, I've clarified that state changes should be categorized as `event.kind:"event"`

webmat: Ok, I like the resulting description for "state", it makes a lot of sense.

I'm back with you and @andrewstucki, let's have state changes be "events" 👍

schemas/event.yml
```diff
@@ -11,10 +11,11 @@
     Examples of log events include a process starting on a host,
     a network packet being sent from a source to a destination,
     or a network connection between a client and a server being initiated or closed.
-    A metric is defined as an event containing one or more numerical or
-    categorical measurements and the time at which the measurement was taken.
-    Examples of metric events include memory pressure measured on a host,
-    or vulnerabilities measured on a scanned host.
+    A metric is defined as an event containing one or more numerical measurements
```
dainperkins: Should we add periodic? (e.g. metric is defined as an event containing one or more numerical or categorical periodic measurements...)

schemas/event.yml
```diff
@@ -68,19 +69,31 @@
             It is used to represent events that indicate that something happened.
         - name: metric
           description: >
-            This value is used to indicate that this event that a numeric measurement
-            was taken at given point in time.
+            This value is used to indicate that this event describes a numeric measurement
```
dainperkins: same as above - add periodic in to differentiate a periodic measurement of some value vs. a state at a given time

schemas/event.yml
```diff
@@ -68,19 +69,31 @@
...
+            the state of a TCP connection (open, closed, fin_wait, etc.), the state of a host with respect
+            to a software vulnerability (vulnerable, not vulnerable), and the state of a system
+            regarding compliance with a regulatory standard (compliant, not compliant).
+
+            Note that an event that describes a change of state would not use `event.kind:state`,
+            but instead would use 'event.kind:event' since a state change fits the more general
+            event definition of something that happened.
+
+            State events are often collected on a predictable frequency, such as once
+            every few seconds, once a minute, once an hour, or once a day.
```
dainperkins: will "often collected on a predictable frequency" suggest to casual users that a one off, or manually initiated ad hoc state query (vulnerability scan for instance) should go somewhere else?  
should it say something like:
            State events are often collected on a predictable frequency, such as once
            every few seconds, once a minute, once an hour, or once a day, **but can 
            also be used to store ad-hoc state queries as performed**

schemas/event.yml
MikePaquette: I think no.  While metrics are "often" collected on a periodic basis, our definition allows for metrics that are taken on an ad hoc, or irregular schedule.

This is clarified in the event.kind section.

schemas/event.yml
MikePaquette: Again, both state and metric event kinds can occur at a fixed period, or on an ad hoc, or irregular basis.

schemas/event.yml
MikePaquette: We added "often" for exactly that reason - to allow for ad hoc or manual measurements to still be included in this event kind, but because we are writing a spec, I like your suggestion for explicit clarity.  Will add.

schemas/event.yml
dainperkins: Ok - I'm good with that

schemas/event.yml
dainperkins: good with it here too

PR #759, opened 2020-02-24
Enhance description of event.outcome field

MikePaquette:
Since publishing ECS 1.4, where the ECS categorization fields were introduced, the `event.outcome` field has generated the most questions about its definition, use, and consistency.

This PR significantly expands the definitions and example usage of event.outcome based on the feedback from the community.

No field names or allowed values are changed with this PR.

### DIFF
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -244,14 +244,24 @@
     - name: outcome
       level: core
       type: keyword
-      short: The outcome of the event. The lowest categorization field in the hierarchy.
+      short: The outcome of the event. The lowest level categorization field in the hierarchy.
       description: >
         This is one of four ECS Categorization Fields, and indicates the
         lowest level in the ECS category hierarchy.
 
-        `event.outcome` simply denotes whether the event represent a success or a failure.
-        Note that not all events will have an associated outcome. For example, this field is
-        generally not populated for metric events or events with `event.type:info`.
+        `event.outcome` simply denotes whether the event represents a success or a failure from
+        the perspective of the entity that produced the event.
+
+        Note that when a single transaction is described in multiple events, each event may
+        populate different values of `event.outcome`, according to their perspective.
+
+        Also note that in the case of a compount event (a single event that contains multiple logical events),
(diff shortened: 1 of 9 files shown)


schemas/event.yml
```diff
@@ -261,15 +271,16 @@
...
         - name: unknown
           description: >
             Indicates that this event describes only an attempt for which the result
-            is unknown. For example, if the event contains information only about a
-            request in an entity transaction that usually results in a response,
-            populating `event.outcome:unknown` is appropriate.
+            is unknown from the perspective of the event producer. For example, if the
+            event contains information only about a request in an entity transaction that
+            usually results in a response, populating `event.outcome:unknown` in the request
+            event is appropriate.
```
webmat: We should clarify the description for "unknown" a bit, in my opinion. Here's an attempt:

> For example, if the event contains information only about the request side of a transaction that results in a response, populating `event.outcome:unknown` in the request event is appropriate.

We could also reinforce the distinction between absent and unknown, here as well. We don't need to reiterate all of [this last paragraph](https://github.com/elastic/ecs/pull/759/files?file-filters%5B%5D=.yml#diff-fd197febfa7b490a2fa19f86a95f6d52R262-R264), but perhaps just:

> This field should be left absent when an outcome doesn't make logical sense for the event.

MikePaquette: Thanks, updated accordingly.

schemas/event.yml
```diff
@@ -244,14 +244,24 @@
...
-        `event.outcome` simply denotes whether the event represent a success or a failure.
-        Note that not all events will have an associated outcome. For example, this field is
-        generally not populated for metric events or events with `event.type:info`.
+        `event.outcome` simply denotes whether the event represents a success or a failure from
+        the perspective of the entity that produced the event.
+
+        Note that when a single transaction is described in multiple events, each event may
+        populate different values of `event.outcome`, according to their perspective.
+
+        Also note that in the case of a compount event (a single event that contains multiple logical events),
```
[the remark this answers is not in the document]
MikePaquette: fixed, thanks.

Issue #749, opened 2020-02-15
Mime Types 


neu5ron:
I was wondering if we could discuss/add `mime_type` as a nested field.
There is a bit discussed on this already over at #554, however since it started as HTTP specific header fields, I think it may have gotten lost in the mix and mime type's deserve their own issue :)

I think this would be a pretty quick win that would benefit setting up some nice analytics and just all around schema for a lot of logs.

a really easy "analytic" is something like `file.mime_type:"application/x-dosexec" AND NOT file.extension:(exe OR dll OR msi)`

Some type of logs and things that would benefit:
- Suricata HTTP
- Suricata Alert
- Yara and other file scanning
- AV, EDR, and or Sandbox
- Zeek HTTP
- Zeek SMTP
- Zeek File
- Zeek FTP
- Anything else containing HTTP/SMTP logs (Squid, PaloAlto, etc...)

Miss you @webmat ;)

webmat:
Ah right, good point. Where would you see adding a `.mime_type` field?

* `file.mime_type`
* `http.request.mime_type`
* `http.response.mime_type`
* Somewhere under `email.*` when we get around to it, of course
* Anywhere else?

Is it worth breaking up the mime_type into type, subtype & parameter to discrete fields, you think?

BTW you're still hanging around the cool places, and that's what counts ;-)

![French Mime](https://media.giphy.com/media/8P1oO2JbrZK2uSYnL6/giphy.gif)

neu5ron:
I think all those fields are perfect start. 
Seems good to me!

webmat:
Solved by #760

webmat:
Just realized #760 only added `file.mime_type`, but not the http request/response ones.

@ebeahan Let's see if we can add `http.request.mime_type` and `http.response.mime_type` for 1.6, WDYT?

ebeahan:
Sounds good 👍

ebeahan:
Reading over the past discussion here, #760, and #554, appears only the `file.mime_type` field was added due to the ongoing conversations around how to best handle HTTP headers in ECS such as `content-type`.

I do see a case for populating a dedicated `http.response.mime_type` field alongside the nested HTTP header fields. The first extracts the MIME type (that happens to be transmitted in the `Content-Type` header) while the second captures all HTTP headers from the request/response (which would capture the `content-type` as well, but not in a specialized field).

[...]

PR #760, opened 2020-02-24
Add initial stab at mime_type for file objects

dcode:
Address #749 file objects.

While "Content-Type" in HTTP headers and related protocols utilize
MIME type descriptions, they are offered as a hint to the other
end of the connection so they know how to handle the data. However,
in the context of a file, the MIME type should be more authoritative
in that tools like Zeek, Suricata, and other tools that use
libmagic or yara to detect filetypes provide information based upon
a match to a pattern of bytes.

For #554, I think retaining the headers as `Content-Type` makes sense
as this is the configured value set by the server (or client). I think
it is useful to compare what was advertised vs what was actually sent, 
e.g. the header value in `http.response.content-type` vs `file.mime_type`
of the file transferred.


schemas/file.yml
```diff
@@ -4,169 +4,178 @@
...
+
+        Note that not all filesystems keep track of access time.
+
+    - name: mime_type
+      level: extended
+      type: keyword
+      short: Media type of file, document, or arrangement of bytes.
+      description: >
+        MIME type should identify the format of the file or stream of bytes using
+        [IANA official types](https://www.iana.org/assignments/media-types/media-types.xhtml),
```
webmat: Please replace the Markdown link with an asciidoc link.

```suggestion
        https://www.iana.org/assignments/media-types/media-types.xhtml[IANA official types],
```

We even normally try to avoid asciidoc link, as the descriptions are used elsewhere as well. But this is one case where the link is just too long for it to look good without a real link :-)

webmat:
Ping @neu5ron

webmat:
Ping @neu5ron

PR #761, opened 2020-02-24
Add network category and related types

MikePaquette:
This PR adds a new value of event.category:"network" and the corresponding values of `event.type` to be able to represent all network events and metrics in ECS.

- [x] `event.category:"network"`
- [x] `event.type:"allowed"`
- [x] `event.type:"denied"`
- [x] `event.type:"connection"`
- [x] `event.type:"protocol"`


dainperkins:
Are we putting flow to be under connection? Or would that be dependent on event.kind?

webmat:
I think we should add "allowed" and "denied" to category "authentication" as well. WDYT?

dainperkins:
authentication should be success, failure - allowed/denied is authorization (which iirc we are typically reporting on under the resource that access was attempted on - e.g. file access allowed / denied)

MikePaquette:
@dainperkins re:

> Are we putting flow to be under connection? Or would that be dependent on event.kind?

Not sure what you mean by "flow" in that question, but indeed anything that would be included in flow analysis would have `event.type:connection`

schemas/event.yml
```diff
@@ -316,6 +361,17 @@
             that indicate that something was deleted.
             A common example is `event.category:file AND event.type:deletion`
             to indicate that a file has been deleted.
+        - name: denied
+          description: >
+            The denied event type is used for the subset of events within a category that
+            indicate that something was denied. Common examples include
+            `event.category:”network” AND event.type:denied` (to indicate a network
```
webmat: Please make sure to always use straight quotes, not the fancy ones Google Docs and the others use.

```suggestion
            `event.category:"network" AND event.type:denied` (to indicate a network
```

schemas/event.yml
```diff
@@ -344,6 +400,22 @@
             The installation event type is used for the subset of events within a category
             that indicate that something was installed.
             A common example is `event.category:package` AND `event.type:installation`.
+        - name: protocol
+          description: >
+            The protocol event type is used for the subset of events within a category that
+            indicate that they contain protocol details or analysis. Generally, network events
+            that contain protocol details will fall into this subcategory. A common example is
+            `event.category:network AND event.type:protocol AND event.type:connection` AND event.type:end'
```
dcode: Move back tick to end to fix formatting.

```suggestion
            `event.category:network AND event.type:protocol AND event.type:connection AND event.type:end`'
```

docs/field-values.asciidoc
```diff
@@ -341,6 +370,16 @@ The change event type is used for the subset of events within a category that in
 
 
 
+[float]
+[[ecs-event-type-connection]]
+==== connection
+
+Used primary with `event.category:network` this value is used for the subset of network traffic that includes sufficient information for the event to be included in flow or connection analysis. Events in this subcategory will contain at least source and destination IP addresses, source and destination TCP/UDP ports, and will usually contain counts of bytes and/or packets transferred. Events in this subcategory may contain unidirectional or bidirectional information, including summary information. Use this subcategory to visualize and analyze network connections. Flow analysis, including Netflow, IPFIX, and other flow-related events fit in this subcategory. Note that firewall events from many NGFW devices will also fit into this subcategory.  A common filter for flow/connection information would be `event.category:network AND event.type:connection AND event.type:end` (to view or analyze all completed network connections, ignoring mid-flow reports). You can further distinguish connection events using the ECS `event.action` field, populating with values of your choosing, such as "timeout", or "reset".
```
andrewkroh: `s/primary/primarily/`

andrewkroh: NGFW - if this is the first mention this also include Next-Generation Firewall.

docs/field-values.asciidoc
```diff
@@ -361,6 +400,16 @@ The deletion event type is used for the subset of events within a category that
 
 
 
+[float]
+[[ecs-event-type-denied]]
+==== denied
+
+The denied event type is used for the subset of events within a category that indicate that something was denied. Common examples include `event.category:”network” AND event.type:denied` (to indicate a network firewall event for which the firewall disposition was to deny the connection to complete) and `event.category:intrusion_detection AND event.type:denied` (to indicate a network intrusion prevention system event for which the IPS disposition was to deny the connection to complete). You can further distinguish denied operations using the ECS `event.action` field, populating with values of your choosing, such as "blocked", "dropped", or "quarantined".
```
andrewkroh: The help keep people sane 😄 replace the usage of right (`”`) quotes with a plain old quote (`"`). Or just remote that set of quotes since the other term in the same query doesn't use quotes.

MikePaquette: OK, deleted them.

docs/field-values.asciidoc
MikePaquette: good, spelled out for first-time use.

schemas/event.yml
MikePaquette: thanks, fixed.

PR #764, opened 2020-02-26
Add agent.build for extended agent version information

andrewstucki:
This adds a field for extended build information for agents. This would actually be pretty useful in a few scenarios:

1. An agent running in an emulated execution context (i.e. 32bit binary running on a 64bit system)
2. When someone wants to identify a custom build of an agent that otherwise follows the exact same versioning semantics as other agents

The use case from the endpoint side is in identifying exact build configurations of an endpoint that's shipping data to elasticsearch. When we see a buggy payload we want to be able to identify the exact build of the binary even if semver indicates it's from a particular release branch. I could imagine the same thing would be potentially useful in beats.


webmat:
I think this makes sense. The gist is for this field to be free form, correct? E.g. another type of agent, or a custom pipeline could very well populate this only with a Git sha, a 4-5 digit numeric or other kinds of shenanigans?

@ruflin or @urso thoughts on this field?

webmat:
@james-elastic Would this be useful to the Endpoint team? Does someone on the team have thoughts about this?

urso:
When hearing custom build and sha, I rather wonder if it would make sense to have standardized fields, instead of an adhoc string only field. E.g. `build.timestamp`, ` `build.id`, `build.machine`, `build.host`, ...

ruflin:
+1 on the idea of having `.build` or `.build.*` in ECS. 

@urso Perhaps an idea is that we go with `agent.build.id` to get started so we have a prefix available but don't need to get into the details of all the other fields?

urso:
+1 on making `agent.build` extensible.

What will `agent.build.id` actually mean? The example output uses: `metricbeat version 7.6.0 (amd64), libbeat 7.6.0 [6a23e8f8f30f5001ba344e4e54d8d9cb82cb107c built 2020-02-05 23:10:10 +0000 UTC]`. The example reminds me in `user_agent.original`. To be consistent with that we might want to name the field `agent.build.original` :)

ruflin:
I'm on board starting with `.original`.

james-elastic:
@webmat yeah this would be very helpful!

webmat:
Love the idea of following the `.original` naming pattern from User Agent. Thanks for this idea, @urso !

@andrewstucki I think we can move swiftly and rename your field addition to `agent.build.original`, meant to be free form, like you show in the example.

Then in later PRs, as needed we can start adding more precise fields. `agent.build.commit` would be an obvious next field, and perhaps the build date as well.

WDYT about this plan?

andrewstucki:
@webmat sorry, this totally slipped by--I renamed the field to `agent.build.original` and updated the changelog.

schemas/agent.yml
```diff
@@ -23,6 +22,14 @@
 
       example: 6.0.0-rc2
 
+    - name: build.original
+      level: core
+      type: keyword
+      description: >
+        Extended build information for the agent.
+
+      example: metricbeat version 7.6.0 (amd64), libbeat 7.6.0 [6a23e8f8f30f5001ba344e4e54d8d9cb82cb107c built 2020-02-05 23:10:10 +0000 UTC]
```
webmat, line 31: I think we should be explicit here that no format is mandated. So a few things:

- [ ] copy the current first sentence to populate `short:` as is, I think it's great as a one liner.
- [ ] expand `description:` with a second sentence (or para) to the effect that no format is mandated, it's up to the data source to populate with anything they have / need.

CHANGELOG.next.md
```diff
@@ -22,6 +22,7 @@ Thanks, you're awesome :-) -->
 
 * Added `search.*` fields #729
 * Add architecture and imphash for PE field set. (#763)
+* Added `agent.build.*` for extended agent version information. (#764)
```
webmat, line 25:
```suggestion
* Added `agent.build.original` for extended agent version information. (#764)
```

PR #769, opened 2020-03-03
Add environment field

dro25:
Many companies need to specify the environment in the log messages themselves. Service seems like the appropriate place for this field to me. Would you please consider adding an environment field to allow easily isolating logs from the appropriate environment?


dro25:
Closing since it might take time to get the license sanctioned

PR #782, opened 2020-03-11
Quote the `labels` example, to avoid the YAML interpreter interpreting it

webmat:


webmat:
I don't think this warrants a patch release.

I suggest you replace the problem section of the 1.5 YML in Beats with this exact output (https://github.com/elastic/ecs/pull/782/files#diff-5c26e320ef5dca7ad0441c5eeb32dba3R33), @adriansr. This will avoid an unnecessary diff when comes time to introduce 1.6 in Beats. WDYT?

PR #788, opened 2020-03-17
log.syslog can contain multiple objects, hence requires the array definition.

### LINKED ISSUE #780: [1.5] network.inner, observer.ingress and observer.egress object definitions
The following new schema definitions in ECS 1.5:

- `network.inner`
- `observer.ingress`
- `observer.egress`

Exhibit the same format of definition in the `ecs_nested.yml` file as:

- `dns.anwers`
- `log.syslog`

That is, they specify the root object and then subsequent properties off of this root object, for example (truncated YAML sample):

```
ingress:
      dashed_name: observer-ingress
      description: Observer.ingress holds information like interface number and name,
        vlan, and zone information to  classify ingress traffic.  Single armed monitoring
        such as a network sensor on a span port should  only use observer.ingress
        to categorize traffic.
      type: object
ingress.interface.alias:
      dashed_name: observer-ingress-interface-alias
      description: Interface alias as reported by the system, typically used in firewall
        implementations for e.g. inside, outside, or dmz logical interface naming.
      type: keyword
ingress.interface.id:
      dashed_name: observer-ingress-interface-id
      description: Interface ID as reported by an observer (typically SNMP interface
        ID).
      type: keyword
```

In C# it is not possible to have both a `class` and `property` with the same name, so in the ECS .NET generator we have adopted the same strategy for these new properties as `dns.answers` and `log.syslog`.

Does this seem reasonable?

codebrain:
This was locally overridden in the .NET client.

Fixes #780 

@webmat - please can you re-run the generator before approval? Thanks.

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -3014,6 +3014,9 @@ example: `Sep 19 08:26:10 localhost My log`
 type: object
 
 
+Note: this field should contain an array of values.
+
+
 
 
 
--- a/generated/csv/fields.csv
+++ b/generated/csv/fields.csv
@@ -269,7 +269,7 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 1.6.0-dev,true,log,log.origin.file.name,keyword,extended,,Bootstrap.java,The file which originated the log event.
 1.6.0-dev,true,log,log.origin.function,keyword,extended,,init,The function which originated the log event.
 1.6.0-dev,false,log,log.original,keyword,core,,Sep 19 08:26:10 localhost My log,"Original log message with light interpretation only (encoding, newlines)."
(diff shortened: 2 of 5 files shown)


webmat:
Where did you get the impression that `log.syslog` should be an array? I don't think it should be.

codebrain:
Its a good question!

I made it an array in the .NET code last October 31st 2019 in this commit: https://github.com/elastic/ecs-dotnet/commit/69f11e19f1e36d588df829e6bc866fd03794a4ec

I must have seen an example ECS json at the time, and saw that it contained elements in an array, and therefore made the change. I have gone back through my conversation history and cannot seem to find the example again.

Therefore, I think I'll defer to your better judgement here and revert the change and have the .NET client map as a single object and not an array.

Please feel free to close this PR.

webmat:
This one should indeed not be an array. Closing. Thanks Stuart

PR #797, opened 2020-03-23
Normalize DNS related fields across entities

### LINKED ISSUE #728: Standardize Domain representations
Need to double check scope of top level fields, but I think it makes sense to standardize the domain representation between locations (minimally e.g. dns & url, likely client, server etc.) both for the sale of consistent representation, but also for potentially pulling standard features from all domain trackable information.

I think we should consider adding a hostname field as well, within the domain representation (its getting large, but breaking it up seems to be worth the effort)

WIll add a PR shortly

dainperkins:
Adding fields to standardize domain name representations & features for:
dns, client, destination, host, server, source fields.

Closes #728

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Added `search.*` fields #729
+* normalized dns fields across dns, client, destination, host, server, source #
 
 #### Improvements
 
@@ -28,9 +28,6 @@ Thanks, you're awesome :-) -->
 #### Breaking changes
 
 #### Bugfixes
-
-* Quoted the example for `labels` to avoid YAML interpreting it, and having
-  slightly different results in different situations. #782
 
 #### Added
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* normalized dns fields across dns, client, destination, host, server, source #
+* normalized dns fields across dns, client, destination, host, server, source #797
 
 #### Improvements
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* normalized dns fields across dns (question and answers), client, destination, host, server, source #797
+* Added missing fields to normalize DNS-related fields across `dns` (question and answers), `client`, `destination`, `host`, `server`, `source`, and `url` #797
 
 #### Improvements
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* normalized dns fields across dns, client, destination, host, server, source #797
+* normalized dns fields across dns (question and answers), client, destination, host, server, source #797
 
 #### Improvements
 
--- a/schemas/client.yml
+++ b/schemas/client.yml
@@ -74,7 +74,21 @@
         the additional fields of top_level_domain, registered_domain, subdomain, and hostname
         as they are available.
       example: www.east.mydomain.co.uk 
- 
+
+    # Domain Stuff
+    - name: hostname
+      level: core
+      type: keyword
+      short: Hostname of the system.
+      description: >
+        Hostname of the system.
+
+        It normally contains what the `hostname` command returns on the host machine, or the
+        host portion of a fully qualified domain name.
+        
+        For example, the hostname portion of "www.east.mydomain.co.uk " is "www".
+      example: www
+    
     - name: registered_domain
       level: extended
       type: keyword
@@ -82,7 +96,7 @@
       description: >
         The highest registered domain, stripped of the subdomain.
 
-        For example, the registered domain for "foo.google.com" is "google.com".
+        For example, the registered domain for " www.east.mydomain.co.uk " is "mydomain.co.uk".
 
         This value can be determined precisely with a list like the public
         suffix list (http://publicsuffix.org). Trying to approximate this by
@@ -95,12 +109,12 @@
       short: The effective top level domain (com, org, net, co.uk).
       description: >
         The effective top level domain (eTLD), also known as the domain suffix,
-        is the last part of the domain name.
-        For example, the top level domain for google.com is "com".
+        is the last part of the domain name and is typically inclusive of the 
+        top level domain (e.g. "co"), as well as including country code and region codes
+        For example, the top level domain for www.east.mydomain.co.uk is "co.uk".
 
         This value can be determined precisely with a list like the public
-        suffix list (http://publicsuffix.org). Trying to approximate this by
-        simply taking the last label will not work well for effective TLDs such as "co.uk".
+        suffix list (http://publicsuffix.org). 
       example: co.uk
 
     - name: subdomain
@@ -108,32 +122,27 @@
       type: keyword
       short: The subdomain of the domain.
       description: >
-        The subdomain is all of the labels under the registered_domain in a partially qualified
-        domain name, and all of the labels except the hostname in a fully qualified domain name.
+        The subdomain portion of a fully qualified domain name includes all of the names except 
+        the host name under the registered_domain.  In a partially qualified domain, or if the 
+        the qualification level of the full name cannot be determined, subdomain contains all of
+        the names below the registered domain.
         
+        For example the subdomain portion of "www.east.mydomain.co.uk " is "east".
+
         If the domain has multiple levels of subdomain, such as "sub2.sub1.example.com",
         the subdomain field should contain "sub2.sub1", with no trailing period.
       example: east
 
-    - name: hostname
-      level: core
-      type: keyword
-      short: Hostname of the client system.
-      description: >
-        Hostname of the client system.
-
-        It normally contains what the `hostname` command returns on the host machine, or the
-        host portion of a fully qualified domain name.
-      example: www
-
-    - name: country_code
+    - name: is_internationalized
       level: extended
-      type: keyword
-      short: The country code portion of the domain.
+      type: boolean
+      short: Does the domain contain punycode characters.
       description: >
-        The country code portion of the domain.
+        Internationalized domain names may contain non LDH-ASCII character encoding used
+        to represent full unicode characters in the ASCII representation used by DNS. For example the 
+        internationalized country code TLD for Israel is .xn--4dbrk0ce.
 
-      example: uk
+      example: false
 
     # Metrics
     - name: bytes
@@ -150,7 +159,8 @@
       example: 12
       description: >
         Packets sent from the client to the server.
-
+    
+    #nat fields
     - name: nat.ip
       level: extended
       type: ip
--- a/schemas/client.yml
+++ b/schemas/client.yml
@@ -59,19 +59,35 @@
       description: >
         Client domain.
 
+    - name: name
+      level: core
+      type: keyword
+      short: Name of the client system.
+      description: >
+        Name of the client system.
+
+        It can contain what `hostname` returns on Unix systems, the fully
+        qualified domain name, or a name specified by the user. The sender
+        decides which value to use.
+        
+        In the case of fully and partially qualified domain names it is recommended to parse out 
+        the additional fields of top_level_domain, registered_domain, subdomain, and hostname
+        as they are available.
+      example: www.east.mydomain.co.uk 
+ 
     - name: registered_domain
       level: extended
       type: keyword
-      short: The highest registered client domain, stripped of the subdomain.
+      short: The highest registered domain, stripped of the subdomain.
       description: >
-        The highest registered client domain, stripped of the subdomain.
+        The highest registered domain, stripped of the subdomain.
 
         For example, the registered domain for "foo.google.com" is "google.com".
 
         This value can be determined precisely with a list like the public
         suffix list (http://publicsuffix.org). Trying to approximate this by
         simply taking the last two labels will not work well for TLDs such as "co.uk".
-      example: google.com
+      example: mydomain.co.uk
 
     - name: top_level_domain
       level: extended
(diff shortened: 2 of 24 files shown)


webmat:
Tip: I added `Closes #728` to the end of the PR text. This will automatically close the related issue when we merge this. Please do this, when a PR addresses an issue in full. When it addresses part of an issue, just mentioning the is useful as well, as it links the two together.

CHANGELOG.next.md
```diff
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Added `search.*` fields #729
```
webmat, line 19: Please add back this line

CHANGELOG.next.md
```diff
@@ -29,9 +29,6 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
-* Quoted the example for `labels` to avoid YAML interpreting it, and having
-  slightly different results in different situations. #782
-
```
webmat, line 34: Please add back this entry as well

CHANGELOG.next.md
```diff
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Added `search.*` fields #729
+* normalized dns fields across dns (question and answers), client, destination, host, server, source #797
```
webmat, line 19: Please try to follow the boring changelog sentence structure when possible, especially the first word. Here we could say

```suggestion
* Added missing fields to normalize DNS-related fields across `dns` (question and answers), `client`, `destination`, `host`, `server`, and `source` #797
```

schemas/client.yml
```diff
@@ -87,6 +103,38 @@
...
+      example: www
+
+    - name: country_code
+      level: extended
+      type: keyword
+      short: The country code portion of the domain.
+      description: >
+        The country code portion of the domain.
+
+      example: uk
```
webmat, line 136: I'm not sure if we should go all the way to adding `country_code`.

I understand the value, but even just parsing out the eTLD and the registered_domain reliably puts a high burden on implementations, IMO. 

Asking them to also figure out the country code (only when appropriate, so not when it's a ".com", ".org" or ".beer") opens up other cans of worm: what about the region? Being from Quebec Canada I'm thinking of our beloved ".qc.ca" tld and the like.

@andrewkroh What do you think? Does the public suffix list help fill this country code (aka it's not much more of a burden)? Or does it not help with country code?

If we decide to keep it, we should at least say in the description that it should be filled only when there's a country code :-)

schemas/client.yml
```diff
@@ -59,19 +59,35 @@
...
-        The highest registered client domain, stripped of the subdomain.
+        The highest registered domain, stripped of the subdomain.
 
         For example, the registered domain for "foo.google.com" is "google.com".
 
         This value can be determined precisely with a list like the public
         suffix list (http://publicsuffix.org). Trying to approximate this by
         simply taking the last two labels will not work well for TLDs such as "co.uk".
-      example: google.com
+      example: mydomain.co.uk
```
webmat, line 90: Please keep the example aligned across the file when possible. In this case I agree with using .co.uk to make the point about `registered_domain`, but please adjust the description accordingly as well (line 85)

schemas/client.yml
```diff
@@ -59,19 +59,35 @@
       description: >
         Client domain.
 
+    - name: name
```
webmat, line 62: What is this field's relationship to `client.domain`?

schemas/client.yml
```diff
@@ -87,6 +103,38 @@
...
+      short: The subdomain of the domain.
+      description: >
+        The subdomain is all of the labels under the registered_domain in a partially qualified
+        domain name, and all of the labels except the hostname in a fully qualified domain name.
+        
+        If the domain has multiple levels of subdomain, such as "sub2.sub1.example.com",
+        the subdomain field should contain "sub2.sub1", with no trailing period.
+      example: east
+
+    - name: hostname
```
webmat, line 118: What is this field's relationship to `client.subdomain`?

schemas/dns.yml
```diff
@@ -159,6 +179,66 @@
         be the original `question.name` repeated.
       example: www.google.com
 
+    - name: answers.registered_domain
```
webmat, line 182: I'm not sure I would add all of these fields in the DNS answers array of object.

@MikePaquette Do you think we have a need for that?

@leehinman & @andrewkroh What do you think about implementing this on DNS answers? Sounds to me like adding this here may be problematic, considering the different kinds of DNS answers we can get, or even just for performance reasons: a query for a domain name can yield many CNAMEs as an answer, so worst case this could be multiple public suffix list lookups per DNS answer event. WDYT?

CHANGELOG.next.md
dainperkins: so for future clarification - when I update from source, I should **just** add a line to CHANGELOG.next.md?

schemas/client.yml
dainperkins: I feel like parsing issues shouldn't be a consideration in terms of the availability of accurate sub fields.   I'll update the description to say it should only be filled in when there is a country code.

schemas/client.yml
dainperkins: domain is listed under host as being e.g. the windows domain (e.g. netbios name - "CONTOSO");  name is used elsewhere (dns.question.name) as being specific to the domain name...

Up until digging thru this I had expected to lobby for fqdn, but based on the fact that its next to impossible to determine from information gathered about a dns request if a name is partially or fully qualified, I went with the DNS spec as the baseline for consistency

CHANGELOG.next.md
dainperkins: done - updating PR shortly

CHANGELOG.next.md
dainperkins: done - updating PR shortly

schemas/client.yml
dainperkins: I would consider it to be the first in a multi-level name in a p/fqdn 

Thinking about dns analysis, I might like to be able to analyze the host portion of a name (various features could be useful) as well as the [subdomains] as it stands (host + subdomains)

trying to avoid thinking about features etc., and just thinking about e.g. show me every domain name (requested or accessed) for a particular domain, and bucket by subdomain (assuming first name is a host designation - which even if its a subdomain with an a record, its still effectively a specific host...

CHANGELOG.next.md
webmat: Yes, just add your line and don't touch the rest :-)

schemas/client.yml
andrewkroh: The country code TLDs are a subset of all TLDs. And we already have a `top_level_domain` field so this will be duplicated. If is useful to know that a TLD is a country code then maybe that's a boolean flag like `is_country_code` to indicate the `top_level_domain` is a country.

Similarly I had thought other attributes of the domain name could be flags. Like whether it contains puny codes with `is_internationalized` (https://github.com/elastic/beats/issues/5809#issuecomment-451710649).

dainperkins: @andrewkroh that leaves me wondering 2 things (and I'm serious... not just pondering to be annoying, which I may have been known to do on occasion :) 

1) where do we draw the line for cutting up aggregate things (like a fqdn) into its component parts?

tld is part of a fqdn - and we duplicates it, but cc (and apparently ?region code? thanks for that Quebec!) are also part of a  tld...  should we be cutting a fqdn up into all of its component parts? or just the major ones? what criteria should we use for "major"? 

I can't see visualizing cc against tld, but I could potentially see utilizing cc as a feature, or as an influencer / bucket for analysis (region code should result in the creators immediate relegation to their own personal purgatory as far as I am concerned :) 

2) I love the "is_internationalized" flag idea, and think theres probably some others we could consider as atomic features (digit ratio, number of special characters, entropy, levenstein distance to similar popular domains, subdomain count, subdomain length, etc.) but, well where do we draw the line with that too?

schemas/dns.yml
dainperkins: Looking at a packetbeat dns request there's 1 request for www.microsoft.com, which gets answered with  3 CNAMES and an A record.    (reference below). As the A record is effectively a recursive, automatic request - it seems like the name should be parsed out and treated like a straight up dns.question??

> {
>   "data": "www.microsoft.com-c-3.edgekey.net",
>   "name": "www.microsoft.com",
>   "type": "CNAME",
>   "class": "IN",
>   "ttl": "2726"
> },
> {
>   "data": "www.microsoft.com-c-3.edgekey.net.globalredir.akadns.net",
>   "name": "www.microsoft.com-c-3.edgekey.net",
>   "type": "CNAME",
>   "ttl": "21204",
>   "class": "IN"
> },
> {
>   "data": "e13678.dspb.akamaiedge.net",
>   "name": "www.microsoft.com-c-3.edgekey.net.globalredir.akadns.net",
>   "type": "CNAME",
>   "ttl": "900",
>   "class": "IN"
> },
> {
>   "data": "23.33.58.14",
>   "name": "e13678.dspb.akamaiedge.net",
>   "type": "A",
>   "class": "IN",
>   "ttl": "20"
> }

schemas/client.yml
dainperkins: @webmat - did I answer the question?  I'm not 100% sure I understood your original question (as I looked at the other references)

ebeahan:
@dainperkins Have you had a chance to take a look at the feedback here? We would like to get this added in for 1.6.

dainperkins:
@ebeahan,
hadn't heard again from anyone but just pushed updates including single example domains, pulled cc with and identified tld/reg'd domain as being the appropriate place (e.g. **co.uk**, mydomain.**co.uk**)

Added is_internationalized (boolean) 
 and hopefully fixed the changelog.md

I also pulled the breakdown for answers, I'm not 100% sure thats the right way to go, but its easy enough to add back in.

schemas/client.yml
```diff
@@ -59,34 +59,91 @@
...
       example: co.uk
 
+    - name: subdomain
+      level: extended
+      type: keyword
+      short: The subdomain of the domain.
+      description: >
+        The subdomain portion of a fully qualified domain name includes all of the names except 
+        the host name under the registered_domain.  In a partially qualified domain, or if the 
+        the qualification level of the full name cannot be determined, subdomain contains all of
```
antcodd, line 127: I'm not sure of excluding the "hostname" from the subdomain, or assuming it should populate the hostname field.

What happens when the the "hostname" is not the hostname of the machine as with most publicly facing URLs?

For example mail.google.com is unlikely to be a machine whose actual hostname is "mail". It could be more than one machine load balanced where an individual host responded, or more than one subdomain hosted by the same machine. Is this what ``..or if the qualification level of the full name cannot be determined`` is supposed to cover? There seem to be different schools of thought on whether the leftmost part of a domain counts as a hostname, or always a subdomain, or sometimes one or the other depending on the context that can't be determined from just looking at the url.

This is particularly important in fields set groups outside url and dns (like client and server) where you might be recording the machines actual hostname based on network traffic and not the url, but copying from url.domain to server.domain. If you drop mail from the subdomain but have hostname as the real internal IP based one the information is inconsistent.

schemas/client.yml
```diff
@@ -59,34 +59,91 @@
...
+      description: >
+        Name of the client system.
+
+        It can contain what `hostname` returns on Unix systems, the fully
+        qualified domain name, or a name specified by the user. The sender
+        decides which value to use.
+        
+        In the case of fully and partially qualified domain names it is recommended to parse out 
+        the additional fields of top_level_domain, registered_domain, subdomain, and hostname
+        as they are available.
```
antcodd, line 75: How would whether this is a domain name or a custom name be determined? By whether the name contains a dot?

Perhaps "name" should work like "address" and you copy the best available name from more specific fields, or have a custom user supplied one? It feels somewhat overloaded here, it's a shame domain is used for other purposes in some field set groups.

schemas/client.yml
```diff
@@ -59,34 +59,91 @@
...
+      type: keyword
+      short: Hostname of the system.
+      description: >
+        Hostname of the system.
+
+        It normally contains what the `hostname` command returns on the host machine, or the
+        host portion of a fully qualified domain name.
+        
+        For example, the hostname portion of "www.east.mydomain.co.uk " is "www".
+      example: www
```
antcodd, line 90: I'm not sure www is a good example of a hostname, especially outside of the url field set. See my comments on subdomain.

schemas/client.yml
dainperkins: added is_internationalized, removed country_code and added to the tld description ti indicate its the spot to tld/cctld

schemas/client.yml
dainperkins: I believe everything is now www.east.mydomain.com

schemas/client.yml
dainperkins: Like you said, relates back to the subdomain, but I can go back and replace www with something less obviously a virtual name for the sake of the example, I think that makes sense.

schemas/client.yml
dainperkins: So this one is annoyingly specific to a number of criteria... 


- Locality (local hostname vs. host portion of a fqdn)
- Record Type (CName(s) vs A record)
- Truth (dns "host"/"A Record" that resolves to a load balancer, vs round robin dns load balancing to multiple individual hosts)

I was trying to adjust, with minimal breaks, dns info across multiple fields and I think the valid concerns come down to frame of reference.

Organizationally remote from DNS Server:
in the mail.google.com example - I get a CName to googlemail.l.google.com. which resolves to a single IP (as far as I can tell from 1 nslookup) - so regardless of the backend reality, final dns response looks like host.registered_domain.  (theres also no issue with multiple responses mapping the same name to multiple ips (or vice versa)


I'm thinking maybe pulling name/hostname out of the normalization

schemas/client.yml
dainperkins: This one comes down to processing  - I'm honestly not sure what the results would look like for a partial domain in the logstash tld filter (will test and see)

CHANGELOG.next.md
ebeahan: Looks the conflicts with the `CHANGELOG.next` are still outstanding.

schemas/dns.yml
```diff
@@ -122,12 +123,39 @@
...
+        the names below the registered domain.
+        
+        For example the subdomain portion of "www.east.mydomain.co.uk " is "east".
 
         If the domain has multiple levels of subdomain, such as "sub2.sub1.example.com",
         the subdomain field should contain "sub2.sub1", with no trailing period.
-      example: www
+      example: east
+
+    - name: is_internationalized
```
ebeahan, line 137: I like the idea of `is_internationalized`, and I think Andrew's use case from elastic/beats#5809 makes a lot of sense. Do we have a mechanism that will be able to populate this field today or in the near future?

Just wonder if we should perhaps hold it back until we have a more immediate use. Going along with the internationalized domain name use case, it'd be more useful to a user if the punycode-to-unicode translation was captured as well as setting the `is_internationalized` flag.

dainperkins: found a simple regex to check   ^(.+[-_.])??xn--), contains ".xn--" in the registered / tld would also do it

From a machine learning standpoint I think the unicode translation would definitely be interesting (looks like apple, but its not!) but the is_internationalized flag seems easy enough to implement.

dainperkins:
Per ECS Team discussion - closing this PR and opening 2 new PRs to separate:
1) DNS specifics (related to dns name queries observed or logged)
2) Host specifics (related to internal host naming conventions - logical name, system hostname, dns/directory name)

ebeahan:
@dainperkins Is the plan still to split this into two new PRs as mentioned above? If so, can this PR be closed?

dainperkins:
@ebeahan I think its probably for the best to close this one, and I'll tackle the individual items in a future one.

I do think normalizing the dns domain fields throughout would be a good thing, and probably one of those potentially breaking changes to be considered for 2.0

dainperkins:
- closing to split into multiple, scope limited PRs

PR #800, opened 2020-03-25
Changes search result IDs to be array of objects

joshdevins:
Search results were originally modelled as a list of document IDs. When
implementing query logging in Workplace Search, we realized that the
unique key for a document is actually index plus document ID since
queries can be executed across multiple indices and document IDs alone
are required to be unique only within an index. For metrics tasks, the
ID field could have encoded the index and ID, and most machine learning
tasks such as training data for learning-to-rank could also use this
scheme to keep the data structure simple. On the other side, the rank
evaluation API uses two fields to represent the unique keys of a
document. We are changing this schema primarily to be consistent with
other APIs such as rank evaluation, at the cost of adding some schema
complexity and potentially data structure overhead in raw events.

This PR is a request for comments and a discussion point to decide if
this is the right way to proceed vs the way we have currently implemented
this as an array of plain/opaque IDs.

### DIFF
--- a/code/go/ecs/search.go
+++ b/code/go/ecs/search.go
@@ -55,9 +55,20 @@ type Search struct {
 	// in the query response.
 	ResultsTotal int64 `ecs:"results.total"`
 
-	// A list of opaque document IDs representing the results that were shown
-	// to the user. This is effectively the impression list and it's size
-	// should be equal to `results.size`. This field can be empty when there
-	// are no results to return.
-	ResultsIds string `ecs:"results.ids"`
+	// A list of documents representing the results that were shown to the
+	// user. This is effectively the impression list and it's size should be
+	// equal to `results.size`. This field can be empty when there are no
+	// results to return.
+	ResultsDocument map[string]interface{} `ecs:"results.document"`
+
+	// An opaque document ID representing the result that was shown to the
+	// user. This ID should be unique within the index it was retrieved from.
+	// The combination of document ID and index should be globally unique.
+	ResultsDocumentID string `ecs:"results.document.id"`
+
+	// The name of the index that the result document was retrieved from. This
+	// can be considered as optional if all search results are always retrieved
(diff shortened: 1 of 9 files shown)


joshdevins:
For reference, have a look at the [rank evaluation API](https://www.elastic.co/guide/en/elasticsearch/reference/7.6/search-rank-eval.html).

webmat:
This is a breaking change.

ECS is meant to standardize well understood usage patterns only, not experimental work. Please find a way to capture this new structure via new fields that won't cause a breaking change with your initial request.

webmat:
Ha! Never mind, it was merged a little while ago, but I just realized it was still only in master. So disregard the previous comment about breaking changes.

joshdevins:
Yes, I'm hoping to get some discussion on this before 1.6.0 release to avoid having to fix this potential mistake post-release. Waiting also for comments from @ioanatia and @kovyrin . Maybe @cbuescher also wants to comment as he worked on the rank evaluation API.

ioanatia:
Having `search.results.document` makes sense to us since we can add the Elasticsearch index name and be able to capture result data for multi index searches.
About removing `search.results.ids` I don't have a strong opinion.
If we decide to add `search.results.document` to ECS, then in Workplace Search we will switch to using `search.results.document` instead of `search.results.ids` for our analytics events.

joshdevins:
Thanks @ioanatia . I've discussed it with a couple folks and it seems like this is the best way forward. @webmat any chance to get this merged before 1.6.0 to correct the mistake?

joshdevins:
Due to the need to remove the field set for 1.6.0 release, I'm closing this PR for now. When we re-introduce the field set, we can make this change then.

See: https://github.com/elastic/ecs/pull/812

webmat:
Thanks @joshdevins and sorry for that back & forth

PR #802, opened 2020-03-25
Add `log.file.path` to capture the log file path, following Filebeat convention.

### LINKED ISSUE #770: Introduce log.file.path and friends
Filebeat uses `log.file.path` to describe the file from which a log line was harvested. There’s no such ECS field. [This](https://github.com/elastic/ecs/blob/54f600d76cebcfcf121e1666a2ea8f2c72c6b892/use-cases/logging.md) describes `source.path` which also does not seem to exist in ECS. There’s `file.path` which probably is meant to mean that a specific log line contained a reference to a file, rather than that the log line was read from that file. Would be great to get `log.file.path` into ECS.

webmat:
A few points to note about this PR:

* This is not nesting all of `file.*` at `log.file.*`, this is adding a single field only: `log.file.path`. If I remember correctly, it's the first time we do this.
  * If there's a need for more `file.*` fields we could add them, but so far that need has not arisen.
* This approach also made it possible to tailor the field description to this specific use case.
* Where `file.path` has a `text` multi-field which may be useful for threat hunting in arbitrary file paths, `log.file.path` does not. The thinking is that these log files are controlled by operators and aren't "user data" in which we're likely to do threat hunting. This is something we can add later if needed.
* This PR also adjusts the field description for `log.origin.file.name` to disambiguate: `log.origin.file.*` are fields to capture source code file details, in application logging use cases.

Closes #770

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -10,6 +10,8 @@ Thanks, you're awesome :-) -->
 
 ### Schema Changes
 
+* Added `log.file.path` to capture the log file an event came from. #802
+
 #### Breaking changes
 
 #### Bugfixes
--- a/code/go/ecs/log.go
+++ b/code/go/ecs/log.go
@@ -34,6 +34,11 @@ type Log struct {
 	// Some examples are `warn`, `err`, `i`, `informational`.
 	Level string `ecs:"level"`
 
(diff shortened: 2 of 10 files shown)


webmat:
FYI @roncohen @ruflin

Issue #746, opened 2020-02-13
Document the usage of the ECS generator


webmat:
The ECS generator is accruing features that let users generate their artifacts based on their additional custom fields.

We should document the usage of

* `--include` #494
* `--subset` #737
* `--out`
* `--intermediate-only`

Since this is low level advanced stuff, I think we could document this in the github repo in `generated/README.md`, with a mention in the main readme as well.

webmat:
Until I have time to put together coherent docs, I guess we can still drop something I can point to here.

Beats users should disregard this. Beats includes hundreds of other field definitions that aren't in ECS. Follow Beats docs to add custom fields to Beats.

These tools are experimental and should be used for custom indices only.

**Help**

```
python scripts/generator.py --help
```

**Output**

Generate ECS artifacts in a different directory

```
python scripts/generator.py --out ../myproject/ecs/out/
```

**ECS + Custom fields**

Generate ECS artifacts based on ECS + my custom fields (whole directory of files).

```
# single additional yml file
python scripts/generator.py --include ../myproject/custom-fields/acme.yml
# all yml files in a directory
python scripts/generator.py --include ../myproject/ecs/custom-fields/
```

Check out the [schemas/README.md](https://github.com/elastic/ecs/tree/master/schemas#fields-supported-in-schemasyml) or the YAML files in this directory, for the file format of how to put together these YAML files.

**Pick a subset of ECS**

If your index will never populate some of the ECS fields, no need to have these field defs in your mapping. You can trim it down by creating a YAML file that indicates which field sets, or specific fields to include.

```
python scripts/generator.py --subset ../myproject/ecs/subset.yml
```

The structure of this YAML file should be as follow:

```YAML
base:
  fields: "*"
event:
  fields: "*"
host:
  fields:
    name:
      fields: "*"
```

The above means:

* All `base` fields
* All `event.*` fields
* Only `host.name`, out of the `host.*` field set.

**Complete example**

Take only the ECS fields as described in `subset.yml`, add custom fields from `acme.yml`, and output all generated artifacts to `../myproject/ecs/out/`.

```
python scripts/generator.py \
  --include ../myproject/custom-fields/acme.yml
  --subset ../myproject/ecs/subset.yml
  --out ../myproject/ecs/out/
```

Caveats:

* This also generates documentation, which may be useful. But it doesn't forcefully set field levels to "custom". So this may be misleading to end users.
[...]

PR #803, opened 2020-03-25
Handle nestings better and refactor asciidoc generation

### LINKED ISSUE #796: Make "nestings" in ecs_flat.yml support deep nesting/reuse better
Since we've added the ability to nest field sets deeper within the field hierarchy, the "nestings" array in ecs_flat.yml is now misleading, as it doesn't fully capture what is nested where (more specifically how deep).

This has led to this bug creeping into 1.5.0: https://github.com/elastic/ecs/pull/784, where for example the nesting of `interface` under observer was listed as `observer.interface.*` instead of the intended `observer.ingress.interface.*` and `observer.egress.interface.*`.

This bug has not caused problems with the generated Beats field definitions, csv nor the sample Elasticsearch templates, as none of these rely on the "nestings" array.

marshallmain:
Closes #796 

A few related fixes in this PR:
- `nestings` array now contains the full path to reused fields within a fieldset, rather than just containing the name of the reused fieldset.  This better identifies places where a fieldset is reused multiple times within another set.
- Moved the `original_fieldset` logic from into the function where reused fields are converted from references to copies.  This makes the ecs_flat and ecs_nested files more consistent - `original_fieldset` was showing up in reusable fields even at the top level in ecs_flat.  However, it was supposed to be identifying places where a field was actually being reused.
- Refactor ascii doc generation to use intermediate field data structure instead of both the flat and nested structures.  This makes the logic easier to follow imo by reducing the number of transformations the data goes through.

### DIFF
--- a/generated/ecs/ecs_flat.yml
+++ b/generated/ecs/ecs_flat.yml
@@ -102,7 +102,6 @@ as.number:
   name: number
   normalize: []
   order: 0
-  original_fieldset: as
   short: Unique number allocated to the autonomous system. The autonomous system number
     (ASN) uniquely identifies each network on the Internet.
   type: long
@@ -121,7 +120,6 @@ as.organization.name:
   name: organization.name
   normalize: []
   order: 1
-  original_fieldset: as
   short: Organization name.
   type: keyword
 client.address:
@@ -636,7 +634,6 @@ code_signature.exists:
   name: exists
   normalize: []
   order: 0
-  original_fieldset: code_signature
   short: Boolean to capture if a signature is present.
   type: boolean
 code_signature.status:
@@ -653,7 +650,6 @@ code_signature.status:
   name: status
   normalize: []
   order: 4
-  original_fieldset: code_signature
   short: Additional information about the certificate status.
   type: keyword
 code_signature.subject_name:
@@ -666,7 +662,6 @@ code_signature.subject_name:
   name: subject_name
   normalize: []
   order: 1
-  original_fieldset: code_signature
   short: Subject name of the code signer
   type: keyword
 code_signature.trusted:
@@ -681,7 +676,6 @@ code_signature.trusted:
   name: trusted
   normalize: []
   order: 3
-  original_fieldset: code_signature
   short: Stores the trust status of the certificate chain.
   type: boolean
 code_signature.valid:
@@ -696,7 +690,6 @@ code_signature.valid:
   name: valid
   normalize: []
   order: 2
-  original_fieldset: code_signature
   short: Boolean to capture if the digital signature is verified against the binary
     content.
   type: boolean
@@ -3021,7 +3014,6 @@ geo.city_name:
   name: city_name
   normalize: []
   order: 4
-  original_fieldset: geo
   short: City name.
   type: keyword
 geo.continent_name:
@@ -3034,7 +3026,6 @@ geo.continent_name:
   name: continent_name
   normalize: []
   order: 1
-  original_fieldset: geo
   short: Name of the continent.
   type: keyword
 geo.country_iso_code:
@@ -3047,7 +3038,6 @@ geo.country_iso_code:
   name: country_iso_code
   normalize: []
   order: 5
-  original_fieldset: geo
   short: Country ISO code.
   type: keyword
 geo.country_name:
@@ -3060,7 +3050,6 @@ geo.country_name:
   name: country_name
   normalize: []
   order: 2
-  original_fieldset: geo
   short: Country name.
   type: keyword
 geo.location:
@@ -3072,7 +3061,6 @@ geo.location:
   name: location
   normalize: []
   order: 0
-  original_fieldset: geo
   short: Longitude and latitude.
   type: geo_point
 geo.name:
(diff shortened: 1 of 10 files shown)


webmat:

generated/ecs/ecs_nested.yml
```diff
@@ -1351,9 +1351,9 @@ destination:
   group: 2
   name: destination
   nestings:
-  - as
-  - geo
-  - user
+  - destination.as
+  - destination.geo
+  - destination.user
```
webmat: We may not want to address this in this PR, as this is already a great improvement. But here we're also missing `destination.user.group`.

This PR improves the situation when a field set gets nested deeper directly, like `vlan` => `network.inner.vlan`. But it doesn't yet address a nesting within another nesting like `group` => `user.group`, then `user` => `destination.user`...

Not sure if we need to address that for now though. Concretely, this array drives things like the "field reuse" section at the bottom [here](http://ecs_803.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-destination.html#_field_reuse_4), and perhaps simply saying that "user" is nested here is enough.

WDYT?

scripts/generator.py
```diff
@@ -58,7 +58,7 @@ def main():
     csv_generator.generate(flat, ecs_version, out_dir)
     es_template.generate(flat, ecs_version, out_dir)
     beats.generate(nested, ecs_version, out_dir)
-    asciidoc_fields.generate(nested, flat, ecs_version, docs_dir)
+    asciidoc_fields.generate(intermediate_fields, ecs_version, docs_dir)
```
webmat: Why are you switching to passing `intermediate_fields` here?

The goal of the intermediate files (nested and flat) is to make the two structures possible to visualize, when passing it around to the other generators, and also for other external projects to have either file to start from, to generate other artifacts.

I'm all for no longer passing both "nested" and "flat" to asciidoc and passing just one instead, that was temporary a hack :-) But could we improve "nested" with what you're missing, instead of adding a third structure that we're passing around?

generated/ecs/ecs_flat.yml
```diff
@@ -102,7 +102,6 @@ as.number:
   name: number
   normalize: []
   order: 0
-  original_fieldset: as
```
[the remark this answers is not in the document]
marshallmain: Yeah those top level `original_fieldset` keys were added recently as a side effect of https://github.com/elastic/ecs/commit/68afbb82c721206c6473cce4f3ff6186f713ed93 this commit.  I think removing them now is the right move, as nothing should depend on them yet and we probably don't want things to become dependent on them.

generated/ecs/ecs_nested.yml
marshallmain: I like the idea of explicitly listing out all the reused fields so the links to all the reusables in a fieldset are in one place.  I'll add that change in.

scripts/generator.py
marshallmain: Nested and flat are both missing the top level reused fields - i.e. `hash` is reused in `dll` so you can find `dll.hash.md5`, `dll.hash.sha1`, etc but `dll.hash` is not there.  To render the ascii doc reusable sections we look for the short description of the reusable fieldset, but in this example since `dll.hash` isn't in nested or flat we look for `hash` at the top level instead and get the short description there.  Unfortunately, this means if we are using a subset that includes `dll.hash` but doesn't include `hash` at the top level then ascii doc generation errors out.

I'm now realizing that in that example the link to the `hash` fieldset wouldn't work since those fields wouldn't be in the docs at the top level so the ascii docs are going to be broken in some cases with subsets anyway.  I'm not sure how high priority it is to get ascii docs fully working with subsets.

marshallmain: The other reason I switched is that intermediate_fields is already being used as a data structure to generate the nested and flat structures.  My thought is that it will be simpler to use intermediate_fields as the input to all the generators since it has all the information and we can remove the dependency of the other generators on the nested and flat generator.

webmat: In my mind, the asciidoc output when using --include or --subset is fraught with pitfalls and should be ignored.

E.g. in the case of --include, I think we should force set the field level to "custom" for everything that was included. Right now it takes whatever's in the included files as is. This has the potential to be very confusing for end users looking at custom docs.

The way I currently envision the output when using --include and/or --subset is that only the following files make sense to use without huge caveats or adjustments:

* the intermediate YAML files
* the Elasticsearch template
* the CSV

I should probably edit [this comment](https://github.com/elastic/ecs/issues/746#issuecomment-604013230) yet again, with these caveats.

webmat: Sorry I got distracted for a moment.

IMO, fixing asciidoc when using --subset is not a priority :-)

marshallmain: I think that makes sense, excluding the ascii docs when `--subset` is used makes it simpler.

I'd still like to use `intermediate_fields` as the input though so we can move towards using it as the only internal intermediate data structure.  Right now we essentially have 3 internal structures: `intermediate_fields`, `nested`, and `flat`, but `nested` and `flat` are generated from `intermediate_fields` and exposed in the generated docs as well.  In the overall code structure I think it's simpler to render the asciidocs directly from intermediate_fields rather than convert intermediate_fields to the nested/flat format and render the asciidocs from those - that additional flattening step adds complexity.  We can still keep `nested` and `flat` but present them as output docs that are stable to external users.

PR #805, opened 2020-03-25
Shorthand subset notation

marshallmain:
This PR makes the subset notation less cumbersome by including all subfields if the `fields` key is missing.  Instead of writing 
```
field1:
    fields: *
```
every time you want to include all subfields, you can instead write 
```
field1: {}
```
This still leaves the ability to specify other options for the field in the dictionary later on without breaking changes.

### DIFF
--- a/scripts/generators/ecs_helpers.py
+++ b/scripts/generators/ecs_helpers.py
@@ -53,15 +53,24 @@ def safe_merge_dicts(a, b):
 def fields_subset(subset, fields):
     retained_fields = {}
     for key, val in subset.items():
-        # Every field must have a 'fields' key or the schema is invalid
-        if isinstance(val['fields'], dict):
+        # A missing fields key is shorthand for including all subfields
+        if 'fields' not in val or val['fields'] == '*':
+            retained_fields[key] = fields[key]
+        elif isinstance(val['fields'], dict):
             # Copy the full field over so we get all the options, then replace the 'fields' with the right subset
             retained_fields[key] = fields[key]
             retained_fields[key]['fields'] = fields_subset(val['fields'], fields[key]['fields'])
-        elif val['fields'] == '*':
-            retained_fields[key] = fields[key]
     return retained_fields
 
+
+def get_reusable_fields(fields):
+    reusable_fields = {}
+    for key in fields:
+        if 'reusable' in fields[key]:
+            reusable_fields[key] = True
+        if 'fields' in fields[key]:
+            reusable_fields.update(get_reusable_fields(fields[key]['fields']))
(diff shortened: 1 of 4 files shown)


scripts/generators/ecs_helpers.py
```diff
@@ -53,24 +53,24 @@ def safe_merge_dicts(a, b):
 def fields_subset(subset, fields):
     retained_fields = {}
     for key, val in subset.items():
-        # Every field must have a 'fields' key or the schema is invalid
-        if isinstance(val['fields'], dict):
+        # A missing fields key is shorthand for including all subfields
+        if 'fields' not in val or val['fields'] == '*':
```
webmat: Here I'd make the check a little more specific, to fail more explicitly if someone makes a typo.

If it's specifically and empty dictionary (rather than the absence of the key "fields"), then it's the shorthand to grab all subkeys.

Otherwise the following silently gets me all of `host` instead of just `host.name`:

```YAML
base:
  fields: "*"
host:
  feels:
    name:
      fields: "*"
```

I think realistically, users will use this with many field sets (10-15?), so a subtle typo like that could go unnoticed.

marshallmain: That's a good point, how about a list of allowed keys to safeguard against that case instead?  I haven't gotten around to it yet but I'd like to allow `index` as an option for each field and a list of allowed keys would let you do something like
```
host:
  index: false
```
to grab all the host fields with the shorthand notation while also excluding the potential mistyped fields.

webmat: Something more strict like a whitelist of allowed keys in that "subset" structure sounds good, yeah.

I'm not sure I follow what `index` does here, though. Can you clarify?

marshallmain: Sorry I don't know how I missed this, I thought I replied.

`index` here would correspond to the `index` property in the elasticsearch mapping. So this would let us specify fields from the schema that we want to exist in documents but don't need to be indexed. This particular use case may or may not be useful but it's something to consider.

webmat: When saving a document in Elasticsearch, the raw document always remain complete and untouched.

The behaviour of `index:false` on field `myfield` simply does not create a searchable mapping entry for `myfield`. But if you were to read a document that contains `myfield`, it would be present in the `_source` of the doc.

webmat: I think I misread your last comment. Sorry for explaining essentially what you just said 😂 

Let's keep the whitelist to only accept "fields" for now, if you don't mind.

Issue #809, opened 2020-03-27
[ECS] Multiple users in an event proposal


webmat:
After reviewing proposals and discussions, new and old, private and public (#117, #234, #589, #678, elastic/beats#9963, elastic/beats#10111, elastic/beats#10192), here's a high level proposal. I'd like to get feedback on this.

When `user.*` is populated at the root of an event, it's meant to capture the user performing an action, or the only user relevant to an event, when that's the situation.

The goal of this proposal is to solve the following use cases:

1. User A creates or deletes User B (no prior or post state)
2. User A modifies User B (there's a prior and a post state)
3. User A assumes the identity of User B (or fails to)

One thing to note, is that with real-time privilege escalation (e.g. sudo, MacOS modal) or auditing frameworks like auditd, it's useful to be able to identify all 3 of these additional users concurrently, in a single event.

Here are locations I propose we could nest the user schema, to represent various use cases:

* user.effective.*
* user.target.*
* user.new.*

Note that in each of these cases, we are nesting the `user` schema as a new name, similar to what we do with `process.parent`. To give a concrete example the effective user name would be at `user.effective.name` (not at `user.effective.user.name`).

### user.effective.*
 
Used to identify the final identity, when logging in remotely as another user, or escalating privileges.

Simplified privilege escalation example, where "alice" escalates privileges to "root":

```JSON
{
  "user": {
    "name": "alice",
    "effective": {
      "name": "root"
    }
  }
}
```

### user.target.*

Used to identify the user being modified, or targeted by an action.

Simplified user management example, where "root" modifies user "bob":
```JSON
{
  "user": {
    "name": "root",
    "target": {
      "name": "bob"
    }
  }
}
```

### user.new.*

Used to identify what was changed about the user. Ideally this would only capture the difference, not repeat all values.

Simplified user management example, when an existing user is modified, for example changing user name. Here user "bob" is renamed "foo" by "root":
```JSON
{
  "user": {
    "name": "root",
    "target": {
      "name": "bob"
    },
    "new": {
      "name": "foo"
    }
  }
}

```

Note: The creation of a new user would only use `user.target`, not `user.new`. `user.new` should only be used.

### Putting it all together

If it's possible to determine who was the original user prior to privilege escalation, a full event could look like this:

```JSON
{
  "user": {
    "name": "alice",
    "effective": {
      "name": "root"
    },
    "target": {
      "name": "bob"
    }
  }
}
```

The action being observed should be captured in `event.action`.

[...]

PR #811, opened 2020-04-08
Refactor schema_reader.py

webmat:
This is a small cleanup to clarify schema_reader.py. It would need more <3 but
that's a start.

In this PR:

* No changes to the generated artifacts, except for the removal of the `order` attribute.
  This attribute was initially meant to capture and reuse the order in which
  fields are defined in the source yaml file. This was never used. Now it's gone.
* Reorder the code in schema_reader.py to make it easier to follow the logic.
* Move some helper functions to ecs_helpers.py
* Add a few doc strings
* Replace a helper function by proper use of Python, which I learned from @marshallmain's contributions. Thanks Marshall :-)

I'll continue work on #809 on top of this. But I want to separate the work a bit,
since significant underlying improvements to the tooling are still needed, to deliver #809.


marshallmain:
Looks good to me except the ecs_flat.yml and ecs_nested.yml files got deleted, was that intentional?

webmat:
Whoa you had me scared, there for a second 😂

Thanks

Issue #495, opened 2019-07-02
Honor reusable.top_level: false for generated ES templates


andrewkroh:
When setting `reusable.top_level: false` the generated Elasticsearch template still includes the field in the template at the top level. I would not expect this field to be included at the top level of the template when this is `false`.

webmat:
Most "reusable" field sets in ECS are only meant to be used in their nested location (e.g. `source.geo`, `destination.geo`, but not `geo` at the root of a document).

The documentation states this fact correctly, but many of the other generated artifacts still incorrectly declared the fields at the top level. As an example, the sample Elasticsearch templates do have field definitions for `geo` at the root of the document.

This issue actually affects the following artifacts:

* Beats file for ECS field definitions
* The sample Elasticsearch templates
* The CSV file
* The intermediate file `ecs_flat.yml`

I will be addressing this bug shortly for the last 3 artifacts.

In Beats, removing these fields may be seen as a breaking change. @andrewkroh or @leehinman could you let me know what your preference is, here? Happy to remove them in the Beats file, if you're confident that they were never used. For now I will leave them in, to avoid any surprises, but let me know if / when you want them removed, and I'll make it so, in this PR or another.

To complete the picture, here's a list of reusable field sets, and whether they are expected at the root of an ECS document (`top_level: true/false` in the YAML), as of ECS 1.5.0:

| reusable field set | expected at root |
| --- | --- |
| `as` | no |
| `code_signature` | no |
| `geo` | no |
| `group` | yes |
| `hash` | no |
| `interface` | no |
| `os` | no |
| `pe` | no |
| `user` | yes |
| `vlan` | no |

andrewkroh:
I can't guarantee they are not used, but I'm not aware of their usage in Beats. When they are removed we'll run automated tests that check that fields used in modules are in the template. We could run that test before we finalize the change in ECS (really just need the generated ECS fields.yml to drop into a beats PR to have the tests run).

webmat:
Ok thanks. I'll still do these as separate PRs, I think, to avoid slowing down #648 

But happy to provide you with the resulting Beats file today and get that merged quickly, if we're good on the Beats side.

webmat:
I haven't opened a PR yet for it. Happy to fold into #813 if this works well. But you can get the Beats file to test this from this branch [fix-495-beats](https://github.com/webmat/ecs/tree/fix-495-beats).

[...]

PR #813, opened 2020-04-09
Fix bug #495

webmat:
Most reusable field sets in ECS are **not** expected at the root of the documents. As of ECS 1.5.0:

| reusable field set | expected at root |
| --- | --- |
| `as` | no |
| `code_signature` | no |
| `geo` | no |
| `group` | yes |
| `hash` | no |
| `interface` | no |
| `os` | no |
| `pe` | no |
| `user` | yes |
| `vlan` | no |

Unfortunately, ever since the introduction of reusable fields, these field sets remained defined at the root nonetheless, for some of the generated artifacts.

The official documentation correctly mentions which field sets were expected at the root and which weren't. This issue however affected the following artifacts:

* 'ecs_flat.yml'
* the csv
* the Elasticsearch sample templates
* Beats field definition file

This PR does not affect the Beats field definitions for now. Even if they weren't meant to be defined there, we first need confirmation whether they've actually been used.

This PR only fixes the issue for the first 3 artifacts.

Note about 'ecs_nested.yml': this file mixes two types of information. Information about field sets, and information about field definitions. Consumers of this file must make sure to check the attribute `reusable.top_level`. You should skip defining fields at the root for any field sets where `reusable.top_level: false`.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -36,6 +36,9 @@ Thanks, you're awesome :-) -->
 
 #### Breaking changes
 
+* Removed field definitions at the root of documents for fieldsets that
+  had `reusable.top_level:false`. This PR affects `ecs_flat.yml`, the csv file
+  and the sample Elasticsearch templates. #495, #813
 * Removed the `order` attribute from the `ecs_nested.yml` and `ecs_flat.yml` files. #811
 
 #### Bugfixes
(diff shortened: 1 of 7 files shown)


webmat:
For awareness @elastic/ecs-logging. Please make sure to review the note at the end of the PR body.

webmat:
Whoa, just found a subtle bug where our Makefile's empty include would mean we skip the generation of the Beats file and the docs. Thankfully this issue was only introduced recently and didn't affect us.

Fixed in 6382d266 :)

PR #814, opened 2020-04-14
Allowing beats output files to be generated with --include and --subset

jonathan-buttner:
The beats output files are used to create the `fields.yml` mapping file for the endpoint package's mapping located here: https://github.com/elastic/package-registry/tree/master/dev/packages/example/endpoint-1.0.0/dataset/events/fields

The endpoint team needs the ability to generate the beats output while using the `--include` and `--subset` flags because we specify custom fields for our alerts and also leverage the subset functionality to limit the fields from ecs core.

The typical command I use to generate the `fields.yml` file is:
`python scripts/generator.py --out ../gen --include ../endpoint-app-team/custom_schemas --subset ../endpoint-app-team/custom_subsets/elastic_endpoint/events/* ../endpoint-app-team/custom_subsets/*.yml`

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -49,6 +49,7 @@ Thanks, you're awesome :-) -->
   slightly different results in different situations. #782
 * Fix incorrect listing of where field sets are nested in asciidoc,
   when they are nested deep. #784
+* Allow beats output to be generated when using `--include` or `--subset` flags. #814
 
 #### Added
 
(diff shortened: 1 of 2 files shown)


jonathan-buttner:
> Worth noting that the beats generator is hard coded to look for the base fields so if they're not in the subset it will crash in the `beats.generate()` call.

Just so I remember in the future, I think the reason that the `beats` generator succeeds is because all of the endpoint events use at least the `@timestamp` field from base.

Issue #411, opened 2019-04-04
Missing a schema for email


krisATelastic:
Hey team, just working through the schema and there doesn't appear to be a defined schema for email in order to normalise logs from sources like sendmail/exim or devices like a Cisco ESA, etc...

Will something like this be coming in later cuts or would custom fields need to be utilised for this (i.e. create your own).

zlammers:
Agreed, there should be email-specific fields.  In our current schema, I had an email object with these fields:

```
email.source.source
email.source.user
email.source.domain
email.target.target
email.target.user
email.target.domain
email.subject
email.cc
email.bcc
email.content_type
email.reply_to
```

Though I came up with the above, I was never 100% happy with it due to:

1. The schema above assumes single target, which obviously is no good.  Perhaps just simply retaining arrays of all targets, .cc's, and .bcc's is the better method.
2. The repitition of email.source.source and email.target.target -- I should have renamed them to ...email.source.original and email.target.original -- as the (user|domain) fields are created by Logstash off the email seen.

We would keyword all of them, except for subject, which we would retain both keyword and text versions.

krisATelastic:
It's also handy to index things like the message Id, as if multiple SMTP relays are involved, you can also track the email flow as it comes in through say an SMTP proxy, right the way back to Exchange for example.

@MikePaquette what do you think on the above, I can see how we can structure the email addresses themselves, but do we have any guidelines for an email-type schema by chance?

webmat:
Hey folks, I'd say email will definitely be coming at some point, although it hasn't been at the top of our priority list so far. So direct email support in the schema is not imminent either.

My recommendation is therefore to use custom fields for now. Providing expectations / desires / field ideas like @zlammers did above is extremely helpful in the meantime :-) This will help us come up solid out of the box support, when we get around to it.

Here are is another set of fields that was suggested elsewhere, around supporting emails in ECS. I'm recording it for future usage, but haven't had time to investigate more yet:

```
email.to 
email.from 
email.protocol 
email.sbrs 
email.cipher 
email.response 
email.mid 
email.icid 
email.dcid 
email.rid
```

weichea:
+1. Is this being worked on? 

For the to and from field, should we use source.user.email and destination.user.email instead?

anhlqn:
[...]

PR #815, opened 2020-04-17
Initial discussion for email object

dainok:
I would like to start a discussion about adding an object for emails or not.
MTA logs could fit in events, but a more detailed log need a specific object.

Questions:
- Makes sense for you? How do you translate MTA logs to ECS schema?
- Can I define an array of objects? Examples: "to" is an array of dicts (address, full name), and attachments would be an array of files.
- Suggested fields to add?

Feel free to comment.

### DIFF
--- a/schemas/email.yml
+++ b/schemas/email.yml
@@ -1,0 +1,58 @@
+---
+- name: email
+  group: 2
+  title: Email
+  short: Fields describing emails.
+  description: >
+     A file is defined as a set of information that has been created on, or has existed on a filesystem.
+
+     File objects can be associated with host events, network events,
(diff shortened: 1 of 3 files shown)


ebeahan:
Adding email support into the schema is something many users have voiced interest in, and there's been a lot of good discussion in #411.  I'd recommend also sharing any thoughts/expectations/ideas to add to the discussion there as well as those comments are extremely useful!

An addition of this scope would definitely be a great candidate for the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md) we've recently established.

dainok:
Moving to https://github.com/elastic/ecs/issues/411

PR #816, opened 2020-04-17
Standardize additional cloud metadata

graphaelli:
Adds:

* account.name
* project.id
* project.name

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -22,6 +22,7 @@ Thanks, you're awesome :-) -->
 
 * Add architecture and imphash for PE field set. (#763)
 * Added `agent.build.*` for extended agent version information. (#764)
+* Added more account and project cloud metadata. (#816)
 
 #### Improvements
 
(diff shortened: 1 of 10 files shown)


graphaelli:
This does not impact Infra UIs - `project.id` is already supplied by the `add_cloud_metadata` processor in beats, `account.id` is supplied the metricbeat aws module, `project.name` is new.

urso:
+1 on adding project/account information to the cloud namespace.

@kvch @kaiyan-sheng @narph You've been working with Cloud (AWS, Azure, GCP) to some extend. Any other meta-data in mind we would like to add?

webmat:
Thanks @urso. What I'm looking for here is actually any concerns from getting this data from people that know about the implementation of the meta data collector in Beats.

Although if additional fields could be useful, additional PRs are always welcome :-)

kvch:
@webmat It would be great to have appropriate field names for such services. For example, I use `resource.*` fields to identify and describe the states of the AWS/GCP resources from the event. But I understand, it is outside of the scope of this PR.

webmat:
@urso @kvch Thanks for the feedback. So there's no issues in getting this information from the metadata collector?

urso:
I don't think so.

webmat:
Oh wait, getting carried away. Did you want to wait on @simitt's review before I merge, Gil?

graphaelli:
@webmat Changelog pushed, @simitt is on board

PR #818, opened 2020-04-17
Add a protobuf generator to ECS

quixoten:
Not sure if this is useful for the general ECS audience, but I thought I'd try contributing it back just in case. This adds a protobuf generator to elastic common schema. The two major challenges I'm aware of, and how I tried to solve them, are detailed below:

 1. Protobuf does not allow fields to be removed.

    To deal with this, I cache all the generated protobuf fields to a yaml file. Upon next run, if a field is in the cache, but no longer in the schema, it is marked as deprecated in the protobuf definition. If I deleted the `agent.id` field, for example, here's how the protobuf definition changes:

    ```diff
    diff --git a/generated/protobuf/ecs.proto b/generated/protobuf/ecs.proto
    index 3b75872..2136522 100644
    --- a/generated/protobuf/ecs.proto
    +++ b/generated/protobuf/ecs.proto
    @@ -10,7 +10,7 @@ message Document {

         Build build = 1;
         string ephemeral_id = 2;
    -    string id = 3;
    +    string id = 3 [deprecated=true];
         string name = 4;
         string type = 5;
         string version = 6;
    ```

 2. Protobuf does not allow fields to change types.

    There are some exceptions (`int32` to `uint32`), and I've accounted for those, but other type changes are invalid in protobuf. For example, changing the `agent.id` field from `string` to `int32` is invalid in protobuf. Here's how I deal with that:

    The original `string` version of `agent.id` is renamed to `agent.deprecated_id` in the protobuf object and keeps its original protobuf tag. A new protobuf tag is allocated for the `int32` version of `agent.id`.

    ```diff
    diff --git a/generated/protobuf/ecs.proto b/generated/protobuf/ecs.proto
    index 3b75872..a9b731c 100644
    --- a/generated/protobuf/ecs.proto
    +++ b/generated/protobuf/ecs.proto
    @@ -10,10 +10,11 @@ message Document {
 
         Build build = 1;
         string ephemeral_id = 2;
    -    string id = 3;
    +    string deprecated_id = 3 [deprecated=true];
         string name = 4;
         string type = 5;
         string version = 6;
    +    int32 id = 7;
       }
 
       message Client {
    ```

Since protobuf also exports extensions, it might be useful to allocate a block of tags in each message for that purpose, but I haven't added logic for that yet.

For active development, the generated protobuf object will likely get rather messy as fields are added, removed, and updated, but sprawl could be kept to a minimum by regenerating at release time with the cached fields from the previous release.

### DIFF
--- a/scripts/generators/protobuf.py
+++ b/scripts/generators/protobuf.py
@@ -61,23 +61,26 @@ def parse_flat_fields(flat_fields):
         name = segments.pop()
         scope = '.'.join(segments)
 
-        if len(segments) > 0:
+        while len(segments) > 0:
+            parent_key = '.'.join(segments)
             parent_name = segments.pop()
-            fields.setdefault(scope, {
-                'key': scope,
+            parent_scope = '.'.join(segments)
+
+            fields[parent_key] = {
+                'key': parent_key,
                 'name': parent_name,
-                'scope': '.'.join(segments),
+                'scope': parent_scope,
                 'type': camelize(parent_name),
                 'deprecated': False,
-            })
+            }
 
-        fields[key] = {
+        fields.setdefault(key, {
             'key': key,
             'name': name,
             'scope': scope,
             'type': field_type_to_protobuf_type[flat_field['type']],
             'deprecated': False,
-        }
+        })
 
     return fields
 
--- a/scripts/generators/protobuf.py
+++ b/scripts/generators/protobuf.py
@@ -41,8 +41,8 @@ protobuf_legal_type_changes_for = {
 
 
 valid_normalize_settings = [
- [],
- ['array'],
+    [],
+    ['array'],
 ]
 
 
--- a/scripts/generators/protobuf.py
+++ b/scripts/generators/protobuf.py
@@ -40,6 +40,12 @@ protobuf_legal_type_changes_for = {
 }
 
 
+valid_normalize_settings = [
+ [],
+ ['array'],
+]
+
+
 def generate(ecs_flat, ecs_version, out_dir):
     current_fields = parse_flat_fields(ecs_flat)
     cached_file = join(out_dir, 'protobuf', 'cached-fields.yaml')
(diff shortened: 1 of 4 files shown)


scripts/generators/protobuf.py
```diff
@@ -0,0 +1,201 @@
...
+
+from os.path import join, exists
+from generators import ecs_helpers
+
+
+field_type_to_protobuf_type = {
+    'binary': 'string',
+    'boolean': 'bool',
+    'byte': 'int32',
+    'data_nanos': 'long',
```
webmat, line 12:
```suggestion
    'date_nanos': 'long',
```

? :-)

scripts/generators/protobuf.py
```diff
@@ -0,0 +1,201 @@
...
+    'date': 'string',
+    'double': 'double',
+    'float': 'float',
+    'geo_point': 'string',
+    'half_float': 'float',
+    'integer': 'int32',
+    'ip': 'string',
+    'keyword': 'string',
+    'long': 'int64',
+    'object': 'map <string, string>',
```
webmat, line 22: Does this support multiple of object nestings?

quixoten: It doesn't. Currently (once i fixed a bug) there are only two fields in the generated protobuf definition that use `map <string, string>`: `labels`, and `container.labels`. I think the lack of nesting will work for those, but it could potentially be an issue.

quixoten:
> how does a transition like a field becoming an array need to be handled in a Protobuf schema?

I think that could be done with a similar strategy as type changes, for example:

```
string blargh = 1;
```

becomes

```
string deprecated_blargh = 1 [deprecated=true];
repeated string blargh = 2;
```

I didn't realize arrays were being defined through the `normalize` field, so I'll have to figure this out sooner rather than later.

---

> do you foresee other potential impedance mismatches between Elasticsearch semantics and Protobuf?

If ECS started using some of the more complex ES types, they could be challenging to map to protobuf, though not impossible. The protobuf definition could get rather bloated over time from fields being removed, or types changing. I think reserved fields could be a way to mitigate that though, and might be a better strategy than I'm currently using for those changes. Looking at normalize, I see that it's an array, but so far nothing is using more than one normalize value. I'm not sure what it means if there is more than one normalize value, and how it would effect protobuf mappings.

webmat:
You're right that as of now, `normalize` only ever contains `array`. But this is a place where we may eventually add other things, like `lowercase` or other.

quixoten:
going to move this [here](https://github.com/quixoten/ecs-protobuf) for now.

PR #820, opened 2020-04-20
Add flexibility to the field set reuse mechanism

webmat:
New possibilities:

* nest as: lets us nest a field set as another name.
* self-nesting: lets us nest a field set within itself.

Most of the time, self-nesting and "nest as" will be used together. Nothing currently prevents the nesting of a fieldset elsewhere as another name. But this should be used with caution, as it could lead to confusion.

Examples:

* Nesting `process` at `process.parent`
* Nesting `user` at `user.target` and the other locations discussed in #809.

It's important to highlight one detail about self-nesting. Field sets like `user` that are also reused within other field sets (e.g. `destination.user`) and that will soon also have self-nesting (e.g. `user.target`, `user.effective`) will **not** carry the self-nesting in all other places (there won't be a `destination.user.target`).

In addition to improving the reuse mechanism, this PR replaces all of the `process.parent.*` manually duplicated fields. The process of removing this manual duplication exposed a few problems with the prior manual duplication, which are fixed by this PR:

* Small differences had crept up between the `process.*` and `process.parent.*` fields.
* The `pe` field set was nested `process.pe` but not `process.parent.pe`.
* Not an issue per se, but since `process.parent.*` fields were all explicit before, they were all explicitly listed in the asciidoc field listing. They are now indicated only via the normal "field reuse" section at the bottom of the page.

This PR does not address #809. This will be done as a separate PR.

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -6644,7 +6644,7 @@ example: `outside`
 
 ==== Field Reuse
 
-The `vlan` fields are expected to be nested at: `network.vlan`, `network.inner.vlan`, `observer.egress.vlan`, `observer.ingress.vlan`.
+The `vlan` fields are expected to be nested at: `network.inner.vlan`, `network.vlan`, `observer.egress.vlan`, `observer.ingress.vlan`.
 
 Note also that the `vlan` fields are not expected to be used directly at the top level.
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -4402,7 +4402,7 @@ Note also that the `process` fields may be used directly at the top level.
 // ===============================================================
 
 
-| <<ecs-parent,process.parent.*>>
+| <<ecs-process,process.parent.*>>
 | These fields contain information about a process.
 
 // ===============================================================
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -657,7 +657,7 @@ example: `true`
 
 ==== Field Reuse
 
-The `code_signature` fields are expected to be nested at: `dll.code_signature`, `file.code_signature`, `process.code_signature`, `process.parent.code_signature`.
+The `code_signature` fields are expected to be nested at: `dll.code_signature`, `file.code_signature`, `process.code_signature`.
 
 Note also that the `code_signature` fields are not expected to be used directly at the top level.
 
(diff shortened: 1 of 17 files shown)


webmat:
Ah I hadn't noticed the broken asciidoc links that cause docs build to fail. I'll fix that shortly.

The rest can already be reviewed.

webmat:
Ok this is ready for a review.

docs/field-details.asciidoc
```diff
@@ -4630,6 +4373,10 @@ example: `/home/alice`
 
 ==== Field Reuse
 
+The `process` fields are expected to be nested at: `process.parent`.
```
rw-access, line 4376: Have you thought of ways to explain what each nesting means? I think we just copy and paste fieldsets over, but I don't know if we explain what the context of each nesting should mean.

webmat: The new structure for how we specify field reuse will easily allow for that:

```YAML
  reusable:
    top_level: true
    expected:
      - at: process
        as: parent
        description: The process that gave birth to the one detailed at the root of the event.
```

And I think it's very much needed. However I think we should tackle this independently, as this will be a significant endeavour to add all of these contextual definitions and redesign the docs section about field reuse.

webmat:
@rw-access I pinged you on this PR in part because we had forgotten to nest `pe` at `process.parent.pe`. This PR will fix this.

scripts/schema_reader.py
```diff
@@ -147,21 +149,64 @@ def duplicate_reusable_fieldsets(schema, fields_nested):
...
+                nesting_destination = fields_nested[top_level]['fields']
+                for level in split_flat_name[1:]:
+                    nesting_destination = nesting_destination.get(level, None)
+                    if not nesting_destination:
+                        raise ValueError('Field {} in path {} not found in schema'.format(level, nest_at))
+                    if nesting_destination.get('reusable', None):
+                        raise ValueError(
+                            'Reusable fields cannot be put inside other reusable fields except when the destination reusable is at the top level')
+                    nesting_destination = nesting_destination.setdefault('fields', {})
+                nesting_destination[nest_as] = copy.deepcopy(schema)
```
marshallmain, line 175: Inserting a copy of the schema here instead of a reference to the original schema will lead to the same bug that was fixed by #722.  I think to solve this we can make this line `nesting_destination[nest_as] = schema` and change line 177 below to `schema['self_nestings'][self_nesting['as']]['fields'] = schema['fields']`.  This way the both the regular and self reused fieldsets will also receive any modifications to the schema fields that occur after this reusable is inserted.

Then, on line 230 when the fieldsets references are being recursively replaced with independent copies, `self_nestings` can be combined with `fields` to produce the final set of fields.

webmat:
This PR has been superseded by #864

PR #827, opened 2020-04-24
[DOCS] Removes incorrect `nested` datatype definition

### LINKED ISSUE #826: [DOCS] `nested` field datatypes def in FAQ is incorrect
The [ECS FAQ][0] notes that the [`nested` field datatype][1] is not used. However, the definition of the `nested` field datatype is incorrect:

> ECS does not use nested datatypes, which are arrays of objects.

An object datatype, which is used by ECS, can also contain an array of objects as a value. The primary difference between the two is that objects in a `nested` datatype can be queried independently.

However, I'm not sure we need to go into that level of detail here. It would likely suffice to say:

> ECS does not use nested datatypes.

[0]: https://www.elastic.co/guide/en/ecs/master/ecs-faq.html#notation-diff
[1]: https://www.elastic.co/guide/en/elasticsearch/reference/7.6/nested.html

jrodewig:
Removes an incorrect definition of the [`nested` field datatype][0].

An [`object` datatype][1], which is used by the Elastic Common Schema,
can also contain an array of objects as a value.

The primary difference between the `object` and `nested` datatypes is
that objects in a `nested `datatype can be queried independently.

However, that distinction isn't needed or relevant here.

Closes #826

[0]: https://www.elastic.co/guide/en/elasticsearch/reference/master/nested.html
[1]: https://www.elastic.co/guide/en/elasticsearch/reference/master/object.html

### DIFF
--- a/docs/faq.asciidoc
+++ b/docs/faq.asciidoc
@@ -62,8 +62,6 @@ In Elasticsearch, `user` is represented as an {ref}/object.html[object
 datatype]. In the case of the underline notation, both are just
 {ref}/mapping-types.html[string datatypes].
 
-NOTE: ECS does not use {ref}/nested.html[nested datatypes].
-
 [float]
 [[dot-adv]]
 ===== Advantages of dot notation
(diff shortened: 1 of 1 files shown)


webmat:
Thank you for opening this.

Would you mind completely removing the line instead? There's nothing inherently wrong with the nested data type, we may very well end up using it. So this statement is irrelevant.

webmat:
We will should backport this to 1.5, which is the current version of the docs.Not sure it's worth

webmat:
I can take care of the backport

jrodewig:
Thanks @webmat!

PR #840, opened 2020-05-07
Remove guidance to lowercase http.request.method

### LINKED ISSUE #838: Removing guidance to lowercase "http.request.method"
It was recently raised that lowercasing `http.request.method` was a lossy normalization on a field that captures potentially suspicious behaviour.

E.g. an attacker could try to evade detection by using "PoST" instead of "POST".

Fields that currently suggest lowercasing the value are the following:

* `http.request.method`
* `network.type`
* `network.transport`
* `network.application`
* `network.protocol`
* `tls.version_protocol`

Other than the HTTP field for now, our stance is to leave the normalization in place. They're not places where we would usually need to do anomaly detection around letter casing. These values are provided by the data sources themselves. Feedback is welcome on this, however.

So the proposal is to mark this guidance as deprecated, and remove this guidance completely at ECS 2.0.

leehinman:
Remove guidance to lowercase http.request.method

Closes #838

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -33,7 +33,7 @@ Thanks, you're awesome :-) -->
 
 #### Deprecated
 
-* Remove guidance to lowercase `http.request.method` #840
+* Deprecate guidance to lowercase `http.request.method` #840
 
 ### Tooling and Artifact Changes
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -33,6 +33,7 @@ Thanks, you're awesome :-) -->
 
 #### Deprecated
 
+* Remove guidance to lowercase `http.request.method` #840
 
 ### Tooling and Artifact Changes
 
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -14,10 +14,7 @@
       short: HTTP request method.
       description: >
         HTTP request method.
-
-        The field value must be normalized to lowercase for querying. See
-        the documentation section "Implementing ECS".
-      example: get, post, put
+      example: get, POST, Put
 
     - name: request.body.content
       level: extended
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -17,14 +17,14 @@
 
         Prior to ECS 1.6.0 the following guidance was provided:
 
-          The field value must be normalized to lowercase for
-          querying. See the documentation section "Implementing ECS".
+        "The field value must be normalized to lowercase for
+        querying."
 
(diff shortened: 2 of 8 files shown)


schemas/http.yml
```diff
@@ -14,10 +14,7 @@
       short: HTTP request method.
       description: >
         HTTP request method.
-
-        The field value must be normalized to lowercase for querying. See
-        the documentation section "Implementing ECS".
-      example: get, post, put
+      example: get, POST, Put
```
webmat, line 17: I consider this to be a breaking change. 

I think we should put something back that says that we used to recommend lowercasing this field, but we're removing this guidance for now, as it's a lossy normalization on a field where it's worth doing anomaly detection. Original case will be mandated as of 2.0.

schemas/http.yml
```diff
@@ -15,9 +15,16 @@
...
-        the documentation section "Implementing ECS".
-      example: get, post, put
+        Prior to ECS 1.6.0 the following guidance was provided:
+
+          The field value must be normalized to lowercase for
+          querying. See the documentation section "Implementing ECS".
+
+        As of ECS 1.6.0, the guidance is deprecated because the
+        original case of the method may be useful in anomaly
+        detection.  Original case will be mandated in ECS 2.0.0
```
webmat, line 25:
```suggestion
        Prior to ECS 1.6.0 the following guidance was provided:

        "The field value must be normalized to lowercase, for querying."

        As of ECS 1.6.0, the guidance is deprecated because the
        original case of the method may be useful in anomaly
        detection.  Original case will be mandated in ECS 2.0.0
```

Let's fudge the quoting a bit. I think this would work best ^

schemas/http.yml
```diff
@@ -15,9 +15,16 @@
...
+
+      example: get, POST, Put
```
webmat, line 27: Nitpick: since we favour original casing, which will almost always be uppercase here I'd use all caps for each, and add one to make the point about anomaly detection.

```suggestion
      example: GET, POST, PUT, PoST
```

CHANGELOG.next.md
```diff
@@ -33,6 +33,7 @@ Thanks, you're awesome :-) -->
 
 #### Deprecated
 
+* Remove guidance to lowercase `http.request.method` #840
```
webmat, line 36:
```suggestion
* Deprecate guidance to lowercase `http.request.method` #840
```

Sorry last lil change 😬

webmat:
And the doc preview is back: http://ecs_840.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-http.html

PR #844, opened 2020-05-11
ip address locality

dainok:
We found useful having an attribute defining if the IP is local or not. We can use it as a self-describing field:
- internal to the organization
- external
Or it can also be used with:
- public
- private (or rfc1918)
The latter case is useless for IPv6.
Third option could be a bool:
- internal: False|True

Depending on the case, it can be useful having that info as both tag and attribute.

Any thoughts?

### DIFF
--- a/schemas/client.yml
+++ b/schemas/client.yml
@@ -40,6 +40,15 @@
 
         Can be one or multiple IPv4 or IPv6 addresses.
 
+    - name: locality
+      level: extended
+      type: keyword
+      short: Client locality.
+      description: >
+        Identify if the IP address is internal or external to the management
+        system. Could also identify rfc1918 IPv4 addresses. Suggested attributes
+        can be: "public" or "private" (or "rfc1918"), "external" or "internal".
+
     - name: port
       format: string
       level: core
--- a/schemas/destination.yml
+++ b/schemas/destination.yml
@@ -31,6 +31,15 @@
 
         Can be one or multiple IPv4 or IPv6 addresses.
 
+    - name: locality
+      level: extended
+      type: keyword
+      short: Destination locality.
+      description: >
+        Identify if the IP address is internal or external to the management
+        system. Could also identify rfc1918 IPv4 addresses. Suggested attributes
+        can be: "public" or "private" (or "rfc1918"), "external" or "internal".
+
     - name: port
       format: string
       level: core
--- a/schemas/server.yml
+++ b/schemas/server.yml
@@ -40,6 +40,15 @@
 
         Can be one or multiple IPv4 or IPv6 addresses.
 
+    - name: locality
+      level: extended
+      type: keyword
+      short: Server locality.
+      description: >
+        Identify if the IP address is internal or external to the management
+        system. Could also identify rfc1918 IPv4 addresses. Suggested attributes
+        can be: "public" or "private" (or "rfc1918"), "external" or "internal".
+
     - name: port
       format: string
       level: core
--- a/schemas/source.yml
+++ b/schemas/source.yml
@@ -31,6 +31,15 @@
 
         Can be one or multiple IPv4 or IPv6 addresses.
 
+    - name: locality
+      level: extended
+      type: keyword
+      short: Source locality.
+      description: >
+        Identify if the IP address is internal or external to the management
+        system. Could also identify rfc1918 IPv4 addresses. Suggested attributes
+        can be: "public" or "private" (or "rfc1918"), "external" or "internal".
+
     - name: port
       format: string
       level: core


andrewstucki:
To be honest, not entirely sure I see the need for something like this. Why wouldn't looking at the corresponding CIDR blocks for private/loopback/link-local addresses be enough to tell you whether something is "internal" or "external"? Why the need for an additional flag when there's enough information encoded into the IP address itself?

The following CIDR ranges really should be the only things that are ever "internal" addresses by definition:

```
127.0.0.0/8 -- loopback
169.254.0.0/16 -- link-local
10.0.0.0/8 -- private networking
172.16.0.0/12 -- private networking
192.168.0.0/16 -- private networking
::1 -- ipv6 loopback
fe80::/10 -- ipv6 link local
fd00::/8 -- ipv6 private networking
```

dainok:
I got the idea from the filebeat netflow plugin. It's using flow.locality: public|private
I totally agree with you, locality is a derived information, and a good query can substitute. But analysts can find useful filtering for attribute rather than writing a long query including or excluding ip addresses. Possible scenarios: exclude threats from external to internal and focus only on internal to internal threats.

I also agree that public|private is a non sense on IPv6, I would suggest something like "external, hq, campus...". We already added observer.egress|ingress.zone: that's the point of view of the observer, and a packet can flow through many observers. source|destination.locality is "absolute" and related to packet.

Let me know what you think aobut.

dainperkins:
@andrewstucki  & @dainok 

There's really no mutual inclusivity/exclusivity between private (unrouted) and internal.  I've seen plenty of vpn connections to external parties utilizing private addresses on both sides, mpls connections using non routed ips inside the vrfs, and conversely plenty of times connections to "internal resources" are in the cloud & are using public ips for connectivity.

By the same token public/private seems unnecessary (its not indicative of anything useful in and of itself, and is fairly easy to determine from a static definition of subnets)

I keep going back to a few network performance / risk concepts that I've picked up over the years, likely based on e.g. an enrichment process.

Top Level:
 - internal / external (mine / not mine) 

Sub Groups:
 - site (e.g. NYC Office)
 - segment (e.g. vlan 10 client segment)
 - apps (e.g. exchange servers)

Risk Labels
 - risk (e.g. pci, hippa, etc)

Thinking being that the enrichment can tie to cidr or ip, based on a CMDB style index.   A lot of this comes back to a network / application performance perspective (e.g. tracking resets for specific application types, etc.) as well as modeling intra-zone connections for security analysis.

dainok:
@dainperkins @andrewstucki seems the topic is not so popular :)
How do we continue? Should I drop the PR?

andrewstucki:
@dainok I'm fine leaving this open for a bit just, if anything just to discuss a bit more. But I think @dainperkins brings up some good points about "internal"/"external" being in the eye of the beholder (i.e. "internal" systems with public IPs). I'm wondering--does this use-case make more sense just leveraging `labels` or `tags` from the [base ECS field set](https://www.elastic.co/guide/en/ecs/current/ecs-base.html)?

webmat:
Let's leave it open for now please :-) 

Thanks for chiming in frequently, @dainok!

H1L021:
Flow or Source LOCALITY is something useful for me personally in the context of developping rules to detect Internal Vertica or Horizontal Scan activity.
For instance, I found that some companies use public IPs for internal routing specially for firewalls. So if I rely solely on Geo and CIDR plugins I won't detect intern recon acitivity with fidelity and honestly it would be a mess. However I manually add a pipeline with logstash for these particular cases (something that were done in a great way here https://github.com/enotspe/fortinet-2-elasticsearch/blob/master/logstash/50-common_ecs) where also I can add those public IPs and define them as internal something like this : 
```
if [source][ip] {
        cidr {
                add_field => {"[source][locality]" => "private"}
                address => [ "%{[source][ip]}" ]
                network => [ "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "1.1.1.120/24" ]
        }
```

In my context the field source.locality or flow.locality is a must for accurate detection : 
```
"size": 0,
          "query": {
            "bool": {
              "must": [
                {
                  "match": {
                    "tags": "firewall"
                  }
                },
                {
                  "term": {
                    "source.locality": {
                      "value": "private"
                    }
                  }
                },
                {
                  "range": {
                    "@timestamp": {
                      "gte": "now-2m"
                    }
                  }
                }
              ]
            }
          },
          "aggs": {
            "by_src_ip": {
              "terms": {
                "field": "source.ip"
              },
              "aggs": {
                "by_target_ip": {
                  "terms": {
                    "field": "destination.ip",
                    "order": {
                      "unique_port_count": "desc"
                    }
                  },
                  "aggs": {
                    "unique_port_count": {
                      "cardinality": {
                        "field": "destination.port"
                      }
                    }
                  }
                }
              }
            }
          }
```

djptek:
There is a PR open to add `ip_range` to the `enrich` processor which would enable adding this field using Elasticsearch based on indexed data, see: https://github.com/elastic/elasticsearch/pull/65781 

If/when the PR is merged you'd no longer need to store this field since it could be recreated on demand

djptek:
No activity (other than mine) in the last year, closing, please fix failing CI tests and re-open if this is still required

PR #846, opened 2020-05-13
Deprecate event.dataset

ruflin:
With https://github.com/elastic/ecs/pull/845 stream.dataset is introduced. Even though event.dataset is still heavily used, users should start to switch over to using stream.dataset. I expect event.dataset to be remove from ECS in the next major version.

stream.dataset and event.dataset have the same content but are not of the same type. First one is constant_keyword, other one keyword. On the query side this should not make a difference but not sure if it is possible to use alias to reference one to an other.

### DIFF
--- a/generated/ecs/ecs_flat.yml
+++ b/generated/ecs/ecs_flat.yml
@@ -1837,6 +1837,7 @@ event.created:
   type: date
 event.dataset:
   dashed_name: event-dataset
+  deprecated: true
   description: 'Name of the dataset.
 
     If an event source publishes more than one type of log or events (e.g. access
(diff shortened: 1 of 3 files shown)


ruflin:
@webmat Two things

* Same problem with the changelog, where should I put it?
* Does ECS have today the feature of deprecating a field?

neu5ron:
waht....?

defensivedepth:
Hello there!

Which Elastic license are these new ECS fields under?

I'm also trying to wrap my mind around how we would migrate to this new structure for how we are currently using Modules & Dataset?

image

ruflin:
I'm closing this for now as I opened this to have a discussion (but did not label it 🤦 ) and it created more controversy then I wanted.

PR #847, opened 2020-05-13
Deprecate event.module

ruflin:
event.module was introduced together with event.dataset. With https://github.com/elastic/ecs/pull/845 the new stream fields are introduced and module is not used anymore.

When we introduce event.module the assumption was that this is the field used for quering but it turned out in most casese event.dataset is needed and required. I expect event.module to be removed from ECS in the next major version but this does not mean, it cant still be used by Beats for example.

### DIFF
--- a/generated/ecs/ecs_flat.yml
+++ b/generated/ecs/ecs_flat.yml
@@ -2005,6 +2005,7 @@ event.kind:
   type: keyword
 event.module:
   dashed_name: event-module
+  deprecated: true
   description: 'Name of the module this data is coming from.
 
     If your monitoring agent supports the concept of modules or plugins to process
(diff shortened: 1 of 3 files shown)


ruflin:
* Where should I put the changelog?
* Does deprecation exist in ECS today?

webmat:
No specific deprecation support currently exist yet. There's no attribute in the YAML to flag fields that disappear, for example. But it's something we want to add eventually.

However we can still deprecate via a mention in the description field in the meantime.

Changelog entries go to the file `CHANGELOG.next.md`.

ruflin:
I'm closing this for now as I opened this to have a discussion (but did not label it 🤦 ) and it created more controversy then I wanted.

PR #852, opened 2020-05-15
Add x509 to code_signature

andrewstucki:
I'm not entirely sure about other paradigms, but both Mac mach-o binary signatures and Windows pe signatures are in pkcs7 format with embedded x509 cert information. For example, on my machine (with a jank parser I quickly wrote up):

For a mach-o file:

```
➜ ./gocodesign -in /Applications/Microsoft\ Excel.app/Contents/MacOS/Microsoft\ Excel
Subject: CN=Developer ID Application: Microsoft Corporation (UBF8T346G9),OU=UBF8T346G9,O=Microsoft Corporation,C=US
Issuer: CN=Apple Root CA,OU=Apple Certification Authority,O=Apple Inc.,C=US
Issuer: CN=Apple Root CA,OU=Apple Certification Authority,O=Apple Inc.,C=US
Issuer: CN=Developer ID Certification Authority,OU=Apple Certification Authority,O=Apple Inc.,C=US
```
And for a pe I dumped from a windows VM:
```
➜ ./gocodesign -in svchost.exe
Subject: CN=Microsoft Windows Publisher,O=Microsoft Corporation,L=Redmond,ST=Washington,C=US
Issuer: CN=Microsoft Windows Production PCA 2011,O=Microsoft Corporation,L=Redmond,ST=Washington,C=US
Issuer: CN=Microsoft Root Certificate Authority 2010,O=Microsoft Corporation,L=Redmond,ST=Washington,C=US
```

So it seems to me that we should allow for the `x509` fields to be nested under `code_signature` since pkcs7 is a widely used code signing format.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -22,7 +22,7 @@ Thanks, you're awesome :-) -->
 
 * Add architecture and imphash for PE field set. (#763)
 * Added `agent.build.*` for extended agent version information. (#764)
-* Added `x509.*` field set. (#762)
+* Added `x509.*` field set. (#762, #852)
 * Added more account and project cloud metadata. (#816)
 
 #### Improvements
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -709,6 +709,24 @@ Note also that the `code_signature` fields are not expected to be used directly 
 
 
 
+[[ecs-code_signature-nestings]]
+===== Field sets that can be nested under Code Signature
+
+[options="header"]
+|=====
+| Nested fields | Description
+
+// ===============================================================
+
+
+| <<ecs-x509,code_signature.x509.*>>
+| This implements the common core fields for x509 certificates. This information is likely logged with TLS sessions, digital signatures found in executable binaries, S/MIME information in email bodies, or analysis of files on disk. When only a single certificate is logged in an event, it should be nested under `file`. When hashes of the DER-encoded certificate are available, the `hash` data set should be populated as well (e.g. `file.hash.sha256`). For events that contain certificate information for both sides of the connection, the x509 object could be nested under the respective side of the connection information (e.g. `tls.server.x509`).
+
+// ===============================================================
+
+
+|=====
+
 [[ecs-container]]
 === Container Fields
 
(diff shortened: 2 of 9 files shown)


webmat:
Thanks for bringing it up! If this makes sense for the use case I'm all for it.

Code Signatures are currently reused here: `dll.code_signature`, `file.code_signature`, `process.code_signature`, and `process.parent.code_signature`. Is this reuse valid and necessary for all locations? If so, this means the PR adds **96 new fields** in one fell swoop. Ouch! 😄 

Finally, I see we missed a small issue in the original x509 PR. Andrew, since you're piggybacking on the x509 changelog entry, would you be willing to rename the `short_description:` attributes in `schemas/x509.yml` to simply `short:`? Otherwise they don't get picked up and the long description is used, e.g. [here](https://www.elastic.co/guide/en/ecs/master/ecs-field-reference.html) 😬

andrewstucki:
@webmat yeah, I'm fine shoring up the `short_description` stuff. TBH I think that the `code_signature` field makes more sense to be nested under `file` than `process`, `process.parent` and `dll` (and potentially `pe` or other file-format specific fields that we might add later, things like `macho` or `elf`).

To me a `code_signature` in the form of being tied to a `process` is actually describing the underlying signature embedded into the file on disk--same with `dll`--so I would imagine if you're describing the code signature of a process or some library you should pretty much always fill in `code_signature` at `file` (or `pe` if we add it there).

webmat:
I'm curious what folks think here.

> TBH I think that the code_signature field makes more sense to be nested under file than process, process.parent and dll (and potentially pe or other file-format specific fields that we might add later, things like macho or elf).

Should we shore up the sprawl of these fields somehow (likely via a 2.0 breaking change), or are we fine adding all of these fields in all of these locations?

All of these locations are indeed optional to fill (@rw-access). But searching for code signatures is going to be a complicated, given the amount of locations where code signatures could be found, after this.

I wonder if progress has been made on capturing an arbitrary amount of code signatures for the cases where it's warranted?

rw-access:
> TBH I think that the code_signature field makes more sense to be nested under file than process, process.parent and dll (and potentially pe or other file-format specific fields that we might add later, things like macho or elf).

For file events, (the Endpoint) actually populates code_signatures for the executing process and for the file in question, so we do need both locations. Similarly for DLL.

> I wonder if progress has been made on capturing an arbitrary amount of code signatures for the cases where it's warranted?
Endpoint currently uses `file.Ext.code_signatures` and process.Ext.code_signatures`, which is an array/`nested`, because it can gather more than a single signature. They use the `Ext` namespace within file/process for otherwise breaking changes to ECS.


I think the amount of fields introduced is a really good concern. I don't fully understand the impact of the number of fields in a mapping, and what happens if some are never populated. Is that okay if we have a lot of unused fields in a mapping?

ebeahan:
> So it seems to me that we should allow for the x509 fields to be nested under code_signature since pkcs7 is a widely used code signing format.

If there are supporting use cases, this change could still be discussed. However, this PR is stale. Please open a new PR if there's renewed interest.

PR #856, opened 2020-05-20
Add option for custom template and mapping settings

marshallmain:
Little quality of life improvement that lets users override the default template and mapping settings for the generated ES templates. This PR adds 2 new options, `template-settings` and `mapping-settings` that take in filenames and use those json files to populate the template and mapping settings of the generated ES template.

Example usage, run in the ECS directory:
`python scripts/generator.py --mapping-settings custom_mapping.json --template-settings custom_template.json`
This replaces the mapping and template settings with whatever is in custom_mapping.json and custom_template.json, respectively.

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -3,6 +3,7 @@ import glob
 import os
 import schema_reader
 import yaml
+import json
 
 from generators import intermediate_files
 from generators import csv_generator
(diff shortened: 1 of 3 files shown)


scripts/generator.py
```diff
@@ -3,6 +3,7 @@
 import os
 import schema_reader
 import yaml
+import json
```
webmat, line 6: Nit: Please remove this new import, as it's no longer needed :-)

marshallmain: ah good point

PR #865, opened 2020-06-04
Refactor asciidocs to use jinja2 templates

ebeahan:
This PR refactors the `ascii_fields` generator to use [`jinja2`](https://jinja.palletsprojects.com/en/2.11.x/) templating. Motivation is better decoupling the asciidoc formatting from the content generation.

Template changes can be made in the `./scripts/generators/templates` directory without modifying the generator code directly. Formatting was tested by generating docs and ensuring no changes from their existing structure and content.

#### Templating implementation

The `@templated('template_file_name')` decorator is used to inject the additional functionality that renders and returns the template's content to the generator.

```python
@templated('fields_template.j2')
def page_field_index(intermediate_nested, ecs_version):
    fieldsets = ecs_helpers.dict_sorted_by_keys(intermediate_nested, ['group', 'name'])
    return dict(ecs_version=ecs_version, fieldsets=fieldsets)
```

Decorated functions should return a `dict` used to generate the template. When the decorated function returns, the dictionary is passed to the template renderer.

~*NOTE*: Leaving as draft until #864 is merged. I'll update and resolve conflicts accordingly.~ #864 is merged 😎


scripts/generators/asciidoc_fields.py
```diff
@@ -19,22 +58,10 @@ def save_asciidoc(file, text):
 
 # Field Index
 
-
+@templated('fields_template.j2')
```
webmat, line 61: I like the 1:1 mapping between the full templates (fields_template.j2 and field_values_template.j2) and their rendered `.asciidoc` counterpart 👍 

I also like that `field_details/` follows the same 1:1 with its `.asciidoc` counterpart, except it's a directory this time :-)

The following comment is for a subsequent PR perhaps; by no means a blocker for this one. Could we eventually have a master template for `field-details.asciidoc` as well, and call out from this master template to sub-templates as needed? Not sure if it's easy to do with Jinja2.

scripts/generators/asciidoc_fields.py
```diff
@@ -1,15 +1,54 @@
-from os.path import join
+from functools import wraps
+import os.path as path
```
webmat, line 2: Thanks for this small adjustment. The out of the blue `join()` calls always looked strange to me 😂

scripts/generators/templates/field_details/row.j2
```diff
@@ -0,0 +1,14 @@
+
+| {{ flat_name }}
+| {{ description|replace("\n", "\n\n") }}
```
webmat, line 3: This was the only transformation we had in `render_asciidoc_paragraphs()` so that's fine for now.

But if we decide to introduce more transformations again (e.g. support hyperlinks), do you know how to create these Jinja2 modifier functions?

ebeahan: Yes. Outside of the [built-in filters](https://jinja.palletsprojects.com/en/2.11.x/templates/#list-of-builtin-filters), we can also define [custom filters](https://jinja.palletsprojects.com/en/2.11.x/api/#custom-filters) by passing our own defined functions into the Jinja environment declaration.

scripts/generators/asciidoc_fields.py
```diff
@@ -1,15 +1,54 @@
...
+    save_asciidoc(path.join(out_dir, 'field-details.asciidoc'), page_field_details(nested))
+    save_asciidoc(path.join(out_dir, 'field-values.asciidoc'), page_field_values(nested))
 
 # Helpers
 
 
+def templated(template_name):
+    """Decorator function to simplify rendering a template.
+
+    ::param template_name: the name of the template to be rendered
```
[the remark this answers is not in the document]
ebeahan: It is not! Thanks for catching 😄

scripts/generators/asciidoc_fields.py
ebeahan: I agree it will be more consistent to have the master `field_details.j2` file, and I plan on tackling this soon.

Jinja does have features like [template inheritance](https://jinja.palletsprojects.com/en/2.11.x/templates/#template-inheritance) and [include tags](https://jinja.palletsprojects.com/en/2.11.x/templates/#include) to help simplify here. 😃

ebeahan:
@webmat thanks for the feedback! Let me know your thoughts on the additions to `CONTRIBUTING.md`.

PR #866, opened 2020-06-08
Fix some ATT&CK verbiage and trademark

rw-access:
I saw some verbiage for `threat.*` that mentioned the MITRE ATT&CK® framework but needed to be adjusted.

Some of the changes
- Added ®. We were supposed to have ™ before, but early in 2020 they were granted ®.
- Correctly spelled MITRE with all capital letters
- Removed some confusing verbiage like "You can use the Mitre ATT&CK Matrix Tactic categorization." The words "Tactic categorization" are a little confusing and not part of the framework. Instead I went with "You can use a MITRE ATT&CK Matrix® tactic, for example." Also, it's possible to mention tactics without concrete techniques and vice versa, so I removed any language that coupled them together.

### DIFF
--- a/schemas/threat.yml
+++ b/schemas/threat.yml
@@ -4,7 +4,7 @@
   group: 2
   short: Fields to classify events and alerts according to a threat taxonomy.
   description: >
-   Fields to classify events and alerts according to a threat taxonomy such as the Mitre ATT&CK framework.
+   Fields to classify events and alerts according to a threat taxonomy such as the MITRE ATT&CK® framework.
 
    These fields are for users to classify alerts from all of their sources (e.g. IDS, NGFW, etc.) within a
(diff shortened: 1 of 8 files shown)


schemas/threat.yml
```diff
@@ -29,8 +29,8 @@
       type: keyword
       short: Threat tactic.
       description: >
-          Name of the type of tactic used by this threat. You can use the Mitre ATT&CK Matrix Tactic categorization, for example.
-          (ex. https://attack.mitre.org/tactics/TA0040/ )
+          Name of the type of tactic used by this threat. You can use the MITRE ATT&CK® Matrix tactic categorization, for example.
```
webmat, line 32: I would switch the phrasing "use a" instead of "use the" for this sentence as well.

rw-access: thank you! this is why we peer review 😆

PR #868, opened 2020-06-11
Define process.parent via the new field reuse mechanism

webmat:
The `process.parent` fields had initially been defined by duplicating each field explicitly, since the field reuse mechanism didn't support field reuse within the same field set at the time.

Introducing `process.parent.*` by duplicating fields has predictably led to a few small mistakes, despite us squinting very hard and trying to avoid this :-)

* At least one process field was modified without having the same modification applied to its `process.parent` twin.
* Also, the field set `pe` was only reused under `process`, but not `process.parent`. This was not intentional.

This PR removes the field duplication by using the improved reuse mechanism. This means both problems described above are also resolved by this PR.

### DIFF
--- a/code/go/ecs/process.go
+++ b/code/go/ecs/process.go
@@ -31,9 +31,6 @@ type Process struct {
 	// Process id.
 	PID int64 `ecs:"pid"`
 
-	// Process id.
-	ParentPID int64 `ecs:"parent.pid"`
-
 	// Unique identifier for the process.
 	// The implementation of this is specified by the data source, but some
 	// examples of what could be used here are a process-generated UUID, Sysmon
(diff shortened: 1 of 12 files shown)


webmat:
@rw-access When introducing PE, we had discussed that it belonged both under `process` and `process.parent`. But we forgot to do so explicitly, at the time. This PR fixes this. You're still good with this?

rw-access:
Awesome. Before I ✔️, this adds `code_signature` to the parent too, right?

code/go/ecs/process.go
```diff
@@ -31,9 +31,6 @@ type Process struct {
 	// Process id.
 	PID int64 `ecs:"pid"`
 
-	// Process id.
-	ParentPID int64 `ecs:"parent.pid"`
```
rw-access, line 35: Should `Process` be self-referential?
```
      *Process parent `ecs:"process.parent"`
```

webmat: This Golang library is going away from this repo in the future. Does the Endpoint team use it?

rw-access:
Also, I noticed this typo for process.args
```yml
      description: >
        Array of process arguments, starting with the absolute path to the executable.
```

For the first arg, I think it should be passed-through as is, so I think we can remove that note so that it's just this:
```
      description: >
        Array of process arguments.

code/go/ecs/process.go
rw-access: nope, just wondering

smerzlyakov:
Looks not so good to use `process.parent.*` bacause of it do not fir to all scheme. We do not have **parent** base field.
Why not to use `parent.process.*` ? Looks better, because of all sub-fileds are in **process** core field.

PR #869, opened 2020-06-11
Define fields to allow representing multiple users in an event.

### LINKED ISSUE #809: [ECS] Multiple users in an event proposal

webmat:
This PR implements the proposal discussed in #809. In short, we now reuse all `user` fields in 3 new places, in order to allow capturing more relevant users on a given event:

* `user.effective.*` is used to capture the effective user in cases of privilege escalation.
* `user.target.*` is used to capture a distinct user that is affected by an action, like IAM: Alice creates/suspends/deletes Bob.
* `user.changes.*` is used to capture the changes to an existing user. It's worth pointing out that only the attributes that change for the user are expected to be populated here.

For now this PR simply adds the fields. We don't yet have a good way to have either contextual definitions around field reuse, nor a place to document this via free form text, in the user page. This will come as a later addition.

I'm opening this PR in a straighforward manner, by directly introducing these fields. However I'm thinking this major addition is probably a good candidate to move through the RFC process.

What do you think @epixa @ebeahan @tsg?

Closes #809

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -24,6 +24,8 @@ Thanks, you're awesome :-) -->
 * Added `agent.build.*` for extended agent version information. (#764)
 * Added `x509.*` field set. (#762)
 * Added more account and project cloud metadata. (#816)
+* Added `user.effective`, `user.target`, and `user.changes` to capture more details
+  when multiple users are relevant to an event. #869
 
 #### Improvements
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -6526,7 +6526,7 @@ example: `albert`
 
 ==== Field Reuse
 
-The `user` fields are expected to be nested at: `client.user`, `destination.user`, `host.user`, `server.user`, `source.user`.
+The `user` fields are expected to be nested at: `client.user`, `destination.user`, `host.user`, `server.user`, `source.user`, `user.changes`, `user.effective`, `user.target`.
 
 Note also that the `user` fields may be used directly at the top level.
 
@@ -6543,8 +6543,26 @@ Note also that the `user` fields may be used directly at the top level.
 // ===============================================================
 
 
+| <<ecs-user,user.changes.*>>
+| Fields to describe the user relevant to the event.
+
+// ===============================================================
+
+
+| <<ecs-user,user.effective.*>>
+| Fields to describe the user relevant to the event.
+
+// ===============================================================
+
+
 | <<ecs-group,user.group.*>>
 | User's group relevant to the event.
+
+// ===============================================================
+
+
+| <<ecs-user,user.target.*>>
+| Fields to describe the user relevant to the event.
 
 // ===============================================================
 
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -5200,6 +5200,78 @@
       provide an array that includes all of them.'
     type: group
     fields:
+    - name: changes.domain
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: 'Name of the directory the user is a member of.
+
+        For example, an LDAP or Active Directory domain name.'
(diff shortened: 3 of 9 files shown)


webmat:
Ping all of the fine folks who chimed in on #809: @leehinman @neu5ron @janniten @willemdh @rw-access :-)

ebeahan:
+1 for moving this additions through the RFC process 😃

l0x-c0d3z:
I like the approach. I've been having issues mapping IAM audit logs recently. It also sets a good precedent for stuff like 'process.target.*' (for stuff like cross-process access and code injections).

webmat:
Closing in favor of #1066 (see also RFC 0007 stage 3 PR #1017)

PR #872, opened 2020-06-18
Fix up generated go code for arrays

andrewstucki:
I noticed recently when fixing up some packetbeat code that the generated structures for go don't respect array normalization--makes it so we have to do some gross stuff like this as a workaround right now:

```go
type ecsEvent struct {
	ecs.Event `ecs:",inline"`
	// overridden because this needs to be an array
	Category []string `ecs:"category"`
	// overridden because this needs to be an array
	Type []string `ecs:"type"`
}
```

I'll see about adding in some json/`common.MapStr` serializers at some point in a follow-up.

### DIFF
--- a/code/go/ecs/base.go
+++ b/code/go/ecs/base.go
@@ -35,7 +35,7 @@ type Base struct {
 	Timestamp time.Time `ecs:"@timestamp"`
 
 	// List of keywords used to tag each event.
-	Tags string `ecs:"tags"`
+	Tags []string `ecs:"tags"`
 
 	// Custom key/value pairs.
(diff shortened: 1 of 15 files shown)


webmat:
If this Golang library is in use, it should be moved to a different repo. I'll be happy to help do that.

The ECS team doesn't have the resources or expertise to maintain it, and therefore issues raised about this library are not being addressed (see bug reports linked [here](https://github.com/elastic/ecs/pull/628) ).

ebeahan:
Generated go code was removed in https://github.com/elastic/ecs/pull/1567

PR #877, opened 2020-06-25
feat: include alias path when generating template

### LINKED ISSUE #876: Add generator support for alias datatype
**Summary**

Presently, the generator does not include the `path` field in the generated template, meaning that it does not properly support the [alias datatype](https://www.elastic.co/guide/en/elasticsearch/reference/master/alias.html).

**Motivation**:

When migrating to ECS the generator provides a powerful way to generate templates, however, in certain cases it's desirable to be able to create a schema of aliases for deprecated fields to provide backwards-compatibility.

**Detailed Design**:
N/A

rgmz:
This fixes #876 

Done quickly via GitHub web, but let me know if there are any additional changes required (e.g. tests or changelog).

### DIFF
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -34,7 +34,7 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
     allowed_keys = ['name', 'level', 'required', 'type', 'object_type',
                     'ignore_above', 'multi_fields', 'format', 'input_format',
                     'output_format', 'output_precision', 'description',
-                    'example', 'enabled', 'index']
+                    'example', 'enabled', 'index', 'path']
     multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field', 'normalizer', 'ignore_above']
 
     fields = []
@@ -61,14 +61,6 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
 
         if not ecs_field['flat_name'] in df_whitelist:
             beats_field['default_field'] = False
-
-        # Include the `path` attribute for `alias` types.
-        # https://github.com/elastic/ecs/issues/876
-        if ecs_field['type'] == 'alias':
-            if 'path' in ecs_field:
-                beats_field['path'] = ecs_field['path']
-            else:
-                raise ValueError(f'The [path] property must be specified for field [{contextual_name}]')
 
         fields.append(beats_field)
     return sorted(fields, key=lambda x: x['name'])
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -62,6 +62,14 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
         if not ecs_field['flat_name'] in df_whitelist:
             beats_field['default_field'] = False
 
+        # Include the `path` attribute for `alias` types.
+        # https://github.com/elastic/ecs/issues/876
+        if ecs_field['type'] == 'alias':
+            if 'path' in ecs_field:
+                beats_field['path'] = ecs_field['path']
+            else:
+                raise ValueError(f'The [path] property must be specified for field [{contextual_name}]')
+
         fields.append(beats_field)
     return sorted(fields, key=lambda x: x['name'])
 
--- a/scripts/generators/es_template.py
+++ b/scripts/generators/es_template.py
@@ -59,6 +59,11 @@ def entry_for(field):
             ecs_helpers.dict_copy_existing_keys(field, field_entry, ['ignore_above'])
         elif field['type'] == 'text':
             ecs_helpers.dict_copy_existing_keys(field, field_entry, ['norms'])
+        elif field['type'] == 'alias':
+            if 'path' in field:
+                field_entry['path'] = field['path']
+            else:
+                raise ValueError(f'The [path] property must be specified for field [{field["name"]}]')
 
         if 'multi_fields' in field:
             field_entry['fields'] = {}
--- a/scripts/generators/es_template.py
+++ b/scripts/generators/es_template.py
@@ -60,10 +60,7 @@ def entry_for(field):
         elif field['type'] == 'text':
             ecs_helpers.dict_copy_existing_keys(field, field_entry, ['norms'])
         elif field['type'] == 'alias':
-            if 'path' in field:
-                field_entry['path'] = field['path']
-            else:
-                raise ValueError(f'The [path] property must be specified for field [{field["name"]}]')
+            ecs_helpers.dict_copy_existing_keys(field, field_entry, ['path'])
 
         if 'multi_fields' in field:
             field_entry['fields'] = {}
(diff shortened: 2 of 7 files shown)


rgmz:
> x Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [9864f19](https://github.com/elastic/ecs/commit/9864f195a2c0663172a1592c948253c03779bcba)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

I signed the CLA prior to creating this PR, so maybe it takes a bit of time to be recognized.

webmat:
I re-ran our tool to check the CLA, and you signed it correctly indeed.

Thanks for submitting this, we will review shortly

ebeahan:
Thanks again for submitting @rgmz! I see value in adding support for the `alias` type per your description in #876 .

These changes would introduce an optional `path` field set attribute to the schema, which is required when type is specified as `alias`? Is the following example snippet an accurate example?

```yaml
    - name: id
      level: extended
      type: alias
      path: custom.host.id
```

We'll need to consider other aspects for supporting this additional `path` field and the `alias` type (likely others - these were two that came to mind so far):

* The beats generator, `./scripts/generators/beats.py` should also support `path` when type is `alias`.
* The target of `path` is [subject to restrictions](https://www.elastic.co/guide/en/elasticsearch/reference/master/alias.html#alias-targets) which wouldn't be checked or enforced by the ECS tooling (at least initially). We could mark the support for `alias` as experimental and note the limitation, but it could cause some hidden complexity when attempting to apply the mappings.

webmat:
Yeah I support this addition.

This only fixes it for the Go generator. I'd like it if we failed the run if it's missing, in cleaner.py / `field_mandatory_attributes`. Check it out [here](https://github.com/elastic/ecs/blob/master/scripts/schema/cleaner.py#L153-L162)

@rgmz Are you good with adding this? If not, perhaps @ebeahan could pick this up?

ebeahan:
> This only fixes it for the Go generator

ES generator? Or is there a relationship to the Go generator I'm overlooking?

I did see a possible need for introducing a new [type](https://github.com/elastic/ecs/blob/master/scripts/cmd/gocodegen/gocodegen.go#L267-L295) in the Go generator, but since how we're aiming to remove it (#628) in the near future I wasn't concerned. I also did some light testing of the branch and didn't see any immediate issues.

webmat:
Yes the `alias` datatype itself should make it there no problem. But it's a good point to ensure the `path` attribute makes it there too, and that it's not being filtered out.

Note that ECS doesn't make use of the `alias` type, so we should make sure the ES script has test coverage for this as well.

rgmz:
@webmat @ebeahan 

Apologies for letting this stagnate, it's been a busy few months! :)

> These changes would introduce an optional `path` field set attribute to the schema, which is required when type is specified as `alias`? Is the following example snippet an accurate example?
> 
> ```yaml
>     - name: id
>       level: extended
>       type: alias
>       path: custom.host.id
> ```

Correct. The intention is to match the schema to the Elasticsearch usage of `alias`, which you've linked.  

> We'll need to consider other aspects for supporting this additional `path` field and the `alias` type (likely others - these were two that came to mind so far):
> 
>    * The beats generator, `./scripts/generators/beats.py` should also support `path` when type is `alias`.
> 
>    * The target of `path` is [subject to restrictions](https://www.elastic.co/guide/en/elasticsearch/reference/master/alias.html#alias-targets) which wouldn't be checked or enforced by the ECS tooling (at least initially). We could mark the support for `alias` as experimental and note the limitation, but it could cause some hidden complexity when attempting to apply the mappings.

I've updated `./scripts/generators/beats.py` and `scripts/schema/cleaner.py` to support `path` for the `alias` type.

I agree that validating the generated artifacts would be ideal, however, that's probably out of scope for this PR. e.g., you can also generate an invalid template by defining `type: invalid_type123`.

scripts/generators/es_template.py
```diff
@@ -59,6 +59,11 @@ def entry_for(field):
             ecs_helpers.dict_copy_existing_keys(field, field_entry, ['ignore_above'])
         elif field['type'] == 'text':
             ecs_helpers.dict_copy_existing_keys(field, field_entry, ['norms'])
+        elif field['type'] == 'alias':
```
ebeahan, line 62: Similar to the `beats` generator, since we're checking earlier that `path` is present when `type: alias` in `schema.cleaner.field_mandatory_attributes`, I think we're ok to just copy without worrying about an exception.

```python
       elif field['type'] == 'alias':
            ecs_helpers.dict_copy_existing_keys(field, field_entry, ['path'])
```

scripts/generators/beats.py
```diff
@@ -62,6 +62,14 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
         if not ecs_field['flat_name'] in df_whitelist:
             beats_field['default_field'] = False
 
+        # Include the `path` attribute for `alias` types.
+        # https://github.com/elastic/ecs/issues/876
+        if ecs_field['type'] == 'alias':
```
ebeahan, line 67: Instead of handling this here, let's add `path` to the `allowed_keys` list in line 34. When `beats_field = ecs_helpers.dict_copy_keys_ordered(ecs_field, allowed_keys)` is called in line 43, this should take care of the copying the `path` key/value over.

scripts/generators/beats.py
```diff
@@ -62,6 +62,14 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
...
+            if 'path' in ecs_field:
+                beats_field['path'] = ecs_field['path']
+            else:
+                raise ValueError(f'The [path] property must be specified for field [{contextual_name}]')
```
ebeahan, line 71: Since you've added the check in `schema.cleaner.field_mandatory_attributes` looking for `path` when `type: alias`, I don't think we have to worry about checking again here.

ebeahan (changes requested):
Thanks for the continue work here @rgmz ! This will be a nice feature addition.

Outside of the few notes I made below, can you also add:

- [x] A brief entry in the `Supported keys to describe fields` section of  `schemas/README.md` for `path`
- [x] A changelog entry to `CHANGELOG.next.md`

scripts/generators/beats.py
rgmz: Added in https://github.com/elastic/ecs/pull/877/commits/4fba68804468b321e5e12892235134d149ab75c8 :+1:

Given that `path` is (to my knowledge) only valid when `type: alias`, I wasn't sure if it was okay to include it in `allowed_keys`.

scripts/generators/es_template.py
rgmz: Fixed in https://github.com/elastic/ecs/pull/877/commits/231fe9870a9783c5053bfaa29ad6187a32e201f0

rgmz:
> Thanks for the continue work here @rgmz ! This will be a nice feature addition.
> 
> Outside of the few notes I made below, can you also add:
> 
>     * [ ]  A brief entry in the `Supported keys to describe fields` section of  `schemas/README.md` for `path`
> 
>     * [ ]  A changelog entry to `CHANGELOG.next.md`

Done!

(Working from the GitHub UI once again, so let me know if I missed anything.)

ebeahan (approved):
This looks great @rgmz ! Thanks for making those changes.

I'm approving and will handle merging and then backporting 🚢

webmat:
Thanks @rgmz and @ebeahan for getting this to the finish line :-)

Issue #874, opened 2020-06-23
network.forwarded_ip should be an ordered list of IP


dainok:
**Description of the issue:**

network.forwarded_ip is used to map X-Forwarded-For addresses. XFF is an ordered list of IP, because a single request can pass through multiple proxies.

**Any additional context or examples:**

X-Forwarded-For: client, proxy1, proxy2

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

dainperkins:
@dainok, 

Good catch, tho the network.forwarded_ip was meant for NAT translations, which are only locally significant.  

What do you think about an http addition (http.forwarded_for)?

/d

dainok:
http.forwarded_for is the best option I guess, because it's a HTTP header. But at this point I'm not sure what network.forwarded_ip should be.
This bug request is because filebeat-panw module is using network.forwarded_ip to store multiple http x-forwarded-for IP.

dainperkins:
Apologies - I was thinking source.nat.ip, NOT network.forwarded_ip, which is specifically labeled as a proxy forward.  

I still think that x-forwarded-for belongs under http, as in every implementation I can recall a proxied connection is, at ground truth, two totally separate network connections, whereas nat is inline replacement of ip/ports - but the connection is not terminated at any higher layers on the device handling translation.

I also believe, @webmat correct me if I am wrong, array of IPs is fine for an IP field

dainok:
Agree with you but I still don't get when network.forwarded_ip and http.forwarded_for should contains different data.
Try to speculate:
- if http sesssion is observed, then http.forwared_for = network.forwarded_ip
- if L4 traffic is observed, then network.forwarded_ip is set, http.forwarded_for is not.
What do you think about?

ebeahan:
As documented today `network.forwarded_ip` is intended to be the original client host IP of the connection being proxied. My understanding is the device/service/process which is performing the proxying would be the authoritative source for this field (@dainperkins @webmat can help keep me honest if that's inaccurate 😄 ) .

For example - the following flow of a HTTP request:

[client] => [proxyA] => [originServer]

From `proxyA`'s perspective, `network.forwarded_ip` would equal `client` IP.

`proxyA` (if configured to do so) would append `client` IP to an `X-Forwarded-For` HTTP header in the HTTP request proxied to `originServer`.

In a more complex scenario:

[client] => [proxyA] => [proxyB] => [originServer]

[...]

PR #880, opened 2020-06-28
added http.forwarded_for #874

dainok:
Added http.forwarded_for for HTTP X-Forwarded-For header. See https://github.com/elastic/ecs/issues/874

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,7 +16,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Added `http.forwarded_for`. #874
+* Added `http.forwarded_for`, `http.forwarded_host`, `http.forwarded_proto`. #874
 
 #### Improvements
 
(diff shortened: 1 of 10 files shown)


andrewstucki:
Not that I see a need to necessarily introduce them, but wonder why the singling out of `X-Forwarded-For`? If we wanted to introduce defacto standard proxy header support shouldn't we add `X-Forwarded-Host` and `X-Forwarded-Proto` at the same time?

dainok:
@andrewstucki you're right, here the additional fields.

webmat:
Thanks for submitting this, @dainok! I agree we need to address the need for these HTTP headers in ECS.

I think we should focus on adding support for all HTTP headers, rather than the X-Forwarded family however. This has been discussed here https://github.com/elastic/ecs/pull/554 in the past but the discussion petered out at the time.

We'll see if we can get some attention to that.

dainok:
@webmat Ok, let me know if I can do more. Close the PR if you wish.

PR #882, opened 2020-07-01
schemas/tracing.yml: add span.id

axw:
Add `span.id` to the tracing fields. This field may be useful for correlating logs with more specific operations, or where `transaction.id` is unavailable.

The `transaction.id` field is not always available for trace data received from other tracing systems, such as Jaeger and OpenTelemetry. We are guaranteed to have `trace.id` and `span.id`, but not necessarily `transaction.id`. Correlating logs with an entire trace (via `trace.id`) may result in too broad a set of results, compared to correlating with `span.id`.

### DIFF
--- a/code/go/ecs/tracing.go
+++ b/code/go/ecs/tracing.go
@@ -34,4 +34,9 @@ type Tracing struct {
 	// A transaction is the highest level of work measured within a service,
 	// such as a request to a server.
 	TransactionID string `ecs:"transaction.id"`
+
+	// Unique identifier of the span.
+	// A span represents an operation within a transaction, such as a request
+	// to another service, or a database query.
+	SpanID string `ecs:"span.id"`
 }
--- a/code/go/ecs/tracing.go
+++ b/code/go/ecs/tracing.go
@@ -30,12 +30,12 @@ type Tracing struct {
 	// services.
 	TraceID string `ecs:"trace.id"`
 
-	// Unique identifier of the transaction.
+	// Unique identifier of the transaction within the scope of its trace.
 	// A transaction is the highest level of work measured within a service,
 	// such as a request to a server.
 	TransactionID string `ecs:"transaction.id"`
 
-	// Unique identifier of the span.
+	// Unique identifier of the span within the scope of its trace.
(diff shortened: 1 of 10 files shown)


felixbarny:
Does it mean we have to change the query that correlates traces and logs so that we also query for `span.id: $transaction.id` in the Logs UI?

axw:
@felixbarny not sure I follow, so I'll expand a bit on the problem.

OpenTelemetry has no concept of transaction; a trace is just a collection of causally-ordered spans. Spans have a "kind", e.g. Server (i.e. server-side of a request) and Client (client-side of a request). We convert Server kind spans into transactions, but there's no reference to the that span's ID in other service-local spans, aside from its children. Even then we have no way of knowing the parent is the local root.

So for OpenTelemetry, the best we can expect to have is:

 - `trace.id` and `transaction.id` OR
 - `trace.id` and `span.id`

We probably want to extend the Logs UI to query for `span.id: $span.id` if that field exists in the log.

felixbarny:
So far, we only correlate based on the `trace.id`. For the `Trace logs` action, I think we wouldn't want to change this.

When jumping from the logs UI to the trace view, we currently also just consider the `trace.id`. 

So I guess we don't need immediate changes. But we'll have to consider this for a potential logs tab in the transaction view.

code/go/ecs/tracing.go
```diff
@@ -34,4 +34,9 @@ type Tracing struct {
 	// A transaction is the highest level of work measured within a service,
 	// such as a request to a server.
 	TransactionID string `ecs:"transaction.id"`
+
+	// Unique identifier of the span.
```
simitt, line 38: nit: should we rephrase to _Unique identifier of the span per trace_?

axw: I'd be happy to change `transaction.id` and `span.id` descriptions to read

> Unique identifier of the transaction|span within the scope of its trace.

@felixbarny WDYT?

felixbarny: Nice catch Silvia, SGTM

axw: Updated, thanks for the suggestion.

PR #884, opened 2020-07-08
Usage improvements

ebeahan:
Adds `USAGE.md` covering the getting started experience and usage of the ECS tooling. The current `generator.py` arguments are documented along with example usage.

The aim is to simplify the experience for users generating their own artifacts (based off the published or custom schema definitions).

`USAGE.md` is linked from `README.md`

**Caveats**

* Once the changes from #873 are merged, the `--subset` usage will need to be updated to reflect the modified `subset` file format and improvements

Addresses #746

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -1,0 +1,253 @@
+# ECS Tooling Usage
+
+In addition to the published schema and generated artifacts, the ECS repo also contains tools to generate the artifacts based on the current schema. These tools aid in the development of the schema as well as validation in the ECS CI pipeline.
+
+These tools are experimental and should be used for custom indices only.
+
+## Setup and Install
+
+### Prerequisites
+
+ECS requires the following tools:
+
+* [Python 3.6+](https://www.python.org/)
+* [make](https://www.gnu.org/software/make/)
+* [pip](https://pypi.org/project/pip/) package installer
+* [git](https://git-scm.com/)
+
+#### Clone from GitHub
+
+The recommended way to download the ECS tools is by cloning the ECS repo using `git`:
+
+```
+git clone https://github.com/elastic/ecs
+```
+
+Prior to installing dependencies or running the tools, it's recommended to check out the `git` branch for the ECS version being targeted.
+
+**Example**: For ECS `1.5.0`:
+
+```
+$ git checkout 1.5
+```
+
+#### Option 1: Install dependencies via make
+
+Setting up a `virtualenv` (`venv`) can be accomplished by running from the top-level of the ECS repo:
+
+```
+$ make ve
+```
+
+`make ve` will also install all necessary Python dependencies using `pip`.
+
+#### Option 2: Install dependencies via pip
+
+It is recommended to create a `virtualenv` to separate ECS' project dependencies from your Python system dependencies. Activating the `venv` can be done using the activation script:
+
+```
+$ source ./venv/bin/activate
+```
+
+
+Install dependencies using `pip` from within the activated `venv`:
+
+```
+$ pip install -r scripts/requirements.txt
+```
+
+## Getting Started
+
+### Generate Artifacts
+
+The [generator](scripts/generator.py) script generates the various artifacts based on the [current](schemas) ECS schema.
+
+```
+$ python scripts/generator.py
+Loading schemas from local files
+Running generator. ECS version 1.6.0-dev
+```
+
+**Points to note**:
+
+* By default, artifacts are created in the [`generated`](generated) directory and the entire schema is included
+* Also by default, documentation updates will be written to the appropriate file under the `docs` directory. More specifics on generated doc files is covered in the [contributor's file](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md#generated-documentation-files)
+* Each run of the script will rewrite the entirety of the `generated` directory
+* The script will need to be executed from the top-level of the ECS repo
+* The `version` displayed when running `generator.py` is based on the current value of the [version](version) file in the top-level of the repo
+
+Many of these defaults options can be overridden using command-line arguments.
+
+### Generator Options
+
+#### Out
+
+Generate the ECS artifacts in a different directory:
+
+```
+$ python scripts/generator.py --out ../myproject/ecs/out/
+```
+
+Inside the directory passed in as the target dir to the `--out` flag, two directories, `generated` and `docs`, will be created. `docs` will contain three asciidoc files based on the contents of the provided schema. `generated` will contain the various artifacts laid out as in the published repo (`beats`, `csv`, `ecs`, `elasticsearch`).
+
+#### Include
+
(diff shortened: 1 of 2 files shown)


ebeahan:
@jonathan-buttner @marshallmain Would you take a pass and let me know if you have any feedback on the usage docs for some of the `generator` features you've added? (#737 #856 #748 #851)

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+* The script will need to be executed from the top-level of the ECS repo
+* The `version` displayed when running `generator.py` is based on the current value of the [version](version) file in the top-level of the repo
+
+The following options add functionality beyond the defaults.
+
+### Generator Options
+
+#### Out
+
+Generate the ECS artifacts in a different output directory. The argument expects to be passed an existing directory:
```
jonathan-buttner, line 106: The generator script will actually create the path if it doesn't exist: https://github.com/elastic/ecs/blob/master/scripts/generator.py#L44

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+Inside the directory passed in as the target dir to the `--out` flag, two directories, `generated` and `docs`, will be created. `docs` will contain three asciidoc files based on the contents of the provided schema. `generated` will contain the various artifacts laid out as in the published repo (`beats`, `csv`, `ecs`, `elasticsearch`).
+
+#### Include
+
+Use the `--include` flag to generate ECS artifacts based on the current ECS schema field definitions plus provided custom fields:
+
+```
+$ python scripts/generator.py --out ../myproject/ecs/custom-fields/
```
jonathan-buttner, line 119: I think this might have been a copy/paste issue `--out` should be `--include` here.

USAGE.md
ebeahan: Thanks for catching! Fixed.

USAGE.md
ebeahan: Fixed.

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+$ cd ecs
+```
+
+Prior to installing dependencies or running the tools, it's recommended to check out the `git` branch for the ECS version being targeted.
+
+**Example**: For ECS `1.5.0`:
+
+```
+$ git checkout v1.5.0
+```
```
webmat, line 60: Now that we have `--ref` we don't need this step anymore :-)

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+
+**Points to note on the defaults**:
+
+* Artifacts are created in the [`generated`](generated) directory and the entire schema is included
+* Documentation updates will be written to the appropriate file under the `docs` directory. More specifics on generated doc files is covered in the [contributor's file](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md#generated-documentation-files)
+* Each run of the script will rewrite the entirety of the `generated` directory
+* The script will need to be executed from the top-level of the ECS repo
+* The `version` displayed when running `generator.py` is based on the current value of the [version](version) file in the top-level of the repo
+
+The following options add functionality beyond the defaults.
```
webmat, line 100: We could add here that the ECS team maintains the official ECS artifacts by running the script in this way.

Then we can segue from that to explain "and now here's how you add your own fields, or remove the fields you don't need"

USAGE.md
```diff
@@ -0,0 +1,274 @@
+# ECS Tooling Usage
+
+In addition to the published schema and generated artifacts, the ECS repo also contains tools to generate artifacts based on the current published and custom schemas.
```
webmat, line 3: We should expand more here.

Q. Why do people care? 
A.

* ECS has too many fields
* ECS doesn't have their custom fields.

=> Users can use ECS tools to fix both of these things in their project. They can maintain only their customizations and use these tools to produce relevant artifacts for their various data sources.

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+
+#### Out
+
+Generate the ECS artifacts in a different output directory. If the specified directory doesn't exist, it will be created:
+
+```
+$ python scripts/generator.py --out ../myproject/ecs/out/
+```
+
+Inside the directory passed in as the target dir to the `--out` flag, two directories, `generated` and `docs`, will be created. `docs` will contain three asciidoc files based on the contents of the provided schema. `generated` will contain the various artifacts laid out as in the published repo (`beats`, `csv`, `ecs`, `elasticsearch`).
```
webmat, line 112: As soon as either `--include` or `--subset` are used, the asciidoc is not generated, though.

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+
+#### Ref
+
+The `--ref` argument allows for passing a specific `git` tag (e.g. `v.1.5.0`) or commit hash (`1454f8b`) that will be used to build ECS artifacts.
+
+```
+$ python scripts/generator.py --ref v1.5.0
+```
+
+> Note: `--ref` does have a dependency on `git` being installed and all expected commits/tags fetched from the ECS upstream repo. This will unlikely be an issue unless you downloaded the ECS as a zip archive from GitHub vs. cloning it.
```
webmat, line 217: Good to mention 👍

USAGE.md
```diff
@@ -0,0 +1,274 @@
...
+    ],
+    "properties": {}
+}
+```
+
+For `template.json`, the `mappings` object is left empty: `{}`. Likewise the `properties` object remains empty in the `mapping.json` example. This will be filled in automatically by the script.
+
+#### Intermediate-Only
+
+The `--intermediate-only` argument is used for debugging purposes. It only generates the "intermediate files" of ECS without generating the rest of the artifacts.
```
webmat, line 274: We could mention them explicitly here, `ecs_flat.yml` and `ecs_nested.yml`

USAGE.md
webmat: In addition to this, we could specifically call out the two artifacts that I expect users will find most useful. The Elasticsearch template is expected to be most useful for users in general. But Beats contributors may instead be interested in the Beats-formatted YAML field definition files.

ebeahan:
> I'm providing a few thoughts here for a follow-up PR. That seemed the simplest way to provide this feedback

Thanks @webmat for the feedback! I hope this helps clear up some of the common confusion with using the tooling.

I'll work on including your feedback/suggestions in a follow-up PR 😄 . I need to update the documented file format for `--subset` from #873 as well.

USAGE.md
ebeahan: Great point, we should definitely promote `--ref` instead. 

I ran up against an issue (as I was going to update the docs 😄 ) trying to run the generator with `--ref` (#892), so I may hold off removing this until we squash that bug.

USAGE.md
ebeahan: These are great additions. Maybe another nice, future addition is tutorial-like walkthroughs showing end-to-end how to define your own schema customizations, generate artifacts, and then application in Beats or ES.

PR #890, opened 2020-07-20
[RFC] Wildcard Data Type

ebeahan:
### Summary

Phase 0 proposal for adopting the upcoming [wildcard](https://github.com/elastic/elasticsearch/pull/49993) field into ECS.

Phase 0 (Strawperson) Criteria:

- [x] Discuss with domain or subject matter experts the utility of these changes
- [x] Discuss with ECS team whether these changes seem appropriate for ECS


rfcs/0000-wildcard-data-type.md
```diff
@@ -0,0 +1,150 @@
...
+      level: extended
+      type: wildcard
+      short: Full command line that started the process.
+...
+      multi_fields:
+      - type: text
+        name: text
+```
+
+Note the existing `text` data type multi-field will remain to support tokenized searches.
```
webmat, line 51: I would amend this to say the following

```suggestion
Note the existing `text` data type multi-field will remain only if there is a need to support tokenized searches.
```

rfcs/0000-wildcard-data-type.md
```diff
@@ -0,0 +1,150 @@
...
+* Host names
+* Registry data
+
+#### Stack traces
+
+Program stack traces tend to be well-structured but with long text and varied contents. There are too many subtleties and application-specific patterns to map all of them accurately with ECS' field definitions. Better performing wildcard searches can help the user formulate their own queries easier and with a smaller performance hit.
+
+### Command-line execution
+
+Full command-line events often have a series of options or flags which may appear in any arbitrary order. Wildcard searches are often required to successfully identify events regardless of the ordering.
```
webmat, line 86: Actually to work around the ordering of the flags, using the array of keywords `process.args` is much better.

The power of having `wildcard` on `process.command_line` is instead to detect subtle patterns like multiple spaces, detect specific substrings, or specifically to take advantage of wildcard's respect of the order to filter for a flag and a specific argument passed to that flag.

rfcs/0000-wildcard-data-type.md
```diff
@@ -0,0 +1,150 @@
...
+create an option in the tooling to generate OSS compatible mappings that continue supporting the OSS licensed features of Elastic.
+
+## Concerns
+
+<!--
+Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
+-->
+
+As mentioned previously in the proposal, integrating the wildcard data type into ECS will be the first instance of a Elastic licensed feature being adopted by the schema. Until now ECS has relied only on OSS license features, but we do not want to limit ECS from support Elastic licensed features moving forward. ECS will remained OSS license even if the schema implements certain Elastic licensed features. When ECS adopts a feature available only under a license, it will be noted in the documentation.
```
webmat, line 138:
```suggestion
As mentioned previously in the proposal, integrating the wildcard data type into ECS will be the first instance of an Elastic licensed feature being adopted by the schema. Until now ECS has relied only on OSS licensed features, but we do not want to limit ECS from supporting Elastic licensed features moving forward. ECS will remain OSS licensed even if the schema implements certain Elastic licensed features. When ECS adopts a feature available only under a license, it will be noted in the documentation.
```

rfcs/0000-wildcard-data-type.md
```diff
@@ -0,0 +1,150 @@
...
+## People
+
+The following are the people that consulted on the contents of this RFC.
+
+* @ebeahan | author
+
+## Footnotes
+
+* [0] Wildcard queries on `text` fields are limited to matching individual tokens rather than the original value of the field.
+* [1] Keyword fields do not suffer from the tokening problems of `text` fields, but can suffer from slow performance with wildcard searching (especially with leading wildcards).
```
webmat, line 149: The tokenizing of `text` fields is not a bug, it's a feature ;-)

```suggestion
* [1] Keyword fields are not tokenized like `text` fields, so patterns can match multiple words. However they suffer from slow performance with wildcard searching (especially with leading wildcards).
```

rfcs/0000-wildcard-data-type.md
ebeahan: Thanks this is good context. I've updated my wording to better capture this detail.

ebeahan:
Merging to advance to stage 0.

Issue #508, opened 2019-07-29
`services` in organizational hierarchy 


jakauppila:
I'm curious how our organizational hierarchy of applications (or services as defined in ECS) could be translated to ECS to provide some higher level views of what's going on throughout the enterprise.

- Business Unit - Company is comprised of ~12 different lines of business, think 'personal lines' and 'consumer lines' 
- Asset - A collection of applications that comprise a system, each of these have an IT Manager and team associated with them, so typically the go-to place for determining who and how to contact when alerting
- Application - A 'service' as defined by ECS
- Environment - Collection of instantiations of the application; our assets have between 2-35 environments

Today, we capture a few as follows:

- _company.asset_
- _company.env_ 
- _company.appname_
- _company.asset-env-appname_

MikePaquette:
@jakauppila good topic.
We find a wide degree of variability in the way organizations organize their applications, and today there is no specific mapping called out in ECS.

Since this hierarchy is mostly particular to your organization, the most important thing will be consistency of use across your org.  You could always consider using custom fields, but there are several ECS fields that you could use to map this information.

- As you mentioned, your "Company.appname" maps to ECS `service.name`.
- You could map your "Business Unit" to ECS `organization.name`
- You could also consider making use of the ECS `group.*` fields for "Company.Asset"
- Finally you could use the ECS `labels` or `tags` fields to store your environment information, either as key-value pairs, or tags.

jakauppila:
I get that this hierarchy is particular to our org, but are there similar constructs to be considered within ECS especially as mappings to be consumed by Elastic SIEM are involved ?

I'd be slightly hesitant to map "Business Unit" to `organization.name` as that might be reserved for cases of M&A, at which point we would populate that with our company name and could differ with an acquisition. It could make sense if parent was added as discussed in https://github.com/elastic/ecs/issues/135. We'll have to talk internally how to handle that.

How is `group.*` typically mapped? While that could work, it doesn't feel particularly intuitive as a user to see that and make the connection to its meaning.

As for environment information, I'd probably be interested in a top-level field as discussed in https://github.com/elastic/ecs/issues/268. We also utilize environment names and types as mentioned there.

MikePaquette:
@jakauppila 

> I get that this hierarchy is particular to our org, but are there similar constructs to be considered within ECS especially as mappings to be consumed by Elastic SIEM are involved ?

[...]

Issue #268, opened 2018-12-11
Add new top level field "environment"


vbohata:
I propose to add a new top level field named "environment" as discussed here: #143 
This field should be reusable because usually it is needed just top level environment for everything but rarely host and app environment can be different.

I propose following fields:
- environment.name
  (user-defined environment name like "test", "sandbox", "quality", "production")
- environment.type
  (type of this environment like "test", "production")

Examples:
environment.name: sandbox
environment.type: test

environment.name: production
environment.type: production

Maybe there could be also environment.id for env unique identification and environment.kind with values like "onpremise", "cloud".

webmat:
Very interesting. I have always used environment as a single string (production, staging, development). But I've sometimes felt the need for environment types ("like prod").

jakauppila:
We also utilize environment names and types (nonprod/prod) and would see value in this.

SHolzhauer:
In my opinion there is value in this. We use multiple environments too and are already adding fields for this. If this could be incorporated in ECS and down the line usable in the kibana apps that would add amazing value; For example SIEM app to only work with `environment.type:production`

I would suggest to use the (as far as I know) widely used/known DTAP for the environment.type field.

**environment.name**  
User defined string for the environment, `test`, `sandbox`, `windows-development`, etc.  
**environment.type**  
Values: `development`, `test`, `acceptance`, `production`  
**environment.code**  
User defined code for the environment, eg: `tst`, `sbx`, `windev`, `prd`

ebeahan:
@cyrille-leclerc is currently working on an RFC for environment field(s) use cases.

jasonrhodes:
I'm just catching up with this discussion as well as the `service.environment` discussion in that RFC: https://github.com/elastic/ecs/pull/1070

I'm curious about how this "top level environment" field somehow made its way into the "service" block? From an infrastructure metrics perspective, it seems like environments may often live outside of services. If we want to show all hosts or pods based on environment, I don't think we could use `service.environment` for that.

Are there other conversations about environment elsewhere?

cc: @sorantis @kaiyan-sheng @cyrille-leclerc

BenB196:
Looking at this topic, is there any reason, that the `environment` in `service.environment`, can't be added as reusable value for other things like `host.environment` or `observer.environment`? Looking over the current stage 2 rfc, I don't see anything that would explicitly make it not work for other things outside of `service.*`.

cyrille-leclerc:
[...]

PR #891, opened 2020-07-21
[RFC] `service.environment` field

cyrille-leclerc:
Proposal to adopt the `service.environment` attribute to qualify the environment from which an event is emitted (e.g. `production`, `staging`, `qa`, `dev`).

### DIFF
--- a/rfcs/0001-rfc-environment.md
+++ b/rfcs/0001-rfc-environment.md
@@ -1,4 +1,4 @@
-# 0000: Name of RFC
+# 0001: Environment Field
 <!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
 
 - Stage: **0 (strawperson)** <!-- Update to reflect target stage -->
--- a/rfcs/0001-rfc-environment.md
+++ b/rfcs/0001-rfc-environment.md
@@ -1,0 +1,230 @@
+# 0000: Name of RFC
+<!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
+
+- Stage: **0 (strawperson)** <!-- Update to reflect target stage -->
+- Date: **TBD** <!-- Update to reflect date of most recent stage advancement -->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+## Fields
+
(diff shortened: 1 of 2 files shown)


webmat:
@cyrille-leclerc @roncohen 

For the stage 1 proposal I'd like the RFC to be adjusted in a few ways.

Could you also add a link to https://github.com/elastic/ecs/issues/508 in the references? It's not only about environments, but I think it's also relevant.

Please rephrase the proposal a bit more generally, as a proposal to add support for capturing environment information in ECS. APM's current approach is of course a top contender, since it's already in use. But I'd like the proposal to also consider the more flexible approach discussed in #268. In this proposal, the contributor suggests creating this as a field set with a few fields. In the simplest cases where a user has only 2 environments, they could use simply `environment.type: production|staging`. But this approach opens up much more flexibility for other situations. Common examples are production deployments in different settings (multi-tenancy, different locations like "on prem" vs "cloud", different purposes like "main" vs "canary"), etc.

I'd like to loop in someone from the Cloud team and see if more robust support for environment details would be interesting to them.

cyrille-leclerc:
See also 
* https://en.wikipedia.org/wiki/Deployment_environment
* https://github.com/open-telemetry/opentelemetry-specification/pull/606
* https://github.com/open-telemetry/opentelemetry-specification/issues/752

webmat:
Great additional resources, thanks for dropping these notes here.

When do you think you can open the PR for the next stage?

cyrille-leclerc:
> When do you think you can open the PR for the next stage?

I may have changed my mind brainstorming on https://github.com/open-telemetry/opentelemetry-specification/issues/752 .

I'm wondering if we should support defining the "deployment environment" at multiple level: on the host itself and on the "deployed service".

I would like to discuss this with @sorantis

sorantis:
Typically orgs/teams use custom tags for such information, which should be accomplished with labels available in the [base ECS reference](https://www.elastic.co/guide/en/ecs/current/ecs-base.html).

felixbarny:
@cyrille-leclerc do other observability teams plan to adopt `service.environment`?

cyrille-leclerc:
@felixbarny the standardisation of the "environment" qualifier is moving slowly.

Last time I discussed with @sorantis, we saw that 
* most beats (metricsbeat, filebeat) use custom tags for the moment,
* it's not easy for infrastructure to talk of "environment=prod/staging/..." as infrastructure is often multi tenant
* Using "service.*" iis often not what people are looking at to characterize their infrastructure.

Moreover, for application logs, we plan to solve the problem of characterising the environment having the application agent being deeply integrated with filebeat.

I have recently discussed this ECS RFC with @webmat , we are likely to just standardise at the application layer, excluding the infrastructure layer for the moment.

PR #904, opened 2020-07-31
[RFC] Wildcard - stage 1 proposal

ebeahan:
### Summary

Revisions to the stage 0 wildcard adoption for consideration to be accepted as a stage 1 proposal.

- [x] Opened pull request revising the existing strawperson
- [x] Types of fields or changes outlined

[Markdown preview of this RFC](https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-1/rfcs/text/0001-wildcard-data-type.md)

### DIFF
--- a/rfcs/text/0001-wildcard-data-type.md
+++ b/rfcs/text/0001-wildcard-data-type.md
@@ -291,6 +291,10 @@ ECS applies the `ignore_above` setting to keyword fields to prevent strings long
 
 Until now ECS has relied only on OSS licensed features, but ECS will also support Elastic licensed features. The ECS project will remain OSS licensed with the schema implementing Elastic licensed features as part of the specification. When ECS adopts a feature available only under a license, it will be noted in the documentation. ECS plans to provide tooling options which continue to support OSS consumers of ECS and the Elastic Stack.
 
+### Version Compatibility
+
+A data shipper which uses the `wildcard` field type may need to verify that the configured output Elasticsearch destination can support it (>= 7.9.0). For example, if a future version of Beats adopts `wildcard` in index mappings, Beats would may need to gracefully handle a scenario where the targeted Elasticsearch instance doesn't support the data type.
+
 ## People
 
 The following are the people that consulted on the contents of this RFC.
--- a/rfcs/text/0001-wildcard-data-type.md
+++ b/rfcs/text/0001-wildcard-data-type.md
@@ -109,7 +109,7 @@ Mapping definition:
 Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
 -->
 
-Wildcard is well-suited for cases requiring partial matching of string values across long unstructured or semi-structured fields. Often machine generated events, such as logs, metrics, and traces, aren't well suited for the analysis applied on `text` fields. Using keyword allows for exact value searching and introduces filtering, sorting, and aggregations. Keyword fields do also support regex and wildcard queries, however the field's search performance can become vary largely for certain queries (e.g. leading wildcard) and the cardinality of the data.
+Wildcard is well-suited for cases requiring partial matching of string values across long unstructured or semi-structured fields. Often machine generated events, such as logs, metrics, and traces, aren't well suited for the analysis applied on `text` fields. Using `keyword` allows for exact value searching and introduces filtering, sorting, and aggregations. Keyword fields do also support regex and wildcard queries, however the field's search performance can vary a lot, depending on the query (e.g. leading wildcard) and the cardinality of the data.
 
 Often ECS approaches semi-structured fields by breaking their values down into structured ones. These structured fields then enable better use of keyword fields' characteristics. For example, a URL can break down into its constituent parts: scheme, domain,  port, path, etc., and those parts can in turn be mapped to unique fields. However, not all fields are structured enough for this approach.
 
@@ -220,9 +220,9 @@ The goal here is to research and understand the impact of these changes on users
 
 Any component producing data (Beats, Logstash, third-party developed, etc.) will need to adopt the mappings in their index templates. The discussion around the handling of OSS vs. Basic field types between OSS and Basic modules/plugins will be handled outside of this proposal.
 
-### Usage
+### Usage mechanisms
 
-Part of the 7.9 release is the introduction of the first field family[2], `keyword`. Grouping field types by family is intended to eliminate backwards compatibility issues when replacing an old field with a new specialized type on time-based indices (e.g. `keyword` replaced with `wildcard`). The `wildcard` data type will return `keyword` in the output of the field capabilities API call, and this change will enable both types to behave identically at query time. This feature eliminates concerns arising from Kibana's field compatibility checks in index patterns.
+Part of the 7.9 release is the introduction of the keyword family[2]. Grouping field types by family is intended to eliminate backwards compatibility issues when replacing an older field type with a new, more specialized type on time-based indices (e.g. `keyword` replaced with `wildcard`). The `wildcard` data type will return `keyword` in the output of the field capabilities API call, and this change will enable both types to behave identically at query time. This feature eliminates concerns arising from Kibana's field compatibility checks in index patterns..
 
 ### ECS project
 
@@ -234,10 +234,6 @@ ECS is and will remain an open source licensed project. However, there will be f
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
 
-### Licensing
-
-Until now ECS has relied only on OSS licensed features, but ECS will also support Elastic licensed features. The ECS project will remain OSS licensed with the schema implementing Elastic licensed features as part of the specification. When ECS adopts a feature available only under a license, it will be noted in the documentation. ECS plans to provide tooling options which continue to support OSS consumers of ECS and the Elastic Stack.
-
 ### Performance differences
 
 Performance and storage characteristics between wildcard and keyword will be different[3], and this difference may have an impact depending on deployment size and/or the level of duplication in the field data. As part of the transition, fields which were previously indexed keyword will be switched to wildcard. Queries across indices with field names will be necessary. Understanding the differences at both index and query time will be pursued.
@@ -245,6 +241,10 @@ Performance and storage characteristics between wildcard and keyword will be dif
 ### Wildcard field value character limits
 
 ECS applies the `ignore_above` setting to keyword fields to prevent strings longer than 1024 characters from being indexed or stored. While `ignore_above` can be raised, Lucene implements a term byte-length limit of 32766 which cannot be adjusted. Wildcard supports an unlimited max character size for a field value. The `wildcard` field type will still have the `ignore_above` option available, and a reasonable limit may be need applied to mitigate unexpected side-effects.
+
+### Licensing
+
+Until now ECS has relied only on OSS licensed features, but ECS will also support Elastic licensed features. The ECS project will remain OSS licensed with the schema implementing Elastic licensed features as part of the specification. When ECS adopts a feature available only under a license, it will be noted in the documentation. ECS plans to provide tooling options which continue to support OSS consumers of ECS and the Elastic Stack.
 
 ## People
 
--- a/rfcs/text/0001-wildcard-data-type.md
+++ b/rfcs/text/0001-wildcard-data-type.md
@@ -30,7 +30,7 @@ For a field to use wildcard, it will require changing the the field's defined sc
 
 Here's an example of applying this change to the `process.command_line` field:
 
-**Current definition as of ECS 1.5.0**
+**Definition as of ECS 1.5.0**
 
 Schema definition:
 
--- a/rfcs/text/0001-wildcard-data-type.md
+++ b/rfcs/text/0001-wildcard-data-type.md
@@ -234,9 +234,15 @@ ECS is and will remain an open source licensed project. However, there will be f
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
 
+### Licensing
+
 Until now ECS has relied only on OSS licensed features, but ECS will also support Elastic licensed features. The ECS project will remain OSS licensed with the schema implementing Elastic licensed features as part of the specification. When ECS adopts a feature available only under a license, it will be noted in the documentation. ECS plans to provide tooling options which continue to support OSS consumers of ECS and the Elastic Stack.
 
+### Performance differences
+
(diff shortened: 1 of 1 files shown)


ebeahan:
Seeking feedback from folks on the wildcard adoption RFC. I've noted a few areas needing discussion before advancing as stage one: 

- Any concerns with the outlined approach or any critical missing use cases that should be captured?
- Any concerns or implementation challenges overlooked?
- Any field sets not mentioned that should be represented? (Any specifics on particular fields is also welcomed.)

@webmat @jamiehynds @epixa @rw-access @randomuserid @leehinman @jonathan-buttner

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,114 @@ Proposed type change:
...
+A final interesting property of wildcard is its unlimited max character field size. Elasticsearch keyword field default to a max allowed string size of 256 characters which ECS ups to 1024 using `ignore_above`. Keyword can't be increased infinitely because Lucene has a hard set max limit of 32766 bytes for a single term. Due to how wildcard strings are indexed, they do not share this limitation. This makes wildcard the ideal data type for a very large string field (>32kB) that still needs to be indexed.
 
-* File paths
+The next sections details use cases which could benefit wildcard.
+
+### Paths
+
+Flexible nesting of a file path: `file.path:*\\Users\\*\\Temp\\*`
+Match under registry path: `registry.path:\\HKLM\\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\*\Debugger`
+A unique URL path: `url.full:https://api.example.com/account/*/foobar`
```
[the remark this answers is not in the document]
webmat: This applies to lines 138 and 139 below as well (Flexible... and Likewise...). Noting this here, to keep separate from that other thread I'm opening on lines 138-139 :-)

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,114 @@ Proposed type change:
...
+at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1563)
+at org.elasticsearch.client.IndicesClient.getMapping(IndicesClient.java:282)
+
+Caused by: java.lang.invoke.LambdaConversionException: Invalid receiver type interface org.apache.http.Header; not a subtype of implementation type interface org.apache.http.NameValuePair
+at java.base/java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:254)
+at java.base/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:327)
+at java.base/java.lang.invoke.BootstrapMethodInvoker.invoke(BootstrapMethodInvoker.java:127)
+```
+
+If I wanted to look for similar events that also contain the phrase `lambda$performRequestAndParseEntity$9(RestHighLevelClient.java`, I'd need a field that supports searching in the middle of a string. Keyword would perform poorly and text would require rethinking the query to match the analyzer and tokenization applied at index time.
```
webmat, line 175: Let's avoid first person. E.g. 

> When looking for similar events...

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,114 @@ Proposed type change:
...
+
 ### Command-line execution
 
-Building accurate queries again a full command line string field like the `process.command_line` field, also benefit from wildcard improvements:
+The arguments, order of those arguments, and values passed can be arbitrary in a command-line execution. If searching across multiple arguments, retaining their ordering, and/or the argument-value pairing is key to the search criteria, multiple wildcards patterns may be needed in a single query (if the presence of the arguments/values is the only criteria regardless of ordering or pairing, using a structured field such as `process.args` would be preferred). Wildcard searching such an unstructured field indexed as keyword, like `process.command_line`, can cause performance challenges.
```
webmat, line 179: Great info. I would have the whole parens on `.args` as a distinct sentence, rather than in parens.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -115,56 +216,63 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
...
 
 ### Usage
 
-Part of the 7.9 release is the introduction of the first field family[2], `keyword`. Grouping field types by family is intended to eliminate backwards compatibility issues when replacing an old field with a new specialized type on time-based indices (e.g. `keyword` replaced
-with `wildcard`). The `wildcard` data type will return `keyword` in the output of the field capabilities API call, and this change will enable both types to behave identically at query time. This feature eliminates concerns arising from Kibana's field compatibility checks
-in index patterns.
+Part of the 7.9 release is the introduction of the first field family[2], `keyword`. Grouping field types by family is intended to eliminate backwards compatibility issues when replacing an old field with a new specialized type on time-based indices (e.g. `keyword` replaced with `wildcard`). The `wildcard` data type will return `keyword` in the output of the field capabilities API call, and this change will enable both types to behave identically at query time. This feature eliminates concerns arising from Kibana's field compatibility checks in index patterns.
```
webmat, line 225: Good idea to explain field families here. Suggesting a few wording tweaks

```suggestion
Part of the 7.9 release is the introduction of the keyword family[2]. Grouping field types by family is intended to eliminate backwards compatibility issues when replacing an older field type with a new, more specialized type on time-based indices (e.g. `keyword` replaced with `wildcard`). The `wildcard` data type will return `keyword` in the output of the field capabilities API call, and this change will enable both types to behave identically at query time. This feature eliminates concerns arising from Kibana's field compatibility checks in index patterns.
```

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -115,56 +216,63 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
...
 
-ECS-compliant data consumers should not be impacted due to the `keyword` field family[2], however ECS-compliant producers will be required for their field mappings to use `wildcard` where defined by the spec. The ECS project will need to evaluate how best to
-evolve its tooling to support `wildcard`; wildcard will be the first time a Elastic licensed feature is specified in ECS.
-
 ### Ingestion
 
-Any component producing data (Beats, Logstash, third-party developed, etc.) will need to comply with the new mappings.
+Any component producing data (Beats, Logstash, third-party developed, etc.) will need to adopt the mappings in their index templates. The discussion around the handling of OSS vs. Basic field types between OSS and Basic modules/plugins will be handled outside of this proposal.
 
 ### Usage
```
webmat, line 223: We already have another heading named "Usage", this threw me for a bit. Can we rename this heading?

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -115,56 +216,63 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
...
 
 ## Concerns
 
 <!--
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
 
-As mentioned previously in the proposal, integrating the wildcard data type into ECS will be the first instance of an Elastic licensed feature being adopted by the schema. Until now ECS has relied only on OSS licensed features, but we do not want to limit ECS from supporting Elastic licensed features moving forward. ECS will remain OSS licensed even if the schema implements certain Elastic licensed features. When ECS adopts a feature available only under a license, it will be noted in the documentation.
+### Licensing
```
webmat, line 237: Let's start with the technical concerns here, and have the Licensing at the end of this section

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -115,56 +216,63 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
...
 
 ### ECS project
 
-ECS is and will remain an open source licensed project. However, there will be features available under the Elastic license that will benefit the user experience with the Elastic stack and solutions that have a place in the ECS specification. The ECS project's tooling will
-create an option in the tooling to generate OSS compatible mappings that continue supporting the OSS licensed features of Elastic.
+ECS is and will remain an open source licensed project. However, there will be features available under the Elastic license that will benefit the user experience with the Elastic stack and solutions that have a place in the ECS specification. The ECS project's tooling will create an option in the tooling to generate OSS compatible mappings that continue supporting the OSS licensed features of Elastic. The discussion and implementation of this functionality will take place outside this proposal.
 
 ## Concerns
```
webmat, line 231: I think we should add one more concern to the section, perhaps even as the first subsection. A big driver for all of this effort was actually the combination of improving wildcard/regex performance, but also offering a better story around case insensitivity.

The latter is being addressed independently as a regexp query option, and afaik it's not out yet. I think this should get at least a cursory mention here.

(The rest of this comment isn't for the "concerns" section per se, but a continuation of my thoughts surrounding this)

We will probably even take case insensitivity into account in the whole strategy for adoption of `wildcard`:

* not having case insensitivity yet means `.text` multi-fields continue being relevant for a while
* removing `.text` multi-fields is a breaking change, so where this removal is applicable, it be done will be ECS 2.0, to be introduced with Elastic Stack 8.0...

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,114 @@ Proposed type change:
...
+```
+
+Additional cases for wildcard searching against command line executions:
 
-* Multiple spaces
-* Isolating specific substrings
-*
+* Multiple spaces in the command line execution
+* Isolating specific substrings where ordering matters
+* command obfuscation
```
webmat, line 192: I love how the Usage section is broken down into big families of use cases for wildcard.

Perhaps we could add one last section that's not a use case, but that is aimed at helping people make up their mind whether they should use `keyword` or `wildcard`?

And by this I mean including @markharwood's awesome "Comparison with keyword fields" from [the blog post](https://www.elastic.co/blog/find-strings-within-strings-faster-with-the-new-elasticsearch-wildcard-field), as well as the decision tree diagram?

rfcs/text/0001-wildcard-data-type.md
ebeahan: Absolutely 👍

rfcs/text/0001-wildcard-data-type.md
ebeahan: Yes I can see your point; it should be mentioned prominently.

Fact checking my understanding as I re-review past discussion (e.g. https://github.com/elastic/elasticsearch/issues/53603#issuecomment-629280978), would stating `wildcard` will be the preferred type over `keyword` for fields needing frequent case-insensitive queries be accurate? At least from ECS' perspective?

webmat: I think @markharwood would be best to answer this.

The fact that this will be part of the regexp query means it'll be just as functional on `keyword` fields.

However I don't know for a fact whether it will be more performant on `wildcard`. I assume it will, since iirc the ngram backing `wildcard` is lowercased.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,152 @@ Proposed type change:
...
+
+The following are common categories
+
+* File and directory paths
 * URLs
+* Registry data
+
+#### Names
+
+Similar to paths, different names components many need to be searched using one or more wildcard.
```
webmat, line 174: Not sure, but is "many" a typo, there?

```suggestion
Similar to paths, different names components may need to be searched using one or more wildcard.
```

webmat:
> the review comment discussing hostname, email, and username. I'm leaning towards removing all three from that section.

Let's leave it in for now. In stage 2, when we start doing a comprehensive list of all fields we consider migrating, we can come to a decision on this, and remove that section if appropriate.

rfcs/text/0001-wildcard-data-type.md
ebeahan: Yes thank you! 👀

ebeahan:
@cyrille-leclerc @leehinman following up from our discussion last week, would you be able to briefly review and weigh in on the utility of these changes and any concerns or complexity that isn't captured?

rfcs/text/0001-wildcard-data-type.md
webmat: Let's move forward here and assume it will. Let's capture this as a question that needs an answer during stage 2, since this is where we will firm up the precise list of fields to migrate.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -48,46 +81,114 @@ Proposed type change:
...
+* File and directory paths
 * URLs
+* Registry data
+
+### Names
+
+Similar to paths, different names components many need to be searched using one or more wildcard.
+
+Flexible host name searching: `host.name:prod-*-db*`
+Likewise, flexible searching of user names or accounts: `user.email:foo*@example.com`
```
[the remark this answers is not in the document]
webmat: For usernames I think we'll want to stay on `keyword`, because usernames are typically a single word or two, and I'd be surprised that searching for parts of a username is a common thing.

Let's still keep the "Name" section however. I do think some of them may benefit. Upon further reflection, email may be a good fit (e.g. get me all of the gmail addresses: `email:*@gmail.com`), OS names as well, since they can be somewhat messy, and possibly domains names.

I think in stage 2 we should expand this section a bit, and try to capture some of the thinking on which names may benefit most.

webmat (dismissed):
I think we have a good coverage of the potential concerns. Let's check that box :-)

For my part, I'm good with this stage 1 proposal as it is 🚢 

We're only waiting for a quick thumbs up / thumbs down from observability and security

@leehinman and @cyrille-leclerc (or representatives), what we're looking for at stage 1 is simply whether or not we've missed a big concern from your point of view, and whether or not this is heading in the right direction. Stage 2 is when we will hammer out the precise list of fields to migrate, and that's therefore when we will need more thorough reviews. For now I think a cursory review is enough.

rfcs/text/0001-wildcard-data-type.md
ebeahan: Thanks @webmat - captured.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -115,56 +254,70 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
 The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
 -->
 
-ECS-compliant data consumers should not be impacted due to the `keyword` field family[2], however ECS-compliant producers will be required for their field mappings to use `wildcard` where defined by the spec. The ECS project will need to evaluate how best to
-evolve its tooling to support `wildcard`; wildcard will be the first time a Elastic licensed feature is specified in ECS.
-
 ### Ingestion
 
-Any component producing data (Beats, Logstash, third-party developed, etc.) will need to comply with the new mappings.
+Any component producing data (Beats, Logstash, third-party developed, etc.) will need to adopt the mappings in their index templates. The discussion around the handling of OSS vs. Basic field types between OSS and Basic modules/plugins will be handled outside of this proposal.
```
leehinman, line 260: Another important item outside the scope of this proposal is backwards compatibility.  Producers will need to consider the case where the producer has adopted wildcard, but the Elasticsearch instance they are sending data to doesn't support wildcard (example ES 6.8).

leehinman:
+1 for moving to Stage 2

roncohen:
generally LGTM 👍

PR #907, opened 2020-08-04
Add event.reason field

### LINKED ISSUE #613: event reason?
Regarding

`event.category / event.action / event.outcome`

I think we also need `event.reason`

For example:

```
- event.category: user-management
- event.action: user-password-change
- event.outcome: failure
- event.reason: password policy violation
```

ebeahan:
**Summary**

Adds a new field under `event.*`, `event.reason`, which provides a location to capture the reason _why_ an event occurred.

**Example usage**

* A web proxy event shows a denied connection request. `event.action:"request-denied"` describes the action captured and why the action took place is captured in `event.reason:"Blocked site: example.com"`
* An endpoint agent terminates a process on a host system. `event.action:"process-terminated" with reason describing why as `event.reason:"Terminated an unexpected process"` 

**Discussion Points**

Two items I wanted to call to attention: 

1. `event.reason` is currently defined as `type: extended`. This felt most appropriate starting off.
2. A `text` multi-field is included. `event.reason` could possibly hold a longer string (e.g. `File Sharing site URL blocked`).

Closes #613

### DIFF
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -704,3 +704,19 @@
         Alert events, indicated by `event.kind:alert`, are a common use case for this field.
 
       example: https://mysystem.mydomain.com/alert/5271dedb-f5b0-4218-87f0-4ac4870a38fe
+
+    - name: reason
+      level: extended
+      type: keyword
+      short: The reason captured by the event
+      description: >
+        The reason captured by the event.
+
+        This describes the why of a particular action or outcome captured in the event. Where
+        `event.action` captures the action from the event, `event.reason` describes why that action
+        was taken. For example, a web proxy with an `event.action` which denied the request may also
+        populate `event.reason` with the reason why (e.g. `blocked site`).
+      example: "Terminated an unexpected process"
+      multi_fields:
+      - type: text
+        name: text
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -708,9 +708,9 @@
     - name: reason
       level: extended
       type: keyword
-      short: The reason captured by the event
+      short: Reason why this event happened, according to the source
       description: >
-        The reason captured by the event.
+        Reason why this event happened, according to the source.
 
         This describes the why of a particular action or outcome captured in the event. Where
         `event.action` captures the action from the event, `event.reason` describes why that action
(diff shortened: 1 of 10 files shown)


schemas/event.yml
```diff
@@ -704,3 +704,19 @@
         Alert events, indicated by `event.kind:alert`, are a common use case for this field.
 
       example: https://mysystem.mydomain.com/alert/5271dedb-f5b0-4218-87f0-4ac4870a38fe
+
+    - name: reason
+      level: extended
+      type: keyword
+      short: The reason captured by the event
```
webmat, line 711: Perhaps we should phrase this more naturally. E.g. 

"Reason why this event happened, according to the source"

schemas/event.yml
```diff
@@ -704,3 +704,19 @@
...
+        The reason captured by the event.
+
+        This describes the why of a particular action or outcome captured in the event. Where
+        `event.action` captures the action from the event, `event.reason` describes why that action
+        was taken. For example, a web proxy with an `event.action` which denied the request may also
+        populate `event.reason` with the reason why (e.g. `blocked site`).
+      example: "Terminated an unexpected process"
+      multi_fields:
+      - type: text
+        name: text
```
webmat, line 722: Do we expect `event.reason` to be complex often enough to warrant this? Not opposed to this, just wondering out loud :-)

ebeahan: I had a similar internal debate...

I can see value in having the `text` indexing available since `event.reason` could possibly be longer string (e.g. something like [this Okta event](https://github.com/elastic/beats/blob/master/x-pack/filebeat/module/okta/system/test/okta-system-test.json.log-expected.json#L196)), but the extra complexity may not be warranted the majority of the time.

As I type this out, I'm leaning towards removing the multi-field. If a user was trying to search for allowed events across sources using `event.reason.text:allow`, using the categorization fields like `event.type:"allowed" would be the better practice to encourage I think.

webmat: Let's remove it for now. Adding `.text` later is non-breaking, so it'll be fine if it becomes necessary.

Also, I think this will be one of the fields we move to `wildcard` ultimately. I'm sure we'll get to `wildcard` before there's a need to add the `.text` interim solution.

webmat:
CI fails because you're missing a `make` run. I think after `.text` is removed this will be ready.

ebeahan:
@webmat should be good to merge if you wanna give it once last 👀

PR #910, opened 2020-08-06
Use appropriate reserved domain for example URLs

ebeahan:
**Summary**

Two minor documentation changes for the `event.*` field set:

* ~Removed the "live" URL links by adding quotes to the string.~ This is actually incorrect. Our documentation generator is handling this by always wrapping example strings with backticks (`). As long as we ensure the examples don't use potentially valid domains, I don't think we need to worry about being "defensive" for other doc locations which may render a hyperlink for the URLs (e.g. [Beats](https://www.elastic.co/guide/en/beats/filebeat/current/exported-fields-ecs.html)).

* Switched example URLs to use the IANA reserved `*.example.com`

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -40,6 +40,7 @@ Thanks, you're awesome :-) -->
   `log.syslog`, `network.inner`, `observer.egress`, and `observer.ingress`. #871
 * Improved attribute `dashed_name` in `generated/ecs/*.yml` to also
   replace `@` with `-`. #871
+* Updated several URLs in the documentation with appropriate special-use domain. #910
 
 #### Deprecated
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -40,7 +40,7 @@ Thanks, you're awesome :-) -->
   `log.syslog`, `network.inner`, `observer.egress`, and `observer.ingress`. #871
 * Improved attribute `dashed_name` in `generated/ecs/*.yml` to also
   replace `@` with `-`. #871
-* Updated several URLs in the documentation with appropriate special-use domain. #910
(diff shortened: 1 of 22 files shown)


ebeahan:
I noted the other places using real domains:

| Field(s) | Example Value |
| ----- | ------------- |
| `*.registered_domain` |`www.google.com` |
| `package.reference` | `https://golang.org` |
| `tls.client.server_name`, `url.domain` | `www.elastic.co` |
| `x509.issuer.organizational_unit` | `www.digicert.com` |
| `x509.subject.common_name` | `r2.shared.global.fastly.net` |
| `x509.subject.distinguished_name` | `C=US, ST=California, L=San Francisco, O=Fastly, Inc., CN=r2.shared.global.fastly.net` |

While it's top of mind, should we unify all of our example usages to `example.com` for uniformity?

webmat:
Whenever you work on some of these domains, make sure to adjust the domain breakdowns as well. `www.example.com` => `.registered_domain: example.com`, `.top_level_domain: com`

ebeahan:
@webmat this is ready for another 👀 

Should we plan on backporting these changes to 1.5?

CHANGELOG.next.md
```diff
@@ -40,6 +40,7 @@ Thanks, you're awesome :-) -->
   `log.syslog`, `network.inner`, `observer.egress`, and `observer.ingress`. #871
 * Improved attribute `dashed_name` in `generated/ecs/*.yml` to also
   replace `@` with `-`. #871
+* Updated several URLs in the documentation with appropriate special-use domain. #910
```
webmat, line 43:
```suggestion
* Updated several URLs in the documentation with "example.com" domain. #910
```

Let's make the changelog entry a little more plain in what we're doing here :-)

Issue #909, opened 2020-08-05
Update `related.*` field names from singular to plural


ebeahan:
**Description of the issue:**

The `related.*` field set in ECS 1.x uses singular names. As suggested per an additional remark in #863 , using plural names would better align with the [guidelines](https://www.elastic.co/guide/en/ecs/master/ecs-guidelines.html).

**Any additional context or examples:**

`related.hash` => `related.hashes`
`related.ip` => `related.ips`
`related.user` => `related.users`
`related.host` => `related.hosts` (not final)

webmat:
After further thought and discussion, we feel like this inconsistency on a few fields is not worth the pain of a breaking change.

We want to keep our breaking change "budget" for more pressing matters :-)

PR #913, opened 2020-08-06
add related.hosts

### LINKED ISSUE #863: Add a related.host field
**Summary**

Add a `related.host` field to capture hostnames and other host identifiers.

**Motivation**:

We can already define related usernames and IPs for a document, so it seems very natural to also be able to search for related hostnames (some events, like windows eventlog code 4776 don't even provide IPs, but rather workstation names).

**Detailed Design**:

Why `related.host`, and not `related.hostname` or something else?
* The field should hold all types of text-based host identifiers, be that hostnames, dns names, dns aliases, or workstation names.
* This is in line with the `related.user` field, which also doesn't discriminate between full user names, user handles, or even user IDs.
* If you look at the `host` fieldset, the closest thing to my proposed field would be `host.name`. however `related.name` is too ambiguous, so i think `related.host` is a good fit.

A related issue might be the standardization of domain representations: #728 - this issue isn't completed yet, but I don't see it colliding with what I'm proposing here.

PS: If you ask me, the related fields should be plurals (`related.ips, related.users, related.hosts`), because they will often contain arrays (as per the [guidelines](https://www.elastic.co/guide/en/ecs/master/ecs-guidelines.html). But to keep in line with the already existing fields, I'm fine with `related.host`.

ebeahan:
**Summary**

Add a ~`related.host`~ `related.hosts` field. Same intention as the existing `related.*` fields: populate with all hostnames and other host identifiers from the event.

Closes #863

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -27,6 +27,7 @@ Thanks, you're awesome :-) -->
 * Added missing field reuse of `pe` at `process.parent.pe` #868
 * Added `span.id` to the tracing fieldset, for additional log correlation (#882)
 * Added `event.reason` for the reason why an event's outcome or action was taken. #907
+* Added `related.host` to capture all hostnames and host identifiers on an event. #913
 
 #### Improvements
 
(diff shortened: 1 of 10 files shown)


webmat:
cc @dcode also

webmat:
As discussed, let's pluralize this one. This way for 2.0 (see #909) we'll have only 3 fields to migrate, which will be less painful for users.

PR #917, opened 2020-08-11
Add field `user.roles`

### LINKED ISSUE #915: Add a `user.roles` field
**Summary**

When recording events performed by a user under RBAC, it is sometimes useful to record the set of roles the user had at the time the action was performed.

As an example, the Elasticsearch audit logs include this information, as described in [their docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/audit-event-types.html#_extra_audit_event_attributes_for_specific_events)

**Motivation**:

Kibana is revamping its audit logging, and as part of this work, we are transitioning to be fully ECS compliant. Our logs aim to record the set of roles the end-user is assigned, and having a first-class location for this information within ECS would be beneficial to us.

We have a work-in-progress which targets this information under a custom `kibana.user_roles` category, but having this information exist within ECS's `user` category would offer a much nicer experience. We do have other alternatives if you feel this shouldn't be included in ECS, but I wanted to start a discussion to explore our options first.

**Detailed Design**:

* Field names: `user.roles`
* Example values for the fields
```
{
   "user": {
      "name": "elastic",
      "roles": ["superuser"]
   }
}
```
* Suggested appropriate datatypes
I think `keyword` would be an appropriate datatype for this information.

Audit logs generated by both Elasticsearch and Kibana could benefit from this information

webmat:
This array field is meant to capture a list of role names that apply to the user, in an audit log or other RBAC event.

Closes #915

### DIFF
--- a/schemas/user.yml
+++ b/schemas/user.yml
@@ -81,3 +81,20 @@
           Name of the directory the user is a member of.
 
           For example, an LDAP or Active Directory domain name.
+
+    - name: roles
+      level: extended
+      type: keyword
+      normalize:
+        - array
+      short: Array of user roles at the time of the event.
+      description: >
+        Array of user roles or groups, at the time of the event.
+
+        `user.group.*` fields are meant to capture one group's full details,
+        rather than capturing an array of groups associated with a user.
+
+        When it's necessary to capture a list of roles or groups assigned to the
+        user at the time an event or audit log is recorded,
+        use the array field `user.roles`.
+      example: '["kibana_admin", "reporting_user"]'
--- a/schemas/user.yml
+++ b/schemas/user.yml
@@ -87,14 +87,6 @@
       type: keyword
       normalize:
         - array
-      short: Array of user roles at the time of the event.
       description: >
-        Array of user roles or groups, at the time of the event.
-
-        `user.group.*` fields are meant to capture one group's full details,
-        rather than capturing an array of groups associated with a user.
-
-        When it's necessary to capture a list of roles or groups assigned to the
-        user at the time an event or audit log is recorded,
-        use the array field `user.roles`.
+        Array of user roles at the time of the event.
       example: '["kibana_admin", "reporting_user"]'
--- a/code/go/ecs/user.go
+++ b/code/go/ecs/user.go
@@ -45,11 +45,6 @@ type User struct {
 	// For example, an LDAP or Active Directory domain name.
 	Domain string `ecs:"domain"`
 
-	// Array of user roles or groups, at the time of the event.
-	// `user.group.*` fields are meant to capture one group's full details,
-	// rather than capturing an array of groups associated with a user.
-	// When it's necessary to capture a list of roles or groups assigned to the
-	// user at the time an event or audit log is recorded, use the array field
-	// `user.roles`.
+	// Array of user roles at the time of the event.
 	Roles string `ecs:"roles"`
 }
--- a/code/go/ecs/user.go
+++ b/code/go/ecs/user.go
@@ -44,4 +44,12 @@ type User struct {
 	// Name of the directory the user is a member of.
 	// For example, an LDAP or Active Directory domain name.
 	Domain string `ecs:"domain"`
+
+	// Array of user roles or groups, at the time of the event.
+	// `user.group.*` fields are meant to capture one group's full details,
+	// rather than capturing an array of groups associated with a user.
+	// When it's necessary to capture a list of roles or groups assigned to the
+	// user at the time an event or audit log is recorded, use the array field
+	// `user.roles`.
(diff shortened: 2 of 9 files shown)


schemas/user.yml
```diff
@@ -81,3 +81,20 @@
...
+      short: Array of user roles at the time of the event.
+      description: >
+        Array of user roles or groups, at the time of the event.
+
+        `user.group.*` fields are meant to capture one group's full details,
+        rather than capturing an array of groups associated with a user.
+
+        When it's necessary to capture a list of roles or groups assigned to the
+        user at the time an event or audit log is recorded,
+        use the array field `user.roles`.
```
webmat, line 99: @legrego @ebeahan Do you think I'm conflating two things here, by discussing roles and groups?

schemas/user.yml
```diff
@@ -81,3 +81,20 @@
           Name of the directory the user is a member of.
 
           For example, an LDAP or Active Directory domain name.
+
+    - name: roles
```
webmat, line 85: @ebeahan Over on #915 you were floating the idea of other names for the field. I think the simpler `.roles` name would work great specifically because it's not qualifying the field name too much. This way the field may work for more situations (wild guess).

What kind of ambiguities do you think the other suggested names resolve?

webmat:
I like that this addition is a laser focused addition of one field. However I'd be curious to know if there's more fields we may want to capture, that are also important to audit logs mentioning user roles.

schemas/user.yml
legrego: I tend to think of groups as a collection of users, whereas a role is a collection of privileges. Groups are often used to model user roles within an organization though.

In my opinion, I think it makes sense to treat these as separate concepts, and only mention roles here

webmat: Conversely, do you think `group.roles` would be appropriate, then?

I realize ES doesn't have the concept of groups. But in systems that do, roles/privileges can be granted to a group as well.

Note that I'm just trying to make sure we're not missing anything obvious here. But it's probably fine to move forward with adding only this one field, and only add `group.role` later, if it becomes necessary.

schemas/user.yml
ebeahan: I'm going to pull an IAM example from the proposed `user.target` field since I think it helps capture my thinking.

The `user.roles` field values are being populated with two purposes: 

1. the `elastic` user is assigned the `superuser` role. Here `.roles` captures `elastic's` assigned roles at the time of the event.
2. `elastic` is modifying user `bob` and changing `bob's` roles. Roles here captures the changes to bob's user permissions being assigned in the system.

```json
{
  "user": {
    "name": "elastic",
    "roles": [ "superuser" ]
    "target": {
      "name": "bob"
    },
    "changes": {
      "roles": [ "kibana_user", "superuser" ]
    }
  }
}
```

It's only a minor difference though - what roles are currently assumed by a user vs. the assigning of roles to a user.

webmat: So `user.target` should also contain the prior state for the roles. If it had none, it should have an empty array. If it had e.g. "kibana_user" and is being added "superuser", it should look like this:

```JSON
{
  "user": {
    "name": "elastic",
    "roles": [ "superuser" ],
    "target": {
      "name": "bob",
      "roles": [ "kibana_user" ]
    },
    "changes": {
      "roles": [ "kibana_user", "superuser" ]
    }
  }
}
```

schemas/user.yml
ebeahan: I thought about if a `group.roles` field would be appropriate too.

While roles assigned to groups, it's ultimately a user who authenticates and is authorized with the roles group membership grants. So I'm not sure if `group.role` would be as useful right away, but I also agree we could easily add it later if necessary.

schemas/user.yml
ebeahan: Ok. Do you think we should limit (in the field's description) the scope of `user.roles` to just be the assigned roles/privileges of the user performing the event? Or is it ok if in implementation the field might get used to track assignment of those roles too?

schemas/user.yml
legrego: I agree with @ebeahan - we can introduce `group.role` at a later date if the need arises

schemas/user.yml
ebeahan: @webmat made a good point that using categorization fields could help the user filter between the different use cases I had some minor concerns. 👍

schemas/user.yml
webmat: I've completely removed all talk about groups. Actually the field description is only the short description now:

> Array of user roles at the time of the event.

Do you think anything else is needed here?

ebeahan: Concise but clear - I think the short description works.

PR #919, opened 2020-08-12
[1.5] Use appropriate reserved domain for example URLs (#910)

ebeahan:
Backports the following commits to 1.5:
 - Use appropriate reserved domain for example URLs (#910)

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -18,6 +18,8 @@ Thanks, you're awesome :-) -->
 
 #### Improvements
 
+* Updated several URLs in the documentation with "example.com" domain. #910
+
 #### Deprecated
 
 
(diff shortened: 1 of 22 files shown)


webmat:
Let's have the changelog entry in this branch in CHANGELOG.next.md. This way if we decide to cut a patch release to generate an official Beats artifact, we'll have our changelog entry ready for the release notes.

Also some artifacts still need to be updated :-)

ebeahan:
Thanks @webmat - the backport took some `git` detangling, so I figured there'd be some cleanup to do still.

ebeahan:
This PR is ready for one final 👀 .

I noted a couple of additional fields in `tls` and `url` where we can update (currently elastic.co) with `example.com` for consistency. I'll handle those in a separate, later PR and backports.

PR #922, opened 2020-08-12
bumping version for master branch

ebeahan:


webmat:
Still needing a make run :-)

ebeahan:
Thanks :-) Had to rebase with `master` to pick up the most recent changes and then rerun `make`. I believe all should be ✅ once CI finishes.

PR #934, opened 2020-08-14
Improve backport support

ebeahan:
Two updates for improving backporting support:

* Update the backport project config to capture the newer release branches and branching strategy.
* Capture branching strategy and basic tooling usage.

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -1,8 +1,20 @@
 {
   "upstream": "elastic/ecs",
-  "branches": [{ "name": "1.0", "checked": true }],
-  "all": true,
-  "multipleCommits": false,
-  "multipleBranches": true,
-  "labels": ["backport"]
+  "branches": [
+    { "name": "master", "checked": true },
+    { "name": "1.x", "checked": true },
+    "1.6",
+    "1.5",
+    "1.4",
+    "1.3",
+    "1.2",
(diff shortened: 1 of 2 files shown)


.backportrc.json
```diff
@@ -1,8 +1,20 @@
...
+    { "name": "1.x", "checked": true },
+    "1.6",
+    "1.5",
+    "1.4",
+    "1.3",
+    "1.2",
+    "1.1",
+    "1.0"
+  ],
+  "targetPRLabels": ["backport"],
```
webmat, line 14: We're pretty reliable on tagging with `needs_backport` when needed.  Perhaps we could add the following as well?

```suggestion
  "targetPRLabels": ["backport"],
  "prFilter": "needs_backport",
```

webmat (dismissed):
I'm good to try this as is. I've spotted another config we may want to try out. I'm good either way, though. Let's try this out and adjust as needed :-)

.backportrc.json
ebeahan: I like the idea. I adjusted the syntax a bit; `prFilter` let's you define an entire [GitHub query](https://docs.github.com/en/github/searching-for-information-on-github/understanding-the-search-syntax) if we want to get more elaborate in our filtering in the future.

PR #879, opened 2020-06-27
[RFC] Session Fields

dainperkins:
New RFC for Session fields provided.

PR #935, opened 2020-08-17
[RFC] Session RFC PR to move to Stage 2 (Draft)

dainperkins:
Updated verbiage in description and promoted to stage 2 (draft)

[Markdown preview of this RFC](https://github.com/dainperkins/ecs/blob/Session/rfcs/text/0004-session.md)

**TODO**
- [ ] Finish populating the source data section
  - [ ] DB Connection
  - [ ] Web Session
  - [ ] Cloud Admin Session
- [ ] Subject matter experts weighed in on technical utility of field definitions

### DIFF
--- a/rfcs/text/0004-session.md
+++ b/rfcs/text/0004-session.md
@@ -1,8 +1,8 @@
 # 0004: Session
 <!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage -->
-- Date: 7/30/2020 <!-- Update to reflect date of most recent stage advancement -->
+- Stage: **2 (draft)** <!-- Update to reflect target stage -->
+- Date: 8/17/2020 <!-- Update to reflect date of most recent stage advancement -->
 
 <!--
 Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
@@ -88,23 +88,29 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
 
 ## Usage
 
-Session fields are used to describe the sesison attributes of:
- - Client VPN Sessions
- - Network to Network VPN Sessions
- - Network Access Sessions (NAC, WPA, EAP, etc.)
- - Local or remote device login sessions (RDP, ICA, xWindows)
+Session fields are used to describe and track a discrete grouping of interactions, typically bounded by
+authentication or authorization events and tied to a specific user, application, or system component. 
+
+For example:
+ - Network Access Sessions (NAC or Wireless LAN)
+ - Local or remote device login sessions (tty, console, ssh, RDP, ICA, xWindows, etc.)
+ - VPN Sessions (network to network, or client to network)
+ - Local or remote device login sessions (console, tty, RDP, ICA, xWindows, ssh, etc.)
  - Administrative sessions on infrastructure devices
  - Administrative sessions on cloud or application management portals
- - Applications sessions (e.g. sql server odbc session, application access session)
+ - Applications sessions (e.g. sql server odbc session, web cookie based session, application access session)
+ - Cloud API access sessions
 
 
 ## Source data
 Source data expectations include:
  - Wireless Lan Controllers
- - Security appliances
+ - Security appliances (e.g. fw, waf)
  - Network admission control devices
- - Radius / tacacs servers
- - Application server logs
+ - Radius / tacacs servers (802.1x EAP/PEAP aaa)
+ - Application server logs (FTP, MySQL)
+ - Web Server, WAF, or ADC logs (USer or cookie based web ession control)
+ - APM telemetry
 
 Example 1: Meraki 802.1x Logs (WLC)
 * EAP session start)


jonathan-buttner:
@ferullo @rw-access @joe-desimone @jrmolin We're working on adding some session fields to the ECS core schema. I think this could be useful for the session events that the endpoint sends. I don't think we ever nailed down the exact fields that the endpoint would send in our original discussion here: https://github.com/elastic/endpoint-app-team/issues/258 Are there certain changes we'd like to include to cover the information that the endpoint would will send?

Full RFC is here: https://github.com/elastic/ecs/blob/master/rfcs/text/0004-session.md

rw-access:
thanks for the ping @jonathan-buttner.
one thought I have -- `session.type` seems to be higher level than `session.kind` in the hierarchy, but the opposite is true for `event.*` Do you think it would make sense to be consistent in the terminology?

rw-access:
@gabriellandau or @wburgess1 may also have some insights to relevant metadata from an OS perspective for logon sessions. I think it's good that the field set starts out more minimal, with the things that are universally understood when the word "session" is invoked

rw-access:
Another question: (sorry for all the spam)

Do you think this warrants a new `event.category`? What do you have in mind for the category with the examples you have? And how would you categorize a windows logon session (4624 logon event through a 4634 or 4647 logoff)?

dainperkins:
@rw-access 
kind vs. type - good catch, thanks!   I'll update that one to better reflect the typical hierarchy of e.g. event.kind/type.

Looking at the 4624 logon types I think there's definitely potential to add a few examples that I missed in the initial stages.  I'll add the windows examples & the others I have to a gsheet mapping to make sure the coverage is there and post here for review

/d

dainperkins:
@rw-access in looking at the kind/type -

ebeahan:
Looks like this was closed on accident 😄

cyrille-leclerc:
@dainperkins shall we add an Observability member as reviewer of this PR to ensure we observability continue to contribute to this "session" RFC?
cc @andrewvc , @drewpost 
reference https://github.com/elastic/ecs/pull/879

dainperkins:
@cyrille-leclerc I think thats a very good idea - while I tried to make sure I covered the session fields that seemed to make sense from a number of perspectives, having someone with an APM / Observability view would be a huge benefit

rw-access:
@dainperkins I think that your kind/type distinction makes sense to me.
my guess is that the "Close and comment" is just too close to the "Comment" button

webmat (commented):
Thanks Dain.

Let's formalize a few easy tasks that should get done in this PR:

* In the "People" section at the bottom
  - [x] Add Cyrille / Observability as a stakeholder
  - [ ] Add Andrew Cholakian as a subject matter expert. I do want this to work for web sessions if possible.
  - [x] Add Ross Wolf as a subject matter expert for the endpoint POV
- [x] Add link "Stage 2: https://github.com/elastic/ecs/pull/935" in the references at the end

The YAML field definitions is looking great. Please take it out of the Markdown file and instead put it in a subdirectory for RFC 0004, at `rfcs/text/0004/session.yml`.

In doing so, merging this stage 2 PR will make it possible for folks to generate ECS artifacts that include these fields with something like `python scripts/generator.py --include rfcs/text/0004/` (see [USAGE.md](https://github.com/elastic/ecs/blob/master/USAGE.md) for details).

I've tried the above command locally, and I've noticed we're missing short definitions for fields "id", "name" and "authorization", so please add those as well.

-----

That's it for the virtual paperwork 😂 

In terms of specifics on the subject of sessions, here's what I'd like to see added in this PR.

- [ ] My understanding is that both `session.kind` and `session.type` are meant to be values from a pre-determined list. Let's list all acceptable values in the "Usage" section. We can use your initial list as a starting point, and finalize the list with the other subject matter experts. 
- [ ] When you get around to introducing concrete examples, one thing I'd like to see represented is how each kind of sessions should populate all of the session fields. If some attributes are not pertinent for some kinds of sessions, we should start fleshing this out

joshdevins:
Might be good to have someone from Enterprise Search on this. We're in the middle of defining some tracking events and sessions/session IDs has come up.

/cc @RossAllenBell

RossAllenBell:
> Might be good to have someone from Enterprise Search on this. We're in the middle of defining some tracking events and sessions/session IDs has come up.
> 
> /cc @RossAllenBell

Thanks @joshdevins!

Yes, we're going to be tracking user search sessions in Workplace Search soon. If I can help in any way by providing more use case context, please let me know.

dainperkins:
@RossAllenBell - how well do you see the RFC session fields fitting into what you plan on tracking (in concert with the rest of the ECS space as well...)

Do you have any examples I can dig thru?

RossAllenBell:
> @RossAllenBell - how well do you see the RFC session fields fitting into what you plan on tracking (in concert with the rest of the ECS space as well...)
> 
> Do you have any examples I can dig thru?

@dainperkins I expect Workplace Search's use case would primarily make use of the `session.id` field, as we attempt to track analytics data across a user's session when using search in Workplace Search. While we would have data to populate the other fields proposed in this RFC, they would likely always be the same value and potentially not worth storing.

The Workplace Search Analytics project brief: https://docs.google.com/document/d/1k5ttaeVjgnsWJTSJ1s3d0wv2dIV_DVVvKPSBcLfpeNc

An example of data we track today that I would like to add `session.id` to: https://docs.google.com/document/d/1QedgCC8EQLJdPfKzIvxcoCJXbTdfvYT5N_JKXyQWFOE/edit#heading=h.btqtgkfgroil

webmat:
Just dropping a quick note here that I agree with @rw-access' suggestion that we should consider adding `event.category` "session". I think this would make a lot of sense.

webmat:
As agreed out of band, I've adjusted this PR a bit, preparing this for the next round of discussion.

Here's what I did:

* Remove the merge date, which we'll set when the RFC stage 1 PR is merged
* Add the link to this PR in the references
* Move the YAML field definitions to the file `rfcs/text/0004/session.yml`, so we can generate artifacts that include the changes from this RFC.
  * Adjust the YAML to fix errors, and to improve the readability in asciidoc format
* Improve the formatting of the "Source data" section

rfcs/text/0004-session.md
```diff
@@ -1,79 +1,29 @@
...
-This RFC calls for the addition of session fields to describe events related to various types of "sessions" reported by appliances, security devices, systems, management portals, applications, etc.
 
 | Field | Description |
 | ----- | ----------- |
-|session.kind:            | local, remote, network
-|session.authorization:   | user, admin, service
-|session.type:            | system, virtual, application, wired, wireless, vpn
+|session.kind:            | system, application, network
+|session.type:            | [ local, remote, virtual, wired, wireless, vpn ]
+|session.authorization:   | user, admin, service, access
```
webmat, line 21: This compact table is a great way to visualize all of these values together.

I have a few questions:

The value "access" appears to have been dropped out of the values mentioned in the YAML description. Was this intentional? If so, please remove it from here. If it was not intentional, please add it back in the new YAML file at `rfcs/text/0004/session.yml`

I initially thought there was an overlap with values `session.kind: network` and `session.type: remote`. When adjusting the YAML file for readability it occurred to me that `session.kind: network` was likely more to identify network level sessions, rather than *anything that happens over a network*. In other words, a web cookie-based session would not use `session.kind: network`, but `session.kind: application`. Am I understanding this correctly?

However within the different values for `session.type`, I do wonder if there's overlap. It seems like "wired", "wireless" and "vpn" would always be accompanied by value "remote". Is this also how you see this? Note that this is not necessarily a problem. "local" vs "remote" could be useful when plotted against one another, whereas "wired", "wireless" & "vpn" could be useful to break down the kinds of "remote" session types.

One final overlap concern: I do think `session.kind: system` and `session.authorization: service` are almost entirely overlapping, aren't they? Should we get rid of one of them, or am I misunderstanding a subtlety?

rfcs/text/0004-session.md
```diff
@@ -88,62 +38,85 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+EAP session start
+
+`<134>1 1580551704.928047208 my_AP events type=8021x_eap_success radio='1' vap='2' client_mac='12:34:56:78:9A:BC' client_ip='192.168.1.100' identity='JohnDoe' aid='1687088497'`
 
 802.1x EAP De-association Message
-* EAP session end
-    * `<134>1 1580551705.928047208 my_AP events type=8021x_deauth radio='1' vap='2' identity='JohnDoe' aid='1687088497’'
 
-* Note, while there is an association id (session.id) created prior to wpa/802.1x authentication, building the session event from the eap success message allows for easier integration of fields like username, client.ip, etc. in an 802.1x or WPA environment
+EAP session end
```
webmat, line 73: I tried to reformat all of the examples below. The section "example 1" is one where I'm not sure if I did a good job.

Are we missing an example above for "802.1x EAP De-association Message"? Or is this just another way of referencing the "EAP session end"?

rfcs/text/0004-session.md
```diff
@@ -88,62 +38,85 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+`<134>1 1380653443.857790533 MR18 events type=association radio='1' vap='1' channel='2' rssi='23' aid='1687088497'`
+
+Base 802.11 Deassociation Message  (802.11 session end)
+
+`1380653443.857790533 my_AP events type=disassociation radio='1' vap='2' channel='6' reason='8' instigator='2' duration='11979.728000' auth_neg_dur='1380653443.85779053324000' last_auth_ago='5.074000' is_wpa='1' full_conn='1.597000' ip_resp='1.597000' ip_src='192.168.111.251' arp_resp='1.265000' arp_src='192.168.111.251' dns_server='192.168.111.1' dns_req_rtt='1380653443.85779053335000' dns_resp='1.316000' aid='1813578850'`
+
 
-    * Base 802.11 Association:  (802.11 session start)
-        * `<134>1 1380653443.857790533 MR18 events type=association radio='1' vap='1' channel='2' rssi='23' aid='1687088497’
+### Example 2: ASA Admin Login
```
webmat, line 88: Perhaps we could mention "ssh" in the ASA section title as well, to make it stand out for folks looking for an ssh example?

rfcs/text/0004-session.md
```diff
@@ -88,62 +38,85 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+ - VPN Sessions (network to network, or client to network)
+ - Local or remote device login sessions (console, tty, RDP, ICA, xWindows, ssh, etc.)
  - Administrative sessions on infrastructure devices
  - Administrative sessions on cloud or application management portals
- - Applications sessions (e.g. sql server odbc session, application access session)
+ - Applications sessions (e.g. sql server odbc session, web cookie based session, application access session)
+ - Cloud API access sessions
 
 
 ## Source data
```
webmat, line 55: For each of the log examples in the "Source data" section, could you include a breakdown of expected values for the session fields of each log event?

E.g. the ASA login could be something like

* session.kind: network
* session.authorization: admin
* session.type: [remote]

Not sure if one of the values in the ASA example is a session name or session id, but if so, list those as well.

webmat (commented):
Here's a few questions and comments for you @dainperkins, after my adjustments from today.

rfcs/text/0004/session.yml
```diff
@@ -0,0 +1,85 @@
...
+      description: >
+        Session type provides additional details on the scope of the interactions being tracked. Acceptable values are: "local", "remote", "virtual", "wired", "wireless", "vpn".
+
+        "local" is relevant e.g. for serial connections, physical sessions (hands on keyboard).
+
+        "remote" is relevant for any remote connection: direct system access via remote desktop, ssh, etc.
+
+        "virtual" is for VDI, VDA, etc.
+
+        (missing clarifications for: wired, wireless and vpn network connections?)
```
webmat, line 46: Forgot to bring attention to this description.

When I reformatted it, I noticed we were missing definitions for wired, wireless and vpn. Could you add those please?

rfcs/text/0004/session.yml
```diff
@@ -0,0 +1,85 @@
...
+        - array
+      short: 'Session type (array of: local, remote, virtual, wired, wireless and/or vpn).'
+      description: >
+        Session type provides additional details on the scope of the interactions being tracked. Acceptable values are: "local", "remote", "virtual", "wired", "wireless", "vpn".
+
+        "local" is relevant e.g. for serial connections, physical sessions (hands on keyboard).
+
+        "remote" is relevant for any remote connection: direct system access via remote desktop, ssh, etc.
+
+        "virtual" is for VDI, VDA, etc.
```
webmat, line 44: Can we expand on "virtual" what are VDI and VDA?

rfcs/text/0004-session.md
dainperkins: > One final overlap concern: I do think session.kind: system and session.authorization: service are almost entirely overlapping, aren't they? Should we get rid of one of them, or am I misunderstanding a subtlety?

I was trying to find discrete combinations that would uniquely identify <user> logging into a system, and a service running with authentication/authorization included in the "session".  It may be unnecessary (I'll look thru auditbeat service logs and see what they look like...  could just be that a system service is not something that would typically be defined as a session (e.g. starting up a vpn connection on a pfsense linux firewall - presumably the vpn application logs would be sufficient)

> The value "access" appears to have been dropped...

the revamp of the fields seems to lend itself more towards the entity being authorized - authentication would be a part of the overall session, but I don't think it makes sense as a specific session field.  Basically the split becomes: 

- "user interactive session" (user logs in or connects to a service or application to use said service/app)
- "admin interactive session"(user logs in to manage a service or application e.g. root, enable mode in cisco parlance) 
- "Network session" (user or system connects to a monitored network)

Not all of the type values will apply to every session.kind:
Session Kind:
- System (full system access)
  - local (hands on keyboard)
  - remote (bash via ssh)
  - virtual (Citrix -virtual desktop)
- Application
  - remote (e.g. web connection, smb share, ftp)
  - local (possible but I can't think of an example)
  - virtual (citrix published apps)
-Network
  - wired (802.1x network session - e.g. time connected to the LAN)
  - wireless (802.1x network session - e.g. time connected to the WLAN)
  - vpn (user connects to vpn concentrator

I think this one could use some more input- I can see session.type having fixed single values (virtual implies remote), or using an array (vpn would be network & remote, published apps & citrix desktop would be virtual & remote, opening a virtualbox session locally could be local & virtual...)

Does that all make sense?

Wired & wireless however would typically not use remote as the session is for the local user or workstation directly connecting to the local network (as opposed to connecting to another entity over the network)

rfcs/text/0004-session.md
dainperkins: Will check - I think I may have used EAP & 802.11 messages instead of picking one or the other

rfcs/text/0004-session.md
dainperkins: will do

rfcs/text/0004-session.md
dainperkins: I'm actually wondering if virtual needs to be specifically identified... a "remote" "system" session will be telnet, ssh, rdp, or citrix... not sure the virtual type is necessary to split out rdp/citrix

dainperkins: And service as an authorization could also make sense for workstation based network access (802.1x cert auth), network to network vpn, etc...

rfcs/text/0004-session.md
dainperkins: Yeah - probably wouldn't be a bad idea to fill in everything in some additional example.json file(s)?

rfcs/text/0004/session.yml
dainperkins: virtual desktop (rdp/citrix) and virtual desktop applications (citrix published apps - e.g. run just excel via citrix, no actual desktop just the app)

webmat:
I've opened a PR to add the `event.category` "session" https://github.com/elastic/ecs/pull/1049.

I don't think we need to wait for the RFC before introducing the event category itself.

ebeahan:

This PR was initially targeting the now "legacy" stage two, and I'm thinking we update now to target stage 1. Stage 1 (draft) fields will still be added as experimental fields as was done with legacy stage 2 fields.

rfcs/text/0004-session.md
```diff
@@ -1,79 +1,29 @@
...
 ## Fields
-This RFC calls for the addition of session fields to describe events related to various types of "sessions" reported by appliances, security devices, systems, management portals, applications, etc.
 
 | Field | Description |
 | ----- | ----------- |
-|session.kind:            | local, remote, network
-|session.authorization:   | user, admin, service
-|session.type:            | system, virtual, application, wired, wireless, vpn
+|session.kind:            | system, application, network
+|session.type:            | [ local, remote, virtual, wired, wireless, vpn ]
```
andrewvc, line 20: IMHO `kind` vs `type` is really hard to reason about / remember. `type` would be more clear if it was named `connection_type`.

rfcs/text/0004-session.md
```diff
@@ -88,62 +38,85 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
  - Wireless Lan Controllers
- - Security appliances
+ - Security appliances (e.g. fw, waf)
  - Network admission control devices
- - Radius / tacacs servers
- - Application server logs
+ - Radius / tacacs servers (802.1x EAP/PEAP aaa)
+ - Application server logs (FTP, MySQL)
+ - Web Server, WAF, or ADC logs (USer or cookie based web ession control)
+ - APM telemetry
```
andrewvc, line 63: For our RUM agent, running on client websites tracking browsing sessions,  I can only see us setting `session.kind` and `session.id`. Would love feedback from the RFC authors on whether this makes sense.

CC @jahtalab

rfcs/text/0004-session.md
dainperkins: I like the idea, tho my brain says "local isn't a connection!" but maybe thats ok.

dainperkins: I think I've updated these a bit, was hoping to get some feedback before submitting updates.  I'll try and get the revisions up in the near future

rfcs/text/0004-session.md
dainperkins: I think the idea (only filling out relevant fields) makes total sense... tho in respect to this and the previous comment it makes me think I'm missing something that identifies a RUM type session (but then I also don't really know what other fields would be coming in that would make filtering results from a mixed index easy). Unauthenticated web sessions for instance would't need to use session.auth, but after a login could (along with user.name, etc.). I don't see that one as big a deal as e.g. network access logins (vpn, wireless, etc)

jahtalab: To add some context, the RUM sessions are not necessarily authenticated, even until they are complete, e.g. a user session for going through checkout.

jahtalab (commented):
Regarding the Real User Monitoring sessions, I think there are some fields that can be used from this RFC, specifically using kind to `application` and using `session.id` and maybe `session.name`. Although, It's not clear to me from the RFC if the kind `application` is meant to be used for business related sessions, e.g. a user completing an action through multiple web pages. These business sessions could be different to the cookie based web sessions that is mentioned in the examples, e.g. a user might complete multiple business sessions while still authenticated with the same cookie. I think adding some examples would help clarify this.  

Also, the RUM sessions would probably require some extra fields as well. These can be under `session` field but they are not necessarily  a good fit for ECS, since they are only relevant to RUM.

rfcs/text/0004-session.md
andrewvc: @jahtalab do we have any notion of auth no? AFAIK we don't, and it's unlikely given the extreme variety of auth methods.

rfcs/text/0004-session.md
andrewvc: Sure, I would say that local is a connection, connecting to local host is a connection, just within a single kernel.

rfcs/text/0004-session.md
jahtalab: @andrewvc , We don't detect authentication per se, but there's a configuration for customers to add the user (username, id, ..) in the RUM agent. We can assume the user is authenticated (with some method) but it doesn't have to be.

rfcs/text/0004-session.md
```diff
@@ -1,79 +1,29 @@
 # 0004: Session
 <!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage -->
-- Date: 7/30/2020 <!-- Update to reflect date of most recent stage advancement -->
+- Stage: **2 (draft)** <!-- Update to reflect target stage -->
```
epixa, line 4:
```suggestion
- Stage: **2 (candidate)** <!-- Update to reflect target stage -->
```

dainperkins:
Since we seem to be a bit stalled I'm working on a table with specific examples and what fields seem relevant (given the discussions above). 

One item that I wonder about is the utility of defining session fields regarding access method (for instance remote system & app sessions will always have a source/destination component - do we need to have a separate field or is searching session.kind: system, source/destination.ip or network.protocol good enough?)

dainperkins:
@jahtalab @epixa @rw-access @jonathan-buttner @cyrille-leclerc @andrewvc @ebeahan @RossAllenBell 

I've updated the fields with general examples here:
https://docs.google.com/spreadsheets/d/1_PMkCb9S3HAF82VacpIBg3VhYPcIRePtT1eJoSPOG54

Would be great if I could get some feedback on the comments sheet (I can redo the RFC with the new organization and parse a bunch of examples pretty quickly if this looks good for a baseline (I also left two columns for any suggestions of things that should be in from the start)

session.id                       ( optional - source provided or hash )
session.kind                   [ system, network, application ] 
session.type                   [ open with suggestions ] ( basically a general description of the session)
                                        ( local/remote console, remote desktop, *management, NAC PSK|Host|User, VPN SSL|PSec
                                        ( auth, audio, video, meeting, webinar, etc) 
                                        1 special case - management should be used to differentiate users managing service & users 
                                        consuming services
session.name                 logical name, & or o11y service related 
session.authorization    psk, host, user, service, token
                                        (psk, service, token might be redundant

andrewvc:
While there's a lot of good stuff here the  kind and type fields are a serious problem. We should not move forward till they are either improved or removed.

These words are synonyms and should not be used in conjunction. There is a mention of switching kind to category but this suffers from the same problem. 

I guess I'm unclear on what these are for in the examples given as well.  I would have trouble giving a clear definition of either in a  sentence or two 

Would there be a problem with using a single generic 'type' field and largely leaving it to users to fill in what they find useful

dainperkins:
Effectively the break out is for visualizing sessions of different kinds (no pun intended) with type management being of particular importance in a siem use case.

In a nutshell - 
Kind is a generic category indicating what type of session (logging into a system, gaining network access, or application access) While most of the type labels could be derived from e.g. protocol information, it seemed like the trade off to not have to search on e.g. "session.kind: system & network.protocol: https" or similar would be worth it

If you check the ECS event fields:

>  .kind gives high-level information about what type of information the event contains, without being specific to the contents of the event.
> .type represents a categorization "sub-bucket" that, when used along with the event.category field values, enables filtering events down to a level appropriate for single visualization

so it does follow the idiom... not that I'd say no to better categorization terms, but it does seem to scan overall

andrewvc:
Thanks for the background @dainperkins , I wasn't aware we had prior art here. That said, reading through the [category fields](https://www.elastic.co/guide/en/ecs/current/ecs-category-field-values-reference.html) docs they don't seem like a good fit here. They seem explicitly designed for events. If you look at the docs for [kind](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-kind.html) I would think that all sessions would have either `state` or `event` as types.

That said, the docs for ECS category fields are very confusing IMHO, and replicating them would seem to me to only replicate that confusion. 

Here's a thought, given the generic nature of this, we could name these things `context.group` and `context.name`, so you'd get:
```
{
  "context": {
     "group": "network",
     "name": "wireguard"
  }
}
```

The specific words of `context`, `group` and `name` are unimportant except for the fact that name is obviously more specific than group, and nesting them within an object makes their relationship as a hierarchy more obvious.

dainperkins:
I was just referring to the use of kind and type nested under event as a categorization mechanism for sessions as well.  I can see the point of differentiating the field names under session, however I also think it makes sense to re-use existing hierarchical categorization schemes within ECS (kind, type, category, etc.) for ease of use.   

I'll give some more thought to the session categorization, and take a look at whatever other feedback comes.

ebeahan:
> That said, the docs for ECS category fields are very confusing IMHO

@andrewvc I don't want to deviate this conversation, but if you have any specific feedback, feel free to reach out. 😃

ebeahan:
Since a bit of time has past since this PR was originally open, I'll call out that `event.type: session` was added as part of ECS 1.8.0. Do we see a possible alternative in pairing `event.type: session` with some of the other `event.type` allowed values to achieve similar categorization as proposed here by using `session.kind`, and `session.type`?

Examples:

* Authenticated session: `event.type: [ "authentication", "session" ]`
* Network session: `event.type: [ "network", "session" ]`
* Web session: `event.type: [ "web", "session" ]`

dainperkins:
Maybe its time to go for a good core base rather than a perfect complete implementation...

a) provide basic session fields (name/id) to use with event.type: session
b) take some time to evaluate the above suggestions for discriminating between the various types of sessions users might want to use or report on

Thoughts?

ebeahan:
> Maybe its time to go for a good core base rather than a perfect complete implementation...

++ It's much easier to add later than take away 😄 .

I would like to see some mapped examples. Having some examples of these fields populated with real-world values would help anchor the conversation.

bryce-b:
Mobile agents have definite need for this! This RFC looks very applicable, but should an additional `kind` be added for mobile?

djptek:
A `session` is a sequence or family of `event`s

I think it may be simpler to leave `event.type: session` under event and not involve it further at this stage.

Looking at each of the proposed fields for session:

- `session.kind` - the example values for this field might need to encompass every possibility of parent kinds for families of `event.kind`. I'd suggest to provide the cited examples and word this as *expected*, rather than *allowed* values
- `session.type` - same issue as `session.kind`
- `session. authorization` - LGTM
- `session.name` - LGTM
- `session.id` - LGTM

I'd favour making all fields optional

dainperkins:
Has there been any feedback from the session viewer team?  Sneha might have some feedback based on session viewer (minimally worth a discussion if there's overlap). (@snehsach19)

qcorporation:
@mitodrummer @m-sample Just giving you two visibility from a session viewer perspective

mitodrummer:
> @mitodrummer @m-sample Just giving you two visibility from a session viewer perspective

All good from my end. Session viewer's sessions are linux centric. We use a nested process fieldset to group our process events by session_leader, entry_leader etc.... see: https://www.elastic.co/guide/en/ecs/current/ecs-process.html#ecs-process-nestings

Sounds like this RFC is geared towards a more generic handling of sessions.

mitodrummer:
Ya, I think had this fieldset been available at the time of our RFC, we may have considered using session.type to track the entry type. session.id could have been used for sid, but it seems our widened model approach using the process fieldset covers more ground (w context), since we track two types of sessions, entry_leader and session_leader.

m-sample:
> it seems our widened model approach using the process fieldset covers more ground (w context), since we track two types of sessions, entry_leader and session_leader.

Good point. On Linux, a single remote access by a human may have 2 (or even more) actual Linux sessions associated with it.  This is typically from people running `screen` or `tmux` in their entry session to create a child session that can be re-attached to on a subsequent remote access (commonly done to prevent losing ones work/running-processes if the ssh TCP conn dies)

djptek:
Looking at the original intention of the [session RFC ](https://github.com/elastic/ecs/pull/879#issuecomment-666646811) this was envisioned to support:

- physical user sessions at a desktop
- various kinds of network sessions, including browser sessions

so it's totally appropriate that the [process RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0030/process.yml) has a widened model.

rsk0:
Is there intention to continue trying to develop this RFC for generalized sessions?  It seems like the right field set for our use of web application session.

buzzdeee:
any intention to continue this one?

PR #944, opened 2020-08-19
Add http.[request|response].mime_type

### LINKED ISSUE #749: Mime Types

webmat:
**Discussion points**

- [x] Should we make a concession that a response mime_type can be set based on the `Content-Type` header?
- [x] Should this field use `wildcard` instead of `keyword`?
- [x] Since populating this requires analysis of the body content, I think it may not be applicable to many sources, so I'm suggesting level extended

Closes #749

### DIFF
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -26,6 +26,20 @@
 
       example: GET, POST, PUT, PoST
 
+    - name: request.mime_type
+      level: extended
+      type: keyword
+      short: Mime type of the body of the request.
+      description: >
+       Mime type of the body of the request.
+
+       This value must only be populated based on the content of the request
+       body, not on the `Content-Type` header. Comparing the mime type of a
+       request with the request's Content-Type header can be helpful in detecting
+       threats or misconfigured clients.
+
+      example: image/gif
+
     - name: request.body.content
       level: extended
       type: keyword
@@ -50,6 +64,20 @@
       description: >
         HTTP response status code.
       example: 404
+
+    - name: response.mime_type
+      level: extended
+      type: keyword
+      short: Mime type of the body of the response.
+      description: >
+       Mime type of the body of the response.
+
+       This value must only be populated based on the content of the response
+       body, not on the `Content-Type` header. Comparing the mime type of a
+       response with the response's Content-Type header can be helpful in detecting
+       misconfigured servers.
+
+      example: image/gif
 
     - name: response.body.content
       level: extended
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -28,6 +28,7 @@ Thanks, you're awesome :-) -->
 * Added `span.id` to the tracing fieldset, for additional log correlation (#882)
 * Added `event.reason` for the reason why an event's outcome or action was taken. #907
 * Added `related.hosts` to capture all hostnames and host identifiers on an event. #913
+* Added Mime Type fields to HTTP request and response. #944
 
 #### Improvements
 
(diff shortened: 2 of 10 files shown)


webmat:
@ebeahan @neu5ron In #749 I considered making a concession and perhaps allowing `http.response.mime_type` when the source of the data is trusted. By this I meant only for web server logs, not a network IDS. An operator can likely trust their own servers. Whereas from a NIDS point of view, the responses may be untrusted.

Although at this time, this PR does not allow for that. After further thought, I think that concession would have little utility. And when ECS adds support for HTTP headers, that concession would now have zero utility anymore.

Do you agree with keeping the mime_type strictly based on body content?

webmat:
@simitt Would love to have your thoughts on this, or someone else from APM. I think it's straightforward. Does APM capture Mime Type already? After a quick check on a demo app, it looks like it doesn't.

ebeahan:
> Should we make a concession that a response mime_type can be set based on the Content-Type header?

If we don't make the concession, `*.mime_type` field would be populated when the HTTP request/response is examined to determine the effective MIME type vs the `Content-Type` field? For example, mime sniffing in a browser? Ok - I think I understand better now :smile: .

Now that I have better understanding, I don't think we should make the concession since the support for HTTP headers would address `Content-Type` as you mentioned.

> Should this field use wildcard instead of keyword?

I lean towards `keyword`. I could see leading wildcard queries being common here. However, since it _should_ be also be using a curated [set of values](https://www.iana.org/assignments/media-types/media-types.xhtml) set of values, it should be a lower cardinality field and may perform better as keyword (reference: keyword vs. wildcard comparison table from the wildcard intro [blog](https://www.elastic.co/blog/find-strings-within-strings-faster-with-the-new-elasticsearch-wildcard-field)).

> Since populating this requires analysis of the body content, I think it may not be applicable to many sources, so I'm suggesting level extended

👍

webmat:
> examined to determine the effective MIME type

Yes exactly. And if they differ, it's actually meant to be something to look into. It's a feature not a bug :-)

I also agree with sticking to `keyword`.

schemas/http.yml
```diff
@@ -26,6 +26,20 @@
...
+      short: Mime type of the body of the request.
+      description: >
+       Mime type of the body of the request.
+
+       This value must only be populated based on the content of the request
+       body, not on the `Content-Type` header. Comparing the mime type of a
+       request with the request's Content-Type header can be helpful in detecting
+       threats or misconfigured clients.
+
+      example: image/gif
```
ebeahan, line 41: Great choice for the example value 💯

webmat: My favorite mime type 😃 

gif mime example

simitt:
@webmat APM is not collecting a `mime_type` from the body; it would be stored as part of a request or response header if contained and not configured otherwise.

PR #950, opened 2020-08-21
[RFC] Add host metric fields to ECS

kaiyan-sheng:
This PR is stage 1 of the RFC for adding host metric fields into ECS.

[Preview of the RFC](https://github.com/kaiyan-sheng/ecs/blob/add_host_fields/rfcs/text/0005-host-metric-fields.md)

### DIFF
--- a/rfcs/text/0005-host-metric-fields.md
+++ b/rfcs/text/0005-host-metric-fields.md
@@ -14,10 +14,10 @@ With existing `host.id` and `host.name`, these total 9 fields will become the co
 
 Proposed 7 new fields are:
 * host.cpu.usage
-* host.network.in.bytes
-* host.network.in.packets
-* host.network.out.bytes
-* host.network.out.packets
+* host.network.ingress.bytes
+* host.network.ingress.packets
+* host.network.egress.bytes
+* host.network.egress.packets
 * host.disk.read.bytes
 * host.disk.write.bytes
 
@@ -30,13 +30,13 @@ This RFC calls for the addition of host fields to collect basic monitoring metri
 
 | field | type | description |
 | --- | --- | --- |
-| `host.cpu.usage` | scaled_float | Percent CPU used with scaling_factor of 1000. This value is normalized by the number of CPU cores and it ranges from 0 to 1. For example: For a two core host, this value should be the average of the 2 cores, between 0 and 1. |
-| `host.network.in.bytes` | long | The number of bytes received on all network interfaces by the host in a given period of time. |
-| `host.network.in.packets` | long | The number of packets received on all network interfaces by the host in a given period of time. |
-| `host.network.out.bytes` | long | The number of bytes sent out on all network interfaces by the host in a given period of time. |
-| `host.network.out.packets` | long | The number of packets sent out on all network interfaces by the host in a given period of time. |
-| `host.disk.read.bytes` | long | The total number of bytes read successfully in a given period of time. |
-| `host.disk.write.bytes` | long | The total number of bytes write successfully in a given period of time. |
+| `host.cpu.usage` | scaled_float (scaling_factor 1000) | Percent CPU used with scaling_factor of 1000. This value is normalized by the number of CPU cores and it ranges from 0 to 1. For example: For a two core host, this value should be the average of the 2 cores, between 0 and 1. |
+| `host.network.ingress.bytes` | long | The number of bytes received (gauge) on all network interfaces by the host in a given period of time. |
+| `host.network.ingress.packets` | long | The number of packets (gauge) received on all network interfaces by the host in a given period of time. |
+| `host.network.egress.bytes` | long | The number of bytes (gauge) sent out on all network interfaces by the host in a given period of time. |
+| `host.network.egress.packets` | long | The number of packets (gauge) sent out on all network interfaces by the host in a given period of time. |
+| `host.disk.read.bytes` | long | The total number of bytes (gauge) read successfully in a given period of time. |
+| `host.disk.write.bytes` | long | The total number of bytes (gauge) write successfully in a given period of time. |
 
 <!--
 Stage 2: Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting.


exekias (approved):

rfcs/text/0005-host-metric-fields.md
```diff
@@ -26,6 +26,17 @@ Proposed 7 new fields are:
 <!--
 Stage 1: Describe at a high level how this change affects fields. Which fieldsets will be impacted? How many fields overall? Are we primarily adding fields, removing fields, or changing existing fields? The goal here is to understand the fundamental technical implications and likely extent of these changes. ~2-5 sentences.
 -->
+This RFC calls for the addition of host fields to collect basic monitoring metrics from a host or VM such as CPU, network and disk.
+
+| field | type | description |
+| --- | --- | --- |
+| `host.cpu.pct` | scaled_float | Percent CPU used. This value is normalized by the number of CPU cores and it ranges from 0 to 1. |
```
andrewkroh, line 33: It's important to specify the `scaling_factor` so that we know how the value should be interpreted.

rfcs/text/0005-host-metric-fields.md
```diff
@@ -26,6 +26,17 @@ Proposed 7 new fields are:
...
+| `host.network.in.bytes` | long | The number of bytes received on all network interfaces by the host in a given period of time. |
```
andrewkroh, line 34: In the observer namespaces, ECS uses "ingress" and "egress". For consistency we might want to consider uses those terms.

rfcs/text/0005-host-metric-fields.md
andrewkroh, line 33: ECS guidelines recommend not using abbreviated terms like `pct`.

cyrille-leclerc: Can we document how this cpu percentage works on a server with multiple CPU cores.
Example, If I have a server with 12 CPU cores, will this "CPU percentage" max to `1200%` or to `100%`?

rfcs/text/0005-host-metric-fields.md
cyrille-leclerc: Can we document the nature of this counter. Is it a monotonic counter that is sometimes reset (e.g. system restart)?

rfcs/text/0005-host-metric-fields.md
jonathan-buttner: I believe the way that the endpoint team has done this in the past is the 12 CPU cores has a max of `1200.0123...` right @ferullo ?

kaiyan-sheng: Thanks for the input! This `host.cpu.pct` will be the normalized value ranges from 0 to 1 only. If the server has 12 CPU cores, then this normalized value will be `total cpu percentage`/12.

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: @cyrille-leclerc These values will actually be gauges. For example, `host.network.in.bytes` will be the total bytes received aggregated among all network interfaces in 10 seconds. Next collection period, this value will change, might go up or down depends on the network usage of the next 10 seconds.

rfcs/text/0005-host-metric-fields.md
ebeahan: Would that make `scaling_factor: 100`?

This would be the first time a `scaled_float` is defined in ECS, and we'll need to add support for the defining `scaling_factor` in the schema. I don't see any issue with the addition - just noting that dependency.

rfcs/text/0005-host-metric-fields.md
ebeahan: ++ on aligning on `ingress` and `egress`

rfcs/text/0005-host-metric-fields.md
ebeahan: I'd prefer using term `utilization` over `percentage`. I see capturing the CPU % as the way to measure CPU utilization.

ebeahan: Or perhaps to maintain the unit in the field name, something like `usage_percentage`.

kaiyan-sheng: Sounds good! How about changing this field name to `host.cpu.usage_percentage`?

kaiyan-sheng:
> This may be out of scope for this proposal but for `host.cpu` it may also be useful to have these fields:
> 
> field	type	description
> `host.cpu.mean`	a float of some sort	Average CPU load since reboot or since the program observing the CPU was started
> `host.cpu.latest`	a float of some sort	Average CPU load for some time interval
> `host.cpu.histogram`	histogram	This field defines an elasticsearch histogram field (https://www.elastic.co/guide/en/elasticsearch/reference/current/histogram.html#histogram)
> The `histogram` field would allow visualizations to plot the cpu over time easily I think.

@jonathan-buttner These would be great addition but unfortunately we won't be able to collect these metrics for all hosts. That's why these metrics are not included in the schema.

jonathan-buttner:
> @jonathan-buttner These would be great addition but unfortunately we won't be able to collect these metrics for all hosts. That's why these metrics are not included in the schema.

Ah ok 👍

rfcs/text/0005-host-metric-fields.md
cyrille-leclerc: thanks @kaiyan-sheng I was used to ever increasing counter this kind of metric but I don't know what's the state of the art today. 
One benefit of ever increasing counters being that the collector can miss some collects without losing data.

rfcs/text/0005-host-metric-fields.md
```diff
@@ -26,6 +26,17 @@ Proposed 7 new fields are:
...
+| `host.network.in.packets` | long | The number of packets received on all network interfaces by the host in a given period of time. |
+| `host.network.out.bytes` | long | The number of bytes sent out on all network interfaces by the host in a given period of time. |
+| `host.network.out.packets` | long | The number of packets sent out on all network interfaces by the host in a given period of time. |
+| `host.disk.read.bytes` | long | The total number of bytes read successfully in a given period of time. |
```
ebeahan, line 38: For the `host.disk.[read|write].bytes` metric fields, is the idea to sum the total reads/writes  across all disks mounted on the system for the given time period?

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: @ebeahan I was thinking to keep `scaleing_factor` as default value. So value 0.12 for example will be stored into ES as 0.12 itself. Maybe I should just use `float` here instead? Thanks!

webmat: On the `scaling_factor`, it's a required parameter for this datatype, there's no default value here.

Note that the source doc is not affected by it, and would always contain the full float value.

A scaling factor of 100 would mean 0.123 (for 12.3%) would only store 0.12 in the index, for aggregations & so on. If we want to aggregate on full digit percents, a scaling_factor of 100 is appropriate.

@kaiyan-sheng or @cyrille-leclerc is 100 appropriate for this use case? Who can chime in on how observability intends to query this field?

<details>
<summary>Experiment with scaled_float</summary>

I was hoping that `scaled_float` being backed by a long meant it wouldn't have float artifacts, but with the simple test below, the last two documents ingested show up with `...0000001` in the aggregation 🤷‍♂️ 

```
PUT sfloat-1
{ "mappings" : { "properties": {
  "pct" : { "type" : "scaled_float", "scaling_factor": 100 }
}}}
PUT sfloat-1/_doc/1
{ "pct": 0.12 }
PUT sfloat-1/_doc/2
{ "pct": 0.1234 }


PUT sfloat-2
{ "mappings" : { "properties": {
  "pct" : { "type" : "scaled_float", "scaling_factor": 10000 }
}}}
PUT sfloat-2/_doc/1
{ "pct": 0.12 }
PUT sfloat-2/_doc/2
{ "pct": 0.1234 }

GET sfloat-*/_search
GET sfloat-*/_search
{ "size": 0, "aggs" : {
    "percentages" : { "terms" : { "field" : "pct" }
}}}
```

</details>

webmat (commented):
Thanks @kaiyan-sheng for this subsequent PR, and for the follow-up so far.

First here's a bit of virtual paperwork around this stage 1 RFC :-)

- [ ] Could you update to "Stage: 1 (proposal)" at the top of the Markdown?
- [x] Please add Cyrille as the sponsor representing Observability, in the "People" section at the bottom.
- [x] Please add Carlos as a subject matter expert.

I'm also commenting on some of the discussion below.

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: It is. The host.disk.* metrics will also be aggregated across all disks, similarly host.network.* metrics are also a sum across all interfaces.

rfcs/text/0005-host-metric-fields.md
```diff
@@ -41,11 +52,23 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
 ## Source data
 
 <!--
 Stage 1: Provide a high-level description of example sources of data. This does not yet need to be a concrete example of a source document, but instead can simply describe a potential source (e.g. nginx access log). This will ultimately be fleshed out to include literal source examples in a future stage. The goal here is to identify practical sources for these fields in the real world. ~1-3 sentences or unordered list.
 -->
+* Bare metal
+* VMs
+* AWS EC2 instances
+* GCP compute engines
+* Azure compute VMs
```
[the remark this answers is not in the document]
kaiyan-sheng: Good point! We haven't got to `container` in inventory schema work yet. @exekias Can we potentially treat containers as it's own VMs and report the same set of host metrics?

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: @webmat Thank you for the explanation!! I think I would prefer a scale_factor of 1000 in this case so for CPU usage, we are keeping one decimal when displaying it in a percentage format. For example, I would like to store 0.1234 to be 0.123 in the index so it will show as 12.3% in percentage. But if this is a concern with saving space for storing these values, scale_factor of 100 is also fine for me. @exekias WDYT?

rfcs/text/0005-host-metric-fields.md
sorantis: I think we're confusing two different things here.
The main purpose for this set of host metrics is to align the different resource providers, such as public cloud infrastructures, physical hosts, VMware clusters, other hypervisors, i.e. bring hosts/VMs to a common denominator. The purpose is not to capture high fidelity system metrics because some providers (AWS, GCP, Azure, etc.) simply don't provide these. 
Granted, the Metricbeat system module can do much more than just the proposed CPU metrics and should be used wherever possible. 
Be making these metrics part of ECS we want improve user experience, i.e. we can always show basic health information about any hosts anywhere, and recommend users to enable the system module for enhanced details.

rfcs/text/0005-host-metric-fields.md
```diff
@@ -26,6 +26,17 @@ Proposed 7 new fields are:
...
 Stage 1: Describe at a high level how this change affects fields. Which fieldsets will be impacted? How many fields overall? Are we primarily adding fields, removing fields, or changing existing fields? The goal here is to understand the fundamental technical implications and likely extent of these changes. ~2-5 sentences.
 -->
+This RFC calls for the addition of host fields to collect basic monitoring metrics from a host or VM such as CPU, network and disk.
+
+| field | type | description |
+| --- | --- | --- |
+| `host.cpu.usage` | scaled_float | Percent CPU used with scaling_factor of 1000. This value is normalized by the number of CPU cores and it ranges from 0 to 1. For example: For a two core host, this value should be the average of the 2 cores, between 0 and 1. |
+| `host.network.in.bytes` | long | The number of bytes received on all network interfaces by the host in a given period of time. |
+| `host.network.in.packets` | long | The number of packets received on all network interfaces by the host in a given period of time. |
+| `host.network.out.bytes` | long | The number of bytes sent out on all network interfaces by the host in a given period of time. |
```
ruflin, line 36: We have in ES now `_meta` which also support `byte` as unit: https://github.com/elastic/elasticsearch/pull/61941 Even though it is unfortunate that in older metrics we still have the postfix of `.bytes`, I think all new values we add we should skip the unit postfix and use `_meta` instead. 

@webmat ECS should add support for _meta.

exekias: This is a good point, and I think we should avoid postfixes just for the sake of making the unit explicit. In the case of `system.network.out` object (this metric comes from it) this is what we have today:

```
    "system": {
        "network": {
            "in": {
                "bytes": 37904869172,
                "dropped": 32,
                "errors": 0,
                "packets": 32143403
            }
        } 
     }
```

I wonder how we should differentiate all the different metrics there, specially between bytes and packets

rfcs/text/0005-host-metric-fields.md
exekias: We can try to do some testing and compare sizes, 100 has always looked a bit limiting, some more precision would be good

ruflin: Late to the game, but ++ on using _meta for it.

rfcs/text/0005-host-metric-fields.md
ruflin: Ha, interesting case. I'm tempted to say in this context the unit name might make sense. In any case, we should still add the info to _meta as this is what should be used by Kibana.

Alternative `bytes` could be called `data` but not sure if that is better.

kaiyan-sheng: Good point! Thank you @exekias and @ruflin. In this case, I still feel like `bytes` makes more sense in the metric name instead of `data`(too broad). `bytes` in this case, comparing with `packets`, is identifying the type of this value instead of only being the unit postfix.

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: Done! Changed to `host.cpu.usage` already 🙂 Thank you for all the input!

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: @webmat @cyrille-leclerc Agree on the benefit of ever-increasing counters. We will definitely keep all the counters as they are right now from the system module. We only added extra calculation using these counters to get gauges in system module to match whatever we get from other resource providers such as AWS, Azure, and GCP. Unfortunately, they don't provide ever-increasing counters in their monitoring metrics as @sorantis mentioned above.

Also in the UI side, it's definitely easier to have metrics as gauges than counters.

webmat (commented):
Thanks for the feedback everyone, and thanks for addressing concerns @kaiyan-sheng 

My understanding is that the following concerns have been addressed in the discussions above:

* **type & scaling factor**: if `scaled_float` with a factor of 1000 is what you need, we're good 👍 
* **normalization for multi-core**: how to normalize is clearly captured in the RFC 👍 
* **monotonic counters**: they can't always be captured, so you're going with what you can get across all sources 👍 
* **.bytes suffix**: it looks like a suffix, but it's actually the clearest field name for this. 👍 
  * I've made the same observation on the bytes fields in ECS, like `network.bytes`, `source.bytes` and so on. I agree there isn't another word that more clearly describes the content of the field. So I'm good with keeping the `.bytes` field names.

I'm still seeing two points that haven't been addressed:

- [x] **using "ingress" and "egress" instead of "in" and "out"**
  - It would be good to align with the rest of ECS on the naming here, ideally. We strive for consistency as much as possible, to make the schema predictable and easier to learn.
  - As Andrew mentioned, this wording is used in the observer fields. And in the upcoming ECS 1.7, these are two new expected values for [network.direction](https://www.elastic.co/guide/en/ecs/1.7/ecs-network.html) as well.
  - Are there concerns with using `host.network.ingress.*` and `host.network.egress.*`?
- [ ] **Does this apply to containers?**

Of the two points above, I'd really like to see if ingress/egress is possible.

The second about containers is not a blocker to merge this stage 1 PR IMO. We can always discuss this again if/when necessary.

One nipick:

- [x] The mention of scaling_factor in the `host.cpu.usage` description is confusing. Could you move this to the "type" column instead? Can be as simple as `scaled_float (scaling_factor 1000)`.

kaiyan-sheng:
@webmat Thank you for the summary. I made the change for ingress/egress and scaled_float. Regarding your question about container, we will work on the common schema for containers in the upcoming release so we would like to keep this PR only for host metrics only.

webmat (approved):
Alright, thanks for the quick turnaround :-)

Ack on addressing containers separately 👍 

I'll merge this now. When I've done that, would you mind opening the PR for the next stage of this discussion?

For now no need to overthink this next PR, this RFC looks really good already. You could simply open the PR and set "stage: 2 (draft)" at the top. This way we'll already have a place to capture the next action items and further feedback.

PR #957, opened 2020-08-27
Enterprise Content Field - Stage 1 (formerly known as Object Field)

jonasll:
**Summary:** Revisions to the [Stage 0 Object Field](https://github.com/elastic/ecs/pull/883) strawman, with intention to move to Stage 1.

## Key Changes from Stage 0 to Stage 1
- Suggesting new top-level terminology of Enterprise Content (to align with terminology in other parts of the Elastic Stack and ongoing initiatives)
- Removed hyper-specific fields from Stage 0 proposal (struckthrough in the fields table)
- Expanded intention beyond SaaS providers to on-premise productivity and collaboration enterprise platforms

### Fieldset Naming
While the original strawman suggested `object` as top-level terminology, the generic nature of the term and overlap with other ECS concepts ([as explained by Eric](https://github.com/elastic/ecs/pull/883#issuecomment-656374390)) calls for a more targeted label.

Terminology that has been discussed so far as a way to focus the scope of the fieldset's purpose is as follows:
- Resource
- Enterprise Content
- Enterprise Asset
- Enterprise Data
- Enterprise Item
- Enterprise Entity
- Enterprise
- Record

### Semantic Overlap with Existing Fieldsets

As pointed out in the original strawman, much of the data expected from third-party services can be _creatively indexed_ using the `file` fieldset, as much of the behavior inherent to filesystems also applies to SaaS tools and systems. Unfortunately, storing information for a Jira ticket's update event under the `file` realm is inaccurate and misleading.

That said, many SaaS platforms do present interesting dilemmas given the variety of enterprise data they hold. For example, Google Drive can be used to store a PDF file and a Google Sheet. G Suite treats a Google Sheet as a record, not as a file. Which yields interesting questions:
- Is an event stemming from the update of a PDF file stored on Google Drive covered by this new fieldset, or by the existing `file` fieldset?
- Is an event stemming from the update of a Google Sheet stored on Google Drive covered by this new fieldset, or by the existing `file` fieldset?
- Given the high probability of having both the items living alongside on the system, should the indexed event information as it relates to these items' activity be modeled similarly (source-level modeling), or should their naming be consistent with the inherent nature of the object (item-level naming)?

### Field Value Enforcement/Standardization
The use of a type field (`object.type`) is meaningful in understanding the general conceptual purpose of the entity. For example, a _ticket_ and a _meeting_ are conceptually very different in the real world, but share commonalities for the purpose of representing the information in an index. That said, consistent naming of the values for such a field is key to standardization across sources. Presenting a glossary, either enforced or convention-based, would likely help with ensuring standardization of values across implementations.

### Scope of Initial Fieldset
**This proposal focuses solely on the event-relevant fields for enterprise data and resources.** As ECS continues to evolve, we expect to amend this fieldset with semantically relevant fields for documents themselves, in an effort to standardize concepts across enterprise systems (for example, _resource.thumbnail_ or _enterprise_item.excerpt_).

---

**Criteria for consideration**

- [x]  Opened pull request revising the existing strawperson
- [x]  Identified "sponsor" at Elastic who will participate in RFC process and take ownership of the change after the process completes
- [x]  Types of fields or changes outlined
- [x]  High-level description of examples of usage
- [x]  High-level description of example sources of data
- [x]  Identified potential concerns and implementation challenges/complexity
- [x]  Subject matter experts identified and weighed in on the high level utility of these changes in the pull request
- [ ]  ECS team weighed in on appropriateness of these changes in the pull request

[Markdown preview of this RFC](https://github.com/elastic/ecs/blob/object-stage-1/rfcs/text/0003-object-field.md)


thomheymann:
I noticed this proposal has been updated to use `enterprise_content` instead of `object` field. 

What's the reasoning behind this? It seems overly specific to me. 

I'm very keen on having a field like the one originally proposed here for Kibana audit logging where we need to log that a saved object was accessed or written to. 

In addition to that I think there's a more general use case for having a field that customers can use to log events related to objects/resources as it would be useful for any web app using REST and wanting to log events related to resources operated on.

jonasll:
@thomheymann The original intent of the proposal was to capture enterprise content-type data according to the Stage 0 proposal, which is why it was picked up and internally sponsored by the Enterprise Search team. 

This fieldset is indeed being moved to a more specific name convention for 2 main reasons:
1. The original `object` designation was far too broad and unbound - semantically speaking, it didn't particularly point in one direction or another
2. The Enterprise Search team will be taking steps in expanding this fieldset beyond the current event-focused perspective to entity storage. This means storing actual content (searchable corpus) under the same fieldset.

It seems like what you are describing here indeed overlaps with the Stage 0 proposal, but I suspect it was mainly due to its generic nature, and not its underlying intent. I'd love to hear @webmat 's perspective on how to best move forward given your use case.

qhoxie:
Some thoughts for how this applies to various Enterprise Search use-cases:

- I agree with removing additional_details, we don't need catch-all fields like this as you can just add more ad-hoc fields as needed.
- I think one of the biggest changes to Stage 0 is moving from addressing the object at the top-level to addressing the source and object (`document`) as siblings at the top-level.  I think this is a fair change.  `source` and `service` at the ECS top-level are not ideal for us here, but I could see an argument that we should use those.  It could also make sense since we'll likely be addressing just one source at a time.  If we decide to keep the current direction, it feels like `uri` might belong nested under `document`.
- [Lame naming thought] `owner` feels like it could be restrictive for us and others. Will an owner be critical and known here in most of our cases or would we be better served by `user`?
- [Lame naming thought] `document` feels overloaded with ES documents.  I don't have a great suggestion, but I generally like entity or record.  Record feels like a good term since many popular enterprise systems use it in their docs.

drewgatchell:
Just to remove any doubt on my original intent; this statement by @thomheymann:

> I'm very keen on having a field like the one originally proposed here for Kibana audit logging where we need to log that a saved object was accessed or written to.

Very much aligns with the original intent. In fact if it didn't; this field becomes a half-measure IMO.

To address some other points...

I can't say I'm a fan of the "enterprise_content" name; particularly when there is a notion that this field could also cover IaaS/PaaS logging as well as documented in the original PR comments. I know internally at my company, we settled on `resource` while we continue to prove out how this field works. But I understand where this is coming from in this PR.

`document` seems meh; mostly because I don't particularly like the notion of addressing a GitHub Repo or GitHub Org as a `document`; the terminology seems off; but I honestly don't have a better suggestion if we insist on nesting this content.

I'm unsure of the value of `source` as a sibling to `document`; do logs/events exist where `enterprise_content.source` would not be the same as `source.` or `service.`? In our experience that hasn't happened yet; but we haven't dove into any of the on-prem solutions highlighted in this PR.

To address some of @qhoxie's comments:

> I agree with removing additional_details, we don't need catch-all fields like this as you can just add more ad-hoc fields as needed.

This is a fair point; my original intent was largely around giving a specific place for putting those ad-hoc fields. In fact, in our internal usage of trying to prove out this field, we actually went as far as documenting that for a specific `type` of `resource/object/enterprise_content` you could expect a static set of fields under `additional_details`. I'm not sure I would have necessarily included that in the specification, as that's an internal implementation detail; but I just wanted to provide a frame of reference around that concept.


>it feels like uri might belong nested under document.
I addressed my concerns around `source` above; but I agree with this sentiment. 

>[Lame naming thought] owner feels like it could be restrictive for us and others. Will an owner be critical and known here in most of our cases or would we be better served by user?

So, I specifically used `owner` as I didn't want to induce confusion around `user`. Specifically in terms of "User X modified Resource Y Owned By User Z";

I didn't want someone to populate `enterprise_content.user.id` with the information for User X; as that should go under the root level `User.*` object (as established in the previous PR comments); If we feel like that's a non-issue; than I have no real objections to renaming this field. I will callout that it's still slightly overloaded because on some platforms a `resource` can be owned by Group and not a user; but I haven't put much thought into how to call that out.

>[Lame naming thought] document feels overloaded with ES documents. I don't have a great suggestion, but I generally like entity or record. Record feels like a good term since many popular enterprise systems use it in their docs.

As said above, I agree.. but for slightly different reasons. `Record` is generally ok; but doesn't extend to IaaS/PaaS (which I'm not the best advocate for, since my concern is around SaaS).

jthack:
Mostly just agree @drewgatchell, but I think the plainest way it can be stated is that "not all generic resources are going to be from (or for) the enterprise."

drewgatchell:
@jonasll 

>The Enterprise Search team will be taking steps in expanding this fieldset beyond the current event-focused perspective to entity storage. This means storing actual content (searchable corpus) under the same fieldset.

Is there somewhere we can read up on the why of this decision?

This seems concerning in that from the outside perspective that this doesn't at all align with this statement:

>The Elastic Common Schema (ECS) is an open-source specification, developed with support from the Elastic user community. ECS defines a common set of fields to be used when storing event data in Elasticsearch, such as logs and metrics.
...
The goal of ECS is to enable and encourage users of Elasticsearch to normalize their event data, so that they can better analyze, visualize, and correlate the data represented in their events. ECS has been scoped to accommodate a wide variety of events, spanning:

Is the intent to shift ECS away from just event data? Are there worries that this might muddy the waters and ultimately dilute usage of ECS and just encourage people to choose another schema or go "out of spec"? If there's a better place for this conversation, by all means, let me know.

rfcs/text/0003-object-field.md
```diff
@@ -1,30 +1,33 @@
-# 0003: Object Fields
+# 0003: Enterprise Content Fields (Previously known as Object Fields)
```
ebeahan, line 1: When we get agreement on the renaming, we can also rename the file (e.g. `rfcs/text/0003-enterprise-content.md`)

rfcs/text/0003-object-field.md
```diff
@@ -1,30 +1,33 @@
...
+Numerous SaaS and enterprise content/productivity services provide event data where the event reflects an action (typically `create`, `read`, `update`, `delete`) on what can broadly be considered items/documents/resources/records (originally dubbed "objects"). The are generally considered audit logs, or business analytics events. For some of these services, parts of (or in some cases, the entirety of) the corpus consists of _files_ (in the filesystem sense of the term) wherein the existing ECS `file` fields are adequate. However, most enterprise collaboration, storage, and communication systems (Workday, Salesforce, ServiceNow, Zoom, GitHub, G Suite, Zendesk, Jira, Confluence - Cloud - Server, etc.) present data units as "records", "documents", "tickets", "meetings" and more. ECS doesn't currently account for these entities, and relies on `file` as an option, albeit semantically incorrect and incomplete for the purpose of event tracking.
```
ebeahan, line 10:
```suggestion
Numerous SaaS and enterprise content/productivity services provide event data where the event reflects an action (typically `create`, `read`, `update`, `delete`) on what can broadly be considered items/documents/resources/records (originally dubbed "objects"). They are generally considered audit logs, or business analytics events. For some of these services, parts of (or in some cases, the entirety of) the corpus consists of _files_ (in the filesystem sense of the term) wherein the existing ECS `file` fields are adequate. However, most enterprise collaboration, storage, and communication systems (Workday, Salesforce, ServiceNow, Zoom, GitHub, G Suite, Zendesk, Jira, Confluence - Cloud - Server, etc.) present data units as "records", "documents", "tickets", "meetings" and more. ECS doesn't currently account for these entities, and relies on `file` as an option, albeit semantically incorrect and incomplete for the purpose of event tracking.
```

rfcs/text/0003-object-field.md
```diff
@@ -1,30 +1,33 @@
...
-| `object.owner.id` | keyword | ID of the Object Owner |
-| `object.owner.name` | keyword | Name of the Object Owner |
-| `object.owner.email` | keyword | Email of the Object Owner |
+| `enterprise_content.document.name` | keyword | Name of the Enterprise Content Object |
+| `enterprise_content.document.id` | keyword | ID of the Enterprise Content Object |
+| `enterprise_content.document.type` | keyword | Type of Enterprise Content Object represented (e.g. `record`, `meeting`, `repository`, `organization`, etc.) |
+| `enterprise_content.uri` | keyword | Absolute location of Enterprise Content Object  |
+| `enterprise_content.source.id` | keyword | ID for the Enterprise Content Object source |
+| `enterprise_content.source.name` | keyword | Name for the Enterprise Content Object source |
+| `enterprise_content.owner.id` | keyword | ID of the Enterprise Content Object Owner |
```
ebeahan, line 27: ECS nests `user.*` in several places, and with [multiple user support] we can nest under a different name (nest user fields as `owner.*`).

Perhaps that could make sense here?

rfcs/text/0003-object-field.md
```diff
@@ -1,30 +1,33 @@
...
 | field | type | description |
 | --- | --- | --- |
-| `object.name` | keyword | Name of the Object |
-| `object.id` | keyword | ID of the Object |
-| `object.type` | keyword | Type of Object Represented (e.g. `record`, `meeting`, `repository`, `organization`, etc.) |
-| `object.additional_details` | object | Custom Key/Value pairs representing data relevant to the object |
-| `object.owner.id` | keyword | ID of the Object Owner |
-| `object.owner.name` | keyword | Name of the Object Owner |
-| `object.owner.email` | keyword | Email of the Object Owner |
+| `enterprise_content.document.name` | keyword | Name of the Enterprise Content Object |
```
ebeahan, line 21:
```suggestion
| `enterprise_content.document.name` | wildcard | Name of the Enterprise Content Object |
```

Multiple `.name` fields (`process.name`, `organization.name`, etc.) are being targeted for the initial [wildcard adoption](https://github.com/elastic/ecs/blob/master/rfcs/text/0001-wildcard-data-type.md).

We can discuss more once we have example field values to compare against, but `wildcard` could make sense as the type here as well.

rfcs/text/0003-object-field.md
```diff
@@ -79,7 +88,7 @@ The goal here is to research and understand the impact of these changes on users
...
+An `object` - as initially proposed - is a very broad categorization, in some situations it may be hard to determine when something should be normalized into this field or another field like `file` or `package`. the **Enterprise Content** terminology seeks to address this by focusing on the non-time series nature of the target documents. Additionally given the genericness of this field, a lot of finer details about the object may not be represented by this field, but rather addressed in custom key/value pairs nested under this field. `object.type` conventions will be critical to designate what exactly is represented within this field (e.g. is it a `record` or a `meeting`); some context might be derived from other fields in the event, but that would require some foreknowledge in order to properly search, sort or infer. Designating specific acceptable values for `object.type` could minimize any confusion around the object being represented; however that would require specific updates to ECS to support new object "types".
```
ebeahan, line 91:
```suggestion
An `object` - as initially proposed - is a very broad categorization. In some situations it may be hard to determine when something should be normalized into this field or another field like `file` or `package`. The **Enterprise Content** terminology seeks to address this by focusing on the non-time series nature of the target documents. Additionally given the genericness of this field, a lot of finer details about the object may not be represented by this field, but rather addressed in custom key/value pairs nested under this field. `object.type` conventions will be critical to designate what exactly is represented within this field (e.g. is it a `record` or a `meeting`); some context might be derived from other fields in the event, but that would require some foreknowledge in order to properly search, sort or infer. Designating specific acceptable values for `object.type` could minimize any confusion around the object being represented; however that would require specific updates to ECS to support new object "types".
```

Diff isn't very useful here. Just a couple of minor proofreading items noted while reviewing.

ebeahan (commented):
Thanks for sponsoring this effort, @jonasll ! I have some notes below.

rfcs/text/0003-object-field.md
```diff
@@ -1,30 +1,33 @@
-# 0003: Object Fields
+# 0003: Enterprise Content Fields (Previously known as Object Fields)
 <!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage -->
-- Date: **2020-07-27** <!-- Update to reflect date of most recent stage advancement -->
+- Stage: **1 Proposal** <!-- Update to reflect target stage -->
+- Date: **2020-09-23** <!-- Update to reflect date of most recent stage advancement -->
```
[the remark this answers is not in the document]
jonasll: 👍

rfcs/text/0003-object-field.md
jonasll: 👍

rfcs/text/0003-object-field.md
jonasll: Deferring to @ioanatia and @qhoxie on best approach.

rfcs/text/0003-object-field.md
jonasll: 👍  FYI @ioanatia

rfcs/text/0003-object-field.md
jonasll: Thank you

jonasll:
> This seems concerning in that from the outside perspective that this doesn't at all align with this statement: The Elastic Common Schema (ECS) is an open-source specification, developed with support from the Elastic user community. ECS defines a common set of fields to be used when storing event data in Elasticsearch, such as logs and metrics.

@drewgatchell This was the original intent of ECS, but it was always known that ECS would grow beyond event data. I would let @ebeahan or @webmat (core maintainers) to comment further on whether or not this has risk to impact the adoption of the schema. From my perspective though, it reinforces its value by providing additional tools to make sense of data across use cases -- for example, properly doing threat detection (Security) on enterprise content (Search).

rfcs/text/0003-object-field.md
qhoxie: I'm not sure I completely understand the comment.  Is it that we can have multiple aliases for the `user.*` fields here to connote various roles like owner?  If so that sounds good to me.

webmat:
I won't try to answer all the questions here. I've started such a draft and I'm up to a screenful of text, and I'm only halfway done 😂 

Let's swiftly untie a few knots:

* ECS will grow to help define entity indices -- that don't contain events, but rather regularly updated documents. This will be useful in the security use case (db of users, hosts, IOCs, etc).
* ECS will expand outside of its current focus of security & observability, to search use cases as well. This is where @jonasll and the Enterprise Search team come into play.

How we document this expansion of scope of ECS is outside of the scope of this discussion. But we know we'll have to address this.

Now I see both sides of what's making this second RFC PR contentious.

* The Enterprise Search team will operate on enterprise documents, and wants to ease into ECS by capturing *events* about these enterprise documents. Like an audit log.
* My understanding of the initial stage 0 PR was to introduce a place to capture any auditable *thing* that doesn't have a neat place in ECS (and that may never need one either). This was intentionally very broad. There can potentially be audit logs on all things imaginable, at all kinds of conceptual levels.

There's an overlap with the documents Enterprise Search wants to audit, of course. But the original intent was to keep this auditable "object" mostly unbounded, to simply let folks capture all of their audit logs, whether they're about business docs or records, firewall rules, all the way to the most esoteric things.

I think we should separate these two initiatives again. We can work on both proposals in parallel, without the contention. We can decide later if there's enough similarities to combine them, or come up with guidance on when each is appropriate.

But it's worth pointing out that it's always been a challenge to address for the unbounded auditable object, whether we should populate it, when that auditable also has a neat place in ECS: IAM activity (user), FIM (file), etc. Perhaps the guidance we establish for these other cases would be all that's needed here as well?

What do you think?

rfcs/text/0003-object-field.md
ebeahan: I didn't have [this reference link](https://github.com/elastic/ecs/blob/master/rfcs/text/0007-multiple-users.md) correct before in my original comment. 🤦 

ECS already makes use of `user.*` field reuse, with nestings such as `client.user`, `destination.user`, etc. Recently support was added to not only nest `user.*`, but nest the `user.*` fields under a different name.

The `enterprise_content` fieldset could keep the `enterprise_content.owner.*` naming and nest the  [`user.*` fields](https://www.elastic.co/guide/en/ecs/current/ecs-user.html) underneath: `enterprise_content.owner.id`, `enterprise_content.owner.name`, `enterprise_content.owner.email`, etc.

webmat:
To follow-up @ebeahan's suggestion and mine above, I think the clearest way to separate these two initiatives again might be to have RFC 0003 remain about the unbounded auditable object, and for this RFC to get a new RFC number.

So if we agree to separate these two initiatives, all that's needed in this PR is to:

* reset the RFC number to 0000 in the file name and in the title
  * the best way for git not to get confused here, is actually to create a new file named `0000-enterprise-content.md` and reset the content of `0003-object-field.md` as it was before.
* remove references about renaming or repurposing RFC 0003, as this becomes a new, separate RFC.

jonasll:
I'm not certain I support splitting this into two proposals. If the original proposal was in fact broader than interpreted, then our use case would also benefit from the same set of fields. The semantic storage expansion would simply occur in its own `enterprise_content` fieldset. Therefore, we would drop the specificity here in favor of broader scope.

That said, what I still think is up for discussion is the naming on the fieldset. I feel strongly that _object_ or _resource_ are far too broad. It also appears that _enterprise_content_ is too specific. I'm out of ideas, but will keep thinking.

Finally, given the broader approach, it may be worth considering shifting sponsorship internally. @webmat @ebeahan Thoughts are welcome.

webmat:
Alright, thanks @jonasll, and thanks for your openness to follow what comes out of this RFC, in the Enterprise Search audit events. We can discuss the shift in sponsorship at the next committee meeting. Are you also implying that you'd prefer to leave authorship of this RFC to someone else?

Note that I wanted to take the opportunity of a split to two RFCs to talk about what's needed for the searchable corpus. But we can also talk about this at the meeting.

jonasll:
This RFC is moot if we revert back to a generic approach. The Enterprise Content fieldset for entity/semantic storage will come later, and is not an immediate priority. Therefore, I am suggesting that this specific RFC is closed, and that the original proposal be sponsored by another group with more visibility across use cases.

webmat:
Alright, thanks for clarifying Jonas. I will close this PR for now.

We'll find another author to pick up from the stage 0 again. @drewgatchell or other folks who chimed in, if you'd like to give stage 1 go, you're welcome to do so :-)

PR #964, opened 2020-09-08
Remove `expected_event_types` from protocol

ebeahan:
As an `allowed_value` under `event.type`, `protocol` shouldn't have the `expected_event_types` section defined.

This change will remove from the schema as well as in the ECS documentation.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -14,6 +14,8 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
+* The `protocol` allowed value under `event.type` should not have the `expected_event_types` defined. #964
+
 #### Added
 
 * Added Mime Type fields to HTTP request and response. #944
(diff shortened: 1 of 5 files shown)


webmat (approved):
LGTM 🚢

webmat:
@ebeahan Hey when you have a sec, could you backport this fix to 1.7 as well? I see it's still visible in ECS docs.

ebeahan:
I believe it is already in 1.7? https://github.com/elastic/ecs/commit/3eb6d9962c9a551467389e5cd2e110e5592d2a3a

I didn't double-check before opening up the backport PR 😂 . I'll leave it open while we track this down.

webmat:
I still see the "expected event types" blurb [under the protocol event.type](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html#ecs-event-type-protocol), which shouldn't be present.

Yeah that's my bad, "current" is actually 1.6. This should be backported to 1.6, 1.7 is already done indeed :-)

PR #966, opened 2020-09-09
Add example value composite type checking

### LINKED ISSUE #781: All examples that are a composite type (array, object) should be quoted, to avoid YAML interpreting them
This issue mostly affects objects (like the example for `labels`). But it could also affect arrays, if for example we have an array that contains an unquoted `NO`, which would be interpreted by YAML as False, rather than the string "NO".

This issue not only affects the generated artifacts in the ECS repo, but also artifacts down the line, e.g. Beats starting from `generated/beats/fields.ecs.yml` and in turn generating various other artifacts.

- [x] #782 Immediate need is to fix `labels` to double quote it, an each values inside, to fix an issue in Beats
- [ ] #966 Fix all of the arrays
- [ ] #966 Add a linting rule or fail the build if any `example:` contains directly an array or an object, rather than a scalar.

ebeahan:
#### Summary

Adding a strict mode (`--strict`) check that requires array or object in `example` fields be enclosed with quotes (making strings). If these composite values are not handled as strings, YAML could incorrectly interpret them in generated artifacts or other downstream uses.

#### Changes

* Introduce new function, `schema.cleaner.check_example_value()`, to perform the type checking.
* Add supporting unit tests
* Update both `schemas/README.md` and `USAGE.md` documenting.
* Identified and fixed violating field examples in the schemas.

Closes #781

### DIFF
--- a/generated/csv/fields.csv
+++ b/generated/csv/fields.csv
@@ -126,7 +126,7 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 2.0.0-dev,true,dns,dns.question.subdomain,keyword,extended,,www,The subdomain of the domain.
 2.0.0-dev,true,dns,dns.question.top_level_domain,keyword,extended,,co.uk,"The effective top level domain (com, org, net, co.uk)."
 2.0.0-dev,true,dns,dns.question.type,keyword,extended,,AAAA,The type of record being queried.
-2.0.0-dev,true,dns,dns.resolved_ip,ip,extended,array,"[10.10.10.10, 10.10.10.11]",Array containing all IPs seen in answers.data
+2.0.0-dev,true,dns,dns.resolved_ip,ip,extended,array,"[""10.10.10.10"", ""10.10.10.11""]",Array containing all IPs seen in answers.data
 2.0.0-dev,true,dns,dns.response_code,keyword,extended,,NOERROR,The DNS response code.
 2.0.0-dev,true,dns,dns.type,keyword,extended,,answer,"The type of DNS event captured, query or answer."
 2.0.0-dev,true,ecs,ecs.version,keyword,core,,1.0.0,ECS version this event conforms to.
--- a/generated/csv/fields.csv
+++ b/generated/csv/fields.csv
@@ -117,7 +117,7 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 2.0.0-dev,true,dns,dns.answers.name,keyword,extended,,www.example.com,The domain name to which this resource record pertains.
 2.0.0-dev,true,dns,dns.answers.ttl,long,extended,,180,The time interval in seconds that this resource record may be cached before it should be discarded.
 2.0.0-dev,true,dns,dns.answers.type,keyword,extended,,CNAME,The type of data contained in this resource record.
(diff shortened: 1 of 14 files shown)


generated/csv/fields.csv
```diff
@@ -117,7 +117,7 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 2.0.0-dev,true,dns,dns.answers.name,keyword,extended,,www.example.com,The domain name to which this resource record pertains.
 2.0.0-dev,true,dns,dns.answers.ttl,long,extended,,180,The time interval in seconds that this resource record may be cached before it should be discarded.
 2.0.0-dev,true,dns,dns.answers.type,keyword,extended,,CNAME,The type of data contained in this resource record.
-2.0.0-dev,true,dns,dns.header_flags,keyword,extended,array,"['RD', 'RA']",Array of DNS header flags.
+2.0.0-dev,true,dns,dns.header_flags,keyword,extended,array,"[""RD"", ""RA""]",Array of DNS header flags.
```
ebeahan, line 120: Noting some behavior here - the CSV generator is double quoting string values when the schema specification already contains a quoted value.

schemas/README.md
```diff
@@ -125,7 +125,9 @@ Supported keys to describe fields
   Defaults to the main description when absent.
   If the main description has multiple paragraphs, then a 'short' description
   with no newlines is required.
-- example (optional): A single value example of what can be expected in this field
+- example (optional): A single value example of what can be expected in this field.
+  Example values that are composite types (array, object) should be quoted to avoid YAML interpretation
+  in ECS-generated artifacts and other downstream projects depending on the schema.
```
webmat, line 130: Attention to details ❤️  🙂

schemas/dns.yml
```diff
@@ -54,7 +54,7 @@
         Array of 2 letter DNS header flags.
 
         Expected values are: AA, TC, RD, RA, AD, CD, DO.
-      example: [RD, RA]
+      example: "[\"RD\", \"RA\"]"
```
webmat, line 57: Nit: Another approach when the complex value already contains `"` would have been to wrap in single quotes instead. `'["RD", "RA"]'`. YAML will interpret the outer string and the final result will be exactly the same. Just less backslashes to write :-)

But no need to adjust, this if fine.

schemas/dns.yml
```diff
@@ -205,6 +205,6 @@
         data formats it can contain. Extracting all IP addresses seen in there to
         `dns.resolved_ip` makes it possible to index them as IP addresses, and
         makes them easier to visualize and query for.
-      example: [10.10.10.10, 10.10.10.11]
+      example: "[10.10.10.10, 10.10.10.11]"
```
webmat, line 208: In the previous example, YAML was complementing my laziness and turning each array element into a string automatically.

Now that the string is quoted, we should make sure to double quote each element within, so that the example in the docs looks like a valid JSON snippet.

```suggestion
      example: '["10.10.10.10", "10.10.10.11"]'
```

webmat (commented):
One very minor adjustment on DNS `resolved_ip` example (see comment below) and this will be ready to merge 🚢

generated/csv/fields.csv
ebeahan: This ended up being expected CSV behavior for the `csv.writer()` object 😄 .

ebeahan:
Thanks @webmat for the review! I tidied up that minor adjustment, so I'll go ahead and 🚢

PR #967, opened 2020-09-09
Expand definitions of source and destination field sets

ebeahan:
Update the descriptions for both `source.*` and `destination.*` field sets:

* Explicitly state that `source.*` and `destination.*` are considered the ECS baseline for capturing the details of an event's network source & destination. These field sets should _always_ be populated first.
* The `client.*` and `server.*` field sets should be _also_ be populated for added context if the event identifies those roles in the exchange.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -23,6 +23,8 @@ Thanks, you're awesome :-) -->
 
 #### Improvements
 
+* Expanded field set definitions for `source.*` and `destination.*`. #967
+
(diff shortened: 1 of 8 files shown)


webmat (approved):
LGTM

I'd love it if client/server was a bit more forceful in stating that source/destination should always be populated and are considered the baseline for any network event. This could be addressed in this PR or as a separate PR, however. The PR as it is right now is definitely a good step forward.

Issue #228, opened 2018-12-04
Need related event field?


MikePaquette:
Need an event.related_id, which would include a pointer to earlier related events for an issue (ex. Windows Event Logs).

No. 11 of 16.  This question was asked by a new ECS user,  who is familiar with mapping IT events to data models and use cases in other schemas.  These questions are being posted as a GitHub issue, because a) they may offer valuable insights.  b) we expect that many new users will have similar questions.

MikePaquette:
This is a good idea.
We should discuss whether this fits better under the `related.*` [fields](https://github.com/elastic/ecs/blob/master/README.md#related), where it could be an array that would allow for multiple related events.
Perhaps as `related.event_id`

This might be more consistent with the other uses of `related.*`

webmat:
In my mind, the `related.*` field set is specifically to group all values of one event in one field, to make sure it's easy to find "any event involving a given IP/username/hostname/..."

I would tend to think that the semantics of referring to a previous event are different, however. I think I'd rather see previous events referred to as a field on `event`. Perhaps `event.previous_id` or `event.related_id`. I would love to hear more about the scenario(s) in question where this happens, the semantics of this relationship, see what the data looks like, etc.

For example: is it a chain of relations to the previous event (and the previous event may also optionally point to one previous event)? Is it a growing array of all the previous related events?

I could see partial and final network flows needing this. I wonder if this comes up in other situations as well.

djptek:
This is being discussed here: https://github.com/elastic/ecs/issues/1547 under correlated events, closing

PR #968, opened 2020-09-14
Add related.event field

mbrancato:
This is a non-breaking, minor change to add a new keyword `event` field under the `related` object for tracking related, previous events. While I understand `related.user`, `related.ip`, etc (related but no context like source, destination), one of the first things I wanted to relate events to was other events. For example, and authorization event to an authentication event. Or something as generic as a file closed event related to a file open event. 

My primary use-case is security-related and much more complex where we enrich events with information from previous events. By relating event entries to previous events, this provides an investigative pivot point. Relationships could be based on session ID, or similar information.

Since this is only a minor change and hopefully not controversial, I've decided to add this via PR.

### DIFF
--- a/code/go/ecs/related.go
+++ b/code/go/ecs/related.go
@@ -42,4 +42,7 @@ type Related struct {
 	// All hostnames or other host identifiers seen on your event. Example
 	// identifiers include FQDNs, domain names, workstation names, or aliases.
 	Hosts string `ecs:"hosts"`
+
+	// Previous event IDs that are related to this event.
+	Event string `ecs:"event"`
 }
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -4627,6 +4627,22 @@ A concrete example is IP addresses, which can be under host, observer, source, d
 
 // ===============================================================
 
+| related.event
+| Previous event IDs hat are related to this event.
+
+type: keyword
+
+
+Note: this field should contain an array of values.
+
+
+
+
+
+| extended
+
+// ===============================================================
+
 | related.hash
 | All the hashes seen on your event. Populating this field, then using it to search for hashes can help in situations where you're unsure what the hash algorithm is (and therefore which key name to search).
 
--- a/schemas/related.yml
+++ b/schemas/related.yml
@@ -53,3 +53,12 @@
         identifiers include FQDNs, domain names, workstation names, or aliases.
       normalize:
         - array
+
+    - name: event
+      level: extended
+      type: keyword
+      description: >
+        Previous event IDs that are related to this event.
+      normalize:
+        - array
+


ebeahan:
Hi @mbrancato! Thanks for opening the PR - awesome seeing the generated artifacts and `CHANGELOG.next` entry already included 😄 .

Adding some background that you already partially touched on. The [`related.*` field set](https://www.elastic.co/guide/en/ecs/current/ecs-related.html) as it exists in today is to facilitate pivoting between related records where the data might exist in different fields. Various `related.*` fields are populated with values copied from other fields within the **same** event. This simplifies searching; values from different fields are grouped together in one place (e.g. `related.ip:10.42.42.42`).

A `related.event` field would be a departure from this current convention of populating `related.*` fields with data from the event itself. In most cases, I imagine this type of field would be populated with data coming from **different** events. Is this true in your use cases? Any other details you can provide about the relationships between your events that need capturing?

I think the semantics of a referring to previous events are different enough, placing them underneath `related.*` may not be the best fit. @webmat previously suggested `event.previous_id` or `event.related_id` https://github.com/elastic/ecs/issues/228#issuecomment-445962173. I agree that a field underneath `event.*` may be a better approach. The linked comment also mentions some additional good points on establishing a convention of event chaining vs. an array/list approach.

Since `session IDs` were mentioned, I will call attention to an active discussion around a new top-level field set, `session`, which is actively in the RFC process: #935. The proposed `session.id` field might partially address use cases where discrete events (e.g. network flows) are part of a single, related session.

mbrancato:
Thanks for the feedback @ebeahan  - We're using this to relate events that influenced the creation of an alert. Imagine if a watch triggers when more than 5 errors occur in a 1 minute timespan, the action here would be to create an alert and add the  5 related error event IDs. This way, someone investigating the alert can pivot to those 5 events easily.

I think I mainly view this as another sort of enrichment and want to use it as such. For example, one of the main uses of the related array is you can search for an IP in `related.ip` and get back all events related to that IP address, regardless of where they show up in the event. This use-case of searching drives my need here, not the constraint that the data come from the same event. As this top-level object is described as `To facilitate searching`, I think it too is use-case driven, not constrained to data inside the original event. To me, being related to an event, doesn't necessarily mean that the event contains that information. Otherwise, it might be better to make `event.contains.ip` etc. fields to be a bit more explicit that it is constrained on the event source. I believe it is acceptable to do a hostname lookup enrichment and put the IP in `destination.ip` and `related.ip` - and that is very similar to the use case here.

Take a web log of some sort for a generic example:
`198.51.100.3 - - [1/Jan/2020:12:00:00 +0000] "GET /index.html HTTP/1.1" 200 1024`

I'll refrain from typing up the ECS here, but things we don't see in the original event that might be in the terminal version of the ECS events are: hostnames of the client and server, geo-location of the client and server.

How you get these missing fields is debatable, but let's assume we know the web virtual host / domain name from the log source and that gets added to the event. For the client we do a PTR lookup and add that to the event. We use a database to add the geo information for client and server to the event. All this, IIRC, is common to do with tools like Logstash, but you could argue these lookups are not from the same event.

Once this data is obtained, it would be important the add the hostname of the client and server to `related.hosts`. Then, when an analyst wants to find all the access logs from a specific client by hostname, they can use the `related.hosts` fields to search.

This is similar to what we are doing. We query existing events for context and return a list of related events. In the description for these fields, it also seems like this field is supposed to be used for the result of a search - but what if it is used for the input to a search? In my web example above, an analyst can take the data from `related.hosts` and use that for a new query for the value in `host.name`.

On the linked comment / issue of a previous event ID - our use case here is not sequential, it is simply related. The desire is not to be able to replay something, but it's probably achievable with some work like ordering by timestamp. That said, the goal here is an unordered list of related event IDs.

These aren't session ID related, my example was probably too focused on web stuff.

Issue #948, opened 2020-08-20
Improve documentation to ensure source/destination are populated as a priority


webmat:
The definition of the Client & Server field sets are pretty extensive, whereas Source & Destination's definitions are pretty bare.

We've been encouraging people to populate source & destination as a baseline, and client & server only when relevant or helpful. Elastic Security mostly considers source & destination.

However the ECS documentation doesn't make that obvious. There's only a vague mention about prioritizing source/destination in the client/server definitions. No mention of this in source/destination.

- [x] #967 We should expand the definitions of the source & destination field sets to clearly state that they are the baseline, compared to client & server.
- [x] #969 We could also consider having a standalone documentation page that talks about capturing network related events holistically. It could cover:
  - the src/dst baseline, and showcase when cli/srv are useful (e.g. DNS)
  - it could discuss populating the `network.*` fields
  - it could discuss how `event.category:network` + `event.type:protocol` should always come with `network.protocol:[appropriate protocol name]`. Most category/type pairs are complete on their own. But not the pair network/protocol; it should come with `network.protocol`.

ebeahan:
> We could also consider having a standalone documentation page that talks about capturing network related events holistically

++ I think this makes a lot of sense to capture this and in the future other ECS design patterns.

ebeahan:
@jamesmotherway I've opened https://github.com/elastic/ecs/pull/967 aiming to address the first bullet point, if you have any feedback there.

ebeahan:
Closing as #967 and #969 have both been merged 🎉

PR #969, opened 2020-09-14
Add mapping network event guidance doc

ebeahan:
Add a new documentation page, `Mapping Network Events`, which provides guidance and best practices for mapping network-related events to ECS.

Goals:

* Provide better guidance on the usage of `source/destination` and when `client/server` should also be populated.
* Provide example application of the `network.*` & `related.*` field sets.
* Document that `network.protocol` is always populated when an event is categorized as: `event.category:network` + `event.type:protocol`.

Relates to #948 

[Docs Preview](https://ecs_969.docs-preview.app.elstc.co/diff)


docs/additional.asciidoc
```diff
@@ -2,10 +2,12 @@
 == Additional Information
```
ebeahan, line 2: Had originally placed this new page under `Additional Information`, but I'm feeling it's not the right fit.

Thoughts on a better section?

dainperkins:
I don't quite understand the source/dest vs client/server differentiation you are shooting for (assuming complete session visibility for the purposes of the discussion)

- From a strict TCP standpoint they should be the same (assuming you can determine which side is the syn, and which side is the ack, or which station sent the first packet - which from e.g. a netflow or packetbeat perspective is usually not a problem - aws vpcflow as the exception unless things have changed.  w/o syn/ack or first packet its a miserable process to determine which side is the initiator (aside from e.g. well known ports - but get into mongodb and its literally a toss up as everything is high random ports)

- From a udp standpoint its also going to depend on where the first packet came from, the difference being the 5 tuple isn't correlated with a higher level session, but theres still a dns request (randomPortA to 53) and response (53 -> randomPortA)

Its been a really long time since I looked at netflow5 unidirectional flows, but,  the observer (assuming no asymmetric routing) should still know the source and destination and track the same across src->dst & dst->src connection (both flows should use the same src/dst identifiers)

Is there some use case / corner case I'm missing (or maybe just some esoteric blind spot I can't get past?)

webmat:
@dainperkins what we're trying to do here is simply to more clearly establish that we want source+destination to always be populated. Then when it makes sense / is desired, folks can use client+server.

So it's literally a matter of trying to help make sure people can reliably depend on source+destination to be populated across the hundreds of data sources someone may have to deal with.

Now if you see a problem with the current wording, we can always improve on that.

docs/mapping-network-events.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+    "protocol": "dns",
+    "type": "ipv4",
+    "transport": "udp"
+  }
+----
+
+[float]
+==== Related fields
+
+The `related.ip` field captures all the IPs present in the event in a single array:
```
dainperkins, line 116: maybe call out the specific ips to include in the related.ip field in the example?

> The `related.ip` field captures all the IPs present in the event in a single array, in this case including source.ip, destination.ip, as well as any IPs included in the dns.answer field:

Should we include e.g. pipeline code to show how to determine if a dns.answer is an IP?

docs/mapping-network-events.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+Some events may also indicate each host's role in the exchange: client or server. When this information is available, the <<ecs-client,client>> and <<ecs-server,server>> fields should be used _in addition to_ the `source` and `destination` fields. The fields and values mapped under `source`/`destination` should be copied under `client`/`server`.
+
+[float]
+==== Network event mapping example
+
+Below is a DNS network event. The source device (`192.168.86.222`) makes a DNS query, acting as the client and the DNS server is the destination (`192.168.86.1`).
+
+Note this event contains additional details that would populate additional fields (such as the <<ecs-dns>>) if this was a complete mapping example. These additional fields are omitted here to focus on the network details.
+
+[source,json]
```
dainperkins, line 22: Would it make sense to use an ECS version of the source data?

dainperkins:
@ebeahan & @webmat 
Ok I think I get it - the basis for this is the idea that a given network observer will NOT be indicating the session role of a given unidirectional session?  that is to say we are not treating source and destination pairs as e.g. unidirectional flow indicators with no concept of transport or application session (i,e, tcp: syn|syn-ack or udp: e.g. dns request/response)

for example (stripped of irrelevant data, and no transport or app session - aws vpc flows used to look like this, unsure if they still do)

TCP Example

> Event 1:
> source: 192.168.1.10:42113 -> destination: 192.168.1.50:443. 1 packet, 10 bytes (syn)
> 
> Event 2: 
> source:  192.168.1.50:443 -> destination: 192168.1.10:42113 1 packet 10 bytes (syn-ack)
> 
> Event 3:
> source: 192.168.1.10:42113 -> destination: 192.168.1.50:443. 1 packet, 10 bytes (ack)

or for UDP

> Event 1:
> source: 192.168.1.10:42113 -> destination: 192.168.1.50:53. 1 packet, 10 bytes (dns request)
> 
> Event 2: 
> source:  192.168.1.50:53 -> destination: 192168.1.10:42113 1 packet 10 bytes (dns response)

docs/mapping-network-events.asciidoc
webmat: I don't think we should provide example code here. This would be fine in a blog post, but I think this will be too much of a maintenance burden.

ebeahan: > maybe call out the specific ips to include in the related.ip field in the example?

Good call out. I can definitely add a little more to the explanation.

> Should we include e.g. pipeline code to show how to determine if a dns.answer is an IP?

I approached this as an implementation-neutral reference of sorts. I've kept this focused on what the final mappings should be and avoided being too prescriptive about the how.

docs/mapping-network-events.asciidoc
ebeahan: I chose an example event that wasn't already mapped to ECS trying to capture each steps in the exercise of mapping the event's network-related values.

Would breaking down an event that's already mapped to ECS (e.g. a packetbeat DNS event) for the example be more useful?

webmat:
@ebeahan I think we will need to revisit a lot of the other static asciidoc pages. For example there's overlap between some old pages and newer ones. We'll also need to consider the different personas that consume ECS, and help guide users to the different sections in a more useful manner, right from the top.

But for now, I think we should move this page under "Using ECS", perhaps as the last sub-page. It's a bit weird to move it to this section, but the "Migrating" and "Additional" sections are too far down IMO, and not as up to date. So adding this new page there runs a risk of not being noticed.

ebeahan:
> Ok I think I get it - the basis for this is the idea that a given network observer will NOT be indicating the session role of a given unidirectional session?

The aim is to provide a reference where we can refer users for how network events should be mapped. @webmat captured a list in the second checklist item from #948 which served as the foundation.

webmat (commented):
Great content, this will be helpful :-)

I have a few comments below on content suggestions, let me know what you think.

docs/mapping-network-events.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+    ]
+  }
+----
+
+The <<ecs-related,related fields>> are meant to facilitate pivoting. Since these IP addresses can appear in many different fields (`source.ip`, `destination.ip`, `client.ip`, `server.ip`, etc.), you can search for the IP trivially no matter what field it appears using a single query, e.g. `related.ip:192.168.86.222`.
+
+Network events are not only limited to using `related.ip`. If hostnames or other host identifiers were present in the event, `related.hosts` should be populated too.
+
+[float]
+==== Event fields
```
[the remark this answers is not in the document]
webmat: Perhaps we could also have the `network.*` section and `event.*` section grouped together somehow, as the former is also related to categorization, as you explain wrt `event.category: protocol`.

docs/mapping-network-events.asciidoc
dainperkins: Actually I think the original makes more sense now that I've thought about it.  I think the example data threw me for a loop as any request/response protocol analysis will have clear source/destination linking to client/server..

basically for anything besides unidirectional flows (meraki netflow, aws vpcflow) & broadcast requests (dhcp) copy source/destination to client/server.

for the corner cases just use source/dest.

I think I may be getting wrapped around an axle that is as much about base semantics, as it is about using the data e.g. the network siem page displays...

docs/mapping-network-events.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+  "server": {
+    "ip": "192.168.86.1",
+    "port": 53
+  }
+----
+
+Mapping both pairs of field sets gives query visibility of the same network transaction in two ways:
+
+* `source.ip:192.168.86.222` returns all events sourced from `192.168.86.222`, regardless its role in a session
+* `client.ip:192.168.86.222` returns all events with host `192.168.86.222` acting as a client.
```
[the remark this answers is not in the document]
dainperkins: I figured out where I'm getting bogged down... 

I strongly dislike mixing unidirectional and bidirectional data in one bucket.  I know it completely skews ML analysis (unidirectional features don't work in bidirectional, and vice versa), and I think it plays havoc with things like the network tab on the SIEM if filtering / bidirectional normalization isn't being done.

How to deal with that I'm not really sure...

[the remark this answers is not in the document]
ebeahan: @webmat yes it's good idea to call out that while `source`/`destination` will flip, `client`/`server` assignments will apply throughout the transaction.

@dainperkins I don't think we're introducing any new concepts or guidance on unidirectional vs. bidirectional data here, but trying to better state what the existing guidance already is in a write-up that can live in the ECS docs.

Are these issues something we can improve through better ECS guidance here? Or would that topic be better as a separate discussion?

docs/mapping-network-events.asciidoc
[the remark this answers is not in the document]
ebeahan: I merged the two sections by dovetailing the `network.*` section onto the event categorization portion.

I also realized there could be additional `event.*` fields mapped that this example doesn't cover, so I also made some adjustments to focus on event categorization over the `event.*` field set in general.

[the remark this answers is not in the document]
webmat: Great, thanks for the adjustments

webmat (approved):
LGTM thanks for the adjustments. This is looking great!

PR #971, opened 2020-09-21
Add normalizer multi-field capability

madirey:
This PR adds the capability for custom schemas to define a `normalizer` for a specific `multi_field`.

Related: https://github.com/elastic/endpoint-package/pull/79

### DIFF
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -35,7 +35,7 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
                     'ignore_above', 'multi_fields', 'format', 'input_format',
                     'output_format', 'output_precision', 'description',
                     'example', 'enabled', 'index']
-    multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field']
+    multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field', 'normalizer']
 
     fields = []
     for nested_field_name in source_fields:
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -35,7 +35,7 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
                     'ignore_above', 'multi_fields', 'format', 'input_format',
                     'output_format', 'output_precision', 'description',
                     'example', 'enabled', 'index']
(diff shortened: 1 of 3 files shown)


scripts/generators/beats.py
```diff
@@ -35,7 +35,7 @@ def fieldset_field_array(source_fields, df_whitelist, fieldset_prefix):
                     'ignore_above', 'multi_fields', 'format', 'input_format',
                     'output_format', 'output_precision', 'description',
                     'example', 'enabled', 'index']
-    multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field']
+    multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field', 'normalizer']
```
ebeahan, line 38:
```suggestion
    multi_fields_allowed_keys = ['name', 'type', 'norms', 'default_field', 'normalizer', `ignore_above`]
```

In the schema, `ignore_above: 1024` is set as a default for all the `keyword` fields. For consistency, feels like that setting should copied into `keyword` type multi-fields.

scripts/generators/es_template.py
```diff
@@ -64,6 +64,9 @@ def entry_for(field):
             field_entry['fields'] = {}
             for mf in field['multi_fields']:
                 mf_entry = {'type': mf['type']}
+                mf_normalizer = mf.get('normalizer')
+                if mf_normalizer:
+                    mf_entry['normalizer'] = mf_normalizer
```
ebeahan, line 69:
```suggestion
                if mf['type'] == 'keyword':
                    ecs_helpers.dict_copy_existing_keys(mf, mf_entry, ['normalizer', 'ignore_above'])
```

`ecs_helpers.dict_copy_existing_keys` allows copying in `ignore_above` along with the added `normalizer` setting and will also check if the key exists in the source dict before trying to copy.

ebeahan (commented):
Thanks for the PR to add this capability @madirey !

Can you also add an entry to `CHANGELOG.next.md`?

webmat (approved):
LGTM, thanks Madison :-)

I'll let @ebeahan merge when he's had another look at it.

I would normally ask for tests, but to be fair, pre-existing tests for this part of the code are pretty sparse. So it's fair for us to take that on separately.

ebeahan (approved):
LGTM 👍 - will merge.

madirey:
Thanks @ebeahan !

PR #972, opened 2020-09-21
Use escape character over double quoting in generated CSV

ebeahan:
#### Summary

Adjust the params in the `csv.writer()` object in the CSV generator to use escape `\` as an escape characters for double quotes within an already quoted field. Default behavior generates field values with double double-quotes.

Behavior was noted while working on #966.

#### Example

Current:

```
2.0.0-dev,true,base,labels,object,core,,"{""application"": ""foo-bar"", ""env"": ""production""}",Custom key/value pairs.
```

PR:

```
2.0.0-dev,true,base,labels,object,core,,"{\"application\": \"foo-bar\", \"env\": \"production\"}",Custom key/value pairs.
```

#### Side Effect

One small, side-effect change is example values themselves containing `\` as part of their string values will now be quoted:

Current:

```
2.0.0-dev,true,dll,dll.path,keyword,extended,,C:\Windows\System32\kernel32.dll,Full file path of the library.
```

PR:

```
2.0.0-dev,true,dll,dll.path,keyword,extended,,"C:\Windows\System32\kernel32.dll",Full file path of the library.
```

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -40,6 +40,7 @@ Thanks, you're awesome :-) -->
 #### Improvements
 
 * Field details Jinja2 template components have been consolidated into one template #897
+* Use backslashes as the escape characters over double double-quotes in CSV artifact. #972
 
 #### Deprecated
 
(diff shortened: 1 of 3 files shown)


ebeahan:
Adding backslash escaping into CSV breaks default importing into a spreadsheet app, so I won't move forward pursuing this change for now.

PR #937, opened 2020-08-17
Introduce `--strict` flag

ebeahan:
#### Summary

Adding a `--strict` option to the ECS generator script that performs more strict schema validation during the schema cleaning phase.

#### Usage

Initially, `--strict` will only enforce one validation: short descriptions be less than 120 characters and on a single line.

Example usage:

`$ python scripts/generator.py --strict`

Closes #892

PR #975, opened 2020-09-23
[1.x] Introduce `--strict` flag (#937)

ebeahan:
 - Introduce `--strict` flag (#937)


ebeahan:
Original missed backporting after #937 was merged.

PR #978, opened 2020-09-24
[1.x] Add normalizer multi-field capability (#971)

ebeahan:
 - Add normalizer multi-field capability (#971)


webmat (approved):
LGTM

PR #983, opened 2020-09-29
[1.x] Add mapping network event guidance doc (#969)

ebeahan:
 - Add mapping network event guidance doc (#969)


webmat (approved):
LGTM

PR #984, opened 2020-09-29
Removing unneeded link under `Additional Information`

ebeahan:
Missed removing this asciidoc link on the `Additional Information` page for `ecs-mapping-network-events` when moving section underneath `Using ECS` in #969


webmat (approved):
Ah, I closed the discussion thread on the previous PR, wondering why it didn't show it as outdated 🤦‍♂️ 

Thanks for catching that :-)

PR #988, opened 2020-09-30
Add usage docs section

ebeahan:
#### Summary

Adds a free-text section to the ECS field reference docs for each fieldset, called `usage`. The `usage` directory will allow for any current ECS fieldset to have a paired usage document which will appear as a subsection in the ECS docs table of contents.

#### Format/Appearance

The ECS documentation book will add an additional level of chunking:

Screen Shot 2020-09-30 at 1 21 20 PM

This additional, nested documentation page will only be added when a usage document is created for that particular field set.

#### Implementation Notes

* A new directory named `docs/usage` is being added and includes a `README.md` describing the intent and how-to for the usage asciidoc files.
* The `asciidoc_fields` generator will check the contents of `docs/usage` and create the necessary `include` statements in `docs/field-details.asciidoc`.
* No usage documentation is included; this PR focuses on putting the framework in place. Changes to `docs/field-details.asciidoc` were done to align with the changes in the `field_details.j2` template.
* The field details Jinja2 template has added the `[discrete]` section attribute to allow the necessary `include` statement to work as expected.
* The `Makefile` has been updated to use the higher chunking level: `--chunk=2`

Relates to #943

Issue #943, opened 2020-08-19
Introduce free-form text sections in docs


ebeahan:
**Summary**

Introduce a concept into the ECS repo which captures extended usage examples. These examples would appear in the ECS documentation as a dedicated page per field set.

**Motivation**:

Real-world examples are powerful tools to help users better grasp ECS concepts and how users can better map their custom data sources. In addition to the field specifications, ECS also provides guidance on the preferred design patterns for modeling particular events.

A wealth of examples and past discussion is captured in the repo and on Discuss, but not in the ECS docs. Having such examples incorporated into the docs would provide a single, authoritative location. Restructuring could also allow us to include diagrams, code snippets, and other visual aids.

Issues touching on or providing such examples:

* #516
* #860 
* #809 
* #300 

**Design Proposal**:

* Refactor the existing `schemas` directory
    * As of ECS 1.5.0, the `schemas` directory contains a collection of YAML files and a README.
    * create subdirectories named after the fieldset
    * Each subdirectory contains the schema file for that fieldset. It can also contain an AsciiDoc formatted `usage` or `example` file.

* Add a new page dedicated to the free-form usage and examples docs. The page will be nested underneath the field set reference page.

* The AsciiDoc generator will be updated to support this refactoring in the docs

**To-do**:

- [x] #989 Add `[discrete]` attribution for field details section headers.
    - [x] `[discrete]` will need to be added to the previous ECS version branches to ensure the legacy formatting is retained after changing the chunking level.
- [x] https://github.com/elastic/docs/pull/1970 docs PR to increase chunking level for ECS book from `1` to `2`
- [x] #988 add free-form usage asciidoc per fieldset

ebeahan:
I've pushed a branch that can be pulled down and viewed with `make docs`: https://github.com/ebeahan/ecs/tree/free-text

I had to modify the `make docs` command arguments setting `--chunk=2`. We can investigate the correct way to handle within the docs build, but this adjustment unblocked me enough to get a locally working build.

Here, I focused on the hierarchy structure and each usage subpage is a placeholder. The three mockups are viewed under the `base`, `agent`, and `autonomous system` field reference sections.

1. Fieldset details page remains unchanged. One sub-page with usage details:

* Agent Fields
    * Agent Fields Usage


![Screen Shot 2020-09-18 at 12 29 26 PM](https://user-images.githubusercontent.com/7226265/93627497-a7037680-f9aa-11ea-8ab0-6a5caad4b48a.png)



[...]

PR #989, opened 2020-09-30
Add discrete attribute to field details page headers

ebeahan:
Add the `[discrete]` attribute to section headers in the `scripts/templates/field_details.j2` template file.

Adding `[discrete]` to the appropriate section headers in `docs/field-details.asciidoc` will keep the expected sections on the same pages once the asciidoctor chunking level is increased to support #988.

Relates to #943

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -46,6 +46,7 @@ Thanks, you're awesome :-) -->
 #### Improvements
 
 * Field details Jinja2 template components have been consolidated into one template #897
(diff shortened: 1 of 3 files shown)


webmat (approved):
LGTM 🚢

ebeahan:
Yes, we will need the `[discrete]` tags to get the formatting consistent. I may need to tackle this one branch at a time, since the field details doc would be unique per version based on the current schema at that point in time. 🤯

PR #993, opened 2020-10-02
Introduce experimental artifacts

ebeahan:
Add the `experimental` top-level directory which holds field definitions and artifacts which have advanced to [stage two](https://elastic.github.io/ecs/stages.html) in the RFC process.

The directory structure of `experimental` mirrors the top-level of the repo for consistency. A new directive has been added, `make experimental`, to generate the artifacts in `experimental/generated`.

```
$ make experimental 
build/ve/bin/python scripts/generator.py --include experimental/schemas --out experimental
Loading user defined schemas: ['experimental/schemas']
```


experimental/README.md
```diff
@@ -0,0 +1,19 @@
+# ECS Experimental Definitions
+
+ECS experimental definitions are changes and features which have been merged as [stage two](https://elastic.github.io/ecs/stages.html) in the ECS [RFC process](../rfcs), but they have not yet advanced stage three. These initial field definitions are comprehensive but not necessarily complete. They are also still subject to breaking changes or being removed in any future version.
+
+This directory provides a location for these experimental field definitions and generated artifacts.
```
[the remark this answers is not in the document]
webmat: Here's a take on this intro, WDYT?

> ECS experimental definitions are changes and features which have reached [stage two](https://elastic.github.io/ecs/stages.html) in the ECS [RFC process](../rfcs).
> 
> Stage two changes only appear in the experimental artifacts in this directory, but aren't yet reflected in the official ECS documentation.
> Note that stage three and four proposals do appear in the official ECS documentation.
>
> These experimental changes to ECS are comprehensive but not necessarily final. They are also still subject to breaking changes.

experimental/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Schema Files
+
+The [schemas](./schemas) directory contains the YAML files for the experimental field definitions. These are not complete schemas themselves, but are settings meant to merge with the official schema spec using the `--include` generator flag.
+
+Example usage:
+
+```sh
+$ python scripts/generator.py --include experimental/schemas --out experimental
+```
```
webmat, line 15: The current example is actually what we need to run to maintain this directory.

I would instead document this for folks that want to use the experimental artifacts for their own custom index. So we should refer them to USAGE.md and also give an example of grabbing experimental changes + typical customization. Here's my take on this:

> If you use the ECS generator script as described in [USAGE.md](https://github.com/elastic/ecs/blob/master/USAGE.md) to maintain your custom index templates, here's how you can try these experimental changes in your project:

```sh
$ python scripts/generator.py --include experimental/schemas \
        --include ../myproject/fields/custom/ \
        --out ../myproject/fields/generated
```

> The above would include all experimental changes to ECS along with your custom fields, and output the artifacts in `myproject/fields/generated`.

webmat (commented):
I have a few comments on the text documenting this, noted below.

- [x] Please also add a .gitignore to not publish the Elasticsearch 6 template in the experimental directory. With the introduction of `wildcard`, ES 6 no longer makes sense.
- [ ] Also, we'll wait until RFC 0001 is actually merged to adjust the experimental output.

Other than these details, this looks good to me.

webmat:
One more thing just occurred to me.

- [ ] Let's add the `experimental` make task to the dependencies `make check` that's run by CI.

I don't want this full second run to slow down the default make task. But the above will ensure that the computers catch us forgetting to update the artifacts, when we change the experimental field definitions :-)

experimental/README.md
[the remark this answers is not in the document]
ebeahan: ++

Using the phrasing "reached" stage two over "merged as" is a good change.

.gitignore
```diff
@@ -7,3 +7,7 @@ build
 .idea
 *.iml
 .vscode/*
+
+# experimental exclusions
+experimental/generated/elasticsearch/6
+experimental/generated/docs
```
webmat, line 13: good catch

experimental/README.md
ebeahan: Good call out. Also good steer users towards the usage documentation 😄

ebeahan:
I believe I addressed all the feedback as well as rebased to make CI happy. This should be ready for one final review. 🕵️‍♀️

webmat (approved):
I diff'ed what's in experimental/schemas and the RFCs and confirm they all LGTM. Thanks for importing all of the final adjustments.

Two minor nits on the directory readme and we're good to go

webmat (approved):
🚢

PR #994, opened 2020-10-02
[1.x] Uniformity across domain name breakdown fields (#981)

ebeahan:
 - Uniformity across domain name breakdown fields (#981)


webmat (approved):
🚢

PR #999, opened 2020-10-05
[RFC] E-Mail

jamiehynds:

[RFC Preview](https://github.com/jamiehynds/ecs/blob/jamiehynds-patch-2/rfcs/text/0008-email.md)

### DIFF
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -1,10 +1,10 @@
 # 0008: Email
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **1** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
 - Date: **Oct 5th 2020** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
-This RFC proposes a new top-level field to facilitate email use cases. 
+This RFC proposes a new top-level field to facilitate email use cases.
 
 <!--
 As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
@@ -21,32 +21,35 @@ Stage 0: Provide a high level summary of the premise of these changes. Briefly d
 Stage 1: Describe at a high level how this change affects fields. Which fieldsets will be impacted? How many fields overall? Are we primarily adding fields, removing fields, or changing existing fields? The goal here is to understand the fundamental technical implications and likely extent of these changes. ~2-5 sentences.
 -->
 
+Email specific fields:
+
 | field | type | description |
 | --- | --- | --- |
-| `email.action` | keyword | Action take by the source device, e.g. delivered, blocked, quarantined, deleted |
-| `email.bcc.address` | keyword | Addresses of Bcc's |
-| `email.bcc.domain` | keyword | Domains of the Bcc's |
-| `email.cc.address` | keyword | Addresses of Cc's |
-| `email.cc.domain` | keyword | Domains of Cc addresses |
-| `email.cipher` | keyword | Cipher used e.g. TLS |
-| `email.file.count` | value | Number of attachments included in the message |
-| `email.file.extension` | keyword | Extensions of attachment, e.g. .zip, .docx |
-| `email.file.hash` | keyword | Hash of attachments |
-| `email.file.name` | keyword | File name of attachements |
-| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
+| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
+| `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
 | `email.direction` | keyword | Direction of the message based on the sending and receving domains |
-| `email.from.address` | keyword | Senders email address |
-| `email.from.domain` | keyword | Senders domain |
-| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.sender.address` | wildcard | Senders email address |
+| `email.sender.top_level_domain` | keyword | Senders email address |
 | `email.message_id` | keyword | Internet message ID of the message |
-| `email.process` | keyword | Name of the executable that carried out the transaction, e.g. outlook, sendmail |
-| `email.protocol` | keyword | The email protocol used, e.g. SMTP, IMAP |
-| `email.reply_to.address` | keyword | Reply-to address |
-| `object.return.address` | keyword | The return address for the message |
+| `email.reply_to.address` | wildcard | Reply-to address |
+| `email.return.address` | wildcard | The return address for the message |
 | `email.size` | keyword | Total size of the message, in bytes, including attachments |
-| `email.subject` | keyword | Subject of the message |
-| `email.to` | keyword | Recipieint address |
-| `email.to.domain` | keyword | Recipient domain |
+| `email.subject` | wildcard | Subject of the message |
+| `email.recipients.addresses` | keyword | Recipient addresses |
+| `email.domains` | keyword | domains related to the email |
+
+
+Other ECS fields used together with email usecases:
+| field | description |
+| --- | --- |
+| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email tok to send from source to destination etc |
+| `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
+| `network.protocol` | Type of email protocol used |
+| `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
+
+
 
 ## Usage
 
@@ -112,8 +115,8 @@ People
 
 The following are the people that consulted on the contents of this RFC.
 
-Jamie Hynds | author
-TBD | Sponsor
+Marius Iversen | Author
+Jamie Hynds | Sponsor
 
 <!--
 Who will be or has been consulted on the contents of this RFC? Identify authorship and sponsorship, and optionally identify the nature of involvement of others. Link to GitHub aliases where possible. This list will likely change or grow stage after stage.
@@ -136,7 +139,7 @@ e.g.:
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
-* Stage 0: https://github.com/elastic/ecs/pull/NNN
+* Stage 0: https://github.com/elastic/ecs/pull/999
 
 <!--
 * Stage 1: https://github.com/elastic/ecs/pull/NNN
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -45,6 +45,8 @@ Other ECS fields used together with email usecases:
 | field | description |
 | --- | --- |
 | `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
+| `event.start` | When the email event started
+| `event.end` | When the email event ended
 | `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
 | `network.protocol` | Type of email protocol used |
 | `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -2,7 +2,7 @@
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
 - Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **Oct 5th 2020** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Date: **2020-11-30** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 This RFC proposes a new top-level field to facilitate email use cases.
 
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -31,11 +31,14 @@ Email specific fields:
 | `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
 | `email.direction` | keyword | Direction of the message based on the sending and receving domains |
 | `email.sender.address` | wildcard | Senders email address |
-| `email.sender.top_level_domain` | keyword | Senders email address |
+| `email.sender.domain` | wildcard | Domain of the sender |
+| `email.sender.top_level_domain` | keyword | Top level domain of the sender |
+| `email.sender.registered_domain` | wildcard | Registered domain of the sender |
+| `email.sender.subdomain` | keyword | Subdomain of the sender |
 | `email.message_id` | keyword | Internet message ID of the message |
 | `email.reply_to.address` | wildcard | Reply-to address |
 | `email.return_path.address` | wildcard | The return address for the message |
-| `email.size` | keyword | Total size of the message, in bytes, including attachments |
+| `email.size` | long | Total size of the message, in bytes, including attachments |
 | `email.subject` | wildcard | Subject of the message |
 | `email.recipients.addresses` | keyword | Recipient addresses |
 | `email.domains` | keyword | domains related to the email |
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -25,8 +25,8 @@ Email specific fields:
 
 | field | type | description |
 | --- | --- | --- |
-| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
-| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.bcc.addresses` | `wildcard[]` | Addresses of Bcc's |
+| `email.cc.addresses` | `wildcard[]` | Addresses of Cc's |
 | `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
 | `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
 | `email.direction` | keyword | Direction of the message based on the sending and receving domains |
@@ -40,8 +40,8 @@ Email specific fields:
 | `email.return_path.address` | wildcard | The return address for the message |
 | `email.size` | long | Total size of the message, in bytes, including attachments |
 | `email.subject` | wildcard | Subject of the message |
-| `email.recipients.addresses` | keyword | Recipient addresses |
-| `email.domains` | keyword | domains related to the email |
+| `email.recipients.addresses` | `keyword[]` | Recipient addresses |
+| `email.domains` | `keyword[]` | domains related to the email |
 
 
 Other ECS fields used together with email usecases:
@@ -122,12 +122,12 @@ Stage 4: Document any new concerns and their resolution. The goal here is to eli
 Stage 4: Identify at least one real-world, production-ready implementation that uses these updated field definitions. An example of this might be a GA feature in an Elastic application in Kibana.
 -->
 
-People
+## People
 
 The following are the people that consulted on the contents of this RFC.
 
-@p1llus | Author
-@jamiehynds | Sponsor
+* @p1llus | Author
+* @jamiehynds | Sponsor
 
 <!--
 Who will be or has been consulted on the contents of this RFC? Identify authorship and sponsorship, and optionally identify the nature of involvement of others. Link to GitHub aliases where possible. This list will likely change or grow stage after stage.
@@ -146,11 +146,19 @@ e.g.:
 
 <!-- Insert any links appropriate to this RFC in this section. -->
 
+- [Hubspot](https://legacydocs.hubspot.com/docs/methods/email/email_events_overview)
+- [O365 Message Tracing](https://docs.microsoft.com/en-us/exchange/monitoring/trace-an-email-message/run-a-message-trace-and-view-results)
+- [Postfix](https://nxlog.co/documentation/nxlog-user-guide/postfix.html)
+- [Barracuda](https://campus.barracuda.com/product/emailsecuritygateway/doc/12193950/syslog-and-the-barracuda-email-security-gateway/)
+- [Forcepoint](https://www.websense.com/content/support/library/email/v85/email_siem/siem_log_map.pdf)
+- [Mimecast](https://www.mimecast.com/tech-connect/documentation/tutorials/understanding-siem-logs/)
+- [Proofpoint](https://help.proofpoint.com/Threat_Insight_Dashboard/API_Documentation/SIEM_API)
+
 ### RFC Pull Requests
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
-* Stage 0: https://github.com/elastic/ecs/pull/999
+* Stage 1: https://github.com/elastic/ecs/pull/999
 
 <!--
 * Stage 1: https://github.com/elastic/ecs/pull/NNN
--- a/rfcs/text/0008-email.md
+++ b/rfcs/text/0008-email.md
@@ -34,7 +34,7 @@ Email specific fields:
 | `email.sender.top_level_domain` | keyword | Senders email address |
 | `email.message_id` | keyword | Internet message ID of the message |
 | `email.reply_to.address` | wildcard | Reply-to address |
-| `email.return.address` | wildcard | The return address for the message |
+| `email.return_path.address` | wildcard | The return address for the message |
 | `email.size` | keyword | Total size of the message, in bytes, including attachments |
 | `email.subject` | wildcard | Subject of the message |
 | `email.recipients.addresses` | keyword | Recipient addresses |
@@ -44,7 +44,7 @@ Email specific fields:
 Other ECS fields used together with email usecases:
 | field | description |
 | --- | --- |
-| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email tok to send from source to destination etc |
+| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
 | `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
 | `network.protocol` | Type of email protocol used |
 | `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
@@ -92,6 +92,12 @@ The goal here is to research and understand the impact of these changes on users
 <!--
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
+Current concerns or topics still being discussed from stage 1:
+
+- Whether we want to add specific fields for email protocols, either as a root field or nested under email.* (SMTP, IMAP, POP etc).
+- Need to make sure that the ECS fieldset for email catches all common usecases, for example spam, metrics and deliverables and logging.
+- Whether we want to create a new event.category field (email) and which event.type it should be combined with.
+- The email RFC will be the first ECS fieldset that uses the flattened datatype (for attachments), need to ensure that there will be major issues related to this.
 
 <!--
 Stage 2: Document new concerns or resolutions to previously listed concerns. It's not critical that all concerns have resolutions at this point, but it would be helpful if resolutions were taking shape for the most significant concerns.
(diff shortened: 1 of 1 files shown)


rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.from.address` | keyword | Senders email address |
+| `email.from.domain` | keyword | Senders domain |
+| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.process` | keyword | Name of the executable that carried out the transaction, e.g. outlook, sendmail |
+| `email.protocol` | keyword | The email protocol used, e.g. SMTP, IMAP |
+| `email.reply_to.address` | keyword | Reply-to address |
+| `object.return.address` | keyword | The return address for the message |
```
vpiserchia, line 45: should not be email.return.address?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+
+| field | type | description |
+| --- | --- | --- |
+| `email.action` | keyword | Action take by the source device, e.g. delivered, blocked, quarantined, deleted |
+| `email.bcc.address` | keyword | Addresses of Bcc's |
+| `email.bcc.domain` | keyword | Domains of the Bcc's |
+| `email.cc.address` | keyword | Addresses of Cc's |
+| `email.cc.domain` | keyword | Domains of Cc addresses |
+| `email.cipher` | keyword | Cipher used e.g. TLS |
+| `email.file.count` | value | Number of attachments included in the message |
```
vpiserchia, line 32: why not allowing multiple files as in an array ?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.process` | keyword | Name of the executable that carried out the transaction, e.g. outlook, sendmail |
+| `email.protocol` | keyword | The email protocol used, e.g. SMTP, IMAP |
+| `email.reply_to.address` | keyword | Reply-to address |
+| `object.return.address` | keyword | The return address for the message |
+| `email.size` | keyword | Total size of the message, in bytes, including attachments |
+| `email.subject` | keyword | Subject of the message |
+| `email.to` | keyword | Recipieint address |
+| `email.to.domain` | keyword | Recipient domain |
```
vpiserchia, line 49: email.to is already a field of type keyword, how can you define email.to as an object having domain as subfield? sorry for the silly question

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.file.extension` | keyword | Extensions of attachment, e.g. .zip, .docx |
+| `email.file.hash` | keyword | Hash of attachments |
+| `email.file.name` | keyword | File name of attachements |
+| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.from.address` | keyword | Senders email address |
+| `email.from.domain` | keyword | Senders domain |
+| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.process` | keyword | Name of the executable that carried out the transaction, e.g. outlook, sendmail |
```
vpiserchia, line 42: MTA?

vpiserchia (commented):
A series of question

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.protocol` | keyword | The email protocol used, e.g. SMTP, IMAP |
+| `email.reply_to.address` | keyword | Reply-to address |
+| `object.return.address` | keyword | The return address for the message |
+| `email.size` | keyword | Total size of the message, in bytes, including attachments |
+| `email.subject` | keyword | Subject of the message |
+| `email.to` | keyword | Recipieint address |
```
webmat, line 48: `email.to.address`

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/NNN
```
webmat, line 139:
```suggestion
* Stage 0: https://github.com/elastic/ecs/pull/999
```

rfcs/text/0008-email.md
jeffrysleddens, line 48: I think 'recipient' is a more common term when it comes to email than 'to'.
Also the schema should support multiple recipients.

rfcs/text/0008-email.md
BenB196: You have .address, but don't have .domain for email.reply_to and object.return, should be constant and have .domain as an additional field.

BenB196, line 45: .address exits for email.reply_to and object.return, but not .domain, should be constant and have .domain as an additional field.

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.cc.address` | keyword | Addresses of Cc's |
+| `email.cc.domain` | keyword | Domains of Cc addresses |
+| `email.cipher` | keyword | Cipher used e.g. TLS |
+| `email.file.count` | value | Number of attachments included in the message |
+| `email.file.extension` | keyword | Extensions of attachment, e.g. .zip, .docx |
+| `email.file.hash` | keyword | Hash of attachments |
+| `email.file.name` | keyword | File name of attachements |
+| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.from.address` | keyword | Senders email address |
```
jeffrysleddens, line 38: The schema should support and distinguish both the envelope/smtp 'from' and the header/mime 'from'.

rfcs/text/0008-email.md
vpiserchia: this really depends if you want to capture the semantics of the envelope/smtp and/or the email headers. This applies to other fields as well (to, cc, bcc, from)

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+
+<!--
+Stage 2: Identifies scope of impact of changes. Are breaking changes required? Should deprecation strategies be adopted? Will significant refactoring be involved? Break the impact down into:
+ * Ingestion mechanisms (e.g. beats/logstash)
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
```
BenB196, line 87: When looking at the current fields provided, one of my concerns is it appears that they don't fit well with the rest of ECS. I think this can be partially fixed with the use of aliases, though, I don't believe aliases are standard/common in ECS.

Examples:

email.from -> source
email.to|cc|bcc -> destination
email.file -> file
email.latency -> event.duration

rfcs/text/0008-email.md
vpiserchia: file already exists as ECS field, it would be great to reuse it

rfcs/text/0008-email.md
webmat: I agree there's nuances in recipients, and in how closely we want to represent each protocol.

After looking at a few raw emails, I'm starting to think we should keep the design of this field set pretty high level.

If capturing each protocol's nuances is desired or necessary, then I think we should consider working on a specific breakdown per protocol, as a separate step.

So I think the guiding principle we could use here is to capture the commonalities in `email.*`, even if it means merging some concepts, sometimes.

One point you're raising is pretty interesting, however. Should we capture each "type" of recipient in a different field, or in one parent field with an additional label to indicate which type of recipient this is?

The current proposal takes the former approach:

```
{ "email": {
  "to": [ {"address": "alice@example.com", ...} ],
  "cc": [ {"address": "bob@example.com", ...} ],
} }
```

The "recipient" suggestion would look like:

```
{ "email": {
  "recipients": [
    { "type": "to", "address": "alice@example.com", ...},
    { "type": "cc", "address": "bob@example.com", ...},
  ]
} }
```

🤔

webmat (commented):
Excellent start, @jamiehynds.

Thanks everyone so far for chiming in. Looks like there's a lot of interest in getting support for email in ECS 😃 

I think there's a lot of straightforward suggestions already made (in my review and prior reviews) that we could apply to the current RFC.

There's also many big discussions that don't have an obvious answer. Those should be captured in the "Concerns" section, so that these problem statements are part of the RFC document itself, and that we carry to the next RFC stages. I've tried to point them out in my review.

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.file.count` | value | Number of attachments included in the message |
+| `email.file.extension` | keyword | Extensions of attachment, e.g. .zip, .docx |
+| `email.file.hash` | keyword | Hash of attachments |
+| `email.file.name` | keyword | File name of attachements |
+| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.from.address` | keyword | Senders email address |
+| `email.from.domain` | keyword | Senders domain |
+| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.message_id` | keyword | Internet message ID of the message |
```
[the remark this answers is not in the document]
vpiserchia: nevertheless the message_id captures the uniqueness of a mail.
I can see that different mail servers have specific ways of building this Message ID and could be interesting (for identification purposes) capturing such behaviour (and spot the anomalies). With this said, a multi-field mapping would make sense here:

```
 | `email.message_id` | keyword | Internet message ID of the message |
| `email.message_id.text` | text | Internet message ID of the message for full text search |
```

rfcs/text/0008-email.md
vpiserchia: I really like this one:

`{ "email": {
  "to": [ {"address": "alice@example.com", ...} ],
  "cc": [ {"address": "bob@example.com", ...} ],
} }`

this avoid the need for nested objects. Also open to a new one in the "related" field:

`related.emails: ["alice@example.com", "bob@example.com"]`

rfcs/text/0008-email.md
SHolzhauer: Why add a field for this, i.m.o this can be captured in `process.*` fields. Specifically the `process.name` field.

This would be in line with [the statement](https://github.com/elastic/ecs/pull/999/files#r512252692) on `email.action`

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+<!--
+As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
+Feel free to remove these comments as you go along.
+-->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+## Fields
```
ebeahan, line 18: Another item that came to mind, and I think makes sense to capture for further discussion later:

Would `email` be worth considering as an additional allowed value for `event.category`?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
+# 0008: Email
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **1** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **Oct 5th 2020** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+
+This RFC proposes a new top-level field to facilitate email use cases.
```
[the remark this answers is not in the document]
P1llus: @webmat Would there be any reason to discuss having the "protocols" under email rather than a top field level?

We could for example threat smtp, imap and pop3 as we do with for example "user.*", where it is expected to be nested under email.* but could meet other usecases moving forward.

rfcs/text/0008-email.md
P1llus: I would say +1 on this.

@ebeahan if we would create this new `event.category` we should also update which `event.type` we can combine with it and maybe see if we need new ones?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Fields
+
+<!--
+Stage 1: Describe at a high level how this change affects fields. Which fieldsets will be impacted? How many fields overall? Are we primarily adding fields, removing fields, or changing existing fields? The goal here is to understand the fundamental technical implications and likely extent of these changes. ~2-5 sentences.
+-->
+
+| field | type | description |
+| --- | --- | --- |
+| `email.action` | keyword | Action take by the source device, e.g. delivered, blocked, quarantined, deleted |
```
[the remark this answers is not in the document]
P1llus: I agree with this as well, we should reuse the ECS fields we already have when we can.

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.bcc.address` | keyword | Addresses of Bcc's |
+| `email.bcc.domain` | keyword | Domains of the Bcc's |
+| `email.cc.address` | keyword | Addresses of Cc's |
+| `email.cc.domain` | keyword | Domains of Cc addresses |
```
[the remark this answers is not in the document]
P1llus: @webmat I think it would make more sense to add `related.email` and maybe also `related.domain` instead. It would conform more to the current structure of how we handle arrays like this without nesting too much.
What do you think?

rfcs/text/0008-email.md
P1llus: How would we distinguish which hash is for which file, and the same with size?

Looking at the comment above about changing file to attachments, we will still need to for example have a list of objects if we want to keep track about size, hash, extension belonging to a single file/attachment.
This kinda goes against the current features we normally support (lists have very limited possibilities in terms of visualizations and SIEM/Alert rules)

Are we thinking all of these fields should just be an array?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,144 @@
...
+| `email.file.hash` | keyword | Hash of attachments |
+| `email.file.name` | keyword | File name of attachements |
+| `email.file.size` | keyword | Total size of all attachements in bytes |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.from.address` | keyword | Senders email address |
+| `email.from.domain` | keyword | Senders domain |
+| `email.latency` | keyword | The time, in milliseconds, the delivery attempt took |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.process` | keyword | Name of the executable that carried out the transaction, e.g. outlook, sendmail |
+| `email.protocol` | keyword | The email protocol used, e.g. SMTP, IMAP |
```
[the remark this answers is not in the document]
P1llus: +1 for this.

rfcs/text/0008-email.md
P1llus: I agree with @SHolzhauer on this, I think the name of the process should follow the current ECS format.

When it comes to #2 from @webmat it is something that might be useful in a separate field.

rfcs/text/0008-email.md
P1llus: From an ingest perspective and a visualization perspective I think its better to keep the "to, cc, bcc" field structure over type.
The point from @vpiserchia is a good one, because once we start having list of objects we will have issues with the internal structure when it comes to visualizations, parsing and aggregations.

rfcs/text/0008-email.md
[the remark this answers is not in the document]
ebeahan: Under the current guidance with the `related.*` fields, email addresses are captured in `related.user` and domains in `related.hosts`.

Rather we continue with this guidance or adjust and add additional `related.*` fields is a good conversation to have. 🤔

P1llus:
@ebeahan @webmat 

Just added the first iteration of the changes we discussed, hopefully this commit should cover all the above comments for now, some which we could investigate for later.
I added one more to the list, which was `attachments_count`, outside of the `attachments` flattened fields. This will help for visualisations and parser writing.

Could you please let me know if my commit covered all of the topics we wanted to cover for now?

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,147 @@
...
+| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
+| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
+| `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.sender.address` | wildcard | Senders email address |
+| `email.sender.top_level_domain` | keyword | Senders email address |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.reply_to.address` | wildcard | Reply-to address |
+| `email.return.address` | wildcard | The return address for the message |
```
andrewstucki, line 37: What about renaming this to `return_path` -- it's a bit more descriptive of what I think you're actually going for here.

rfcs/text/0008-email.md
```diff
@@ -21,32 +21,35 @@ Stage 0: Provide a high level summary of the premise of these changes. Briefly d
...
-| `email.to.domain` | keyword | Recipient domain |
+| `email.subject` | wildcard | Subject of the message |
+| `email.recipients.addresses` | keyword | Recipient addresses |
+| `email.domains` | keyword | domains related to the email |
+
+
+Other ECS fields used together with email usecases:
+| field | description |
+| --- | --- |
+| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email tok to send from source to destination etc |
```
ebeahan, line 47:
```suggestion
| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
```

rfcs/text/0008-email.md
```diff
@@ -112,8 +115,8 @@ People
 
 The following are the people that consulted on the contents of this RFC.
 
-Jamie Hynds | author
-TBD | Sponsor
+Marius Iversen | Author
```
ebeahan, line 118: We've been using GitHub handles in the `People` section as a convention for RFCs. Here's an [example](https://github.com/elastic/ecs/blob/master/rfcs/text/0001-wildcard-data-type.md#people).

Can we do the same here as well?

rfcs/text/0008-email.md
```diff
@@ -1,10 +1,10 @@
 # 0008: Email
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **1** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 4: Thanks for updating 👍

ebeahan (commented):
Thanks @P1llus!

I think we should still list some of the discussed points in the `Concerns` section. We don't have to have resolutions at this point. We simply want to capture them now for further discussion later:

* The two items Mat captured in an earlier review thread: https://github.com/elastic/ecs/pull/999#discussion_r512244335
* Possibility of needing an email `event.category` and possibly new `event.type` values.
* Maybe worth noting that if the flattened data type is used for `email.*` fields, it would be the first time using this type in ECS?

rfcs/text/0008-email.md
P1llus: I agree on this point as well, will update it in the next commit.

P1llus:
Adding in the latest changes @ebeahan :+1:

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,155 @@
...
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.sender.address` | wildcard | Senders email address |
+| `email.sender.top_level_domain` | keyword | Senders email address |
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.reply_to.address` | wildcard | Reply-to address |
+| `email.return_path.address` | wildcard | The return address for the message |
+| `email.size` | keyword | Total size of the message, in bytes, including attachments |
+| `email.subject` | wildcard | Subject of the message |
+| `email.recipients.addresses` | keyword | Recipient addresses |
+| `email.domains` | keyword | domains related to the email |
```
BenB196, line 41: This field really feels like it should be part of the related fields. Something like related.domains (though it currently doesn't exist, so it might be worth keeping here)

P1llus: Yeah this field is the outcome of a current discussion we have. Instead of having domain fields for bcc, cc, recipients etc, we decided currently to have them all as an array under one field. This might change in the upcoming stages. Thanks for the pointer, always happy to get feedback

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,155 @@
...
+
+| field | type | description |
+| --- | --- | --- |
+| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
+| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
+| `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.sender.address` | wildcard | Senders email address |
+| `email.sender.top_level_domain` | keyword | Senders email address |
```
BenB196, line 34: Some additional fields that might be useful for the emails.sender fields:
- email.sender.domain | keyword | Domain of the sender
- email.sender.registered_domain | keyword | Registered domain of the sender (this is useful when dealing with reporting that involves multiple sub-domains sending emails)

These fields are useful for pivoting across large amounts of emails, as well as in aggregations for reporting. (email.sender.domain currently falls under email.domains, but capturing direction in reports is not possible via a related field)

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,155 @@
...
+| `email.message_id` | keyword | Internet message ID of the message |
+| `email.reply_to.address` | wildcard | Reply-to address |
+| `email.return_path.address` | wildcard | The return address for the message |
+| `email.size` | keyword | Total size of the message, in bytes, including attachments |
```
BenB196, line 38: I can see email.size being more useful as a long rather than a keyword. I don't see many use-cases where you need to search for a specific email size (where use of keyword would outweigh use of long), compared to use-cases for aggregating and reporting based on a numeric type.

rfcs/text/0008-email.md
BenB196: Yeah, my concern mainly with a related field is that you lose the directionality of the value. Which might be useful for some use-cases.

rfcs/text/0008-email.md
ebeahan: Thanks @BenB196 - Good point to note, and thanks for the feedback about the fields being useful for pivoting and aggs.

There was some work done recently (https://github.com/elastic/ecs/pull/981) to ensure the domain name breakdown fields (`*.domain` + `*.registered_domain`, `top_level_domain`, and `*.subdomain`) were consistent across field sets. I expect we'll want to apply that same uniformity here.

rfcs/text/0008-email.md
P1llus: We do have the `email.direction` though, would that be sufficient? We would calculate the direction before moving the different domains into `email.domains` for example.

rfcs/text/0008-email.md
P1llus: That is a typo from my side, thanks for spotting it!

rfcs/text/0008-email.md
P1llus: I agree with both. This is something we had been discussing as well for email, though the discussion was currently around top_level_domain only.
Since sender is a single address, I do see the reasons why we should have maybe a separate domain field (outside of `email.domains`).

I will add in some additions based on the comment from @ebeahan as well, and we could use that as a basis for any upcoming discussions as it does not need to be set in stone just yet.

rfcs/text/0008-email.md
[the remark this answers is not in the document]
andrewstucki: Quick question about these comments. Is email authentication (i.e. SPF/DKIM/DMARC) in scope for this RFC? Mainly wondering because they're really a product of alignment on Return-Path (MAIL FROM), From, and DKIM-Signature mail headers. Not that I recommend all implementations that can capture these headers attempt to validate SPF/DKIM/DMARC, but it's totally possible given only the message content (and some fun DNS lookups).

I think that they fall under a pretty different realm than say spam/reputation scoring, analytics reports, MTA actions, or even email protocols that were mentioned.

rfcs/text/0008-email.md
dainperkins: latency might be redundant depending on the specific action being recorded, but I wouldn't equate email.to|from with source and destination (or client/server) network entities

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,158 @@
...
+
+Email specific fields:
+
+| field | type | description |
+| --- | --- | --- |
+| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
+| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
+| `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
```
dainperkins, line 32: What do we for see as values for this?, and from which address fields (to, cc, bcc) would it be categorized on?  
It seems to me like something that could potentially be difficult to implement, and not sure of the value for visualizations (but I could easily be missing something obvious, its been one of those days...)

rfcs/text/0008-email.md
webmat: Let's mention this in the RFC "Fields" section.

webmat (commented):
This RFC has by far surpassed the criteria for stage 1. We have a new author (thanks for picking it up, Marius) and a sponsor. And the fields of the proposal are looking really good already.

Of course on such an interesting subject, I couldn't resist providing some detailed technical feedback below.

But I think we should merge this as a stage 1 after a few minor adjustments which I'll outline in the body of this review, and we should address the rest of the technical feedback in a stage 2 PR.

The only points I'd like to address before merging this at stage 1 are:

- [ ] adjusting the "People" heading
- [ ] making the list of people into a bullet list
- [ ] adjusting the "stage 1" link at the very bottom
- [ ] and populating the "references" section

Everything else in my review can be addressed in the stage 2 PR. If you have time to incorporate more feedback in your next session on this, all the better. But feel free to just address the points above so we can move forward.

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,158 @@
...
+Stage 4: Document any new concerns and their resolution. The goal here is to eliminate risk of churn and instability by ensuring all concerns have been addressed.
+-->
+
+## Real-world implementations
+
+<!--
+Stage 4: Identify at least one real-world, production-ready implementation that uses these updated field definitions. An example of this might be a GA feature in an Elastic application in Kibana.
+-->
+
+People
```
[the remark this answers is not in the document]
webmat: Additional note. @jamiehynds discussed with @marc-gr and Marc will take over as the sponsor for this RFC. Could you mark him as such in the People section, please?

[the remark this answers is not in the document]
webmat: Sorry, disregard this comment about the sponsor. I misunderstood what Jamie said.

Jamie remains the sponsor for this one :-)

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,158 @@
...
+Email specific fields:
+
+| field | type | description |
+| --- | --- | --- |
+| `email.bcc.addresses` | wildcard | Addresses of Bcc's |
+| `email.cc.addresses` | wildcard | Addresses of Cc's |
+| `email.attachments_count` | long | A field outside the flattened structure to control how many attachments are included in the email |
+| `email.attachments` | flattened | A flattened field for anything related to attachments. This allows objects being stored with all information for each file when you have multiple attachments |
+| `email.direction` | keyword | Direction of the message based on the sending and receving domains |
+| `email.sender.address` | wildcard | Senders email address |
```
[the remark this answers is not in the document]
P1llus: I think its worth exploring in the upcoming stage for sure if that is appropriate.

P1llus:
Added the required changes you mentioned + a few more tweaks based on some changes that could wait for stage 2 @webmat . Anything else we should do at this point?

webmat (approved):
Alright, thanks for the adjustments, @P1llus! I'll adjust the date and merge.

rfcs/text/0008-email.md
```diff
@@ -0,0 +1,166 @@
+# 0008: Email
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **Oct 5th 2020** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```
webmat, line 5:
```suggestion
- Date: **2020-11-30** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```

Issue #420, opened 2019-04-09
SIP package


jiriatipteldotorg:
Folks,

may I propose including a SIP package (RFC3261). I suggest it would include the fields as bellow, modelled after the HTTP package. What's different:
- SIP requests have many more URLs in them. Therefore the suggestions `Use the url field set to store the url of the request.` only applies to the request-URI which represents the URI in the request as received (SIP proxy elements may also alter it during forwarding). For all the other URIs, I included only the unparsed form `full`.
- other fields that offer themselves for use in SIP events are `event.duration` for reporting SIP call length, `user_agent.*` for SIP User Agent (example: AVM FRITZ!Box Fon WLAN 7270 v3 74.06.05 (Apr  9 2014), `host.name` for receiving SIP server's DNS name, `host.id` for SIP server's unique identification, `host.ip` for receiving server's IP address, `client.ip` and `client.port` for client's IP:port, `user.id` for digest authentication identity, `client.geo.*` for client's geographic info.
- `client.transport` would be useful but doesn't exist (#418 )
- `event.action` seems the proper candidate for values describing the actual nature of an event like `call-attempt`(failed INVITE request), `call-start` (succesful INVITE request), `call-stop` (received or initated BYE), `register-new` for a newly added contact binding, `register-del` for removed contact binding, `register-expired` for one that has expired.
- in SIP, response elements often mirror their request counterparts. Where identical for requests and response, I didn't prefix the field name with request/response -- I hope that's the proper use of the ECS conventions.

There are couple of fields which would be helpful too but I don't see them as SIP specific -- see the list bellow, merely for sake of context, if there is interest I can create separate issues for them.
- session id. It is often useful to correlate some events belonging to each other -- for example a SIP call's id may be used to correlate events describing start and stop of a VoIP call, or a shopping session id may be used to correlate additions to and removals from a shopping basket.
- for use with SIP proxy elements it would be useful to describe in events how such a proxy element describes changes to forwarded requests: most importantly how it is changing the URI, where it is forwarding a request to, and if a response was generated locally or proxied. However this doesn't seem SIP-specific, I think the same data would be useful for example for HTTP proxy servers like nginx or ha-proxy ,
[...]

Issue #418, opened 2019-04-09
request for a new transport field in source / client / server


jiriatipteldotorg:
Folks,

I'm wondering if you could consider including a new field, `transport`, in the source / client / serve specifications.

There are several application protocols, for example SIP or DOTS, that use multiple transport protocols (UDP, TCP, SCTP). The field would allow to discriminate flows over different transport protocols.

The actual specification could be the same as in `network`.

```
    - name: transport
      level: core
      type: keyword
      short: Protocol Name corresponding to the field `iana_number`.
      description: >
        Same as network.iana_number, but instead using the Keyword name of the
        transport layer (udp, tcp, ipv6-icmp, etc.)

        The field value must be normalized to lowercase for querying. See
        the documentation section "Implementing ECS".
      example: tcp
```

webmat:
Hi! Thanks for the suggestion.

Could you flesh out an example (e.g. with simple JSON objects) that would show how you envision this to work?

Another thought, I assume you're talking about wrapping one protocol into another? Wouldn't it be better to e.g. define `network.transport` as the wrapping protocol, and a second field under `network` to represent the embedded protocol?

If we were to add `transport` to the network endpoints (src/dst or cli/srv), what would be the semantics? Always fill both `source.transport` == `destination.transport`? Or are there cases where only one side would be filled?

PR #1014, opened 2020-10-06
[RFC] SIP Fields

dainperkins:
initial SIP Field RFC

[Markdown preview of this RFC](https://github.com/dainperkins/ecs/blob/SIP-RFC/rfcs/text/0008-sip-fields.md)


CHANGELOG.next.md
```diff
@@ -22,6 +22,7 @@ Thanks, you're awesome :-) -->
 * Added `threat.technique.subtechnique` to capture MITRE ATT&CK® subtechniques. #951
 * Added `configuration` as an allowed `event.category`. #963
 * Added network directions ingress and egress. #945
+* Added SIP RFC #tbd
```
ebeahan, line 25: Don't worry about adding CHANGELOG entries for RFC PRs.

Once the RFC has advanced to a later stage, the PR which adds the field definitions to the schema can be listed.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+https://tools.ietf.org/id/draft-ietf-sipclf-format-05.html
+https://www.sipforum.org/
+
+https://github.com/elastic/ecs/issues/420
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 237:
```suggestion
* Stage 0: https://github.com/elastic/ecs/pull/1014
```

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+<!--
+Stage 4: Document any new concerns and their resolution. The goal here is to eliminate risk of churn and instability by ensuring all concerns have been addressed.
+-->
+
+## Real-world implementations
+
+Packetbeat Implementation (packet/protocol analysis)
+Zeek Implementation (packet/protocol/connection analysis)
+Cisco Call Manager (Log ingestion)
+Mictosoft Lync (Log Analysis)
```
ebeahan, line 196:
```suggestion
Microsoft Lync (Log Analysis)
```

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
+# 0000: SIP Fields
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 4: Between the discussions we've had already and the tremendous work that @dainperkins has already invested, I suggest targeting stage 1 here.

What do others think?

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+## Usage
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+Typical implementations will utilize these fields to describe and normalize the various stages of a SIP/SDP based communcations mechanism.  Additional considerations including call analytics, fraud detection, troubleshooting, and threat detection have been identified as additional considerations.
+
+## Source data
+
+Source Data will come from packet/protocol analysis from endpoints (e.g. Packetbeat) or network observers (e.g. Zeek/Corelight & Suricata), logs from SIP Servers (e.g. Cisco Call Manager, Microsfot Lync), or logs SiP aware perimeter devices (e.g. Palo Alto NGFW).
```
ebeahan, line 143:
```suggestion
Source Data will come from packet/protocol analysis from endpoints (e.g. Packetbeat) or network observers (e.g. Zeek/Corelight & Suricata), logs from SIP Servers (e.g. Cisco Call Manager, Microsfot Lync), or logs from SIP-aware perimeter devices (e.g. Palo Alto NGFW).
```

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
+
+Normalization, and the degree of normalization, of SIP URI fields may be an issue for discussion based on the potential implementation of ingesting SIP call records for the purposes of review for e.g. various types of communications fraud (e.g. should PSTN numbers be normalized with international dial codes, should implementaiton include capabilities to define internal call plans for more effective analysis, etc.)
+
+Normalization of SIP/SDP and real time communication protocol connections may require the definition of a field similar to network.community_id to allow for the trackign of the full scope of a connection. Additionally the ininital SDP setup phase often includes
```
ebeahan, line 171:
```suggestion
Normalization of SIP/SDP and real time communication protocol connections may require the definition of a field similar to network.community_id to allow for the tracking of the full scope of a connection. Additionally the initial SDP setup phase often includes
```

ebeahan (commented):
Thanks @dainperkins for putting this together!

I made a quick editorial pass with some notes, but I plan on taking a deeper look at the fields you've proposed.

CHANGELOG.next.md
webmat: Indeed, please remove the changelog line, @dainperkins . No need for those in RFCs.

rfcs/text/0008-sip-fields.md
webmat: Agreed with stage 1 from the start

ebeahan, line 4:
```suggestion
- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```

ebeahan (commented):
I took another 👀 and noted a few items.

One overall observation: reviewing the various headers a SIP message can contain recalled [past discussions](https://github.com/elastic/ecs/pull/554#issuecomment-534279863) around what better guidance ECS could provide around HTTP headers.

Perhaps the SIP fields could be another opportunity to discuss guidance around these type of request/response "boxes" to capture all these headers.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+| sip.via.sent_by.address       	| 192.168.1.2                                                                                                                                                                                                         	| keyword     	|
+| sip.via.sent_by.port          	|                                                                                                                                                                                                                     	| long        	|
+| sip.via.received.address      	| 80.230.219.70                                                                                                                                                                                                       	| keyword     	|
+| sip..via.rport                	| 5060                                                                                                                                                                                                                	|             	|
```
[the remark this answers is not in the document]
dainperkins: I think that's a good idea - I keep getting wrapped around the axle looking @ tcp headers (syn, ack, etc.) and icmp from the standpoint of unidirectional vs bidirectional src/dst... I think a larger discussion would be a great idea.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+| **SIP Response**              | SIP/2.0 200 OK                               	|           	|
+| sip.code                  	| 200                                          	| keyword   	|
+| sip.status                	| ok                                           	| keyword   	|
+| sip.version               	| 2                                            	| keyword   	|
+|                           	|                                              	|           	|
+| **SIP Headers**               |                                              	|           	|
+| sip.accept                	| application/sdp                              	| keyword   	|
+| sip.allow[]               	| REGISTER, INVITE, ACK, BYE                   	| keyword[] 	|
+| sip.call_id               	| 1-1966@10.0.2.20                             	| keyword   	|
+| sip.content_length        	| 0                                            	| integer   	|
```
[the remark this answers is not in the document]
dainperkins: not really...per packet integer would work - per 1m flow or conection might not

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+| sip.private.username      	| 35104723                                     	| keyword   	|
+| sip.supported[]           	| timer, path, replaces                        	| keyword[] 	|
+| user_agent.original       	| FreeSWITCH-mod_sofia/1.6.12-20-b91a0a6~64bit 	| keyword   	|
+| user_agent.original.text  	| FreeSWITCH-mod_sofia/1.6.12-20-b91a0a6~64bit 	| text      	|
+|                           	|                                              	|           	|
+| **SIP Headers CSEQ**        	| 68 invite                                    	|           	|
+| sip.cseq.code             	| 68                                           	| integer   	|
+| sip.cseq.method           	| invite                                       	| keyword   	|
+|                               |                                               |               |
+| **SIP Headers Via**              | Ex 1: SIP/2.0/UDP 192.168.1.2;received=80.230.219.70;rport=5060;branch=z9hG4bKnp112903503-43a64480192.168.1.2 EX2: SIP/2.0/UDP 10.0.2.20:5060;branch=z9hG4bK-1966-1-0                                               	|             	|
```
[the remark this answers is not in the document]
dainperkins: yeah theres a possibility of multiple proxies in the mix.  ordering would be nice, but I'm unsure what it looks like

jiriatipteldotorg:
Folks, my apologies for having missed that this conversation -- I started it in April 2019 #420 and didn't notice it became active now. And thank you for the effort placed in it -- there has been quite some diligence in it. Several comments on it bellow mostly driven by concern around level of context aggregation (quick example in numbers: one logon-loggoff event per SIP phone and day with aggregation, 2880 events per phone/day due to use of SIP re-registers as NAT keep-alives at 30 sec interval). Before that a quick self-intro: I'm running a SIP service (iptel.org) powered by SER/Kamailio which I have co-authored as well, so I hope it is not an exaggeration to state I've had collected quite some SIP mileage. Also my current company has a product (SBC) that generates SIP events as described in #420 and we have built a prototypical `sipbeat` -- based on `packetbeat` yet with numerous further refinements such as adjustments to specifics of underlying transport protocol and actual SIP context. It is running now on iptel.org service where it has been running for years. (see https://monitor.frafos.com for some visualizations)

Generally, the original proposal in #420 was based on our perspective that for security analysis it is important to capture the semantics of SIP traffic  more than replicating the raw  SIP message elements in JSON.  Even perhaps at the (unnecessary) risk of aggregating some details out. Another concern we have had is that  in the SIEM context without reasonable aggregation, SIP volume attacks would be amplified by the induced JSON event volume. Alone in the aggregated form shown in #420 we produce about 125MB of index volume per 1000 subscribers a day. That would increase by several orders of magnitude without aggregation. 

`Semantics` means here the aggregation of protocol elements into events describing their logic and purpose-- not strictly identical in the SIP protocol design. The most obvious example is that with aggregation for a working call we can generate as few as `call-start` and `call-stop` events. If we were to "clone" the SIP behaviour at transaction layer (i.e. with SIP processing limited to pairing requests-responses without dialog correlation), we would have 1) an INVITE transaction event for the unauthenticated call attempt 2) another INVITE transaction for an attempt to authenticate 3) an ACK transaction (by RFC3261 an e2e ACK while having the same CSeq may or may not visit the same path as INVITE and will have a different transaction id), 4) optionally an in-dialog transaction such as INVITE that's actually called "re-INVITE" 5) a BYE transaction. Also with transactional layer, additional correlation and post-processing must be performed by the event consumer because most of subsequent analyzes require a whole view of a call and data such as call duration, or QoS stats if available. 

An even significantly bigger gap occurs with registrations. ,For registrations we report on "log in" and "log-out": something that say for desktop phones happens once a day, for mobile SIP UAs when you loose connectivity. However reporting every REGISTER transaction would often mean in the field to generate an event for every SIP phone every 30 seconds because that's how often re-REGISTERs come to a registrar in order to keep NAT binding alive for incoming SIP traffic. That's often burden already for the SIP protocols (a reason why SBCs have implemented often a "registrar cache"), amplifying the stress by generating JSON events would be quite a challenge.  All of it seems even harder without transaction aggregation on a message-by-message basis because then answers come as separate events additionally as well.

I'm glad to share more details on the actual protocol elements yet I think what's worth the first consideration is the context level at which to operate. 

Here at least few details I've noticed quickly in
https://github.com/elastic/ecs/blob/0244f82a88681bf6789f84bc23b7bfe230caebe1/rfcs/text/0008-sip-fields.md
please bear with me I may be missing some ECS conventions 
- clarification: first of all I assume the current draft operates at a message-by-message level. MEaning that for a simple call the following events would be generated: INVITE w/credentials, 401, ACK, INV/w/credentials, 100, 180, 200, ACK, BYE, 200
- in examples the video SDP lines actually include audio data
- cross-protocol data elements: I think it would be worth documenting how to use non-SIP-fields for SIP traffic, such as `host.ip` for receiving server's IP, `client.ip` for the sender. Perhaps for sake of better correlation `user.id` could be taken for SIP digest authentication identity. 
- link to transport. `client.transport` would be useful but doesn't exist (#418)

rfcs/text/0008-sip-fields.md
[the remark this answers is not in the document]
jiriatipteldotorg: examples of such bellow. ordering matters. The topmost via is the most recent along the source-destination path. Some other headers may also occur multiple times, either as multiple headers, or multiple header values, or combination of both: `Record-Route` and `Route` often. 

```
U 192.168.0.85:5060 -> 192.168.0.155:6060
INVITE sip:5c47b8a83af57797@192.168.0.155:6060;transport=udp SIP/2.0.
Record-Route: <sip:192.168.0.85;ftag=1B8F4BD4-5FA0B47F000630A1-E25B7700;avp=Jr0DBwBhY2NvdW50AwB5ZXMDBgBzdGltZXICADkwAwkAZGlhbG9nX2lkFgAwNGQzLTVmMjNkNDk2LTMxYTgzODNj;lr=on>.
Via: SIP/2.0/UDP 192.168.0.85;branch=z9hG4bK94b5.5043d304.0.
Via: SIP/2.0/UDP 192.168.0.155:6060;branch=z9hG4bK~tW~Hade;rport=6060.
From: "jiri" <sip:jiri@iptel.org>;tag=1B8F4BD4-5FA0B47F000630A1-E25B7700.
To: <sip:103@frafos.net>.
CSeq: 11 INVITE.
Call-ID: 38E5D275-5FA0B47F000630AB-E25B7700.
Max-Forwards: 4.
Supported: ice, outbound, replaces, timer.
Session-Expires: 90.
Allow: INVITE,ACK,CANCEL,BYE,UPDATE,MESSAGE,OPTIONS,REFER,INFO.
User-Agent: JsSIP 3.0.12.
X-Org-ConnID: 94.142.239.227-20201103023807-cMigNfdo+tg2VUv.
X-Call-ID: qfg0157hu3mujhmuufii.
Content-Type: application/sdp.
Content-Length: 4642.
Contact: <sip:1B8F4BD4-5FA0B47F000630A1-E25B7700@192.168.0.155:6060;transport=udp>.
P-hint: usrloc applied.
P-hint-DBG: forward-sbc.
.
v=0.
o=- 1185881956307229883 2 IN IP4 192.168.0.155.
s=-.
...
```

dainperkins:
Wow - great information!  Thanks for jumping in (my SIP experience is obviously very network centric (not to mention Cisco Call Manager and a couple of SIP providers)

On the registration volume - is there an easy way to determine a discrete identifier that would work across most or all of the likely sessions (source/dest ip seem reasonable until NAT comes into play, I'm not sure if sessions will typically end up using different source ports with multiple devices connecting thru Nat) as a way to roll-up the data that would be considered repetitive for a session?

The video codec in the audio data is probably just a copy/paste error on my part.

Your assumption on the basis (message by message) is correct - I took this from a network capture perspective, and tried to grab all of the things that I could recall troubleshooting in a corporate environment (e.g. codec specifics relating to WAN ckts being overrun). The additional fields (source/destination.ip, port, user, etc.) will definitely be added in to get a full scope, I'm also hoping to use the session fields (another RFC) to work towards some of the data normalization you identified above.

webmat:
@jiriatipteldotorg Thank you so much for this added context, this is priceless!

I'm very much out of my depth when it comes to SIP. However I wanted to follow up on one specific thing you mentioned:

> documenting how to use non-SIP-fields for SIP traffic

Yes, this RFC should cover this. If some SIP fields can be mapped to existing ECS fields, we should call out to them here; and we'll have that in the documentation, when the addition is made to ECS. In your implementation, could you list all of these mappings? I see you've named a few already, but if there's more, we'd love to hear it :-)

Note on the `host.ip` & `client.ip` mentions: in ECS, we try to emphasize the following pairs: source & destination and/or client & server. Here I'll defer to you on which pair is most appropriate to use for SIP.

dainperkins:
@jiriatipteldotorg

Getting back to this - on the log based session (think multiple registration / reregistration events for the same user / sip end point) I am wondering if there are fields missing - or if selectively utilizing the fields provided (plus source, destination, user, network, session, etc) would fit the bill.

If there's anything missing we'll need to get that in, but I'm hoping we can leverage a subset of the full packetbeat targeted fields for the log entries - and leverage some combination of selective logging and pipeline logic to help keep the noise down.

jiriatipteldotorg:
> Wow - great information! Thanks for jumping in (my SIP experience is obviously very network centric (not to mention Cisco Call Manager and a couple of SIP providers)
> 
> On the registration volume - is there an easy way to determine a discrete identifier that would work across most or all of the likely sessions (source/dest ip seem reasonable until NAT comes into play, I'm not sure if sessions will typically end up using different source ports with multiple devices connecting thru Nat) as a way to roll-up the data that would be considered repetitive for a session?

Morning (sorry for late reply -- the issue notifications ended in Junk :-( )

I would recommend {AoR,source_IP, Contact_URI_without_port_and_parameters} as aggregation key. 

registration background: there can be multiple registrations for a SIP AoR so one can have multiple contacts like jiri@10.0.0.1 and jiri@192.168.0.1 registered with an AoR, or even worse they can have the same private IP addresses if behind distinct NATs with overlapping private-side address space. Various misbehaviours in SIP implementations appear that make aggregations harder: some SIP phones do change their port number on every registration (too many events if we include port numbers in keys) and some change CallId. So for practical reasons (not governed by the SIP spec) an aggregation key for SIP registrations is not too specific (no port numbers, no CAllId) and captures each binding using the pair {AoR,Contact_without_port}. I would recommend to also add  source_ip added to disriminate call agents registered from behind NATs with overlapping private address space. This may generate additional records when the NAT public IP address changes (e.g. during forcible IP address change as some ISPs practice) but I think it is fair enough -- all in all we get then reports when UA's connectivity changes as seen from outside even if the client us unaware of it. 


Few more numbers: in average one gets at a public SIP service about 3GB of raw SIP PCAP a day for 1000 subs. The most active contributions to the numbers are the aforementioned SIP re-registrations and busy hour call attempts. (I don't have a specific  break-down of that). The aggregated events cost us 125 MB index space for the 1000 subs a day.



> 
> The video codec in the audio data is probably just a copy/paste error on my part.
> 
> Your assumption on the basis (message by message) is correct - I took this from a network capture perspective, and tried to grab all of the things that I could recall troubleshooting in a corporate environment (e.g. codec specifics relating to WAN ckts being overrun). The additional fields (source/destination.ip, port, user, etc.) will definitely be added in to get a full scope, I'm also hoping to use the session fields (another RFC) to work towards some of the data normalization you identified above.

I don't have hard data for corporate environments (I could try to get some) -- I would expect them to be more moderate (probably NAT refreshes are not used, there are fewer offending sources, even though some features like SUB-NOT-based status updates can generate lot of traffic too).

jiriatipteldotorg:
> @jiriatipteldotorg Thank you so much for this added context, this is priceless!
> 
> I'm very much out of my depth when it comes to SIP. However I wanted to follow up on one specific thing you mentioned:
> 
> > documenting how to use non-SIP-fields for SIP traffic
> 
> Yes, this RFC should cover this. If some SIP fields can be mapped to existing ECS fields, we should call out to them here; and we'll have that in the documentation, when the addition is made to ECS. In your implementation, could you list all of these mappings? I see you've named a few already, but if there's more, we'd love to hear it :-)
> 
> Note on the `host.ip` & `client.ip` mentions: in ECS, we try to emphasize the following pairs: source & destination and/or client & server. Here I'll defer to you on which pair is most appropriate to use for SIP.

At least my sense is too that for sake of cross-protocol correlation which is presumably a key property of ECS, it makes sense to reuse some generic structures wherever posssible. In our implementation we are using `client` and `server` (see examples bellow). I'm not sure I captured the nuance between source/destination and client/server -- I remember I was confused by what's the difference ... is it perhaps reporting on replies separately as opposed to reporting on whole request-response transactions that makes the difference? (in which case a response will have same client/server properties as request and swapped source/destination compared to request?)

Others fields one may consider to reuse
- `event.duration` when a call is terminated to state how long it was (if known by the event sender)
- `user_agent` as shown bellow (similar to HTTP)
- `user.id` for SIP digest identity (may be a differnt value than what's see in From URI)
- `url.original` for SIP request URI. Note that should be an event reported by a proxy element that typicall changes the request URI, it would also make sense to have some `url.outoging`. Typically case: an INVITE arrives at a proxy with jiriatoffice@iptek.org, the proxy makes a user location lookup and forwards the request with changed request URI jiri@191.192.193.194 (learned during SIP registration previously). 

Some event examples as we produce them below. I've planned to release our implementation since long ago but didn't come to it, unfortunately -- hopefully, I can gather myself to do it soon.

```
{
    "agent": {
        "id": "07d7ab3f-1c10-4879-ab68-c999d6f979a2",
        "ephemeral_id": "fe5e84cd-4b87-42b8-adfc-b04b0a34ec4e",
        "version": "0.6.13 sipcallmon 0.6.12",
        "hostname": "ip-172-31-44-93.eu-west-1.compute.internal",
        "name": "ip-172-31-44-93.eu-west-1.compute.internal",
        "type": "sipcmbeat"
    },
    "sip": {
        "response": {
            "status": 401
        },
        "from": "sip:7000@3.250.25.65",
        "call_id": "952659369-1792449447-297086673",
        "contact": "sip:7000@156.96.128.224:51541",
        "sip_reason": "Unauthrorized",
        "request": {
            "method": "INVITE"
        },
        "to": "sip:270046842002343@3.250.25.65"
    },
    "client": {
        "transport": "udp",
        "port": 51541,
        "ip": "156.96.128.224"
    },
    "server": {
        "port": 5060,
        "ip": "172.31.44.93"
    },
    "uri": {
        "original": "sip:270046842002343@3.250.25.65"
    },
    "type": "auth-failed"
}
```

```
{
    "client": {
        "port": 5060,
        "ip": "194.166.186.25",
        "transport": "udp"
    },
    "uri": {
        "original": "sip:iptel.org"
    },
    "host": {
        "name": "sip0.iptel.org"
    },
    "server": {
        "port": 5060,
        "ip": "212.79.111.155"
    },
    "type": "reg-del",
    "ecs": {
        "version": "1.0.0"
    },
    "agent": {
        "version": "7.1.0",
        "hostname": "sip0.iptel.org",
        "ephemeral_id": "a254663e-1ba7-4c28-a9b1-aa3307d1f161",
        "id": "f839018b-ef29-4596-91a2-6fdb0a0c625a",
        "type": "sipcmbeat"
    },
    "@timestamp": "2020-11-11T13:32:43.712Z",
    "sip": {
        "from": "sip:donttellme@iptel.org",
        "call_id": "UkmowGWuK0I4Qosy@10.0.0.2",
        "to": "sip:donttellme@iptel.org",
        "sip_reason": "OK",
        "response": {
            "last": 200
        },
        "request": {
            "method": "REGISTER"
        },
        "contact": "sip:donttellme@199.116.184.25:5060"
    },
    "user_agent": {
        "original": "BSIP-OEM 0.63.035"
    }
}
```

jiriatipteldotorg:
Some other thing: SIP extensions. There are horribly many of them. I would at least suggest to include
- https://tools.ietf.org/html/rfc7989 -- `Session-ID` that can help in some correlation cases (when a B2BUA transforms the "usual" dialog identifiers so an incoming call looks different from the outgoing)
- https://tools.ietf.org/html/rfc3325 -- `P-Asserted-Identity ` -- like if there were not enough SIP URIs, this one adds more and is used often
- https://tools.ietf.org/html/rfc3581 -- `rport` -- Via header-field fix of the previously NAT-unaware protocol design
- some often used additional methods: REFER (call tranfers) RFC3515, INFO (sometimes used for DTMF) RFC 6086, SUB-NOT (RFC6665), UPDATE (RFC3311)
- some often used additional header fields: Reason (good for diagnostics) RFC3326

note that in addition to SIP/SIPS protocol discriminators, `tel` is also used (https://tools.ietf.org/html/rfc3966)

an overview  of related specs was published as RFC5411 but it is dated 2009 -- some documents are not included and some are included but didn't get adopted in the field.
https://tools.ietf.org/html/rfc5411
PAul Jones' packetizer list is not annotated but more up-to-date: https://www.packetizer.com/ipmc/sip/standards.html

jiriatipteldotorg:
> @jiriatipteldotorg
> 
> Getting back to this - on the log based session (think multiple registration / reregistration events for the same user / sip end point) I am wondering if there are fields missing - or if selectively utilizing the fields provided (plus source, destination, user, network, session, etc) would fit the bill.
> 
> If there's anything missing we'll need to get that in, but I'm hoping we can leverage a subset of the full packetbeat targeted fields for the log entries - and leverage some combination of selective logging and pipeline logic to help keep the noise down.

I think the proper level of aggregation is `SIP binding` -- that tells you that a phone logged on or off. AoR would be IMO too coarse as it would fail to capture bheaviour of multiple phones registered for the same AoR. I think a proper aggregation key  for a SIP binding is  {AoR, Contact-without-port-and-params, source-ip} -- see a separate answer for the rationale. 

Not sure what would be the best mapping to existing ECS fields but to capture the situation of a phone logged on/off, at least the To URI and Contact should be provided so that one knows both the affected user and the device in question. Plus the usual `source`, `destination`, `user-agent` always useful.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+<!--
+Stage 4: Document any new concerns and their resolution. The goal here is to eliminate risk of churn and instability by ensuring all concerns have been addressed.
+-->
+
+## Real-world implementations
+
+Packetbeat Implementation (packet/protocol analysis)
+Zeek Implementation (packet/protocol/connection analysis)
+Cisco Call Manager (Log ingestion)
+Microsoft Lync (Log Analysis)
```
webmat, line 196: Here we're looking to list the real world implementation of using the SIP ECS fields. You can keep Packetbeat here, since the goal is for it to be adjusted to match the ECS fields.

The other ones should be removed for now, though. They're already correctly mentioned in the "data sources" section.

Note that when the Zeek SIP parser is adjusted to follow ECS we can list it here as well. But hopefully it won't be adjusted until this RFC reaches at least stage 2, when fields will be more stable.

webmat (commented):
This proposal is really far along, the list of fields looks really good already for a stage 1 proposal 👍 

In order to merge as stage 1, I'd like the PR to be adjusted in a few ways (see corresponding review comments):

- [x] remove the changelog entry
- [x] change the order of the "type" and "Example" columns for readability
- [x] capture [this sample event](https://github.com/elastic/ecs/pull/1014/files#r516386512) in the RFC for future reference
- [ ] fix the typo in "Concerns"
- [x] adjust the content of "Real-world Implementations"
- [x] add the subject matter entry
- [x] Add @marc-gr as the sponsor for this RFC
- [ ] turn the reference section into bullet points

My review contains a lot more technical feedback as well. You can either address right away, or we can carry over these discussions and finalize a stage 2 PR.

I'd like to merge this stage 1 PR next week if possible, so please feel free to only address the smaller list of points above for now. We can finalize on the technical feedback in a stage 2 PR.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+<!--
+Stage 4: Identify at least one real-world, production-ready implementation that uses these updated field definitions. An example of this might be a GA feature in an Elastic application in Kibana.
+-->
+
+## People
+
+The following are the people that consulted on the contents of this RFC.
+
+* @DainPerkins | Author
```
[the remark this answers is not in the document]
webmat: @jamiehynds just discussed with @marc-gr, and Marc will the sponsor for this RFC. Could you add him here as well please?

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting.
+-->
+
+<!--
+Stage 3: Add more real world example source documents so we have at least 2 total, but ideally 3. Format as described in stage 2.
+-->
+
+## Scope of impact
+
+No impact expected as SIP fieldsets are new, and will not impact any existing fields.
```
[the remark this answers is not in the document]
ebeahan: It looks like yes, the Packetbeat SIP/SDP support is out in 7.10 as beta ([reference](https://www.elastic.co/guide/en/beats/packetbeat/current/configuration-protocols.html)).

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+| sip.auth.uri.original.text    	| sip:192.168.1.2                                                                                                                                                                                                     	| text        	|
+| sip.auth.uri.scheme           	| sip                                                                                                                                                                                                                 	| keyword     	|
+| sip.auth.uri.host             	| 192.168.1.2                                                                                                                                                                                                         	| keyword     	|
+| sip.auth.uri.port             	| n/a                                                                                                                                                                                                                 	| long        	|
```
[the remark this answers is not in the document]
dainperkins: No, just that thee example didn't list one - I left it blank in the update

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Fields
+
+| Sip Fields                	| Example                                      	| type      	|
+|---------------------------	|----------------------------------------------	|-----------	|
+|                           	|                                              	|           	|
+| **SIP Request**               | INVITE sip:test@10.0.2.15:5060 SIP/2.0       	|           	|
+| sip.type                  	| request / response                           	| keyword   	|
+| sip.method                	| invite                                       	| keyword   	|
+| sip.uri.original          	| test@10.0.2.15:5060                          	| keyword   	|
```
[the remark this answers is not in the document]
dainperkins: did it for all of the uri fields, and removed the text multifield.

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
...
+Stage 2: Identifies scope of impact of changes. Are breaking changes required? Should deprecation strategies be adopted? Will significant refactoring be involved? Break the impact down into:
+ * Ingestion mechanisms (e.g. beats/logstash)
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
+
+Normalization, and the degree of normalization, of SIP URI fields may be an issue for discussion based on the potential implementation of ingesting SIP call records for the purposes of review for e.g. various types of communications fraud (e.g. should PSTN numbers be normalized with international dial codes, should implementaiton include capabilities to define internal call plans for more effective analysis, etc.)
```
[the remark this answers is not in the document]
dainperkins: fixed

rfcs/text/0008-sip-fields.md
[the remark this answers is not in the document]
dainperkins: added both

rfcs/text/0008-sip-fields.md
dainperkins: commented out

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,239 @@
...
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+Typical implementations will utilize these fields to describe and normalize the various stages of a SIP/SDP based communcations mechanism.  Additional considerations including call analytics, fraud detection, troubleshooting, and threat detection have been identified as additional considerations.
+
+## Source data
+
+Source Data will come from packet/protocol analysis from endpoints (e.g. Packetbeat) or network observers (e.g. Zeek/Corelight & Suricata), logs from SIP Servers (e.g. Cisco Call Manager, Microsoft Lync), or logs from SIP-aware perimeter devices (e.g. Palo Alto NGFW).
+<!--
```
webmat, line 135:
```suggestion

See this example of [raw SIP header](0008/Sip-via-ordering-example.txt).

<!--
```

webmat (commented):
Thanks for all of the adjustments, @dainperkins 👍 

One small thing left and we can merge: can you link the sample SIP example document from the main RFC document? This will ensure it's discoverable to folks who don't know about the corresponding directory structure and stuff.

You can use my suggestion as is, or reword if you think it's useful to discuss the VIA header ordering stuff.

Once this is in, we can merge and quickly get going with the stage 2.

dainperkins:
In looking thru the items above I've compiled the following as potential tasks to consider for moving forward to stage 3

**Sip RFC Stage 3 To Dos:**
- [ ] Review additional RFCs for key fields (e.g. SIP session https://tools.ietf.org/html/rfc7989)
- [ ] Detail non SIP fields typically utilized and when they should be utilized (e.g. src/dest will always be used, top level User might be useful)
- [ ] Determine potential additions to session RFC (e.g. session.id, session.remote-id) to better fit top level Session & SIP requirements
- [ ] Add reference pcaps, log entries, and related parsing examples

ebeahan:
> In looking thru the items above I've compiled the following as potential tasks to consider for moving forward to stage 3

Is this PR still targeting stage one? I believe after the remaining requested adjustments were made, this PR would advance to stage 1 by merging it. Discussions could then continue in a new PR for stage two.

No harm in adding the notes for future reference, of course 😄 .

webmat:
Yes this still targets stage 1. Right after this we'll open a stage 2 PR. I think the "stage 3" mention was just a mistake :-)

webmat (dismissed):
Thanks for all of the work on this, Dain!

rfcs/text/0008-sip-fields.md
```diff
@@ -0,0 +1,242 @@
+# 0000: SIP Fields
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **TBD** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```
webmat, line 5:
```suggestion
- Date: **2020-12-04** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```

webmat (approved):

PR #1015, opened 2020-10-06
[RFC] Wildcard - stage 3 proposal

ebeahan:
### Summary

Revisions to the stage two wildcard RFC for consideration of advancement for stage three.

Opening this PR with minimal changes. Feedback and concerns as the wildcard changes are adopted can be captured in this PR and incorporated into the proposal as appropriate.

- [x] Opened pull request for this candidate revising the existing draft

[Markdown preview of this RFC](https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-3/rfcs/text/0001-wildcard-data-type.md)


rw-access:
I saw a few typos. These say  "leading wildcard" but the example is a trailing wildcard: `foo*`
https://github.com/elastic/ecs/blob/e9b9681581755783ca4ccf8a0ab9f57ae9f30c81/rfcs/text/0001-wildcard-data-type.md#L148-L149

Inconsistent escaping, should do  `\\` for all escaped backslashes.
https://github.com/elastic/ecs/blob/e9b9681581755783ca4ccf8a0ab9f57ae9f30c81/rfcs/text/0001-wildcard-data-type.md#L182

Similarly, I think these should be escaped with `\` not `/`. I don't think this is valid syntax:
https://github.com/elastic/ecs/blob/e9b9681581755783ca4ccf8a0ab9f57ae9f30c81/rfcs/text/0001-wildcard-data-type.md#L241

Worth double checking if this is valid syntax. I'm not 100%  sure if `\` can escape whitespace. 
`process.command_line:*f\ foo* and process.command_line:*/b\ bar*`

webmat:
We should remove `user.domain`, in line with not including `host.domain`

webmat:
We could add a link to the wildcard docs in the references https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#wildcard-field-type

webmat:
Whoa, I just came across this `process.thread.name`:

```
"process.thread.name":"TC: 14:05:53 Stage: Initializing Startup subsystem; TeamCity Server (version 2020.1.4 build 78906 started 2020-10-21 14:05:53.422), node id: MAIN_SERVER"
```

And here I thought people were as imaginative as me, calling their threads "thread1", "thread2", "worker1" and so on.

I think we should make this field wildcard as well.

ebeahan:
@webmat - I, as well, must not be very imaginative 😂 

I've captured `process.thread.name` in the list.

webmat:
Just like I mentioned in today's meeting, I think we should make obvious that this RFC is solely about the migration of pre-existing `keyword` fields to `wildcard`.

Using `wildcard` right away on net new fields is totally fine, IMO. WDYT?

ebeahan:
Yes agreed! I've actually made the suggestion to consider `wildcard` for net new fields in other proposals, so I'm glad we're aligned 😉 😂 

I will adjust the RFC's messaging to make it clear this RFC is documenting the initial proposal focused on existing fields, but any net new fields should consider using `wildcard` whenever appropriate.

ebeahan:
I made some adjustments to the title and intro to try and better focus on the migration over wildcard's general presence in ECS: [`3f2eaaa` (#1015)](https://github.com/elastic/ecs/pull/1015/commits/3f2eaaa57b3f1ee7a1a85df01484157d47d57a4c)

I'm happy to flesh this out with more detail if that feels appropriate.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -1,12 +1,14 @@
-# 0001: Wildcard Field Adoption into ECS
+# 0001: Wildcard Field Migration
 <!--^ The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC, taking care not to conflict with other RFCs.-->
 
-- Stage: **2 (draft)** <!-- Update to reflect target stage -->
+- Stage: **3 (draft)** <!-- Update to reflect target stage -->
```
webmat, line 4:
```suggestion
- Stage: **3 (candidate)** <!-- Update to reflect target stage -->
```

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -33,13 +35,13 @@ For a field to use wildcard, it will require changing the the field's defined sc
...
 | [`pe`](0001/pe.yml) | `pe.original_file_name` |
-| [`process`](0001/process.yml) | `process.command_line`<br> `process.executable`<br> `process.name`<br> `process.title`<br> `process.working_directory`<br> |
+| [`process`](0001/process.yml) | `process.command_line`<br> `process.executable`<br> `process.name`<br> `process.thread.name`<br> `process.title`<br> `process.working_directory`<br> |
 | [`registry`](0001/registry.yml) | `registry.key`<br> `registry.path`<br> `registry.data.strings` |
 | [`server`](0001/server.yml) | `server.domain`<br> `server.registered_domain` |
 | [`source`](0001/source.yml) | `source.domain`<br> `source.registered_domain` |
 | [`tls`](0001/tls.yml) | `tls.client.issuer`<br> `tls.client.subject`<br> `tls.server.issuer`<br> `tls.server.subject` |
 | [`url`](0001/url.yml) | `url.full`<br> `url.original`<br> `url.path`<br> `url.domain`<br> `url.registered_domain` |
-| [`user`](0001/user.yml) | `user.name`<br> `user.full_name`<br> `user.email`<br> `user.domain` |
+| [`user`](0001/user.yml) | `user.name`<br> `user.full_name`<br> `user.email` |
```
webmat, line 44: I think it wasn't removed in the YAML file, please remove it there as well :-)

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -429,26 +431,44 @@ Stage 2: Document new concerns or resolutions to previously listed concerns. It'
 
 Some fields require flexibility in how users search. Their content is messy (e.g. user-agent) or popular for threat hunters (e.g. file paths and names, command line processes), and a single character in the opposite casing can bypass a detection today for `keyword` fields. The `wildcard` field provides improved performance of leading `wildcard` and `regex` term-level queries, but is also a step towards case-insensitive search support in Elasticsearch. As Elasticsearch moves forward towards introducing a case-insensitive query option [3], ECS considers the fields adopting `wildcard` to be popular candidates for case-insensitive searching once the feature is available.
 
-One remaining question is if `wildcard` performs better vs. `keyword` for the upcoming case-insensitive query option.
+#### Resolution
+
+The `case_insensitivity` query parameter is an expected feature in Elasticsearch 7.10 with support for the `regexp`, `term`, `prefix`, and `wildcard` query types[4]. Since this is a query parameter, both `keyword` and `wildcard` types will both be supported, and each type's noted [performance characteristics](#comparison-with-keyword) will be consistent.
```
webmat, line 436: We don't need both of these "both"s ;-)

```suggestion
The `case_insensitivity` query parameter is an expected feature in Elasticsearch 7.10 with support for the `regexp`, `term`, `prefix`, and `wildcard` query types[4]. Since this is a query parameter, both `keyword` and `wildcard` types will be supported, and each type's noted [performance characteristics](#comparison-with-keyword) will be consistent.
```

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -429,26 +431,44 @@ Stage 2: Document new concerns or resolutions to previously listed concerns. It'
...
 
 ### Performance differences
 
-Performance and storage characteristics between wildcard and keyword will be different[4], and this difference may have an impact depending on deployment size and/or the level of duplication in the field data. As part of the transition, fields which were previously indexed keyword will be switched to wildcard. Queries across indices with field names will be necessary. Understanding the differences at both index and query time will be pursued.
+Performance and storage characteristics between wildcard and keyword will be different[5], and this difference may have an impact depending on deployment size and/or the level of duplication in the field data. As part of the transition, fields which were previously indexed keyword will be switched to wildcard. Queries across indices with field names will be necessary.
```
webmat, line 440: > Queries across indices with field names will be necessary.

I don't get what you're trying to convey here.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -429,26 +431,44 @@ Stage 2: Document new concerns or resolutions to previously listed concerns. It'
...
+
+The ECS team will not maintain a second OSS-compatible set of ECS field definitions. However a feature has been added to the ECS generator script allowing for OSS-compatibility versions of the ECS-maintained artifacts to be generated by users. When the generator is called with `--oss`, ECS fields using Basic data types will fallback to their compatible OSS types (if an OSS fallback type is available).
 
 ### Version Compatibility
 
 A data shipper which uses the `wildcard` field type may need to verify that the configured output Elasticsearch destination can support it (>= 7.9.0). For example, if a future version of Beats adopts `wildcard` in index mappings, Beats would may need to gracefully handle a scenario where the targeted Elasticsearch instance doesn't support the data type.
 
+#### Resolution
+
+Ongoing discussions are talking place to determine how to best address for Beats and Logstash.
```
webmat, line 470: That's too much talking 😂 

```suggestion
Ongoing discussions are taking place to determine how to best address for Beats and Logstash.
```

webmat (commented):
Noted a few things around the recent changes 👍

rfcs/text/0001-wildcard-data-type.md
ebeahan: I removed the YAML file entry in [`f51ecf8` (#970)](https://github.com/elastic/ecs/pull/970/commits/f51ecf876366abc5f0f600cee187b63fde098aa9), but I overlooked the entry in the table.

rfcs/text/0001-wildcard-data-type.md
ebeahan: Some of the edits and shuffling I did clearly left this unfinished 😂 . Improved in the latest revision.

ebeahan:
PR to apply `wildcard` migration changes in progress #1098

ebeahan:
Including `event.original` as a candidate field for `wildcard` may need reconsideration.

The `event.original` field in ECS is currently `keyword` with indexing disabled. Valid concerns were raised (https://github.com/elastic/ecs/pull/970#issuecomment-702813957) to continue disabling indexing on `event.original` to avoid any significant impact on indexing. Some data sources will produce original events that are very, very large. 

Wildcard is part of the `keyword` family. Unlike other `keyword` data types, it doesn't support disabling either indexing or doc values by design. Trying to set either option to false will be rejected by the API:

```
PUT wildcard-index
{
  "mappings": {
    "properties": {
      "wildcard": {
        "type": "wildcard",
        "index": false
      }
    }
  }
}

{
  "error" : {
    "root_cause" : [
      {
        "type" : "mapper_parsing_exception",
        "reason" : "The field [wildcard] cannot have index = false"
      }
    ],
    "type" : "mapper_parsing_exception",
    "reason" : "Failed to parse mapping [_doc]: The field [wildcard] cannot have index = false",
    "caused_by" : {
      "type" : "mapper_parsing_exception",
      "reason" : "The field [wildcard] cannot have index = false"
    }
  },
  "status" : 400
}
```

A simple path may be to omit `event.original` as a candidate for `wildcard` and have it remain `keyword` with `index: false`.

@webmat @leehinman Any thoughts?

webmat (commented):
I think we're good on the current list of fields to migrate. This LGTM other than a few small adjustments.

Nitpick: we're missing `organization.name` in the table :-)

I'm good with reverting `event.original` to `keyword` with `index: false` for now, as the way forward. We could adjust the field definition in #1098 to say it's currently not indexed, and that if users want to override this and make it indexed, they should consider `wildcard`.

ebeahan:
Latest round of changes:

* Added missing `organization.name` to the fields table
* Removed `event.original` as a wildcard candidate field
* Corrected formatting and incorporated peer suggestions.

webmat (dismissed):
This looks good, let's change that date and merge 👍

RFC 0001 is also the first to reach stage 3 🎉 😃

PR #1016, opened 2020-10-07
Clarify that file extension should exclude the dot.

webmat:
It was only implicit via the example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 #### Bugfixes
 
 * The `protocol` allowed value under `event.type` should not have the `expected_event_types` defined. #964
+* Clarify the definition of `file.extension` (no dots). #1016
 
 #### Added
 
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -1604,7 +1604,7 @@
       level: extended
       type: keyword
       ignore_above: 1024
-      description: File extension.
+      description: File extension, excluding the dot.
       example: png
     - name: gid
       level: extended
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -1606,8 +1606,8 @@
       ignore_above: 1024
       description: 'File extension, excluding the leading dot.
 
-        If the file name has multime extensions (example.tar.gz), this field should
-        contain all extensions without the leading dot (tar.gz).'
+        Note that when the file name has multiple extensions (example.tar.gz), only
+        the last one should be captured ("gz", not "tar.gz").'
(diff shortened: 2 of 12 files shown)


generated/beats/fields.ecs.yml
```diff
@@ -1604,7 +1604,7 @@
       level: extended
       type: keyword
       ignore_above: 1024
-      description: File extension.
+      description: File extension, excluding the dot.
```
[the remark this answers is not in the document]
webmat: Oh great point (pun intended).

I will address in the long description, yes

webmat:
@ebeahan WDYT of that new wording for multiple extensions?

rw-access:
It might be nontrivial for data sources to tell the difference between an extra extension vs just part of the filename. I always assumed this was just the outermost extension. I think it's debatable if double extensions are even a thing, since I don't think many operating systems recognize more than one.

One example where multiple extensions could arrive at the wrong conclusion: `FancyPants.2020.10.07.exe`

ebeahan:
@webmat I like the new wording, but @rw-access also brings up a valid point.

How the different routine/methods in the various languages/tools typically handle parsing file extensions should be considered. Looks like both [go](https://play.golang.org/p/OxCg0gCRsDt) and [node](https://nodejs.org/api/path.html#path_path_extname_path) both consider the extension the substring starting at the final `.` to the end of the string.

If most of our own data sources will typically only populate the final extension, perhaps it's best ECS aligns on that convention as well.

webmat:
Indeed, thanks for chiming in @rw-access

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 #### Bugfixes
 
 * The `protocol` allowed value under `event.type` should not have the `expected_event_types` defined. #964
+* Clarify the definition of `file.extension` (no dots). #1016
```
webmat, line 13: @ebeahan Since we'll backport this all the way back to 1.6/current, I'm shoving this PR's changelog directly in the 1.7 changelog for now.

I'll add a separate entry in 1.6 on the backport.

WDYT of this approach?

A bug fix changing the structure of the events (type change, typo fix) that we'd backport would warrant a point release. But clarifying the description to make explicit what was hidden in the example doesn't warrant a point release, IMO.

However I think it's important to make the change visible to anyone looking at past changelogs.

ebeahan: I agree with this approach of modifying `CHANGELOG.md` directly and then adding a separate entry for the `CHANGELOG.md` in the 1.6 branch.

Also agree that since the change only improves the description around what value was already expected, this doesn't warrant the point release.

ebeahan (approved):
LGTM

PR #1017, opened 2020-10-07
[RFC] Multiple users in an event, stage 3 PR

webmat:
Stage 3 criteria:

- Completed field definitions
  - [x] These are field reuses, so there's no new definitions per se.
  - [x] #1066 How to use each appearance of the user fields will be documented in a usage section for the user field set, via a distinct PR.
- [x] Included multiple real world example source documents.
  - This was done in the stage 2 PR, with the help of many contributors ❤️ 
  - Please let us know if you think an important usage example is missing
- [x] Existing or newly raised questions and concerns are addressed.
  - [x] Deprecating host.user fields.

Additional TODOs

- [x] #1066 Open PR effecting these changes in official ECS fields, removing them from the experimental artifacts.

[Preview of the RFC](https://github.com/webmat/ecs/blob/rfc/0007/stage-3/rfcs/text/0007-multiple-users.md)

### DIFF
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -765,7 +765,7 @@ e.g.:
 
 * Stage 2: https://github.com/elastic/ecs/pull/914
   * Stage 2 correction: https://github.com/elastic/ecs/pull/996
-* Stage 3: https://github.com/elastic/ecs/pull/NNNN
+* Stage 3: https://github.com/elastic/ecs/pull/1017
 
 Note: This RFC was initially proposed via a PR that targeted stage 2,
 given the amount of discussion that has already has happened on this subject.
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -636,14 +636,6 @@ and the assumed role in the `userIdentity`. This makes it easy to keep track of 
 the real user at `user.*` and the escalated privileges at `user.effective.*` in
 all subsequent activity after privilege escalation.
 
-<!--
-Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting.
--->
-
-<!--
-Stage 3: Add more real world example source documents so we have at least 2 total, but ideally 3. Format as described in stage 2.
--->
-
 ## Scope of impact
 
 ### New fields for IAM
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -1,8 +1,8 @@
 # 0007: Multiple users in an event
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **2 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **2020-10-02** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Stage: **3 (candidate)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **TBD** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 Many log events refer to more than one user at the same time.
 Examples of this are remote logons as someone else, user management and privilege escalation.
@@ -765,6 +765,7 @@ e.g.:
 
 * Stage 2: https://github.com/elastic/ecs/pull/914
   * Stage 2 correction: https://github.com/elastic/ecs/pull/996
+* Stage 3: https://github.com/elastic/ecs/pull/NNNN
 
 Note: This RFC was initially proposed via a PR that targeted stage 2,
 given the amount of discussion that has already has happened on this subject.
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -663,12 +663,12 @@ These came up while working on this RFC; this is not guidance that was given
 in the past. Data sources that populate these fields will need to be revisited
 and adjusted accordingly.
 
-<!-- TODO
+### host.user fields are deprecated for removal
 
-Depending on the outcome of the discussion on `host.user.*`, mention it here.
-It's currently listed in the concerns below.
+Seeing no use in the wild, it was decided to remove the reuse of the user fields at `host.user.*`.
+We will start by deprecating them in ECS 1.8, and will remove them at the next major version.
 
--->
+Please let us know before the next major ECS release if you disagree with this, and share how you're using them.
 
 ## Concerns
 
@@ -678,11 +678,11 @@ In past discussions and recent research, we have not identified a clear purpose
 for the user fields nested at `host.user.*`.
 
 We are considering deprecating these fields with the intent to remove them completely.
-Please let us know if you disagree with this, and share how you're using them.
 
 #### Resolution
 
-No resolution yet.
+They will be marked as deprecated starting with ECS 1.8, and will be removed in
+the next ECS major release.
 
 ### Documenting the purpose of each usage of the user fields
 
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -2,7 +2,7 @@
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
 - Stage: **3 (candidate)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **TBD** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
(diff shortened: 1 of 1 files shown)


P1llus:
As discussed with @webmat here is an example from a MySQL Audit log point of view, where you can have one user account being used for authentication, while another account is used for the activity after the initial authentication.

Dataset:
```
"class": "table_access",
	"event": "insert",
	"connection_id": 16,
	"account": {
		"user": "audit_test_user2",
		"host": "hades.home"
	},
	"login": {
		"user": "audit_test_user2",
		"os": "",
		"ip": "192.168.2.5",
		"proxy": ""
	},
```
The description of the **User** Object
`user`
A string representing a `user` name. The meaning differs depending on the item within which user occurs:
Within `account` items, `user` is a string representing the user that the server authenticated the client as. This is the `user `name that the server uses for privilege checking.
Within `login` items, `user` is a string representing the user name sent by the client.
Field documentation for audit logging: 
https://dev.mysql.com/doc/refman/5.7/en/audit-log-file-formats.html#audit-log-file-json-format

webmat:
I've updated the body of the issue with the criteria for merging as stage 3.

A few things to note:

* Having not heard any pushback on eventual removal of the `host.user.*` fields, I've gone ahead and adjusted the text assuming that this is our plan. Check out the RFC text for the details. Now's a good time to chime in if anybody disagrees. This was the only concern left to address.
* The criteria for the amount of examples for stage 3 were already met in the stage 2 RFC.

This may be the first stage 3 RFC we merge. With fields moving out of stage 2 (experimental) to stage 3 (beta in mainline ECS), this means close knowledge of the ECS build process is now required. Moreover, the current way of handling experimental changes is manual, and bound to change. It's not expected that RFC authors need to be familiar with the internals of the ECS build.

Therefore I propose that stage 3 RFCs (and stage 4s that come with field changes) be accompanied by a sister PR that effects the changes, authored by an ECS core team member. Both should be merged at the same time, when both are approved and ready. This decouples the simple RFC process of editing a markdown file from the intricacies of the ECS build process.

I'll start preparing the sister PR. When that PR is ready, in my opinion this RFC will be ready to be merged as stage 3. I'm shooting to finalize this next week.

Now's a good time for a last look, @jonathan-buttner @ebeahan @leehinman @willemdh @janniten @rylnd @MikePaquette 

cc @epixa on process adjustment.

webmat:
Merging this stage 3 RFC may be blocked by https://github.com/elastic/ecs/pull/1051. Not sure if this will handle beta field reuse.

ebeahan:
> Merging this stage 3 RFC may be blocked by #1051. Not sure if this will handle beta field reuse.

Ah yes I hadn't considered that need. Will reassess.

dainperkins:
This might be a much longer discussion (re: source/destination vs. client/server event scope) but would it make sense to remove source/destination.user in favor of client/server.user, given the current guidelines for source/destination fields?

webmat:
@dainperkins What guidelines are you referring to? The guidance is that source & destination should be populated as much as possible, as a baseline for network stuff. Then if client/server makes sense, those should be populated too.

This means source/destination.user should also be populated as much as possible as well.

webmat:
@ebeahan Let's have that discussion on #1051 however. There's other things I wonder about 🤔

rfcs/text/0007-multiple-users.md
```diff
@@ -678,11 +678,11 @@ In past discussions and recent research, we have not identified a clear purpose
 for the user fields nested at `host.user.*`.
 
 We are considering deprecating these fields with the intent to remove them completely.
-Please let us know if you disagree with this, and share how you're using them.
 
 #### Resolution
 
-No resolution yet.
+They will be marked as deprecated starting with ECS 1.8, and will be removed in
```
ebeahan, line 684: ++ to proceeding as documented here.

ebeahan (approved):
Everything is looking great here @webmat. 

> Therefore I propose that stage 3 RFCs (and stage 4s that come with field changes) be accompanied by a sister PR that effects the changes, authored by an ECS core team member. Both should be merged at the same time, when both are approved and ready. This decouples the simple RFC process of editing a markdown file from the intricacies of the ECS build process.

No objections to keeping the proposal doc PR separate from the PR that actually implements the proposed changes. Should the sibling PR # should be captured in the proposal markdown prior to merging?

I expect we'd be open to the highly motivated author who wishes to also author the sibling PR. However, I also understand that case to be the rare exception, and the ECS core team is absolutely willing to translate the proposal into field definitions. 😄

dainperkins:
@webmat 
> @dainperkins What guidelines are you referring to?
https://www.elastic.co/guide/en/ecs/1.7/ecs-mapping-network-events.html#_source_and_destination_baseline

Because source / destination can be unidirectional I'm not sure it makes sense to add user information in field sets that, anecdotally anyway, are there for a conglomeration of unidirectional and bidirectional basic network layer info.  

User info is always going to be protocol related which suggests client/server, bidirectional records

Also effective visualization of combined unidirectional and bidirectional network data requires normalization to one or the other... which suggests it might not be worth the additional overhead for repetition of fields that don't seem to fit the mold

webmat:
@dainperkins Yeah it's a bigger discussion if we want to reframe the meaning of source/destination vs client/server. Right now our recommendation is not to interpret them quite this way. This goes beyond the scope of this RFC. I would love it if captured these thoughts in an issue, however.

webmat:
Further note on the process discussion above. We've discussed this out of band, and we all agree to do the decoupling and have two PRs, ideally quite close to one another. But we're not going to enforce that the second one is opened and approved before merging an RFC PR. We could decide to merge the RFC before the second PR exists.

Also the second PR doesn't *have* to be authored by an ECS core team member. If an RFC author is willing to work on that RFC, this is welcome. This was meant more as a statement that whoever the author is, they can depend on the ECS team to take care about the plumbing, if they don't want to.

webmat:
PR to apply these changes is in progress at #1066

webmat:
PR to make these changes come to life is now "feature-complete" https://github.com/elastic/ecs/pull/1066. Contributors familiar with this RFC will have a sense of deja-vu 😂  Some new contents in there as well, though.

This will be our first use of the new mechanism that lets us add a supplemental documentation page for a field set.

willemdh:
> feedback welcome

Well, glad to see this going forward. This should be the base on which UEBA functionalities should be added in Elastic Security. What I'm somehow missing in this whole story is how a username is supposed to be indexed and if there is room for:
- the raw username for the related log
- the ad logon name
- the ad sam account name
- the domain of the related user
- possible related email addresses for the user

I've seen users such as:
- pieter.janssens
- jansspi
- pieterja
- DOMAINX\jansspi
- pieter.janssens@domain.be
- pieter.janssens@alias.be

All of the above 'could' be the same actual user. Imho there should be room for these kind of user info to prevent issues like I'm trying to escalate with host.name shortname and fqdn. It should also be clear from the start in what form user.name should be indexed preferably.

webmat:
@willemdh Good point.

As you point out, the same person can have many different identifiers according to different systems, or even different events types from the same system.

Laying out all of the correlations a UEBA product/feature has to do, in order to help build a precise picture of user identity across data sources is out of scope here.

But I agree it would be helpful to add a section clarifying some subtleties around capturing user identifiers. Here's a few points I could see tackle in there:

* If the only identifier is an email address, which field(s) do we put it in?
* If the identifier includes the domain name, should we parse it, out or leave the domain in, such as `DOMAINX\jansspi`?

Do you have other questions you think should be addressed? Other than these questions that come to mind based on your list, I think the other subtleties are better left to a full-fledged UEBA correlation engine. WDYT?

willemdh:
@webmat

> I think the other subtleties are better left to a full-fledged UEBA correlation engine.

My thought => Multiple NG SIEMs have UEBA functionalities these days, I hope Elastic sees that evolution. There should be a module which allows to index active directory and Elastic should work towards correlating users with the AD index. There should be a Users tab in Kibana Security (Next to Hosts and Network).

But my UEBA dreams do not have to be a show stopper to get this RFC going forward. I think the updated user object has been discussed throughly and is definitely a step in the good direction. If one day more fields are required to get sth UEBA like started, a new object could be created, such as ad_user.* or sth more applicable.

Thanks a lot for all the effort everyone put in this.

ebeahan (dismissed):

webmat:
@willemdh UEBA is floating in the air, but I can't give specifics there.

I will add a section about user identifiers, and try to nudge implementers to avoid problems like you've raised for hostnames. However I'll do this over on https://github.com/elastic/ecs/pull/1066, the PR that makes this proposal take shape in the main ECS documentation.

rfcs/text/0007-multiple-users.md
```diff
@@ -1,8 +1,8 @@
 # 0007: Multiple users in an event
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **2 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **2020-10-02** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Stage: **3 (candidate)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **TBD** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```
webmat, line 5:
```suggestion
- Date: **2020-11-11** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
```

PR #1022, opened 2020-10-08
[1.6] Clarify that file extension should exclude the dot. (#1016)

webmat:
Backports the following commits to 1.6:
* Clarify that file extension should exclude the dot. (#1016)


ebeahan (approved):
LGTM

PR #1024, opened 2020-10-08
[1.x] Add usage docs section (#988)

ebeahan:
 - Add usage docs section (#988)


CHANGELOG.next.md
```diff
@@ -28,6 +28,8 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
+* Added ability to supply free-form usage documentation per fieldset. #988
```
webmat, line 31: Didn't notice earlier, but since we're backporting to 1.7, we should directly add this to the 1.7 changelog.

webmat (approved):
LGTM

CHANGELOG.next.md
ebeahan: Ah yes - thanks! Will adjust.

PR #1032, opened 2020-10-16
Adds container.is_vulnerable

robgil:
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? N (bugged)

# Description
[FedRAMP Guidance](https://www.fedramp.gov/assets/resources/documents/DRAFT_FedRAMP_Vulnerbility_Scanning_Requirements_for_the_Development_and_Use_of_Containers.pdf) is now pushing to maintain an inventory of containers and introduce additional assurance requirements that containers are not Vulnerable.

This field is intended to be used by in conjunction with Vulnerability Scanners (Twistlock, Nessus, etc) in order to log which containers have been identified as containing vulnerable code. 

# Extended Thoughts
Technically this could apply to instances, vms, hosts, and devices as well. Any time an asset is identified as vulnerable. 

Could this be done at a higher level context like [vulnerability](https://github.com/elastic/ecs/blob/master/schemas/vulnerability.yml)? Potentially, but that top level field would need to be re-worked a bit to be intelligent enough to know whether its the container or host that is vulnerable. Simple boolean at the host.* or container.* level would suffice to meet the FedRAMP control. `vulnerability.*` still makes sense to clarify what was found, but it's unclear as to what asset it might associate (host, vm, container, device). 

## make generate bugs
`make generate` fails in a few places on Linux. This may be a Linux vs. Mac issue.

The code I'm referring to is..
```
# Generate Go code from the schema.
.PHONY: gocodegen
gocodegen:
    find code/go/ecs -name '*.go' -not -name 'doc.go' | xargs rm
    cd scripts \
      && $(FORCE_GO_MODULES) go run cmd/gocodegen/gocodegen.go \
            -version=$(VERSION) \
            -schema=../schemas \
            -out=../code/go/ecs
```

The `rm` command is missing an operand. This might work on mac, but not linux. 
```
$ find code/go/ecs -name '*.go' -not -name 'doc.go' | xargs rm
rm: missing operand
Try 'rm --help' for more information.
```

## make setup hang
`make setup` is also hanging on the beats module
```
$ make setup
cd scripts && GO111MODULE=on go mod download
go: finding github.com/elastic/beats v7.0.0-alpha1.0.20181217223741-58573a9f3f15+incompatible
```

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -28,7 +28,7 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
-* Added ability to supply free-form usage documentation per fieldset. #988
+* Added container.is_vulnerable
 
 #### Improvements
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -791,6 +791,14 @@ Note: this field should contain an array of values.
 type: object
 
 
+| extended
+
+// ===============================================================
+
+| container.is_vulnerable
+| Is the container vulnerable
+
+type: boolean
 
 
 
--- a/schemas/container.yml
+++ b/schemas/container.yml
@@ -50,3 +50,10 @@
       object_type: keyword
       description: >
         Image labels.
+
+    - name: is_vulnerable
+      level: extended
+      type: boolean
+      description: >
+        Identified as vulnerable.
+


webmat:
Hey @robgil long time no see :-) 👋 Sorry it took a while to get to your issue.

Thanks for sending his along.

Our CI runs on Linux, so the problem is likely elsewhere 😉 

Based on the `make setup` failure, I'm wondering if you have a very old Beats repo checked out in your Go modules directory. Is that possible? Because this seems to be referencing a 7.0.0 alpha release, which is super old. If that's the case, I'm not much of a Go developer myself, but I think you could either go in there and do a `git pull` or simply delete the directory (if you don't need it).

If the Go beats module thing is not the problem, perhaps `make clean` followed by `make setup` could clean up some old cruft in the ECS build directory.

Then you should have some uncommitted changes to send along.

A few questions about this attribute:

* how/when would this field be populated?
* Is this field intended to be used on all events emitted by a host, or is this for a host inventory index, where you have 1 doc per host?

robgil:
Hey @webmat! Long time indeed! 

Yeah, I was thinking this would be populated by vuln scanning tools and vuln scanning reports. Could also be populated if ECS is being used as a host inventory. I could see this also falling in to an extended schema. 

For FedRAMP, IL4, etc, vuln scans need to be aggregated and reported. Being able to normalize this data across nessus, rapid7, twistlock, OpenSCAP, Sysdig Falco, and so on is really the use case. 

I know we're talking about vulns here, but this also applies to CIS/STIG validation as well. So something like `container.is_compliant` or `container.is_cis`. 

Has anyone else brought this up yet?

ebeahan:
Pinging a few folks who have been recently involved in related efforts - @shimonmodi @peasead @P1llus

peasead:
Hi, @robgil 

Would this be an enrichment then? Enriching `container.*` (or `host.*` for that matter) and `vulnerability.*` documents to populate a boolean `*.is_vulnerable` in an enriched docment?

robgil:
@peasead if you mean "enrichment" as a "user defined field", then no. I think this should either be core or extended schema.

webmat:
I think the gist of the question was more along the lines of: would this attribute be populated on a host entity index (one document per host), or would it be added to all events emitted by that host, via some form of enrichment?

ebeahan:
Ping @cyrille-leclerc @jamiehynds - this might relate to recent Snyk discussions.

djptek:
Hi @robgil - can you merge latest master and then our CI will run this through the updated build process

robgil:
@djptek is this an accepted PR? If so, would be happy to rebase/fix for build/merge to master.

djptek:
Hi @robgil I think there are still some open questions - request to merge was so that it could run through CI (currently showing a fail) and then we could look into those

See conflict around `CHANGELOG.next.md`

ebeahan:
A challenge here is that PR proposes a single field, `container.is_vulnerable`, but there are larger use cases to consider.

@robgil stated it well earlier in the discussion:

> For FedRAMP, IL4, etc, vuln scans need to be aggregated and reported. Being able to normalize this data across nessus, rapid7, twistlock, OpenSCAP, Sysdig Falco, and so on is really the use case.

I'd rather take a step back and approach the larger problem instead of adding a field or two which may or may not adequately address the greater use case (and risk 1. introducing user confusion and/or 2. breaking changes if those fields need to be removed later on).

I for developing a comprehensive approach, the RFC process will be the best direction. But an issue can be a great way to brainstorm more and gather additional feedback first.

Thoughts?

ebeahan:
This would be better discussed either in an issue or through the RFC process to tackle the larger use case.

I'm going to close the PR for now.

PR #1036, opened 2020-10-19
Add TLDR example to USAGE.md and full example in `usage-example/`

webmat:
This is 99% documentation.

There's one small adjustment to the generator.py script, to avoid creating an empty `docs/` directory, since we don't generate it when `--include` or `--subset` are being used.


webmat:
@ebeahan I almost decided to create the directory with the example as a personal gist. But this may not stand the test of time as well.

I wouldn't want this subdirectory to be auto-generated on each build, so right now it's a one off addition. When we make changes relevant to managing custom fields, we can simply run the command in `usage-example/README.md` again and commit the changes. WDYT?

ebeahan (approved):
> There's one small adjustment to the generator.py script, to avoid creating an empty docs/ directory, since we don't generate it when --include or --subset are being used.

👏 

> When we make changes relevant to managing custom fields, we can simply run the command in usage-example/README.md again and commit the changes. WDYT?

👍 - I think that's a fine place to start.

PR #1040, opened 2020-10-20
Add event.category registry

### LINKED ISSUE #1019: Add registry specific event.category value
**Summary**

We do not have a specific category for registry (for example for winlogbeat registry events)

It seems endgame and endpoint are using already an `event.category = registry` value, since it is not listed as an allowed value we might want to consider to add it or either find a different solution and move all our products to use it.

**Motivation**:

We went through https://discuss.elastic.co/t/winlogbeat-sysmon-registry-events-missing-event-category/251282 which pointed the lack of categorisation for winlogbeat sysmon events.

webmat:
This PR adds the allowed value "registry" to `event.category`. It was brought up recently that Endpoint was already using this value.

This value was in previous discussions of categorization. It wasn't included in the initial release of categorization simply because we didn't have data sources that would use it at the time.

[Doc preview](https://ecs_1040.docs-preview.app.elstc.co/diff)

Closes #1019

### DIFF
--- a/docs/field-values.asciidoc
+++ b/docs/field-values.asciidoc
@@ -303,7 +303,7 @@ access, change, end, info, start
 [[ecs-event-category-registry]]
 ==== registry
 
-Use this category of event for events related to the Windows registry.
+Having to do with settings and assets stored in the Windows registry. Use this category to visualize and analyze activity such as registry access and modifications.
 
 
 *Expected event types for category registry:*
--- a/docs/field-values.asciidoc
+++ b/docs/field-values.asciidoc
@@ -144,6 +144,7 @@ that will require subsequent breaking changes.
 * <<ecs-event-category-network,network>>
 * <<ecs-event-category-package,package>>
 * <<ecs-event-category-process,process>>
+* <<ecs-event-category-registry,registry>>
 * <<ecs-event-category-web,web>>
 
 [float]
@@ -296,6 +297,18 @@ Use this category of events to visualize and analyze process-specific informatio
 *Expected event types for category process:*
 
 access, change, end, info, start
+
+
+[float]
+[[ecs-event-category-registry]]
+==== registry
+
+Use this category of event for events related to the Windows registry.
+
+
+*Expected event types for category registry:*
+
+access, change, creation, deletion
 
 
 [float]
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -279,7 +279,8 @@
             - start
         - name: registry
           description: >
-            Use this category of event for events related to the Windows registry.
+            Having to do with settings and assets stored in the Windows registry.
+            Use this category to visualize and analyze activity such as registry access and modifications.
           expected_event_types:
             - access
             - change
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -277,6 +277,14 @@
             - end
             - info
             - start
+        - name: registry
+          description: >
+            Use this category of event for events related to the Windows registry.
+          expected_event_types:
+            - access
+            - change
+            - creation
+            - deletion
         - name: web
           description: >
             Relating to web server access. Use this category to create a dashboard of
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -16,6 +16,8 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
+* Added `event.category` allowed value "registry". #1040
+
 #### Improvements
 
 #### Deprecated
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -1597,7 +1597,7 @@ Note: this field should contain an array of values.
 
 *Important*: The field value must be one of the following:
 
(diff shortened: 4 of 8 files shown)


webmat:
@jonathan-buttner @paulewing @rw-access Folks, this PR adds `event.category` "registry" as an allowed value.

I'd like one of you to confirm whether you agree with the associated list of expected `event.type` values. Does this match what Endpoint is populating?

Examples:
* Are we missing any (e.g. "info"?)
* Are any of them unneeded (e.g. is it possible to detect "access" to Windows Registry keys?)

webmat:
@andrewkroh @marc-gr Hey is Winlogbeat already populating `event.category` "registry", on the Sysmon events?

webmat:
cc @willemdh thanks for raising this in Discuss ❤️

schemas/event.yml
```diff
@@ -277,6 +277,14 @@
             - end
             - info
             - start
+        - name: registry
+          description: >
+            Use this category of event for events related to the Windows registry.
+          expected_event_types:
```
rw-access, line 283: thanks for the add.
I think this seems like a good start. And I would err on the side of omitting `info` like you did. It's always easier to add later than it is to remove.

These values may not be specific enough, because they don't differentiate between key vs value changes. But I think 99% of the time, we're looking at changed values, not keys. So this is okay in my opinion.

rw-access (approved):
thanks @webmat!

schemas/event.yml
```diff
@@ -277,6 +277,14 @@
...
+            - access
```
rw-access, line 284: I don't remember the distinction between access/read, but there is at least one example that comes to mind. Retrieving the permissions or contents of the SAM key is helpful telemetry for endpoint telemetry, so I think this one makes sense, too.

marc-gr:
> @andrewkroh @marc-gr Is Winlogbeat already populating `event.category` "registry" as well?

@webmat not yet iirc

schemas/event.yml
webmat: There's actually no `event.type` value "read". There's only "access". So no distinction to make, just use "access" :-)

schemas/event.yml
webmat: @MikePaquette Did you have a specific use in mind for "info"? This proposal is not yet adding it as an expected value.

One way Auditbeat uses "info" is to dump a list of all running processes, to establish a baseline. Then it reports process start/end as things happen.

I could see a similar use for dumping a set of registry keys (hopefully not the whole hive); then monitor for changes with creation/change/deletion. Was that the idea?

webmat: @rw-access On specificity wrt key vs value activity, I'm not sure we should change the allowed values in `event.type`.

I would think `event.action` can be used to capture whether it's a change to a value or a data blob?

schemas/event.yml
webmat: @rw-access Although to get back to my initial question, I guess I was missing the word "even": is it *even* possible to detect registry key access?

If the registry doesn't offer the ability to be notified of a read or access event, we could remove the value "access".

rw-access: there is an ability to detect registry reads, so I think we can keep **access**.

jonathan-buttner (dismissed):
Thanks for adding Mat! I think the types you have defined make sense 👍

schemas/event.yml
webmat: Perfect, thanks for confirming 🙇

docs/field-values.asciidoc
```diff
@@ -298,6 +299,18 @@ Use this category of events to visualize and analyze process-specific informatio
 access, change, end, info, start
 
 
+[float]
+[[ecs-event-category-registry]]
+==== registry
+
+Having to do with settings and assets stored in the Windows registry. Use this category to visualize and analyze activity such as registry access and modifications.
```
webmat, line 306: @MikePaquette Not word for word our description from way back when, but here's what I'd be going for now.

ebeahan: Comparing this description to the draft in the public ECS categorization, I think limiting the category's intent to solely the Windows registry makes good sense.

ebeahan (dismissed):

docs/field-values.asciidoc
webmat: I don't think there's anything else that has a registry, is there? If I'm wrong on that, then perhaps we should expand that again.

rw-access: there are package registries. I can see how an individual package could make sense in ECS, but not as much for a package registry

ebeahan: Container registries were another one. But no, I don't think we need to expand the category's intent.

webmat: Yes my question wasn't so much about usage of the word registry in other contexts. More around are there other OSes that have a weird settings database/object storage thingy called a registry.

But yeah, being reminded about these other current day uses of the word registry reinforces my resolve on the wording changes :-) So let's go with that

ebeahan (dismissed):
LGTM

ebeahan (approved):

PR #1041, opened 2020-10-20
New default make task that generates main and experimental artifacts.

webmat:
This PR is also changing the order of the 'generate' task: it now starts with the new generator, then runs the legacy scripts.

After this, by default just running `make` will build all artifacts.

During active development, if a faster build is needed, `make generator` builds only the main artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 #
 # Variables
 #
-.DEFAULT_GOAL    := generate
+.DEFAULT_GOAL    := all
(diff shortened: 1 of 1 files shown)


ebeahan (approved):
LGTM 👍

webmat:
This could be backported all the way back to 1.7, if it's helpful for the 1.7 release.

ebeahan:
++ I think that's a reasonable idea.

PR #1045, opened 2020-10-22
Correct experimental usage example

ebeahan:
Additional, documentation was added to `USAGE.md` demonstrating using `--include` with multiple target directories.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -190,7 +190,9 @@ For example, if we defined the following schema definition in a file named `mypr
 Multiple directory targets can also be provided:
 
 ```
-$ python scripts/generator.py --include ../myproject/custom-fields-A/  ../myproject/custom-fields-B --out ../myproject/out/
+$ python scripts/generator.py \
+    --include ../myproject/custom-fields-A/  ../myproject/custom-fields-B \
+    --out ../myproject/out/
 ```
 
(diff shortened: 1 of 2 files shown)


USAGE.md
```diff
@@ -187,6 +187,12 @@ For example, if we defined the following schema definition in a file named `mypr
         Unique identifier of the widget.
 ```
 
+Multiple directory targets can also be provided:
```
webmat, line 190: Ah, I guess I was misunderstanding how to properly pass in multiple directories. I knew the `action=append` thing would make it possible to pass in the same flag multiple times, but I didn't realize we could just pass in space-separated paths as well.

USAGE.md
```diff
@@ -187,6 +187,12 @@ For example, if we defined the following schema definition in a file named `mypr
...
+
+```
+$ python scripts/generator.py --include ../myproject/custom-fields-A/  ../myproject/custom-fields-B --out ../myproject/out/
```
webmat, line 193: For readability I would put one per line here instead, to make sure the proper way to do it stands out

```suggestion
$ python scripts/generator.py \
    --include ../myproject/custom-fields-A/  ../myproject/custom-fields-B \
    --out ../myproject/out/
```

webmat (dismissed):
Nice, thanks for fixing this

webmat (approved):
LGTM

PR #1048, opened 2020-10-26
Change the index pattern in the sample template.

webmat:
Logstash with ECS support creates a template that targets the index pattern `ecs-logstash-*`.

A user experimenting with our sample template (which previously targeted `ecs-*`) and using Logstash in ECS mode on the same cluster would likely have issues. Indeed, both templates would match their index name `ecs-logstash-*`.

I'm changing our sample template's pattern to `try-ecs-*`, which will avoid the above issue. It will also reinforce that when used as is, this template is appropriate for experimentation, more than production use.

I've added a bit of text in these instructions to better explain how to use this template, and where to go if users want to customize that template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,11 @@ All notable changes to this project will be documented in this file based on the
 
 ### Tooling and Artifact Changes
 
+#### Breaking changes
+
+* Changed the index pattern of the sample Elasticsearch template from `ecs-*` to
+  `try-ecs-*` to avoid conflicting with Logstash' `ecs-logstash-*`. #1048
+
 #### Bugfixes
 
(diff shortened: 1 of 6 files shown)


ebeahan (approved):
Good adjustment 👍

webmat:
I'll mark that as a breaking change in the tooling section. WDYT?

Also, I'd backport this to 1.7, so that folks playing with this after the 1.7 release will avoid this problem.

ebeahan:
> I'll mark that as a breaking change in the tooling section. WDYT?

Yes, I think it's a good idea to call attention to this as breaking for anyone who may be consuming the templates directly or generating with the defaults.

I saw the `1.7.0` label and agree, feels sensible to include.

ebeahan (approved):
LGTM 👍

PR #1049, opened 2020-10-26
Add event.category session.

webmat:
Adding this category has been mentioned in the "session" RFC (#935).

I'm not sure we need to wait for the RFC to introduce the category itself, however. This was one of the early categories we were considering, and is already in use in places like Elastic Endpoint.

### DIFF
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -277,6 +277,15 @@
             - end
             - info
             - start
+        - name: session
+          description: >
+            The session category is applied to events and metrics regarding logical persistent connections to hosts and services.
+            Use this category to visualize and analyze interactive or automated persistent connections between assets.
+            Data for this category may come from Windows Event logs, SSH logs, or stateless sessions such as HTTP cookie-based sessions, etc.
+          expected_event_types:
+            - start
+            - end
+            - info
         - name: web
           description: >
             Relating to web server access. Use this category to create a dashboard of
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -10,6 +10,8 @@ Thanks, you're awesome :-) -->
 
 ### Schema Changes
 
+* Added `event.category` "session". #1049
+
 #### Breaking changes
 
 #### Bugfixes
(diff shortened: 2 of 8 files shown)


schemas/event.yml
```diff
@@ -277,6 +277,15 @@
             - end
             - info
             - start
+        - name: session
+          description: >
+            The session category is applied to events and metrics regarding logical persistent connections to hosts and services.
+            Use this category to visualize and analyze interactive or automated persistent connections between assets.
+            Data for this category may come from Windows Event logs, SSH logs, or stateless sessions such as HTTP cookie-based sessions, etc.
+          expected_event_types:
+            - start
```
ebeahan, line 286: Could `creation` and `deletion` have a place here? For instance, web server which is creating or deleting user sessions?

webmat: Interesting question. In thinking of an answer, I've gone back to our definitions for these 4 event types, and I think we could tighten them up accordingly.

I would see that as just the start/end of the session. In most cases these sessions are reported by the server side (ssh, Windows logons).

Here's how I see the distinction between the two pairs:

* Creation/deletion is about something happening to a tangible resource. "change" is also usually relevant to that resource.
* Start/end is about something that is starting, and is only going to last for a certain amount of time.

I could see how there could be a grey area between creation/deletion vs start/end, in that sessions are often backed by resources such as temp files, records such as cookies & stuff. But I think for a session, the time aspect is more important, and the resources backing it are an implementation detail.

So I would think start/end make the most sense.

Am I missing something what you had in mind with creation/deletion here?

ebeahan: I didn't have a firm use case for `creation`/`deletion` either, and I agree it falls into a gray area. As I was thinking through how a web application handle session state, I thought in terms of the app creating sessions at session start and then deleting at session end.

As @webmat described, the backend session management is really just an implementation detail of sessions starting or ending. 👍 for `start`/`end`.

ebeahan (approved):
LGTM

PR #1050, opened 2020-10-26
Add wildcard type support to go code generator

ebeahan:
#### Summary

Add support for the `wildcard` data type for the Go code generator.

#### Why is this change needed?

When `wildcard` appears once or more in the field definitions under `schemas`, the `gocodegen` Makefile directive will cause a `make` run to fail with the exception:

```
cd scripts \
          && GO111MODULE=on go run cmd/gocodegen/gocodegen.go \
                -version=2.0.0-dev \
                -schema=../schemas \
                -out=../code/go/ecs
no translation for wildcard (field executable)
exit status 1
make: *** [gocodegen] Error 1
```

### DIFF
--- a/scripts/cmd/gocodegen/gocodegen.go
+++ b/scripts/cmd/gocodegen/gocodegen.go
@@ -274,7 +274,7 @@ func goDataType(fieldName, elasticsearchDataType string) string {
 	}
 
 	switch elasticsearchDataType {
-	case "keyword", "text", "ip", "geo_point":
+	case "keyword", "wildcard", "text", "ip", "geo_point":
 		return "string"
 	case "long":
 		return "int64"


scripts/cmd/gocodegen/gocodegen.go
```diff
@@ -274,7 +274,7 @@ func goDataType(fieldName, elasticsearchDataType string) string {
 	}
 
 	switch elasticsearchDataType {
-	case "keyword", "text", "ip", "geo_point":
+	case "keyword", "wildcard", "text", "ip", "geo_point":
```
webmat, line 277: I think we should take the opportunity to also add `version` as well, here. WDYT?

ebeahan: Good idea!

Does using `string` work here for the `version` data type or is there a more appropriate type to use?

cc @andrewkroh

ebeahan: And perhaps `constant_keyword` could be another addition here.

andrewkroh: Yeah, string seems like the best option for version. Another option would be a struct that has parameters that match up with the version data type (but I've never seen the version data type in ES so I have no idea what it looks like).

webmat: `version` is another type in the keyword family. So you can safely cram all manners of weird/incorrect version strings in it. But well-formatted ones will be queryable in new interesting ways.

So I think the simple answer is to have it equivalent to a Go string.

I don't object to supporting `constant_keyword` here. But let me open a door for a moment🚪

`constant_keyword` will be tricky to add to "main ECS", because using it means any index not using it correctly will have ingestion screech to a halt when sending a second value for the field.

I'm thinking it will have to be some sort of explicit opt-in, for folks who know they are adopting that specific strategy to splitting indices. Its usage will work much better with composable templates.

But we can close that door 🚪 and add `constant_keyword` as well 😂

ebeahan: Both `version` and `constant_keyword` have been added.

We've had some recent discussions around `version` being a `keyword` specialization type instead of a member of the `keyword` family. If that changes feelings around including it here (for now), I'm ok to remove.

webmat: Let's leave it in for now. I'm not sure if what we saw in the field caps API call is an oversight or if it's here to stay. If it's here to stay, it's likely because the query possibilities are significantly greater with `version` (pun intended).

But the structure of the data itself is best captured as a string in the various programming languages.

webmat (approved):
LGTM

PR #1058, opened 2020-10-28
Add constant_keyword to OSS fallback types list

ebeahan:


webmat (approved):
Thanks for fixing that oversight :-)

PR #1063, opened 2020-10-29
Add --ref support for experimental artifacts

ebeahan:
#### Summary of Changes

When `--ref` argument is set and the `experimental/schemas` directory is loaded with `--include`, these changes ensure the experimental field definitions are pulling from the target git ref and not the local file system.

#### Background

`--ref` is restricted to loading the field definitions from the `schemas` directory. Custom field definition can be supplied but will always load from the local file system. This works well for users managing independent custom definitions outside of the ECS field definitions.

The `experimental/schemas` directory was introduced to hold ECS field definition changes considered experimental, but not beta or GA. Users can use `--include experimental/schemas` if they wish to generate artifacts using these experimental fields.

A problem is introduced if a user wishes to generate artifacts released for a specific version of ECS and include the experimental artifacts from that same version.

#### Example

A user, starting from scratch by cloning the ECS repo and installing dependencies, attempts to generate artifacts for a specific version:

```
$ git clone https://github.com/elastic/ecs
$ cd ecs
$ make ve
$ python scripts/generator.py --ref v1.7.0 --include experimental/schemas --out ../my-project/
```

If the `v1.7.0` exists, `--ref` will properly checkout that git tree and load the field definitions in the `schemas` directory. However, `--include` loads custom field definitions from the user's local file system. The user would have inadvertently merged `v1.7.0` fields with the latest experimental fields in the repo's default branch.

#### Proposed Solution

Add logic in `schema.loader.load_schemas` for when a `--ref` is provided and `--include` includes `experiment/schemas`, the contents of `experiment/schemas` is loaded from the target git ref. Other custom field definitions will continue to be loaded from the local file system.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -277,7 +277,7 @@ It's also possible to combine `--include` and `--subset` together! Do note that 
 
 #### Ref
 
-The `--ref` argument allows for passing a specific `git` tag (e.g. `v.1.5.0`) or commit hash (`1454f8b`) that will be used to build ECS artifacts.
+The `--ref` argument allows for passing a specific `git` tag (e.g. `v1.5.0`) or commit hash (`1454f8b`) that will be used to build ECS artifacts.
 
 ```
 $ python scripts/generator.py --ref v1.5.0
@@ -285,16 +285,10 @@ $ python scripts/generator.py --ref v1.5.0
 
 The `--ref` argument loads field definitions from the specified git reference (branch, tag, etc.) from directories [`./schemas`](./schemas) and [`./experimental/schemas`](./experimental/schemas) (when specified via `--include`).
 
-Here's an example of building artifacts based on ECS `v1.6.0` and adding custom fields from the `myproject` directory:
-
-```
-$ python scripts/generator.py --ref v1.6.0 --include ../myproject/fields/custom`  --out ../myproject/out
-```
-
 Here's another example loading both ECS fields and [experimental](experimental/README.md) changes *from branch "1.7"*, then adds custom fields on top.
 
 ```
-$ python scripts/generator.py --ref v1.7.0 --include experimental/schemas --out ../myproject/out
+$ python scripts/generator.py --ref 1.7 --include experimental/schemas ../myproject/fields/custom --out ../myproject/out
 ```
 
 The command above will produce artifacts based on:
@@ -302,10 +296,6 @@ The command above will produce artifacts based on:
 * main ECS field definitions as of branch 1.7
 * experimental ECS changes as of branch 1.7
 * custom fields in `../myproject/fields/custom` as they are on the filesystem
-
-```
-$ python scripts/generator.py --ref 1.7 --include experimental/schemas ../myproject/fields/custom --out ../myproject/out
-```
 
 > Note: `--ref` does have a dependency on `git` being installed and all expected commits/tags fetched from the ECS upstream repo. This will unlikely be an issue unless you downloaded the ECS as a zip archive from GitHub vs. cloning it.
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -425,7 +425,7 @@ Use cases and examples in this section detail more advanced usage of the ECS too
 
 #### Using ref and include with experimental fields
 
-The `-ref` argument only loads field definitions from the [`./schemas`](./schemas) subdirectory in the target `git` ref. This allows for any custom fieldsets to be merged using `--include`.
+The `--ref` argument only loads field definitions from the [`./schemas`](./schemas) subdirectory in the target `git` reference. This allows for merging custom fieldsets (outside the git ref tree) using `--include`.
 
 Here's an example merging ECS `v1.6.0` fields with custom fields maintained in a separate `myproject` directory:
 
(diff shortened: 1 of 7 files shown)


scripts/schema/loader.py
```diff
@@ -68,13 +77,18 @@ def load_schema_files(files):
...
-            fields_nested = ecs_helpers.safe_merge_dicts(fields_nested, new_fields)
+
+    # Handles case if target dir doesn't exists in git ref
+    if ecs_helpers.path_exists_in_git_tree(tree, target_dir):
+        for blob in tree[target_dir].blobs:
+            if blob.name.endswith('.yml'):
+                new_fields = read_schema_blob(blob, ref)
+                fields_nested = ecs_helpers.safe_merge_dicts(fields_nested, new_fields)
+    else:
+        raise KeyError(f"Target directory './{target_dir}' not present in current git ref!")
```
webmat, line 91: Let's make that error just a tad more verbose :-)

```suggestion
        raise KeyError(f"Target directory './{target_dir}' not present in git ref '{ref}'!")
```

scripts/tests/unit/test_schema_loader.py
```diff
@@ -124,6 +139,43 @@ def test_nest_schema_raises_on_missing_schema_name(self):
         with self.assertRaisesRegex(ValueError, 'incomplete.yml'):
             loader.nest_schema([{'description': 'just a description'}], 'incomplete.yml')
 
+    def test_load_schemas_from_git(self):
```
webmat, line 142: Thanks for adding a high level test for loading from git 👍

webmat (commented):
I like that this is a straightforward fix to this problem, this could have gone in so many different directions.

My spidey sense is a bit leery because behind the scenes we're making `--include` work differently for one specific directory. This might be hard to convey in the Arg Parser description.

Perhaps we can try is to make a quick mention of this on the `--ref` arg?

We definitely need to explain this in the Usage section for `--ref` and `--include`.

After adding these two mentions above, and a suggested tweak to the exception we raise (below), I think we'll be good to move forward.

ebeahan:
> After adding these two mentions above, and a suggested tweak to the exception we raise (below), I think we'll be good to move forward.

Thanks @webmat! I made the suggested adjustments updated the usage documentation and argument descriptions.

After some back and forth, I landed on placing the new usage details underneath `--ref` since I think this is the best logical location.

USAGE.md
```diff
@@ -281,6 +283,26 @@ The `--ref` argument allows for passing a specific `git` tag (e.g. `v.1.5.0`) or
...
+$ python scripts/generator.py --ref v1.6.0 --include ../myproject/fields/custom`  --out ../myproject/out
+```
+
+There one is one exception supporting the [experimental](experimental/README.md) fields. If using `--ref` and including the `experimental/schemas` directory, the experimental fields will also load from the specified git ref. This example would merge ECS v1.7.0 fields with the v1.7.0 experimental fields:
+
+```
+$ python scripts/generator.py --ref v1.7.0 --include experimental/schemas --out ../myproject/out
+```
+
+Custom fields can also be passed using `--include` with the experimental ones. This command merges v1.7.0 fields, v1.7.0 experimental fields, and custom fields from `../myproject/fields/custom`:
```
webmat, line 300: I like this explicit call out, but given the rewritten intro paragraph I'm suggesting above, I would move it below the example, as bullet points.

This way we can spell out what's happening in a bit more details. Perhaps something like this:

```suggestion
The command above will produce artifacts based on:

* main ECS field definitions as of branch 1.7
* experimental ECS changes as of branch 1.7
* custom fields in `../myproject/fields/custom` as they are on the filesystem
```

scripts/generator.py
```diff
@@ -63,7 +63,8 @@ def main():
 
 def argument_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument('--ref', action='store', help='git reference to use when building schemas')
+    parser.add_argument('--ref', action='store', help='Loads fields definitions from `./schemas` subdirectory from specified git reference. \
+                                                       Use with `--include experimental/schemas` to additionally load experimental fields from the target ref.')
```
webmat, line 67: Tough thing to message in this cramped space :-) What about this?

```suggestion
                                                       Note that "--include experimental/schemas" will also respect this git ref.')
```

webmat (commented):
Looking good!

My review is mostly about how we message this tricky situation. Let me know what you think of these suggestions.

The implementation looks good, and thanks for all the tests <3

USAGE.md
```diff
@@ -281,6 +283,26 @@ The `--ref` argument allows for passing a specific `git` tag (e.g. `v.1.5.0`) or
...
+
+```
+$ python scripts/generator.py --ref v1.7.0 --include experimental/schemas ../myproject/fields/custom --out ../myproject/out
```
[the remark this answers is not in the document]
ebeahan: Nice - good substitution 👍

ebeahan:
Thanks @webmat - I've your suggestions.

After all the rephrasing/rewording, the exact lines you suggested removing shifted quite a bit. Let me know if the latest revisions weren't quite what you envisioned.

webmat (approved):
Thanks for the adjustments! LGTM

PR #1066, opened 2020-10-30
Changes related to merging RFC 0007 at stage 3

webmat:
Effects the changes from #1017. To be merged at the same time, or after.

[Preview the new user "usage" page](https://ecs_1066.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-user-usage.html):

TODO

- [x] Categorization section
- [x] Notes about reuse (as a different name) of user fields within user. Namely that these don't get carried around when nesting user fields elsewhere.
- [x] Add a section about deprecating usage of `host.user.*`
- [x] I'm also considering adding a TOC at the top of this page, to give folks a sense of the sections contained within this humongous page.
- [x] Wait for RFC stage 3 to be merged #1017
- [x] Identify these new field reuses as beta
  - [x] Wait for #1051  to be merged

**Questions**

- [x] Towards the end of this page, I link to the RFC in the ECS repo, for folks who want to see mapping examples. I'm trying to avoid making this page 2x longer still. But will users (not ECS contributors) be lost, when they land in the RFC?
- [ ] Out of scope for this PR, but I think that the listing of user fields [in the field reuse section](https://ecs_1066.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-user.html#ecs-user-nestings) is a bit off. Perhaps we should mention that these are actually the user fields (right now it's designed to list other field sets nested here), or perhaps we should redesign this a bit. Note that this also affects the `process.parent` reuse.


webmat:
This is not the only thing in this PR, but here's the [preview of the new user usage page](https://ecs_1066.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-user-usage.html)

webmat:
This is ready for a review.

The "subtleties" section makes me want to cry a little 😢 😂

ebeahan:
Responding to the open questions:

> Towards the end of this page, I link to the RFC in the ECS repo, for folks who want to see mapping examples. I'm trying to avoid making this page 2x longer still. But will users (not ECS contributors) be lost, when they land in the RFC?

I agree it could cause some confusion, but I think the confusion would be minor. Perhaps we can adjust the wording in the link from "RFC" to something more general like "feature proposal", for users not familiar with an RFC document?

The concern over page length is understandable, but I can also see the mappings being a useful reference for users. With the sample mappings external to the docs, it might make discoverability more challenging (an extra click, not searchable from the docs site).

> Out of scope for this PR, but I think that the listing of user fields in the field reuse section is a bit off. Perhaps we should mention that these are actually the user fields (right now it's designed to list other field sets nested here), or perhaps we should redesign this a bit. Note that this also affects the process.parent reuse.

I made this observation too recently. In #1051, I'm proposing introducing a `beta` attribute to the reusable `expected` object. Perhaps we can also consider adding a secondary description attribute as well?

ebeahan:
> The "subtleties" section makes me want to cry a little

Yes this is a challenging area to detail!

I don't think it's necessary for this PR, but in the future perhaps we can add some sort of visual as another way of explaining the concepts.

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+[discrete]
+[[ecs-user-usage-user-at-root]]
+====== User fields at the Root of an Event
+
+The user fields at the root of an event are used to capture the user
+performing the main action described by the event. This is especially important
+when there's more than one user present on the event. `user.*` fields at the root
+of the event represent the user performing the action.
+
+In many cases, events that only mention one user should populate the user fields
```
ebeahan, line 121: A high-level summary of the guidance covered could be a useful companion quick-reference to the more detailed explanations. I could see it both early in the doc as well as towards the end as a summary.

Examples of quick guidance:

* Always populate `user.*` in addition to more purpose-specific fields when one or more users are present in an event. If more than one user, the user performing the main action populates `user.*`. 
* Any users present in the event should also be captured in `related.user`
* etc. etc.

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+    "user": {
+      "name": "deus"
+    },
+    "ip": "10.42.42.43"
+  },
+  "related": { "user": ["alice", "deus"] }
+}
+-----------
+
+Whenever an event source populates the `client` and `server` fields in addition
```
ebeahan, line 177: I could see a second example which builds on the first, showing both `client`/`server` + `source`/`destination` fields all populated in the single event. WDYT?

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+* A user changing identity on a host.
+** Examples: sudo, su, Run as.
+* Running a program as a different user. Examples:
+** A trusted user runs a specific admin command as root via a mechanism such as the Posix setuid/setgid.
+** A service manager with administrator privileges starts child processes as limited
+    users, for security purposes (e.g. root runs Apache HTTPD as user "apache")
+
+In cases where the event source only gives information about the effective user
+and not who requested different privileges, the `user` fields at the root of the
+event should be used instead.
```
ebeahan, line 202:
```suggestion
event should be used instead of `user.effective`.
```
nit: Makes explicitly clear to use `user` over `user.effective` in this case.

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+
+[source,json]
+-----------
+{
+  "user": {
+    "name": "alice",
+    "id": "1001",
+    "effective": {
+      "name": "root",
+      "id": "1"
```
ebeahan, line 214: Not related to the review at all, but seeing `root` and `id` value of 1 got me off down the rabbit hole.

It has absolutely no bearing; I'm just hardwired to expect a UID/GID of zero. 😂

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+[discrete]
+[[ecs-user-usage-combining]]
+====== Combining IAM and Privilege Change
+
+We've covered above how `user.target` and `user.changes` can be used at the same time.
+If privilege escalation is also present in the same IAM event, `user.effective`
+should of course be used as well.
+
+Here's the "rename" example from the IAM section above. In the following example,
+we know "alice" is escalating privileges as "root", in order to modify user "bob":
```
ebeahan, line 299:
```suggestion
we know "alice" is escalating privileges to "root", in order to modify user "bob":
```

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+Two examples of this are:
+
+* reusing `group` in `user`, resulting in the `user.group.*` fields, or
+* reusing `user` in `destination`, resulting in the `destination.user.*` fields,
+  which also include `destination.user.group.*`.
+
+The `user` fields can also be reused within `user` as different names,
+representing the role of each relevant user. Examples are the `user.target.*` or `user.effective.*` fields.
+
+It's important to note, that contrary to the `group` fields,
```
ebeahan, line 337: Can `group` be reused where `user` is a nesting?

For example, would `user.effective.group.*` be valid?

docs/usage/user.asciidoc
```diff
@@ -0,0 +1,396 @@
...
+pivoting. For example, if you have a suspicion about user "bob.barker", searching
+for this name in `related.user` will give you all events related to this user, whether
+it's the creation / rename of the user, or events where this user was active in a system.
+
+[discrete]
+[[ecs-user-usage-mappings]]
+===== Mapping Examples
+
+For examples of mapping events from various sources, you can look at
+https://github.com/elastic/ecs/blob/master/rfcs/text/0007-multiple-users.md#source-data[RFC 0007 in section Source Data].
```
ebeahan, line 389: It's neither in-scope nor a blocker for this PR, but I'd like to add project-wide ECS attributes to our docs for commonly (or potentially common) reused strings such as the ECS repo GitHub URL.

We will appreciate it when/if we need to update the path to only have the single place 😂

ebeahan (commented):
I added a handful of minor notes, but I think overall this is great @webmat! Having a consolidated reference is going to be a great resource!

I'll hold off on giving the final 👍  until we have the beta labeling support from #1051 finalized and the appropriate changes incorporated here.

docs/usage/user.asciidoc
webmat: Oops yeah, the idea isn't to troll anyone with this 😂  Got confused with PID 1, here I think. I'll fix

docs/usage/user.asciidoc
webmat: Yes, but only if it's relevant to the privilege escalation / change / target user.

- [ ] I'll rephrase to explicitly give an example of how group is carried everywhere (including in a self-reuse), and how the self-reuses don't.

docs/usage/user.asciidoc
webmat: Yeah very good idea. It will be useful when we tackle inclusive naming.

docs/usage/user.asciidoc
webmat: Good idea!

docs/usage/user.asciidoc
webmat: Here I think I would err on the side of "terseness", if we can call such a long page terse. If the idea is to reinforce this pattern, perhaps we can instead link to the page on mapping network events?

docs/usage/user.asciidoc
webmat: WDYT about this new presentation? https://ecs_1066.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-user-usage.html#ecs-user-usage-reuse-subtleties

webmat:

docs/field-details.asciidoc
```diff
@@ -6604,14 +6604,16 @@ Note also that the `user` fields may be used directly at the root of the events.
 // ===============================================================
 
 
-| <<ecs-user,user.changes.*>>
-| Fields to describe the user relevant to the event.
+| <<ecs-user,user.changes.*>>| beta:[ Reusing the user fields in this location is currently considered beta.]
+
+Fields to describe the user relevant to the event.
```
ebeahan, line 6609: > Out of scope for this PR, but I think that the listing of user fields in the field reuse section is a bit off. Perhaps we should mention that these are actually the user fields (right now it's designed to list other field sets nested here), or perhaps we should redesign this a bit. Note that this also affects the process.parent reuse.

I agree that the wording is a bit off for these reuses.

ebeahan (commented):
The adjustments are looking great!

> Towards the end of this page, I link to the RFC in the ECS repo, for folks who want to see mapping examples. I'm trying to avoid making this page 2x longer still. But will users (not ECS contributors) be lost, when they land in the RFC?

I think linking to the RFC is acceptable for curious, interested users.

I responded to the second question in my notes below.

docs/usage/user.asciidoc
```diff
@@ -337,21 +337,38 @@ we know "alice" is escalating privileges to "root", in order to modify user "bob
...
+Most cases of field reuse in ECS are reusing a field set inside a different field set.
 Two examples of this are:
 
 * reusing `group` in `user`, resulting in the `user.group.*` fields, or
 * reusing `user` in `destination`, resulting in the `destination.user.*` fields,
   which also include `destination.user.group.*`.
 
 The `user` fields can also be reused within `user` as different names,
-representing the role of each relevant user. Examples are the `user.target.*` or `user.effective.*` fields.
+representing the role of each relevant user. Examples are the `user.target.*` or `user.effective.*` fields. However it's important to note that `user` fields reused within
```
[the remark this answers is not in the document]
webmat: Yeah good idea!

docs/usage/user.asciidoc
webmat: I feel like this is not a blocker to this PR. WDYT?

Perhaps we could note down some of the points for this quick guidance in a separate issue, and address separately?

ebeahan: Agreed, it's not a blocker here. We can capture and address later on 😄 .

docs/usage/user.asciidoc
webmat: I've added a link at the end of this paragraph

docs/field-details.asciidoc
ebeahan: Captured in #1174 to address in the future.

ebeahan (approved):
Thanks for the iterations here! LGTM 👍

PR #1068, opened 2020-11-02
[1.x] Change the index pattern in the sample template.

webmat:

* Change the index pattern in the sample template. (#1048)


ebeahan (approved):
LGTM 👍

PR #1070, opened 2020-11-03
[RFC] Service Environment Field - stage 1

cyrille-leclerc:
Improvement of the RFC for the Service Environment Field

[RFC Preview](https://github.com/elastic/ecs/blob/service-environment-field-rfc-improvement/rfcs/text/0002-rfc-environment.md)

### DIFF
--- a/rfcs/text/0002-rfc-environment.md
+++ b/rfcs/text/0002-rfc-environment.md
@@ -381,3 +381,13 @@ Stage 4: Identify at least one real-world, production-ready implementation that 
 * ? | sponsor
 * @exekias | subject matter expert
 * ? | grammar, spelling, prose
+
+## References	
+
+<!-- Insert any links appropriate to this RFC in this section. -->	
+
+### RFC Pull Requests	
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->	
+
+* Stage 0: https://github.com/elastic/ecs/pull/891
--- a/rfcs/text/0002-rfc-environment.md
+++ b/rfcs/text/0002-rfc-environment.md
@@ -123,36 +123,16 @@ Note the support in Heartbeat of `service.name` is waiting for https://github.co
 Note that all the Beats come with a different example field to define the environment: `fields.env` (see filebeat.yml, metricbeat.yml, auditbeat.yml or heartbeat.yml).
 This `fields.env` field is misaligned with this desire of standardisation because it cannot be standardised in ECS due to the following ECS rules:
 * The namespace `fields.*` is not accepted in ECS, this namespace is dedicated to non standardised fields.
-* ECs don't use abbreviations and `env` is the abbreviation of `environment`.
+* ECS don't use abbreviations and `env` is the abbreviation of `environment`.
 
 Note that the `fields.env` field is just an example and it not likely to be used very broadly because for infrastructure elements such as servers, the delineation between environments is often difficult to establish as servers are frequently simulatenaously running production and non production workloads.
 
-We propose to tackle the question of standardisation environment for infrastructure later.
-
-
-We don't need to evolve Filebeat and Metricbeat configurations initially.
-
-Note: we may want to improve the configuration of Filebeat and Metricbeat later down the road to ease the specification `service.environment` for specific files or metrics that can be associated with a specific service/application.
-Today, such configuration is cumbersome requiring to specify a "processor". 
-
-Example of verbose Filebeat configuration today to define `service.name=www-frontend` and `service.environment=production`:
-
-```
----
-filebeat.inputs:
-  - type: log
-    enabled: true
-    paths:
-      - /var/log/www-frontent/www-frontend.log
-processors:
-  - add_fields:
-      when:
-        ... todo find the syntax to filter on file path
-      target: 'service'
-      fields:
-        name: www-frontend
-        environment: production
-```
+We propose to tackle later the questions of
+* Standardisating the characterization of the environment for infrastructure later.
+* Specifying the `service.environment` and service.name for service/application related 
+   * Log files collected by filebeat: application log file collected on disk...
+   * Metrics collected by metricbeat: application metrics collected via Prometheus, Jolokia, JMX...
+For the metricbeat and filebeat problems described above, the solution used by Heartbeat HTTP monitors (see above) is an interesting path forward.
 
 ## Source data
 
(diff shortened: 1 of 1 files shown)


rfcs/text/0002-rfc-environment.md
```diff
@@ -47,73 +49,118 @@ Stage 2: Include new or updated yml field definitions for all of the essential f
...
+    environment: staging
+  urls: ["https://www.staging.my-ecommerce-company.com/status"]
+  schedule: '@every 10s'
+  check.response.status: 200
+  timeout: 5s
+``` 
+
+Note the support in Heartbeat of `service.name` is waiting for https://github.com/elastic/beats/pull/20330
+
+### Usage with Filebeat and Metricbeat
```
webmat, line 121: This section spends a lot of time on what's effectively an example of how to use the "fields" Beats config option. I don't think anyone is actually pushing for this to be adopted, so I suggest we don't mention it at all in this section, but we flesh out this point in the "concerns" section.

So I think we could rewrite this section with a flow somewhat like this:

* contrary to APM and Hearbeat, Filebeat and Metricbeat have very little semantic knowledge out of the box about services, so they rarely populate these fields.
* However in cases where Fb/Mb are monitoring services that are related to services also monitored by APM/Heartbeat, users may benefit from correlating them.
* The way to do so is to populate `service.*` fields explicitly in the relevant Fb/Mb events.

WDYT? Does this follow the intent of the addition of this field?

rfcs/text/0002-rfc-environment.md
```diff
@@ -47,73 +49,118 @@ Stage 2: Include new or updated yml field definitions for all of the essential f
...
       target: 'service'
       fields:
-        environment: staging
+        name: www-frontend
+        environment: production
 ```
 
-
-
 ## Source data
```
webmat, line 157: Since this proposal is about correlating events from different sources via the service fields and the proposed `service.environment` field, I think we should use this section to give relevant event snippets that demonstrate this.

Could you grab one or a few sample APM events, and a sample Heartbeat event in JSON format? Then we can adjust their `service` fields to fit with the example you've been building in the "Usage" section.

And finally to demonstrate the correlation with Metricbeat or Filebeat, we could do the same with one of their events and also adjust its `service` fields to demonstrate how each event across the stack could be correlated. If needed I can help you gather a Filebeat example.

rfcs/text/0002-rfc-environment.md
```diff
@@ -138,29 +185,27 @@ The goal here is to research and understand the impact of these changes on users
 
 ## Concerns
 
- * Beats (Filebeat, Metricbeat, Heartbeat...) document in their configuration files (filebeat.yml...) an alternate non standardizable field `fields.env` illustrated with the demo value `staging`.
+<!--
+Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
+-->
+
+ * Beats (Filebeat, Metricbeat, Heartbeat...) document in their configuration files (filebeat.yml, metricbeat.yml...).
```
webmat, line 192: A format that's been working well in other RFCs is to capture concerns in a format of concern + resolution.

Since some of the concerns are complex (multi-line, code snippets, etc) I don't think the bullet points work well. Perhaps you could separate each concern with a `###` heading instead? This way you're free to have paragraphs and examples as needed. You can check out how I've done it [in this RFC](https://github.com/elastic/ecs/blob/master/rfcs/text/0007-multiple-users.md#concerns) for an example.

rfcs/text/0002-rfc-environment.md
```diff
@@ -180,72 +225,11 @@ Stage 4: Document any new concerns and their resolution. The goal here is to eli
...
-
-```
 
 ## People
 
 
 * @cyrille-leclerc | author
 * ? | sponsor
-* @roncohen | subject matter expert
+* @carlos | subject matter expert
```
webmat, line 234: Please use Carlo's Github handle @exekias here :-)

rfcs/text/0002-rfc-environment.md
```diff
@@ -180,72 +225,11 @@ Stage 4: Document any new concerns and their resolution. The goal here is to eli
...
-
-## References
-
-<!-- Insert any links appropriate to this RFC in this section. -->
-
-### RFC Pull Requests
-
-<!-- An RFC should link to the PRs for each of it stage advancements. -->
-
-* Stage 0: https://github.com/elastic/ecs/pull/891
```
webmat, line 246: Please add back that "References" section.

Note also that we'll want to add the link to this PR in the "RFC Pull Requests" section, below the "stage 0" entry: 

```
* Stage 1: https://github.com/elastic/ecs/pull/1070
```

rfcs/text/0002-rfc-environment.md
```diff
@@ -138,29 +185,27 @@ The goal here is to research and understand the impact of these changes on users
...
 
- * This topic has already been discussed multiple times
-    * https://github.com/elastic/ecs/issues/268 Add new top level field "environment" #268
-    * https://github.com/elastic/ecs/issues/704 New field: organization.environment #704
-    * https://github.com/elastic/ecs/issues/143 agent.environment and service.environment #143
 
-<!--
-Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
--->
+ * OpenTelemtry has standardised `deployment.environment`, refering to [Wikipedia: Deployment Environment](https://en.wikipedia.org/wiki/Deployment_environment). The benefit of `deployment.environment` is that it works better for the characterization of the infrastructure (e.g. physical server, vm): https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/deployment_environment.md
```
webmat, line 207: A few typos 👮‍♂️ 

```suggestion
 * OpenTelemetry has standardized `deployment.environment`, referring to [Wikipedia: Deployment Environment](https://en.wikipedia.org/wiki/Deployment_environment). The benefit of `deployment.environment` is that it works better for the characterization of the infrastructure (e.g. physical server, vm): https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/deployment_environment.md
```

webmat (commented):
Thanks for opening this stage 1 PR, @cyrille-leclerc. Here's some preliminary feedback.

I like that you're keeping the focus on `service.environment`, after all 👍  I still think it's worth going through the RFC process for this though, since this is meant for correlation across so many different parts of our product.

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: I have updated the RFC accordingly.

rfcs/text/0002-rfc-environment.md
```diff
@@ -47,73 +49,118 @@ Stage 2: Include new or updated yml field definitions for all of the essential f
...
     enabled: true
-    json.keys_under_root: true
-    json.overwrite_keys: true
     paths:
-      - /usr/local/var/log/my-shopping-cart/anti-fraud.log
+      - /var/log/www-frontent/www-frontend.log
 processors:
   - add_fields:
+      when:
+        ... todo find the syntax to filter on file path
```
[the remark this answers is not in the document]
cyrille-leclerc: problem solved.

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: Problem solved

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: Problem solved

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: Problem solved

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: Problem solved.

cyrille-leclerc:
@webmat I have updated the RFC. Thank you very much for your review.

ebeahan (commented):
Thanks for the revisions @cyrille-leclerc! This is shaping up well.

I believe the one outstanding item is identifying a sponsor. Has a sponsor been considered?

rfcs/text/0002-rfc-environment.md
webmat: Can you add the link to the stage 1 PR as well? 🙏

rfcs/text/0002-rfc-environment.md
```diff
@@ -138,29 +327,33 @@ The goal here is to research and understand the impact of these changes on users
...
-    * Most events can eventually be attached to a `service`
+### "Service" namespace not suited to characterise infrastructure component
 
- * This topic has already been discussed multiple times
-    * https://github.com/elastic/ecs/issues/268 Add new top level field "environment" #268
-    * https://github.com/elastic/ecs/issues/704 New field: organization.environment #704
-    * https://github.com/elastic/ecs/issues/143 agent.environment and service.environment #143
+Using the namespace `service` for `service.environment` could sound awkward to some low level messages (e.g. system events). The rationale is that `service.environment` will only be used to qualify components of the application layer (ie services, application, function) but will not be used to qualify the environment of an infrastructure component (server, virtual machine...).
+
+Note that the `environment` field is a good candidate to be reused in other namespace than the `service.*` to cover the Infrastructure use cases.
```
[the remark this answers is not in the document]
cyrille-leclerc: Good catch on updating the `tags` fields.

rfcs/text/0002-rfc-environment.md
cyrille-leclerc: recursive link :-) I didn't think about it in this case but you are right :-)

cyrille-leclerc:
@ebeahan I have identified the sponsor. The subject matter expert will be @sqren from the APM team and the sponsor will be @exekias from the metrics / infrastructure monitoring team.

sqren (approved):

sqren (approved):
lgtm. But I'll leave the final review to @webmat and @ebeahan

ebeahan (dismissed):
Thanks, @cyrille-leclerc, for the adjustments and iteration!

I'll set the advancement date and merge.

Issue #1039, opened 2020-10-20
Extend file.pe


peasead:
**Summary**

The Portable Executable (PE) sub-field, of the `file` top-level fieldset, can be updated to include more file attributes to aid in malware analysis.

Because this is extending an existing sub-field vs. creating a top-level fieldset, I wanted to make an Issue first to determine if this required an RFC.

**Motivation**:

In creating a VirusTotal Filebeat module, we identified opportunities to extend the `file.*` top-level fieldset.

**Detailed Design**:

```
- name: file.pe
  type: group
  fields:
    - name: flattened
      release: beta
      type: group
      fields:
        - name: debug
          default_field: false
          description: >
            Debug information, if present
          type: flattened
        - name: import_list
          default_field: false
          description: >
            List of all imported functions
          type: flattened
        - name: sections
          default_field: false
          description: >
            Data about sections of compiled binary PE
          type: flattened
        - name: resource_details
          default_field: false
          type: flattened
          description: >
            If the PE contains resources, some info about them
        - name: resource_languages
          default_field: false
          type: flattened
          description: >
            Digest of languages found in resources. Key is language (as string) and
            value is how many resources there are having that language (as integer)
        - name: resource_types
          default_field: false
          type: flattened
          description: >
            Digest of resource types. Key is resource type (as string) and value
            is how many resources there are of that specific type (as integer)
        - name: packers
          default_field: false
          type: flattened
          description: >
            Identifies packers used on Windows PE files by several tools and AVs.
            Keys are tool names and values are identified packers, both strings.
            see `file.pe.packers` for merged list of packers from all tools.
      overwrite: true
    - name: exports
      type: keyword
      description: >
        List of symbols exported by PE

    - name: creation_date
      default_field: false
      description: >
        extracted when possible from the file's metadata. Indicates when it was
        built or compiled. It can also be faked by malware creators.
      type: date
    - name: authentihash
      description: >
        Authentihash of the PE file.
      type: keyword
    - name: compile_timestamp
      description: >
        Compile timestamp of the PE file.
      type: date

    - name: packers
      type: keyword
      description: >
        Merged list of all detected packers by all tools used. See `file.pe.flattened.packers`.

    - name: main_icon.hash
      type: group
      description: >
        Hashes of embedded program icon
      fields:
        - name: dhash
          description: >
            Difference Hash for a given PE file.
          type: keyword
          release: beta
        - name: md5
          type: keyword
          description: >
            MD5 hash of raw icon data
[...]

PR #1071, opened 2020-11-03
Extend file.pe Fieldset

peasead:
## Issue

Resolves https://github.com/elastic/ecs/issues/1039

- If submitting code/script changes, have you verified all tests pass locally using `make test`? Yes
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? Yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? Yes

[RFC Preview](https://github.com/peasead/ecs/blob/file.pe-extend/rfcs/text/0000-extend-file-pe.md)


CHANGELOG.next.md
```diff
@@ -21,6 +21,8 @@ Thanks, you're awesome :-) -->
 
 #### Improvements
 
+* Extended `file.pe`. #1071
+
```
webmat, line 25: We can remove that. During the RFC process there's no official changes to advertize

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,226 @@
...
+        transformations have occurred, which would change more traditional hash values.
+
+        Learn more at https://www.fireeye.com/blog/threat-research/2014/01/tracking-malware-import-hashing.html.
+      example: 0c6803c4e922103c4dca5963aad36ddf
+
+    - name: architecture
+      level: extended
+      type: keyword
+      description: CPU architecture target for the file.
+      example: x64
```
webmat, line 61: All of these fields are already in `schemas/pe.yml`, so we can remove them from this file.

This file is either supplemental fields, or we can also do overrides to existing fields when appropriate. But this RFC is only for new fields, so keep the `- fields:` above, and delete all the way to the `architecture` field.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,226 @@
...
+  group: 2
+  description: These fields contain Windows Portable Executable (PE) metadata.
+  type: group
+  reusable:
+    top_level: false
+    expected:
+      - file
+      - dll
+      - process
+  fields:
```
webmat, line 13: Since this file is for supplemental changes, we can omit most of this. Let's just make sure the structural details are in place, and get rid of the rest, like this:

```suggestion
- name: pe
  type: group
  fields:
```

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
+# 0000: Extend the PE subfield of the File fieldset
+
+- Stage: **0 (strawperson)**
```
webmat, line 3: As we've discussed today, we can already target stage 1 here as well

```suggestion
- Stage: **1 (proposal)**
```

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/1071
```
webmat, line 1071:
```suggestion
* Stage 1: https://github.com/elastic/ecs/pull/1071
```

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+
+## People
+
+The following are the people that consulted on the contents of this RFC.
+
+* @peasead | author
+* @devonakerr | sponsor
+* @dcode, @peasead | subject matter expert
+
+## References
```
webmat, line 1063: All the links are fine right in the body of the text above, but please make sure to list them all in the References section as well

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
+# 0000: Extend the PE subfield of the File fieldset
```
webmat, line 1: The "PE" field set actually appears under file, dll and process.

So we can simply call this:

```suggestion
# 0000: Extend the PE field set
```

or similar

webmat (commented):
This is looking really good.

For stage 1 we don't need to have all fields hashed out. So let's discuss them a bit, but if any of the discussions on the field end up being thorny, we can capture these concerns in the RFC and this shouldn't be a blocker for stage 1.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+
+## Fields
+
+**Stage 0**
+
+This RFC is to create 25 additional sub-fields within the `file.pe` fieldset.
+
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| file.pe.authentihash | keyword | Authentihash of the PE file. |
```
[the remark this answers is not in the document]
webmat: Forgot to add the examples, brb 🏃

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,168 @@
...
+      description: >
+        List of symbols exported by PE
+
+    - name: creation_date
+      level: extended
+      short: Build or compile date.
+      description: >
+        Extracted when possible from the file's metadata. Indicates when it was
+        built or compiled. It can also be faked by malware creators.
+      type: date
```
webmat, line 82: When adding examples here, it's as simple as adding another attribute called `example:`.

```suggestion
      type: date
      example: "2020-11-05T17:25:47.000Z"
```

If the example is a composite value like an array or a JSON blob, make sure you quote it.

If the example is a string, quoting is optional.

If the example is a float or int, don't quote it.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+| file.pe.compile_timestamp | date | Compile timestamp of the PE file. |
+| file.pe.compiler_product_versions | keyword | Version of the compiler. |
```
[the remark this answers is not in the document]
dcode: here's a sample of what is in there today, as returned by VT. I'm not a PE expert, so this can probably get cleaned up, but it presumably means something to somebody. @rw-access, @Samirbous  do you have any input here?

```json
        "compiler_product_versions": [
          "id: 259, version: 27412 count=18",
          "id: 253, version: 23601 count=2",
          "id: 260, version: 27412 count=36",
          "id: 261, version: 27412 count=208",
          "id: 262, version: 27412 count=1",
          "id: 257, version: 27412 count=35",
          "[---] Unmarked objects count=841",
          "id: 260, version: 28920 count=18",
          "id: 259, version: 28920 count=11",
          "id: 261, version: 28920 count=466",
          "id: 261, version: 29111 count=15",
          "id: 255, version: 29111 count=1",
          "id: 151, version: 0 count=1",
          "id: 258, version: 29111 count=1"
        ]
```

another, more meaningful example:

```json
        "compiler_product_versions": [
          "[IMP] VS2008 SP1 build 30729 count=3",
          "[---] Unmarked objects count=10",
          "id: 174, version: 40219 count=11",
          "[LNK] VS2010 SP1 build 40219 count=1"
        ]
```

rfcs/text/0000-extend-file-pe.md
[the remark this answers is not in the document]
peasead: Thanks @webmat 

I'll get those filled in.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+| file.pe.sections | flattened | Data about sections of compiled binary PE |
+| file.pe.resource_details | flattened | If the PE contains resources, some info about them |
+| file.pe.resource_languages | flattened | Digest of languages found in resources. Key is language (as string) and value is how many resources there are having that language (as integer) |
+| file.pe.resource_types | flattened | Digest of resource types. Key is resource type (as string) and value is how many resources there are of that specific type (as integer) |
+| file.pe.packers | flattened | Identifies packers used on Windows PE files by several tools and AVs. Keys are tool names and values are identified packers, both strings. See `file.pe.packers` for merged list of packers from all tools. |
+| file.pe.machine_type | keyword | Machine type of the PE file. |
+| file.pe.main_icon.hash.dhash | keyword | Difference Hash for a given PE file. |
+| file.pe.main_icon.hash.md5 | keyword | MD5 hash of raw icon data |
+| file.pe.overlay.chi2 | float | Chi2 information of the PE file. |
+| file.pe.overlay.entropy | float | Entropy information of the PE file. |
```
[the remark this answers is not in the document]
dcode: It is calculated. In this case calculated by VT, but many PE analyzers also do this. The chi2 calculation is also useful, as it provides a sense of frequency distribution of data.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| file.pe.authentihash | keyword | Authentihash of the PE file. |
+| file.pe.compile_timestamp | date | Compile timestamp of the PE file. |
+| file.pe.compiler_product_versions | keyword | Version of the compiler. |
+| file.pe.creation_date | date | Extracted when possible from the file's metadata. Indicates when it was built or compiled. It can also be faked by malware creators. |
+| file.pe.entry_point | long | Entry point of the PE file. |
+| file.pe.exports | keyword | List of symbols exported by PE |
+| file.pe.debug | flattened | Debug information, if present |
+| file.pe.import_list | flattened | List of all imported functions |
```
dcode, line 23: I'd really like to present a common interface to imports across PE, MachO, and ELF. A project called LIEF does just this with respect to symbols and a base [abstract symbol](https://lief.quarkslab.com/doc/latest/api/python/abstract.html#symbol) definition. Basically, I propose a superset of what we have today from each of these three types from VT. Tools will populate the objects with information available and `file.*.imported_symbols` would be a `nested` list of objects with: `name`, `type`, `library_name` at a minimum. These 3 fields would capture everything VT provides today, but a lot of other frameworks (including LIEF) provide additional info, such as library offset, etc. This could offer a good future expansion of the fieldset. 

See additional discussion here: https://github.com/elastic/beats/pull/21815#discussion_r515411414

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+| file.pe.sections | flattened | Data about sections of compiled binary PE |
```
dcode, line 24: Similiar to the imports, sections are common to all executable/library types. Moving this data out of a flattened field gives us more usefulness of it.

Additional discussion: https://github.com/elastic/beats/pull/21815#discussion_r515410709
LIEF Abstract definition for comparision: https://lief.quarkslab.com/doc/latest/api/python/abstract.html#section

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,1076 @@
...
+-->
+
+**Stage 2**
+
+### VirusTotal Filebeat Module
+
+```json
+# spellchecker: disable
+- name: virustotal
+  description: >
```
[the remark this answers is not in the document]
peasead: Yeah, I think I'll remove this for this stage right now and re-add them as appropriate once the VT module gets solidified.

rfcs/text/0000-extend-file-pe.md
peasead: Is this how you'd like to present this; or do you want to extend the values into nested fields? I can adjust that table (which may be a bit stale).
```
    - name: sections
      level: extended
      description: >
        Data about sections of compiled binary PE
      type: keyword
      example: { "chi2" : 3027194, "virtual_address": 8192, "entropy": 6.24, "flags": "rx", "name": ".text", "raw_size": 198144 }
```

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      example: { "chi2": -1, "filetype": "English text", "entropy": 0, "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", "lang": "CHINESE SIMPLIFIED" }
+
+    - name: resource_languages
+      level: extended
+      type: keyword
+      short: List of resource languages.
+      description: >
+        Digest of languages found in resources. Key is language (as string) and
+        value is how many resources there are having that language (as integer)
+      example: { "ENGLISH US": 1, "CHINESE SIMPLIFIED": 760 }
```
andrewstucki, line 58: Could you go through and update these examples? Pretty sure as this is a `type: keyword` the example is actually:

`[ "ENGLISH US", "CHINESE SIMPLIFIED" ]`

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      description: >
+        Hash of the PE header.
+      example: 5aa1aa0f2b4be70397a1e9e2b87627cd
+
+    - name: entry_point
+      level: extended
+      description: >
+        Entry point of the PE file.
+      format: string
+      type: long
```
andrewstucki, line 128: Seeing as we're really talking about a memory address here, would it make sense to just consider this a `keyword` with a hex string? Do we have any other fields currently that are akin to this? I would imagine it's not likely to be doing arithmetic on these in some sort of query.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      example: 25856
+
+    - name: machine_type
+      level: extended
+      description: >
+        Machine type of the PE file.
+      type: keyword
+      example: "Intel 386 or later, and compatibles", "AMD AMD64", 332, 34404
+
+    - name: overlay
```
andrewstucki, line 138: I'm not entirely sure this belongs here. From what I know "overlay" information (stuff appended to the end of the file) can be in lots of different file formats--so wondering if this belongs to a more general `file` field rather than being in `pe` itself.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      example: "2020-11-05T17:25:47.000Z"
+
+    - name: compiler_product_versions
+      level: extended
+      type: keyword
+      description: >
+        Version of the compiler.
+      example: VS98 (6.0) build 8168
+
+    - name: rich_pe_header_hash
```
andrewstucki, line 116: Since rich headers are structured (even though undocumented) and can contain compiler-level information. Does it make sense to move this into its own set of fields (i.e. make this `pe.rich_header.hash.md5`)?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      example: ac9555d914bbb112ecc5f15bb9887ca8371f493ab0941344e976bb8410c8aa78
+
+    - name: compile_timestamp
+      level: extended
+      description: >
+        Compile timestamp of the PE file.
+      type: date
+      example: "2020-11-05T17:25:47.000Z"
+
+    - name: compiler_product_versions
```
andrewstucki, line 109: How about `compilers`?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+          example: b806e17c8e330d82
+
+        - name: md5
+          level: extended
+          type: keyword
+          description: >
+            MD5 hash of raw icon data.
+          example: 6d1cae6272afbb88876ed6476b990d8c
+
+    - name: debug
```
andrewstucki, line 25: this should be a structured set of fields rather than a keyword, no?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
+---
+- name: pe
+  type: group
+  fields:
+    - name: main_icon
```
andrewstucki, line 5: `icon` v. `main_icon`?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+      level: extended
+      type: object
+      description: >
+        Hashes of embedded program icon.
+      fields:
+        - name: dhash
+          level: extended
+          type: keyword
+          description: >
+            Difference Hash (dhash) for a given PE file.
```
andrewstucki, line 15: Can the description mention that this is about image-based visual differential hashing for the icon itself (not the executable code)?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,193 @@
...
+          example: b806e17c8e330d82
+
+        - name: md5
```
andrewstucki, line 18: Would it make sense to have the `icon` info contain a wider variety of subfields such as things like file type or size in bytes? Mainly just wondering, because if so, I feel like we should follow the form of other ECS fields here and move this to something like `pe.icon.hash.md5`

rfcs/text/0000-extend-file-pe.md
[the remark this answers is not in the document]
andrewstucki: @webmat this is generally the [Shannon Entropy](https://en.wikipedia.org/wiki/Entropy_(information_theory)) of a blob of data. I'm thinking it might be worth linking to an article on the algorithm or even just this wiki article, but it's a well-known algorithm, so a link is probably enough.

rfcs/text/0000-extend-file-pe.md
andrewstucki: wondering if we could get `sections` specified as a `nested` field in this PR and then maybe that could lead us to a discussion of the various overlaps in subfields to drive a conversation around extraction into a known/documented common field set.

rfcs/text/pe/pe.yml
dcode: In this case, it provides a summary of each of the resource languages and VT does return it as presented here. that said, I think it's not incredibly useful in this format. A list of keywords as suggested by @andrewstucki would give us a single point to search to see if _any_ resource matches a given language set. But then, we could also do that with the resource list and a nested query. In that case, this could be eliminated altogether or be restructured as:

```json
[ {"name": "ENGLISH US", "count": 1 }, {"name": "CHINESE SIMPLIFIED", "count": 760}]```

Without one or the other, you'll have a field name explosion to account for all possible localizations.

File: rfcs/text/pe/pe.yml (hunk 16 above)
Comment (dcode) [reply rfcs/text/pe/pe.yml:128]: I agree. Either case presents issues though. If you're going to use a keyword, it should be as a hex string, which is consistent with most analysis tools. In this case, VT (and I'm sure other tools), provides the address as an integer. I think it'll have to resort to a script processor doing a string format call in your favorite ETL (ruby, javascript, or painless). The one possible use-case I see for integer representation of an address like this would be for a range query.... but I've never done a range query for an entry point.

File: rfcs/text/pe/pe.yml (hunk 21 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:5]: Not an expert, but I think you can include multiple icons for a file. The `MAINICON` is the one that is used as the actual program icon.

That said, I think it's simpler with just `icon`. If you agree, I'll make those changes.

@dcode for any input.

File: rfcs/text/pe/pe.yml (hunk 23 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:18]: yeah, `pe.icon` is already a group object, so adding `hash.` for different algos (to include `dhash`) would work.

File: rfcs/text/pe/pe.yml (hunk 20 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:45]: yeah, and I think the same could be said for
* import_list
* sections
* resource_details
* resource_languages
* resource_types
* packers

How do I represent that properly?

File: rfcs/text/pe/pe.yml (hunk 15 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:78]: @dcode I didn't know those were counts, what is it telling us that we'd not to represent it with?
```
[ "ENGLISH US", "CHINESE SIMPLIFIED" ]
```

File: rfcs/text/pe/pe.yml (hunk 20 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:45]: or do you mean?
```
    - name: debug
      level: extended
      type: object
      description: Debug information, if present
      fields:
        - name: offset
           level: extended
           type: keyword
           description: description
           example: 1296336
...
```

Comment (peasead) [reply rfcs/text/pe/pe.yml:45]: or do you mean?
```
    - name: debug
      level: extended
      type: object
      description: Debug information, if present
      fields:
        - name: offset
           level: extended
           type: keyword
           description: description
           example: 1296336
...
```

Comment (andrewstucki) [reply rfcs/text/pe/pe.yml:45]: so, there are basically two mechanisms AFAIK for making lists of structured sets of fields -- you'd declare the fields as either `nested` or `flattened`. [`nested`](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) causes some performance penalties due to the way it indexes additional documents and has some limitations on the number of nested objects that a single document can contain (1000 by default, though I doubt we'd hit that). Alternatively you can use [`flattened`](https://www.elastic.co/guide/en/elasticsearch/reference/current/flattened.html) which is an Elastic-licensed only thiing. It keeps the mapping sizes under control, but has the limitation that it essentially makes each subfield a `keyword` field.

You'll likely want to go through each of the fields that can be a "list of objects" and assess whether it makes sense to be `nested` or `flattened` or just a list of keywords (maybe `packers` if you're only putting the name of the packer there?).

File: rfcs/text/0000-extend-file-pe.md (hunk 13 above)
Comment (peasead) [reply rfcs/text/0000-extend-file-pe.md:24]: Is this the way you'd prefer to represent this?
```
    - name: sections
      level: extended
      type: object
      description: Data about sections of compiled binary PE
      fields:
        - name: chi2
           level: extended
           type: keyword
           description: description
           example: 3027194
...
```

File: rfcs/text/pe/pe.yml (hunk 17 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:196]: Fair point. I can remove this and suggest it for the `file.` fieldset so it can be used for any architecture.

File: rfcs/text/pe/pe.yml (hunk 15 above)
Comment (andrewstucki) [reply rfcs/text/pe/pe.yml:78]: @peasead yeah, so those are the rolled up totals of the language code for each resource, retrieved from traversing the resource table. Each leaf node of the resource table corresponds to a single embedded resource and has parent nodes that specify the language of each of the given resources. It's basically just an aggregate of the individual resources that look something like this on their own:

https://github.com/andrewstucki/fingerprint/blob/625f616f4551c33d5e984ab28a0cc48e793c7fad/fixtures/calc.exe.fingerprint#L508-L512

I was more or less commenting on the description of the field and the type. It's currently a keyword, so I would think this would just be a list of all seen language codes acting as a "digest". That said, I'm not quite sure we actually need this anyway, since the data would exist on all of the individual resources anyway and I'm not sure that ECS is where we'd want to create extra fields just for pre-rolled aggregates.

Comment (dcode) [reply rfcs/text/pe/pe.yml:64]: I vote drop this field entirely since it can be equally aggregated with nested agg

Comment (peasead) [reply rfcs/text/pe/pe.yml]: Done.

File: rfcs/text/pe/pe.yml (hunk 20 above)
Comment (peasead) [reply rfcs/text/pe/pe.yml:45]: Opted for `nested`; LMK if you think that `flattened` makes more sense.

File: rfcs/text/0000-extend-file-pe.md (hunk 13 above)
Comment (peasead) [reply rfcs/text/0000-extend-file-pe.md:29]: Did this as `nested`, just like `debug`.

File: rfcs/text/0000-extend-file-pe.md (hunk 12 above)
Comment (andrewstucki) [reply rfcs/text/0000-extend-file-pe.md:28]: Was thinking about this more @dcode -- might want to keep this as `type: flattened` field, mainly to avoid a glut of potential sub documents if this field was `nested`. I can imagine a binary or dynamic library that imports more than 1000 symbols. Maybe we could just figure out a way of documenting, "only use these sub fields in this flattened field"?

Also maybe rename to `imports`?

File: rfcs/text/pe/pe.yml [hunk 24]
```diff
@@ -0,0 +1,228 @@
...
+          example: 198144
+
+    - name: resource_details
+      level: extended
+      type: keyword
+      description: >
+        If the PE contains resources, some info about them
+      example: { "chi2": -1, "filetype": "English text", "entropy": 0, "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", "lang": "CHINESE SIMPLIFIED" }
+
+    - name: resource_types
```
andrewstucki, line 138: pretty sure we can drop this, just like the language aggregation -- it's just a roll-up of all the seen resource types from the details

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,228 @@
...
+          type: long
+          format: bytes
+          example: 198144
+
+    - name: resource_details
+      level: extended
+      type: keyword
+      description: >
+        If the PE contains resources, some info about them
+      example: { "chi2": -1, "filetype": "English text", "entropy": 0, "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", "lang": "CHINESE SIMPLIFIED" }
```
andrewstucki, line 136: What about making this `nested` and explode it into structured fields (also rename `resources`)?

```
resources.chi2
resources.filetype
resources.entropy
resources.hash.sha256
resources.language
resources.type <-- this is new, it's what gets rolled up below (i.e. RT_VERSION, etc.)
```

peasead: Done

rfcs/text/pe/pe.yml
peasead: Sorry, I don't quite understand.

Remove `resource_types`? I moved that under `resources` and renamed it to just `types`.

peasead: Just pinging on this @andrewstucki 
> Sorry, I don't quite understand. 
>
> Remove resource_types? I moved that under resources and renamed it to just types.

andrewstucki: So, just like the resource language roll-ups we previously dropped, this is just an aggregation of the counts of individual resources from the PE resource table. Since we already have type information at the individual resource level, my thought is that there's no need for the roll-up in the schema.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,257 @@
...
+          type: keyword
+          example: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+
+        - name: language
+          level: extended
+          description: Language identification.
+          type: keyword
+          example: "CHINESE SIMPLIFIED"
+
+        - name: types
```
andrewstucki, line 167: This will only be a single type, see the structure here: https://docs.microsoft.com/en-us/windows/win32/menurc/resourceheader?redirectedfrom=MSDN

There should only be a single language id and type associated with a single resource ([here's](https://docs.microsoft.com/en-us/windows/win32/menurc/resource-types) the list of pre-defined resource types)

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+| file.pe.debug.timestamp | date | Timestamp of the debug information. |
+| file.pe.imports | flattened | List of all imported functions |
+| file.pe.sections | nested | Data about sections of compiled binary PE |
+| file.pe.sections.chi2 | long | Chi-square probability distribution. |
+| file.pe.sections.virtual_address | long | Virtual address available to the file. |
+| file.pe.sections.entropy | float | Measurement of entropy randomness in the file. |
+| file.pe.sections.flags | keyword | Section flags of the file. |
+| file.pe.sections.name | keyword | Section names of the file. |
+| file.pe.sections.raw_size | long | Size of the section or the dize of the initialized data on disk. |
+| file.pe.resources | nested | If the PE contains resources, some info about them |
```
andrewstucki, line 36: This looks like it needs to be updated with the schema of the fields in the nested document (i.e. `file.pe.resources.language`, etc.)

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+| file.pe.compiler.version | keyword | Version of the compiler. |
+| file.pe.compiler.name | keyword | Name of the compiler. |
+| file.pe.creation_date | date | Extracted when possible from the file's metadata. Indicates when it was built or compiled. It can also be faked by malware creators. |
+| file.pe.entry_point | keyword | Entry point of the PE file. |
+| file.pe.exports | keyword | List of symbols exported by PE |
+| file.pe.debug | nested | Debug information, if present |
+| file.pe.debug.offset | keyword | Debug offset information. |
+| file.pe.debug.size | keyword | Size of the debug information. |
+| file.pe.debug.type_str | keyword | Format of the debug information. |
+| file.pe.debug.type | keyword | Information type generated by the debug options. |
```
andrewstucki, line 26: I think that this should be combined with `type_str` -- `type_str` looks like the human readable translation based off of macro names in `winnt.h`

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
+# 0000: Extend the PE field set
+
+- Stage: **1 (proposal)**
+- Date: **TBD**
+
+The Portable Executable (PE) sub-field, of the `file` top-level fieldset, can be updated to include more file attributes to aid in file analysis. This additional document metadata can be used for malware research, as well as coding and other application development efforts.
+
+## Fields
+
+This RFC is to create 25 additional sub-fields within the `file.pe` fieldset.
```
rw-access, line 10: This includes updates to these as well, right?
* `dll.pe`
* `process.pe`

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,257 @@
...
+            Version of the compiler.
+          example: 11.0.0
+
+    - name: rich_header
+      level: extended
+      description: >
+        Rich Header information for the PE file.
+      type: object
+      fields:
+        - name: hash
```
rw-access, line 232: Does it make sense to reuse the existing `hash` fieldset here?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,257 @@
...
+              level: extended
+              type: keyword
+              description: >
+                MD5 hash of the header for the PE file.
+              example: 5aa1aa0f2b4be70397a1e9e2b87627cd
+
+    - name: entry_point
+      level: extended
+      description: >
+        Entry point of the PE file.
```
rw-access, line 248: I think it's worth adding that this is the relative byte offset to the base of the file/image.

rfcs/text/pe/pe.yml
andrewstucki: Agreed, and I think that was the point of adjusting this to be an object too. Any chance you know how to document that in the RFC doc @rw-access ?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,257 @@
...
+          level: extended
+          type: date
+          description: Timestamp of the debug information.
+          example: "2020-11-05T17:25:47.000Z"
+
+    - name: imports
+      level: extended
+      type: flattened
+      description: List of all imported functions
+      example: { "library_name" : "mscoree.dll", "imported_functions" : "GetFileVersionInfoSizeA" }
```
rw-access, line 85: I think YAML actually will interpret this as an inline object
```suggestion
      example: '{ "library_name" : "mscoree.dll", "imported_functions" : "GetFileVersionInfoSizeA" }'
```

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+| file.pe.sections.virtual_address | long | Virtual address available to the file. |
+| file.pe.sections.entropy | float | Measurement of entropy randomness in the file. |
+| file.pe.sections.flags | keyword | Section flags of the file. |
+| file.pe.sections.name | keyword | Section names of the file. |
+| file.pe.sections.raw_size | long | Size of the section or the dize of the initialized data on disk. |
+| file.pe.resources | nested | If the PE contains resources, some info about them |
+| file.pe.machine_type | keyword | Machine type of the PE file. |
+| file.pe.hash.dhash | keyword | Difference Hash (dhash) to find files with a visually similar icon or thumbnail. |
+| file.pe.hash.md5 | keyword | MD5 hash of raw icon data |
+| file.pe.packers | keyword | Merged list of all detected packers by all tools used. |
```
rw-access, line 40: I don't see this one in the pe.yml file.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Fields
+
+This RFC is to create 25 additional sub-fields within the `file.pe` fieldset.
+
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| file.pe.authentihash | keyword | Authentihash of the PE file. |
+| file.pe.compile_timestamp | date | Compile timestamp of the PE file. |
+| file.pe.compiler | keyword | Compiler information. |
```
rw-access, line 16: seems odd to have this as a `keyword` but to have compile.version underneath. I don't know how you'd ever submit valid JSON to do this

rw-access: should this just be `group`?

rfcs/text/pe/pe.yml
andrewstucki: I'm not entirely sure how to document possible fields in a `flattened` type, but these should definitely be an object with queryable fields. Not sure if it makes sense to include an `example` in a `flattened` field either.

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+| file.pe.creation_date | date | Extracted when possible from the file's metadata. Indicates when it was built or compiled. It can also be faked by malware creators. |
+| file.pe.entry_point | keyword | Entry point of the PE file. |
+| file.pe.exports | keyword | List of symbols exported by PE |
+| file.pe.debug | nested | Debug information, if present |
+| file.pe.debug.offset | keyword | Debug offset information. |
+| file.pe.debug.size | keyword | Size of the debug information. |
+| file.pe.debug.type_str | keyword | Format of the debug information. |
+| file.pe.debug.type | keyword | Information type generated by the debug options. |
+| file.pe.debug.timestamp | date | Timestamp of the debug information. |
+| file.pe.imports | flattened | List of all imported functions |
```
rw-access, line 28: would a multi-valued keyword make sense for this? windebug notation? like `advapi32!RegSetValueExW`

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+| file.pe.sections.chi2 | long | Chi-square probability distribution. |
+| file.pe.sections.virtual_address | long | Virtual address available to the file. |
+| file.pe.sections.entropy | float | Measurement of entropy randomness in the file. |
+| file.pe.sections.flags | keyword | Section flags of the file. |
+| file.pe.sections.name | keyword | Section names of the file. |
+| file.pe.sections.raw_size | long | Size of the section or the dize of the initialized data on disk. |
+| file.pe.resources | nested | If the PE contains resources, some info about them |
+| file.pe.machine_type | keyword | Machine type of the PE file. |
+| file.pe.hash.dhash | keyword | Difference Hash (dhash) to find files with a visually similar icon or thumbnail. |
+| file.pe.hash.md5 | keyword | MD5 hash of raw icon data |
```
rw-access, line 39:
```suggestion
| file.pe.hash | group | Nested use of the `hash.*` fieldset
```

andrewstucki: This doesn't incorporate `dhash` by nesting, but I think you can just move that to the icon fields? Since that's what the value is actually calculated over?

rw-access: I think we would want to add `dhash` to the hash field set in addition?

rfcs/text/0000-extend-file-pe.md
andrewstucki: That seems kind of janky to query, no? Assuming you wanted to query individually by all binaries that imported a particular function or imported a particular DLL

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,142 @@
...
+
+The Portable Executable (PE) sub-field, of the `file` top-level fieldset, can be updated to include more file attributes to aid in file analysis. This additional document metadata can be used for malware research, as well as coding and other application development efforts.
+
+## Fields
+
+This RFC is to create 25 additional sub-fields within the `file.pe` fieldset.
+
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| file.pe.authentihash | keyword | Authentihash of the PE file. |
```
rw-access, line 14: Should these all be `file.pe` or just `pe` since `pe.*` is not a top-level field but is reused in other contexts, besides just file (`process` and `dll`)

rfcs/text/0000-extend-file-pe.md
andrewstucki: I'd try and avoid that, `dhash` is very much tied to image formats, and is more of an image fingerprinting mechanism than a generic, reusable hash function. Just like `tls.client.ja3` and `pe.imphash` I would try and keep it closer to the entity that it's describing since it's "entity-specific" -- otherwise you get weird things like potentially having data chunks from other field sets completely unrelated to images all of a sudden having some notion of perceptual image hashing.

rfcs/text/pe/pe.yml
ebeahan: Yes, the quotes are necessary - good catch @rw-access!

I don't think any `flattened` example will be comprehensive by its nature, but I think it's useful to have a simple example (e.g., what's done with the [labels](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-labels) field) showing the field's intended value is an object.

rfcs/text/0000-extend-file-pe.md
ebeahan: `pe.*` is nested under `dll.*` and `process.*`. Any updates applied to the top-level field set will also follow under the nested uses.

rfcs/text/0000-extend-file-pe.md
ebeahan: Correct, these should be added under the `pe.*` field and then will be nested under `file`, `process`, and `dll`.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,257 @@
...
+            Digest of resource types. Key is resource type (as string) and value
+            is how many resources there are of that specific type (as integer)
+          example: { "RT_VERSION": 1, "RT_MANIFEST": 1 }
+
+    - name: exports
+      level: extended
+      type: keyword
+      description: >
+        List of symbols exported by PE
+      example: DllInstall, DllRegisterServer, DllUnregisterServer
```
ebeahan, line 181: Is this field intended to be an array?

rfcs/text/0000-extend-file-pe.md
peasead: I'm sorry @andrewstucki do you mean combine `.debug.type_str` into `.debug.type`?

rfcs/text/pe/pe.yml
peasead: Good callout @rw-access and @andrewstucki 

I should also remove it here and use the existing `hash` fieldset, right?

https://github.com/elastic/ecs/blob/4fa79d5321ba6674d652c292131f2bf5f47a48f9/rfcs/text/pe/pe.yml#L11-L43

rfcs/text/pe/pe.yml
peasead: Updated.

rfcs/text/0000-extend-file-pe.md
peasead: fix't

Good catch.

rfcs/text/0000-extend-file-pe.md
peasead: fixed to `nested`. It is listed as `object` in the `pe.yml` file.

rfcs/text/0000-extend-file-pe.md
peasead: Is there a recommended change?

rfcs/text/0000-extend-file-pe.md
peasead: I made an update to the RFC doc. `file.pe.icon` was missing entirely from the table on the doc.

I'm waiting on a response here https://github.com/elastic/ecs/pull/1071#discussion_r548076877 but, if `file.pe.icon.hash.{md5,sha1,sha256}` stays, I'll add it to the table as well. I'm assuming it will be removed and rely on the existing `hash.*` fieldset, but I'll wait.

rfcs/text/0000-extend-file-pe.md
peasead: Fix't

rfcs/text/pe/pe.yml
peasead: Yeah, I think that makes sense.

So, would it be structured like this?
```
example: { "DllInstall", "DllRegisterServer", "DllUnregisterServer" }
```

ebeahan: The example value can be quoted and escaped to avoid unexpected YAML interpretation.

Listing `array` under `normalize` declares that a field's value is always expected to be an array, even if there's only a single value.

```yaml
    - name: exports
      level: extended
      type: keyword
      description: >
        List of symbols exported by PE
      example: "[\"DllInstall\", \"DllRegisterServer\", \"DllUnregisterServer\"]"
      normalize:
        - array
```

rfcs/text/pe/pe.yml
ebeahan: If the intent is still to reuse `hash.*`, let's remove the now unneeded hash fields and note the intent to nest `hash.*` fields in the doc. We can sort out the declaration in the field definitions in the next stage's PR.

ebeahan (commented):
I took another pass. The main item I see remaining before advancing is capturing the outstanding questions/concerns within the RFC doc for future discussion.

@andrewstucki @rw-access any other feedback here?

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,263 @@
...
+      level: extended
+      description: >
+        Machine type of the PE file.
+      type: keyword
+      example: "Intel 386 or later, and compatibles", "AMD AMD64", 332, 34404
+
+    - name: packers
+      level: extended
+      description: >
+        List of packers and tools used.
```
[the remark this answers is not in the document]
peasead: Yep, I'll get that updated.

rfcs/text/pe/pe.yml
```diff
@@ -0,0 +1,263 @@
...
+        - name: language
+          level: extended
+          description: Language identification.
+          type: keyword
+          example: "CHINESE SIMPLIFIED"
+
+        - name: type
+          level: extended
+          type: keyword
+          short: List of resource types.
```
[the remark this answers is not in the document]
peasead: Yes, I'll update that as well.

rfcs/text/pe/pe.yml
peasead: Yep, I'll update that.

rfcs/text/pe/pe.yml
peasead: Thanks, fix't.

rfcs/text/0000-extend-file-pe.md
peasead: Ping

andrewstucki:
@ebeahan / @peasead -- sorry about the delay, I'll take another pass at the pe/macho/elf fields and see if anything jumps out to me

rfcs/text/0000-extend-file-pe.md
```diff
@@ -0,0 +1,150 @@
...
+| pe.compiler | nested | Compiler information. |
+| pe.compiler.version | keyword | Version of the compiler. |
+| pe.compiler.name | keyword | Name of the compiler. |
+| pe.creation_date | date | Extracted when possible from the file's metadata. Indicates when it was built or compiled. It can also be faked by malware creators. |
+| pe.entry_point | keyword | Relative byte offset to the base of the PE file. |
+| pe.exports | keyword | List of symbols exported by PE |
+| pe.debug | nested | Debug information, if present |
+| pe.debug.offset | keyword | Debug offset information. |
+| pe.debug.size | keyword | Size of the debug information. |
+| pe.debug.type_str | keyword | Format of the debug information. |
```
andrewstucki, line 25: I'm not sure about this v. `type` -- it looks like `type_str` is just the "normalized" value of `type` with the name found in source macros. Personally I'd prefer just dropping `type_str` and storing the normalized type in `type`, but if we want maybe we can hash that out in a further step of the RFC? @ebeahan thoughts?

andrewstucki (commented):
Generally this looks pretty good to me as far as the initial fields to fight over are concerned 😅 

I'll leave it to @ebeahan to opine on whatever additional formatting for the RFC doc itself might need to be done prior to merge.

rfcs/text/0000-extend-file-pe.md
andrewstucki: sorry, yes, just realized I made the same comment on my re-review just now

rfcs/text/0000-extend-file-pe.md
ebeahan: Based on @andrewstucki explanation, `type` seems sufficient, but I can't speak authoritatively in this context if `type_str` is more idiomatic or accurate of what's being captured in the field.

I do agree that it doesn't have to be fully decided at this stage. It can be captured as an outstanding concern in the doc for further discussion at a later stage.

rfcs/text/0000-extend-file-pe.md
peasead: Done.

rfcs/text/0000-extend-file-pe.md
peasead: @andrewstucki yeah, you'd mentioned this above and I'd not fixed it yet. That's done now. I did as you suggested 👍

peasead:
@ebeahan just checking on this to see what the next steps are.

ebeahan:
I made some minor housekeeping edits to the field definitions:

* `pe.exports` - Since the description indicates this can be a list, I made the explicit.
* `pe.resources.type` - Change `example` value to use array.
* `pe.machine_type` - Switched to one example value since this field doesn't appear to require an array.
* `pe.packers` - Also set `example` value to be an array.

Unless there are any issues with these changes, I think we're good for this proceeding with advancing 🎉

peasead:
Thanks @ebeahan glad to be moving this along!

Thanks for those changes. Looks good.

ebeahan (approved):
Made the last set of changes prior to merging:

* Assigned RFC number and advance date
* Updated the stage name from `proposal` to `draft` to align with the new [stages](https://elastic.github.io/ecs/stages.html)
* Updated the markdown comments as well to align with the new stages (primarily removing the legacy stage 4 sections)

ebeahan:
@peasead Now that this is merged, I recommend opening the stage 2 PR even if only the stage number is updated. By having that PR open in advance, we have a spot to capture feedback or have more discussion.

PR #1073, opened 2020-11-04
Docs: Add ifeval for 7.10 release

EamonnTP:
### Summary

This PR fixes a documentation link that will break when 7.10 becomes the `current` stack version. It conditionally changes the link location of the Logs Monitoring Field Reference page based on the stack version defined in github.com/elastic/docs ([here](https://github.com/elastic/docs/blob/master/shared/versions/stack/current.asciidoc)).

```asciidoc
ifeval::["{branch}"=="7.9"]
{logs-guide}/logs-fields-reference.html[Log Monitoring Field Reference]
endif::[]
ifeval::["{branch}"!="7.9"]
{observability-guide}/logs-app-fields.html[Log Monitoring Field Reference]
endif::[]
```
This ensures the link continues to work for users who click it before and after the 7.10 release.

Tested by building locally with `:branch:` overridden to `7.9`, `7.10`, and `master`.

### Target

This PR will need to be backported to `1.7`, `1.6`, `1.5`, and `1.x`.

### Related issues

https://github.com/elastic/observability-docs/issues/215
https://github.com/elastic/docs/pull/1994

### Follow ups

After the release of 7.10, I will backport a fix that removes the conditional code.

PR #1082, opened 2020-11-04
[1.5] Prepare link to Logs docs changing with the 7.10 release in "getting-started" (#1073)

webmat:
Backports the following commits to 1.5:

* Prepare link to Logs docs changing with the 7.10 release in "getting-started" (#1073)

### DIFF
--- a/docs/using-getting-started.asciidoc
+++ b/docs/using-getting-started.asciidoc
@@ -285,5 +285,10 @@ Here are some examples of additional fields processed by metadata or parser proc
 We've covered at a high level how to map your events to ECS. Now if you'd like your events to render well in the Elastic
 solutions, check out the reference guides below to learn more about each:
 
-* {logs-guide}/logs-fields-reference.html[Logs Monitoring Field Reference]
+ifeval::["{branch}"=="7.9"]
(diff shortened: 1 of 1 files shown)


webmat:
@ebeahan This backport is failing because of the chunking change we did recently. Can you recall what needs to be done here?

These doc backports all the way back to the bronze age happen every once in a while. I wonder if break will happen every time we backport a link fix to 1.5?

ebeahan:
I've identified the issue; I missed including the necessary generator changes when I added the `[discrete]` headers in the 1.5 branch.

webmat:
Yeah I wonder if you had done those manually, to avoid having to modify multiple different past versions of the scripts.

If that was the case, this isn't a blocker for this backport. But we should consider fixing the script anyway, if it's a small fix. I wouldn't want to wonder why the build is failing every other release, when we backport fixes to broken links to older versions of the doc :-)

ebeahan:
I had the changes done locally, but I overlooked committing them 🤕 

Script fix: https://github.com/elastic/ecs/pull/1089

ebeahan (approved):
LGTM

Issue #1096, opened 2020-11-05
Create file.mach-o


peasead:
**Summary**

The Mach Object (Mach-O) sub-field does not yet exist for the `file` top-level fieldset. This can be created to include more file attributes to aid in malware and file analysis.

**Motivation**:

In creating a VirusTotal Filebeat module, we identified opportunities to extend the `file.*` top-level fieldset with the creation of the Mach-O sub-field.

**Detailed Design**:

|   Name                                     |   Type     |   Description                                                               |
|--------------------------------------------|------------|-----------------------------------------------------------------------------|
|   file.mach-o.cpu                          |   object   |   CPU information for the file.                                             |
|   file.mach-o.cpu.architecture             |   keyword  |   CPU architecture target for the file.                                     |
|   file.mach-o.cpu.byte_order               |   keyword  |   CPU byte order for the file.                                              |
|   file.mach-o.cpu.subtype                  |   keyword  |   CPU subtype for the file.                                                 |
|   file.mach-o.cpu.type                     |   keyword  |   CPU type for the file.                                                    |
|   file.mach-o.headers                      |   object   |   Header information for the file.                                          |
|   file.mach-o.headers.commands             |   object   |   Header load commands for the file.                                        |
|   file.mach-o.headers.commands.number      |   long     |   Number of load commands for the Mach-O header.                            |
|   file.mach-o.headers.commands.size        |   long     |   Size of load commands of the Mach-O header.                               |
|   file.mach-o.headers.commands.type        |   keyword  |   Type of the load commands for the Mach-O header.                          |
|   file.mach-o.headers.magic                |   keyword  |   Magic field of the Mach-O header.                                         |
|   file.mach-o.headers.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
|   file.mach-o.segments                     |   object   |   Segment information for the file.                                         |
|   file.mach-o.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
|   file.mach-o.segments.name                |   keyword  |   Name of this segment.                                                     |
|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
|   file.mach-o.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
|   file.mach-o.segments.sections            |   object   |   Section information for the segment of the file.                          |
[...]

PR #1077, opened 2020-11-04
File.elf create

peasead:
[RFC Preview](https://github.com/peasead/ecs/blob/file.elf-create/rfcs/text/0000-create-file-elf.md) 

Resolves https://github.com/elastic/ecs/issues/1062

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)? Yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)? Yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? Yes
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? Yes
- Is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed. Yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? Yes

PR #1097, opened 2020-11-05
File.macho create

peasead:
[RFC Preview](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/0000-create-file-macho.md)

Resolves https://github.com/elastic/ecs/issues/1096

- If submitting code/script changes, have you verified all tests pass locally using `make test`? NA at this stage
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?  NA at this stage
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)?  NA at this stage


andrewstucki:
Not that I'm super excited about it, but--how do you envision working with fat object files with multiple architectures? Just wondering because I would think you'd almost need to structure the entire `macho` field sets as `nested` if you want to handle them... @webmat what are your thoughts on making something that could be as large as a `macho` field set entirely `nested`?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,155 @@
...
+|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   file.mach-o.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   file.mach-o.segments.sections            |   object   |   Section information for the segment of the file.                          |
+|   file.mach-o.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   file.mach-o.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   file.mach-o.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   file.mach-o.signature                    |   object   |   Signature information for the file.                                       |
```
andrewstucki, line 38: I know that there's not necessarily a full corollary with all of the fields that comprise a signed MachO binary, but I'd recommend trying to see if we can use/extend the [`code_signature`](https://www.elastic.co/guide/en/ecs/current/ecs-code_signature.html) fields here--this use case is one of the original reasons for introducing it as a primary field set rather than just having something PE specific.

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,155 @@
...
+|   file.mach-o.headers                      |   object   |   Header information for the file.                                          |
+|   file.mach-o.headers.commands             |   object   |   Header load commands for the file.                                        |
+|   file.mach-o.headers.commands.number      |   long     |   Number of load commands for the Mach-O header.                            |
+|   file.mach-o.headers.commands.size        |   long     |   Size of load commands of the Mach-O header.                               |
+|   file.mach-o.headers.commands.type        |   keyword  |   Type of the load commands for the Mach-O header.                          |
+|   file.mach-o.headers.magic                |   keyword  |   Magic field of the Mach-O header.                                         |
+|   file.mach-o.headers.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
+|   file.mach-o.segments                     |   object   |   Segment information for the file.                                         |
```
andrewstucki, line 28: I would think that the same `entropy` and hash calculations as PE fields would make sense here--also import/export tables.

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,155 @@
...
+|   file.mach-o.signature.team_identifier    |   keyword  |   Team identifier of the code signing certificate.                          |
+|   file.mach-o.signature.sealed_resources   |   long     |   Version of the resource envelope for the code signing certificate.        |
+|   file.mach-o.signature.cms_digest         |   keyword  |   Cryptographic Message Syntax (CMS) hash of the code signing certificate.  |
+|   file.mach-o.signature.cms_digest_type    |   keyword  |   Cryptographic Message Syntax (CMS) type of the code signing certificate.  |
+|   file.mach-o.signature.status             |   keyword  |   Verification information for the code signing certificate.                |
+|   file.mach-o.signature.fingerprint        |   keyword  |   MD5 digest of the der-encoded certificate information.                    |
+|   file.mach-o.executable                   |   object   |   Information about the executable segment for the file.                    |
+|   file.mach-o.executable.segment_base      |   keyword  |   Executable segment base size.                                             |
```
andrewstucki, line 47: wondering why specifically calling out the executable segment here rather than keeping this information for each segment and dropping `file.mach-o.executable`?

rfcs/text/0000-create-file-mach-o.md
peasead: Thanks for the feedback, @andrewstucki 

I looked at [`code_signature`](https://www.elastic.co/guide/en/ecs/current/ecs-code_signature.html) and was able to use `code_signature.status` and remove `file.mach-o.signature.status`. 

Do you think that we should remove `file.mach-o.signature.*` and extend `file.code_signature.*` instead of having signature in two places?

rfcs/text/0000-create-file-mach-o.md
peasead: Apologies, I don't have a great answer here. In looking at the data of a single file, here's what I'm seeing for `executable`:
```
...
"Executable Segment base": "0",
"Executable Segment limit": "49152",
"Executable Segment flags": "0x0",
...
```
Compared to what I'm seeing for the `segments` section:
```
 "segments": [
        {
          "vmaddr": "0x0",
          "name": "__TEXT",
          "vmsize": "0xc000",
          "fileoff": "0x0",
          "filesize": "0xc000",
          "sections": [
            {
              "flags": [
                "SECTION_ATTRIBUTES_USR",
                "SECTION_ATTRIBUTES_SYS"
              ],
              "name": "__text",
              "type": "S_REGULAR"
```

Of note, in the second example (from the same file), `file.mach-o.segments.sections.flags:SECTION_ATTRIBUTES_USR` and `file.mach-o.executable.segment_flags:0x0`...so it appears to represent different parts of the file.

Would you prefer to see something like this?
```
file.mach-o.segments.executable.base
file.mach-o.segments.executable.limit
file.mach-o.segments.executable.flags
```

rfcs/text/0000-create-file-mach-o.md
andrewstucki: Yeah, that would be my suggestion. If possible, try and model some of the sub fields in reusable ways. For example:

I believe`file.mach-o.signature.fingerprint` is related to the x509 cert information embedded in the signature, right? If so, might consider reviving https://github.com/elastic/ecs/pull/852 that I had initially taken a stab at, but then stalled out awhile ago. Additionally for `file.mach-o.signature.team_identifier` is that about the Identifier contained inside the MachO code directory? The description seems to indicate that it's from the code signing cert itself, which I would say could even map to the subject of the cert itself, or a CN in the x509 cert. If it's the code directory identifier, see below around `candidate_cd_hash`.

I don't entirely remember how some of the other pieces of things are encoded in the signature (i.e. `cms_digest`, etc.), but if they're generic ASN.1 attributes, then maybe we could figure out how to model that as a part of generic pkcs7 fields under `code_signature` as well?

For something like `candidate_cd_hash` since this is part of the signature verification process, but not an artifact of a standardized cryptographic signature scheme, I'm wondering if it makes sense just to make a `macho.cdhash` similar to what was suggested for `pe.authentihash`? Similarly for modeling a team identifier pulled from the code directory, something like `macho.signer` or `macho.team_identifier`?

rfcs/text/0000-create-file-mach-o.md
andrewstucki: So, from the looks of:

```
"Executable Segment base": "0",
"Executable Segment limit": "49152",
"Executable Segment flags": "0x0",
```

it seems that what's being described is the offset, size, and flags of the segment with executable code in it, namely the `__TEXT` segment (which will have a number of sections in it). Here's a [reference](https://github.com/aidansteele/osx-abi-macho-file-format-reference#table-1-the-sections-of-a__textsegment).

I would think that it would probably make more sense just to model this as something like:

```
file.mach-o.segments.name = "__TEXT"
file.mach-o.segments.offset = 0
file.mach-o.segments.size = 49152
file.mach-o.segments.flags = []
```

Which would be pretty much identical to with something like a `__DATA` or `__IMPORT` segment.

Also, note that for:

> Of note, in the second example (from the same file), file.mach-o.segments.sections.flags:SECTION_ATTRIBUTES_USR and file.mach-o.executable.segment_flags:0x0...so it appears to represent different parts of the file.

the former is referring to the section flags of the `__text` section inside of the `__TEXT` segment (wonderful names I know), while the latter is referring to the flags of the `__TEXT` segment itself.

rfcs/text/0000-create-file-mach-o.md
peasead: **Fingerprint**
Okay, in doing some research ([1](https://objective-see.com/blog/blog_0x26.html), https://github.com/elastic/ecs/pull/852), I would agree that `file.mach-o.signature.fingerprint` is related to the x509 certificate information. I have removed that.

**Team Identifier**
WRT to `file.mach-o.signature.team_identifier` - _Team identifiers are automatically recorded in signatures when signing with suitable Apple-issued signing certificates_ ([2](https://www.manpagez.com/man/1/codesign/osx-10.10.php)). **IMO, I named this improperly**. It appears Team Identifier and Team ID are not the same things (/smh). The data we had shows `11A1A1AAAA` (or the like), which is actually the Team ID vs. the Team Identifier. What about renaming this to `team_id` and align with what it actually is? `team_identifier` does match up nicely to `x509.subject.common_name` and would make sense to be moved there.

Example: 
```
Identifier:  com.cisco.anyconnect.gui (0x60)
Team ID:     DE8Y96K9QP (0x79)
```
**CMS Digest**
`cms_digest` appears to just fall under the [RFC](https://tools.ietf.org/html/rfc5652) and in browsing through the RFC, it appears to be [ASN.1 attributes](https://tools.ietf.org/html/rfc5652#page-43); so where that would need to go in `code_signature` I can't comment on, but I'd agree it's not a Mach-O-specific attribute, then it shouldn't be nested under `file.mach-o.`. I have removed this.

**CD Hash**
I'd agree how you describe it that we should move it up a tier and out of nested underneath `file.mach-o.signature.`. I'll do that.

**Mach-O Signer/Team Identifier**
I noted this above, but `team_identifier` makes sense to be moved to x509.  **Would you recommend moving `team_id` up to `file.mach-o.team_id` from `file.mach-o.signature.team_id`?**

rfcs/text/0000-create-file-mach-o.md
peasead: I apologize, I don't have entropy or import/export tables data in mach-o.

What am I missing?

rfcs/text/0000-create-file-mach-o.md
peasead: Thanks for the explanation. I like this approach - seems cleaner
```
file.mach-o.segments.name = "__TEXT"
file.mach-o.segments.offset = 0
file.mach-o.segments.size = 49152
file.mach-o.segments.flags = []
```
I'll do this.

rfcs/text/0000-create-file-mach-o.md
peasead: Just pinging on this
> Would you recommend moving team_id up to file.mach-o.team_id from file.mach-o.signature.team_id?

rfcs/text/0000-create-file-mach-o.md
peasead: Just pinging on this @andrewstucki

rfcs/text/0000-create-file-mach-o.md
andrewstucki: Sorry, was out for a couple of weeks and then got wrapped up in some other work, so this fell off my radar. In possible, yes.

My thought is that we already have a `code_signature` field set, so rather than introducing a separate `macho.signature` set of fields we could ideally just use `code_signature` nested under `macho` and move any of the `macho`-specific extensions directly into the `macho` namespace. If that doesn't make sense or is a majority of the fields that would exist under `macho.signature` I'm amenable to an argument that `macho.signature` should be a thing. But overall, I think it makes sense to try and re-use existing fields when we can.

rfcs/text/0000-create-file-mach-o.md
andrewstucki: I'm fine with not including this in the first pass, but I could have sworn that at least VT contained entropy calculations for either segments or sections.

If they're also missing import/export tables in the samples you're working with, I'm fine punting on that until we have a use-case to add them in.

I'm going to try and get around to having some space time and working on the object file parser that I started for beats quite some time back, so if we get this as a foundation, maybe I can extend the schema as needed.

andrewstucki: also, think rather than `object`, you want this to be `nested`, no?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.headers.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
+|   file.mach-o.segments                     |   object   |   Segment information for the file.                                         |
+|   file.mach-o.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
+|   file.mach-o.segments.name                |   keyword  |   Name of this segment.                                                     |
+|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   file.mach-o.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   file.mach-o.segments.sections            |   object   |   Section information for the segment of the file.                          |
```
andrewstucki, line 34: if this only contains keyword subfields, should probably be `flattened`?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   file.mach-o.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   file.mach-o.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   file.mach-o.segments.offset      |   keyword  |   Offset of the segment.                                             |
+|   file.mach-o.segment size     |   keyword  |   Segment limit size.                                            |
```
andrewstucki, line 39:
```suggestion
|   file.mach-o.segments.size     |   long  |   Segment limit size.                                            |
```

Is this supposed to be this?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.segment.flags     |   keyword  |   Segment flags.                                                 |
```
andrewstucki, line 40:
```suggestion
|   file.mach-o.segments.flags     |   keyword  |   Segment flags.                                                 |
```
?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   file.mach-o.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   file.mach-o.segments.sections            |   object   |   Section information for the segment of the file.                          |
+|   file.mach-o.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   file.mach-o.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   file.mach-o.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   file.mach-o.segments.offset      |   keyword  |   Offset of the segment.                                             |
```
andrewstucki, line 38:
```suggestion
|   file.mach-o.segments.offset      |   long  |   Offset of the segment.                                             |
```

Not sure if this should be an actual numeric or if it's like a memory address where you're probably not actually doing queries on it based off of its numeric value?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.headers.commands.size        |   long     |   Size of load commands of the Mach-O header.                               |
+|   file.mach-o.headers.commands.type        |   keyword  |   Type of the load commands for the Mach-O header.                          |
+|   file.mach-o.headers.magic                |   keyword  |   Magic field of the Mach-O header.                                         |
+|   file.mach-o.headers.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
+|   file.mach-o.segments                     |   object   |   Segment information for the file.                                         |
+|   file.mach-o.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
+|   file.mach-o.segments.name                |   keyword  |   Name of this segment.                                                     |
+|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
```
andrewstucki, line 31:
```suggestion
|   file.mach-o.segments.vmsize              |   long  |   Memory size of this segment.                                              |
```

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   file.mach-o.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
```
andrewstucki, line 33:
```suggestion
|   file.mach-o.segments.filesize            |   long  |   Amount of memory to map from the file.                                    |
```

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+|   file.mach-o.headers.commands.type        |   keyword  |   Type of the load commands for the Mach-O header.                          |
+|   file.mach-o.headers.magic                |   keyword  |   Magic field of the Mach-O header.                                         |
+|   file.mach-o.headers.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
+|   file.mach-o.segments                     |   object   |   Segment information for the file.                                         |
+|   file.mach-o.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
+|   file.mach-o.segments.name                |   keyword  |   Name of this segment.                                                     |
+|   file.mach-o.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   file.mach-o.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
```
andrewstucki, line 32:
```suggestion
|   file.mach-o.segments.fileoff             |   long  |   File offset of this segment.                                              |
```

For the offsets are you wanting this to be a `long`?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+
+|   Name                                     |   Type     |   Description                                                               |
+|--------------------------------------------|------------|-----------------------------------------------------------------------------|
+|   file.mach-o.cpu                          |   object   |   CPU information for the file.                                             |
+|   file.mach-o.cpu.architecture             |   keyword  |   CPU architecture target for the file.                                     |
+|   file.mach-o.cpu.byte_order               |   keyword  |   CPU byte order for the file.                                              |
+|   file.mach-o.cpu.subtype                  |   keyword  |   CPU subtype for the file.                                                 |
+|   file.mach-o.cpu.type                     |   keyword  |   CPU type for the file.                                                    |
+|   file.mach-o.headers                      |   object   |   Header information for the file.                                          |
```
andrewstucki, line 21: Is the pluralization of this to account for fat archives? If so I'm assuming you want to actually move that up to the top level -- you generally have things like different cpu architectures that are associated with fat files. Instead, I'd just make this `header` or un-nest everything from under here and each `mach-o` entry be for a separate architecture. I'm assuming we might need to punt on fat architectures on the first pass (@webmat / @ebeahan  : not sure if it's a good idea to make the entire `mach-o` field set `nested`? any insight?).

ebeahan: Is `nested` considered because these objects would need to be queried independently of each other? Nested (along with `flattened`) has been discussed before for various use cases, but ECS hasn't had a field adopt it yet.

How many possible architectures can be associated in this case? A small enough number that naming and nesting them into predictable places could be a consideration?

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
+# 0000: Create the Mach-O sub-field of the File fieldset
+
+- Stage: **0 (strawperson)**
+- Date: **TBD**
+
+Create the Mach Object (Mach-O) sub-field, of the `file` top-level fieldset. This document metadata can be used for malware research, as well as coding and other application development efforts.
```
ebeahan, line 6: Perhaps `macho` could be added as a new top-level field set and nested under `file.*`, similar to the `pe.*` fields?

rfcs/text/mach-o/mach-o.yml
```diff
@@ -0,0 +1,190 @@
+---
```
ebeahan, line 1: Thanks for adding the field definitions file! 💯

rfcs/text/0000-create-file-mach-o.md
andrewstucki: So, @ebeahan , Apple's multi-architecture fat formats are a bit gross. I'm not entirely sure if macho can be a generic container for things like various MIPS architectures, but off the top of my head if we wanted to support what Mac architectures have been around in the past ~10 years (of which I'm sure there's a good amount of floating around out there), we'd likely need to have support for x86 32-bit and 64-bit architectures, ppc 32-bit and 64-bit architectures, and ARM 32-bit and 64-bit architectures.

Due to the ability to compile "fat" binaries that support multiple architectures, all of these could (theoretically) be in a single macho binary. Each would require all of the segments, symbols, headers fields at minimum since they're more or less independent of each other.

In practice, you're probably likely to run into fat binaries with 32-bit and 64-bit x86 and, with some older binaries, ppc64 architectures on a somewhat regular basis. With Apple's move to ARM chips now though, if they push continuing fat file formats, this could shift towards something like 64-bit x86 and ARM.

If we want to punt on the whole multi-architecture question we could possibly just say "forget the nested documents for now" and say that if you're parsing an object file with multiple formats in it, you should just emit 2 events, but IMO that's not exactly ideal.

rfcs/text/0000-create-file-mach-o.md
peasead: No worries at all!

Yeah, that makes sense, so I'll remove it here and make a note to look at adding it to [`code_signature`](https://www.elastic.co/guide/en/ecs/current/ecs-code_signature.html).

rfcs/text/0000-create-file-mach-o.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Fields
+
+**Stage 0**
+
+This RFC is to create the Mach-O sub-field within the `file.` fieldset. This will include 35 sub-fields.
+
+|   Name                                     |   Type     |   Description                                                               |
+|--------------------------------------------|------------|-----------------------------------------------------------------------------|
+|   file.mach-o.cpu                          |   object   |   CPU information for the file.                                             |
```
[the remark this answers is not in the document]
peasead: Yeah, if `-` isn't permitted, I'll change everything to `macho` 👍

rfcs/text/0000-create-file-mach-o.md
peasead: Yep, that's the intention.

I've gone back and cleaned up the `file.mach.*` that was throughout the doc/yml file...I made the same mistake with `pe.*`

rfcs/text/0000-create-file-mach-o.md
peasead: Sorry, I thought that `type: object` was meaning that there were `nested` fields? Is it something different?

rfcs/text/0000-create-file-mach-o.md
peasead: So change `type: object` to `type: flattened` for `macho.segments`?

rfcs/text/0000-create-file-mach-o.md
peasead: > Is the pluralization of this to account for fat archives? If so I'm assuming you want to actually move that up to the top level -- you generally have things like different cpu architectures that are associated with fat files. Instead, I'd just make this `header` or un-nest everything from under here and each `mach-o` entry be for a separate architecture. I'm assuming we might need to punt on fat architectures on the first pass.

Nothing that fancy, just an oversight. Fix't.

WRT to the "fat" binaries, I'm willing to put in the work if someone wants to guide me (or push into the PR themselves). if it's easier to pass for now, that's fine too.

rfcs/text/0000-create-file-mach-o.md
peasead: Done.

rfcs/text/0000-create-file-mach-o.md
peasead: I've made the change to `header` from `headers`.

Was there a decision on nesting/un-nesting?

rfcs/text/0000-create-file-mach-o.md
peasead: Should it be changed to `type: nested`?

rfcs/text/0000-create-file-mach-o.md
ebeahan: Before making that call, I think there's a decision needed if these initial `mach-o` fields will capture header data for universal/fat macho-o binaries?

Once we understand and agree on the structure of possible events we need to model for, I think it'll be easier to agree on data types.

rfcs/text/0000-create-file-mach-o.md
ebeahan: `type: nested` is only necessary if this is intended to use the Elasticsearch nested type data type, but not if the fields are just nested underneath an ECS fieldset (e.g., `macho.segments.name`)

andrewstucki: in the case of segments, we're dealing with multiple segments, not just a single segment with a given name (i.e. no, `segments.name`, instead `segments[0].name`), so it's either going to need to be `nested` or `flattened` to be queryable. Since not all of the attributes (i.e. offsets) are text, `nested` is the way to go.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   macho.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   macho.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   macho.segments.sections            |   object   |   Section information for the segment of the file.                          |
+|   macho.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   macho.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   macho.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   macho.segments.offset      |   keyword  |   Offset of the segment.                                             |
+|   macho.segment size     |   keyword  |   Segment limit size.                                            |
```
andrewstucki, line 39: can you clean this up? think it should be `macho.segments.size` and `segments` should be moved to `nested` both here and in the field definitions.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.segment.flags     |   keyword  |   Segment flags.                                                 |
```
andrewstucki, line 40: `macho.segments.flags`?

rfcs/text/macho/macho.yml
```diff
@@ -0,0 +1,173 @@
...
+          level: extended
+          example: 123456
+
+        - name: flags
+          description: Segment flags.
+          type: keyword
+          level: extended
+          example: 0x0
+
+    - name: cdhash
```
andrewstucki, line 162: this looks like it's missing from the RFC document itself.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.header.magic                |   keyword  |   Magic field of the Mach-O header.                                         |
+|   macho.header.flags                |   keyword  |   Flags set in the Mach-O header.                                           |
+|   macho.segments                     |   flattened   |   Segment information for the file.                                         |
+|   macho.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
+|   macho.segments.name                |   keyword  |   Name of this segment.                                                     |
+|   macho.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   macho.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   macho.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   macho.segments.sections            |   object   |   Section information for the segment of the file.                          |
```
andrewstucki, line 34: this is `flattened` in the field document below.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   macho.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   macho.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   macho.segments.offset      |   keyword  |   Offset of the segment.                                             |
```
andrewstucki, line 38: this doesn't match the type in the field definition below

andrewstucki (commented):
Generally looking fairly good to me. There are a few things to clean up in the RFC document and the one type change for `segments` to `nested` (dealing with multiple segments + have subfields that aren't `keyword`) -- besides that and discussing the multi-arch binaries (which are going to be more common now with the new M1 processors), I feel pretty decent about moving this to the next RFC stage as far as the actual fields are concerned.

rfcs/text/0000-create-file-mach-o.md
peasead: Done.

rfcs/text/0000-create-file-macho.md
peasead: Done

rfcs/text/0000-create-file-macho.md
peasead: Fix't

rfcs/text/macho/macho.yml
peasead: Fix't

rfcs/text/0000-create-file-macho.md
peasead: Fixed.

rfcs/text/0000-create-file-macho.md
peasead: Fixed.

rfcs/text/0000-create-file-mach-o.md
dcode: So, it's probably important to consider multiple headers, unfortunately. It is gross because ideally we could have a similar structure across PE, Mach-O, and ELF...there are some fat binary adaptations for the other 2 platforms, but they're not strictly supported in the same way that the Mach-O format supports them.

Namely, I can imagine a visualization/aggregation where I can organize all my `file` document types by their exectuable header information...  i.e. executable type, cpu architecture, binary flags, etc. That said, the overlap of header specific information probably isn't insanely large. Looking at the [LIEF documentation](https://lief.quarkslab.com//doc/latest/api/python/abstract.html#header) (probably the most complete representation I've found), there's just a few fields in the abstract header, which would probably be useful in keyword arrays rather than nested objects (or both).

peasead: So, based on this discussion and [this article](https://medium.com/@cyrilcermak/exploring-ios-es-mach-o-executable-structure-aa5d8d1c7103#f982) on fat binaries, it looks like they're going to be more prevalent in the future. It sounds like we agree moving to `headers` from `header` is the right approach.

@andrewstucki can you explain this? Do you mean to move this to `file.headers` as an object and things like `ARM` and `ARM64` being value examples?
> you want to actually move that up to the top level -- you generally have things like different cpu architectures that are associated with fat files

@dcode you mentioned that nested or keyword arrays would be helpful...do you have an opinion of which would be _more_ useful?

dcode: So...  after digging deeper, it's worse than that. A fat Mach-o binary is literally two full binaries plus an extra header. That means, there are duplicate headers, sections, imports, exports, etc....  so...  long and short of it, not sure yet

peasead: Looks like we're good to go with `macho.headers` as a nested field type. I've made the change to the docs.

peasead:
@ebeahan @dcode has made the call on the nested field type for `macho.headers`. 

Are there any hanging questions for this PR stage?

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
+# 0000: Create the Mach-O sub-field of the File fieldset
+
+- Stage: **1 (proposal)**
```
ebeahan, line 3:
```suggestion
- Stage: **1 (draft)**
```

Same change to the stage naming, again 😂 . I think this is the last one! 🤞

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+- Date: **TBD**
+
+Create the Mach Object (Mach-O) sub-field, of the `file` top-level fieldset. This document metadata can be used for malware research, as well as coding and other application development efforts.
```
ebeahan, line 6: Can we add the intent to nest `macho` underneath `process`, as well?

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.segments.vmaddr              |   keyword  |   Memory address of this segment.                                           |
+|   macho.segments.name                |   keyword  |   Name of this segment.                                                     |
+|   macho.segments.vmsize              |   keyword  |   Memory size of this segment.                                              |
+|   macho.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   macho.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   macho.segments.sections            |   flattened   |   Section information for the segment of the file.                          |
+|   macho.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   macho.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   macho.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
```
ebeahan, line 36: Similar to what we discussed https://github.com/elastic/ecs/pull/1077#discussion_r572274291, `flattened` cannot have explicit fields mapped underneath.

If the data is expected to be shaped to an expected object format, we'll have to capture this another way.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/1097
```
ebeahan, line 141:
```suggestion
* Stage 1: https://github.com/elastic/ecs/pull/1097
```

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+<!--
+Stage 2: Identifies scope of impact of changes. Are breaking changes required? Should deprecation strategies be adopted? Will significant refactoring be involved? Break the impact down into:
+ * Ingestion mechanisms (e.g. beats/logstash)
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
+
```
ebeahan, line 102: Any known concerns we should capture at this point?

ebeahan (commented):
I made a pass on the `macho` YAML field definition.

The only major change I made was removing the sub-fields from `macho.segments.sections`, due to using `type: flattened`. I also touch on this in another comment below.

rfcs/text/0000-create-file-macho.md
peasead: Do you mean that `macho.segments.sections` needs to be `nested`?

rfcs/text/0000-create-file-macho.md
peasead, line 6:
```suggestion
Create the Mach Object (Mach-O) sub-field, of the `file` or `process` top-level fieldsets. This document metadata can be used for malware research, as well as coding and other application development efforts.
```

devonakerr (dismissed):

rfcs/text/0000-create-file-macho.md
ebeahan: No, I mean if we use `flattened` and are expecting specific keys to be present under the `flattened` parent field, we'll need someone to capture this expectation in ECS.

peasead: It does appear that there are specific keys under `macho.segments.sections.*`. @dcode is this something you can help with?

Here's a reference that could help: https://github.com/aidansteele/osx-abi-macho-file-format-reference#section under `flags`.

dcode: The fields under `macho.segments.sections[*]` is well-formed and consists of `flags` that denote how the memory is handled by the loader, the name of the section, the section name, and the type of section that it is. In ELF and PE, we put section offset information in `file.*.sections[*]`. This data is missing from the current rendering of the schema. Both the section and the segment have offset data. While in macho, sections always exist within a specific segment, they are actually two different views into the executable and are processed differently at runtime.

ebeahan: I suggest we go with using `nested` for `macho.segments.sections.*`. Any objection?

ebeahan:
I think we're nearly there with this one. Here's a summary of what's already completed and what's outstanding for advancing:

- [x] Identified sponsor. @devonakerr is sponsoring and has already reviewed ✅ 
- [x] Subject matter experts identified and weighed in
- [ ] Outlined initial field definitions.*

*I believe the thread [here](https://github.com/elastic/ecs/pull/1097#discussion_r574160603) is the last open discussion point, but I'll make another pass at the current field definitions in the proposal.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+## Fields
+
+**Stage 0**
+
+This RFC is to create the Mach-O sub-field within the `file.` fieldset. This will include 35 sub-fields.
+
+|   Name                                     |   Type     |   Description                                                               |
+|--------------------------------------------|------------|-----------------------------------------------------------------------------|
+|   macho.cpu                          |   object   |   CPU information for the file.                                             |
+|   macho.cpu.architecture             |   keyword  |   CPU architecture target for the file.                                     |
```
ebeahan, line 17: @andrewstucki I did some cross-checking between the proposed field list here and your work in https://github.com/elastic/beats/pull/24195. I pulled those mappings for `macho` into [a gist](https://gist.github.com/ebeahan/41507b47bcadff43f866350634ffa9ae) for easier reference.

100% alignment isn't necessary at all at this point. However, there are enough differences between that approach and what's captured here I think it's worth raising.

@peasead @dcode @devonakerr

rfcs/text/0000-create-file-macho.md
dcode: I agree with nested, but I think segments and sections should ideally match the usage as in ELF. Namely, we pull out sections to `file.macho.sections` and segments stay at `file.macho.segments`. The data source above nests them and it isn't wrong, but they are two different views into the binary. Sections also have physical offset data.

I'm going to resolve this conversation and suggest those changes (I can't seem to do it with this thread in the middle of it).

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+|   macho.segments.fileoff             |   keyword  |   File offset of this segment.                                              |
+|   macho.segments.filesize            |   keyword  |   Amount of memory to map from the file.                                    |
+|   macho.segments.sections            |   flattened   |   Section information for the segment of the file.                          |
+|   macho.segments.sections.flags      |   keyword  |   Section flags for the segment of the file.                                |
+|   macho.segments.sections.name       |   keyword  |   Section name for the segment of the file.                                 |
+|   macho.segments.sections.type       |   keyword  |   Section type for the segment of the file.                                 |
+|   macho.segments.offset      |   long  |   Offset of the segment.                                             |
+|   macho.segments.size     |   keyword  |   Segment limit size.                                            |
+|   macho.segments.flags     |   keyword  |   Segment flags.                                                 |
```
dcode, line 39: This suggestion matches the data interface to ELF, using consistent naming (physical offset vs file offset for example), aligns segments and sections information, and adds `entropy` and  `chi2` values to `macho.sections`.
```suggestion
|   macho.segments                 |   nested   |   Segment information for the file.                   |
|   macho.segments.name            |   keyword  |   Name of this segment.                               |
|   macho.segments.physical_offset |   long     |   File offset of this segment.                        |
|   macho.segments.physical_size   |   keyword  |   Amount of memory to map from the file.              |
|   macho.segments.virtual_address |   keyword  |   Memory address of this segment.                     |
|   macho.segments.virtual_size    |   keyword  |   Memory size of this segment.                        |
|   macho.segments.sections        |   keyword  |   Section names contained in this segment.            |
|   macho.sections                 |   nested   |   Section information for the segment of the file.    |
|   macho.sections.name            |   keyword  |   Section name for the segment of the file.           |
|   macho.sections.flags           |   keyword  |   Section flags for the segment of the file.          |
|   macho.sections.type            |   keyword  |   Section type for the segment of the file.           |
|   macho.sections.physical_offset |   long     |   Section List offset.                                |
|   macho.sections.physical_size   |   long     |   Section List physical size.                         |
|   macho.sections.virtual_address |   long     |   Section List virtual address.                       |
|   macho.sections.virtual_size    |   long     |   Section List virtual size.                          |
|   macho.sections.entropy         |   float    |   Shannon entropy calculation from the section.       |
|   macho.sections.chi2            |   float    |   Chi-square probability distribution of the section. |
```

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Stage 1**
+
+This type of data can be provided by logs from VirusTotal, Reversing Labs, Lockheed Martin's LAIKABOSS, Emerson's File Scanning Framework, Target's Strelka, or other file/malware analysis platforms.
+
+* [VirusTotal API](https://developers.virustotal.com/v3.0/reference)
+* [Emerson FSF](https://github.com/EmersonElectricCo/fsf)
+* [Target Strelka](https://github.com/target/strelka)
+* [Lockheed Martin LAIKABOSS](https://github.com/lmco/laikaboss)
+
```
dcode, line 72:
```suggestion
* [LIEF Analysis Library](https://lief.quarkslab.com/doc/latest/api/python/macho.html)

```

rfcs/text/0000-create-file-macho.md
dcode: I have since updated this PR with a proposed layout that aligns sections and segments with ELF and PE (which doesn't have segments).

ebeahan: Thanks, @dcode. 

There's still some large differences between the implementation in elastic/beats#24195 and what's proposed here.

Again, not something we need to completely resolve now. At a minimum, let's capture the difference in approaches as a `Concern` to make sure we revisit the topic in the next stage's PR.

dcode: @andrewstucki the strategy I'm suggesting here is splitting out fat binary components that can be unified by the same file metadata (namely hashes or other source unique id's). That will allow analyzing all binaries by their sections, segments, symbols, etc in a seamless way, regardless of platform or if their multi-platform. Do you think that can work for your implementation?

peasead: Just pining to see if there was a resolution here.

andrewstucki: Just as a use-case example: let's say we're triggering an alert on a malicious binary that has been compiled for both the new m1 as well as older, intel-based chipsets and we want to add in object-level data about the file. Would we emit two separate alerts, one for each architecture? Seems kind of odd.

peasead:
Moved to new PR https://github.com/elastic/ecs/pull/1346

PR #1098, opened 2020-11-05
Stage 3 changes for wildcard RFC 0001

ebeahan:
Apply changes from stage 3 Wildcard Migration RFC (0001): https://github.com/elastic/ecs/pull/1015

[Docs Preview](https://ecs_1098.docs-preview.app.elstc.co/diff)

TODO:

- [x] #1051 merged
- [x] Add beta attributes based on feature set finalized in #1051 
- [x] Determine if `./use-cases` Markdown files should be updated. Currently causing ❌ in Travis.
- [x] Adjust `event.original` field definition offering an option to use `wildcard` + `index: true`
- [x] #1103 Include `process.thread.name` to `wildcard` changes.
- [x] #1171 ES 6 template should not include `wildcard` and other types not available in ES 6.x.
- [x] #1176 Preserve `doc_values` and `index` settings for ES 6 `error.stack_trace` field.


webmat:
Looking good so far. I've cross-referenced all migrated fields with the RFC, the only one I've spotted missing is `process.thread.name`.

The failure here is just a failure to run `make`, no? :-) But in any case, the use cases directory has to go (https://github.com/elastic/ecs/pull/632). It hasn't been maintained in over a year. So if there's an actual problem with that script, we can fast track that PR, bring it up to date and proceed with the removal, IMO.

ebeahan:
Yeah, the CI failure is due to `make` updating a handful of the `./use-cases` markdown files. Happy to include them here, but I wanted to double-check on taking that path since the use cases directory isn't actively maintained nowadays.

Good catch with `process.thread.name`. I was using `experimental/schemas` as my guide, and sure enough, it's also missing from the current experimental definitions 😞 . I'll address that as well 😄 .

schemas/error.yml
```diff
@@ -40,7 +40,6 @@
     - name: stack_trace
       level: extended
       type: wildcard
-      index: false
```
webmat, line 42: https://webmat.keybase.pub/funny-stuff/dalai_lasers.gif

ebeahan: Yeah, I removed it since it's unnecessary with `type: wildcard`.

However, as I think about this case further, removing `index: false` will mean `error.stack_trace` will be indexed even for `keyword`-fallback cases like the ES 6.x template or `--oss`. 🤔

ebeahan: If we did want to leave `index: false` intact for any non-wildcard use cases, two considerations:

1. How to best capture this in the field definition. Having `type: wildcard` and `index: false` could create misunderstanding and confusion.
 
2. Address that the `index: false` and `doc_values: false` settings are being stripped in the ES 6 template (what originally led me to remove this line).

webmat: It's going to feel a bit dirty, but I wonder if we shouldn't hardcode the details for this fallback, or something.

Doing like `event.original` and leaving it non-indexed in the main schema is a no-go, as `wildcard` on stack traces will be objectively fantastic, for developers.

On the other hand, redoing how basic => oss translation works to better handle this edge case is way out of scope, IMO. This is just a "best effort" equivalency. From ECS 1.8 moving forward, ECS is using basic data types. (<= this "period" should be said out loud). So folks are free to override what's happening with this field however works best for them, if they need to stick to OSS and want to use this field. But really they should upgrade to the free basic option, which supports `wildcard`.

So in short, I think we should:

- leave the main schema using `wildcard` and no `index` attribute
- in the OSS fallback we should hardcode something so that `error.stack_trace` puts this back to `keyword` and `doc_values:false` and leave it at that.

I think this field and event.original were the only two, correct?

ebeahan: > I think this field and event.original were the only two, correct?

Correct.

I agree with the best-effort support around these nuances and favoring the simple workaround for now :-)

ebeahan: A "quick and simple" implementation to address proposed in #1176

ebeahan:
This one is ready for a final look @webmat.

cc @epixa

webmat (approved):
Alright LGTM 🎉

PR #1103, opened 2020-11-10
Add missing `process.thread.name` to experimental definitions

ebeahan:
Addresses issue noted originally in https://github.com/elastic/ecs/pull/1098#issuecomment-724783960.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ All notable changes to this project will be documented in this file based on the
 
 * The `protocol` allowed value under `event.type` should not have the `expected_event_types` defined. #964
 * Clarify the definition of `file.extension` (no dots). #1016
-* Include `process.thread.name` in the experimental definitions. #1053
 
 #### Added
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file based on the
 
 * The `protocol` allowed value under `event.type` should not have the `expected_event_types` defined. #964
 * Clarify the definition of `file.extension` (no dots). #1016
+* Include `process.thread.name` in the experimental definitions. #1053
```
webmat, line 14: I just now noticed that the changelog for the creation of the experimental artifacts was in the tooling section. However it's a big change schema-wise. I think we should move this here.

I'm saying this on this unrelated PR, because I'm not sure I would add a bugfix to something that hasn't been released yet. I was going to suggest to just append this PR number to the experimental artifacts changelog.

Something like the following, in the "Added" section for the schema:

```
* Added a new directory with experimental artifacts, which includes all changes
  from RFCs that have reached stage 2. #993, #1053
```

webmat (commented):
The addition is fine. But I have a tweak to suggest about the changelog.

CHANGELOG.md
ebeahan: ++ I think this is a better change.

webmat (approved):
Alright, thanks! 🚢

Issue #839, opened 2020-05-07
This issue is meant to collect breaking changes we want to do for ECS 8.0


webmat:
Any suggestion for breaking changes should be submitted via their own github issue, not as a comment on this issue.

#### Schema

* #1363 Remove guidance to lowercase `http.request.method`
* https://github.com/elastic/ecs/pull/1567 Remove or relocate the Go library and the code generating it
   * Need to confirm if the library will remain a Beats dependency in future releases.
* https://github.com/elastic/ecs/pull/1580 [RFC 0017](https://github.com/elastic/ecs/blob/master/rfcs/text/0017-remove-log-original.md) Remove `log.original`
* https://github.com/elastic/ecs/issues/1362 Remove `host.user.*`
* https://github.com/elastic/ecs/pull/1596 Remove `process.ppid`, now that we have all of `process.parent.*` (which includes the equivalent field `process.parent.pid`)

#### Tooling

 * #1340 Remove `--oss` flag
 * #632 Remove [`use-cases`](https://github.com/elastic/ecs/tree/master/use-cases) directory and associated code
 * #630 

#### Tooling

 * https://github.com/elastic/ecs/issues/1587 Remove legacy templates as generated artifacts.

### No longer under consideration for 8.0

Items previous listed as `Under consideration` but subsequently removed. May still have value in the future, so capturing below.

* #564 Do we need to change the type of `http.response.status_code` to `keyword`, in order to allow users to bucket HTTP traffic by status code?
* #887 #1309 Introduce the `version` data type for version fields
* https://github.com/elastic/ecs/pull/1341 Add other types of labels, potentially nested within `labels.*`, e.g. `labels.long.*`
* Does migrating `@timestamp` from `date` to `date_nanos` require a breaking change, or are the types compatible? https://github.com/elastic/ecs/issues/1065
    * The two types are compatible, but with 64-bit integers not supported in Javascript, Kibana has several outstanding issues (https://github.com/elastic/kibana/issues/40183, https://github.com/elastic/kibana/issues/41441,  https://github.com/elastic/kibana/issues/88290) that would create a degraded experience if `data_nanos` was the default for `@timestamp`, I think.
    * May still be able to provide guidance for those ECS users who require the additional precision of `date_nanos` and understand the possible side-effects.
* #909 Update singular names to plural in the `related.*` field set. 
    * The cost of introducing this as a breaking change outweighs the inconsistency in the naming. Not pursuing at this time.
[...]

PR #1111, opened 2020-11-11
Add os.type field

### LINKED ISSUE #1110: Add one new OS field with predictable set of values
**Summary**

The current OS fields in ECS were inherited from Beats, and didn't come with very strong guidance on how to populate them. This has led to a few of the fields being used inconsistently across data sources (mainly os.platform, os.family and to a lesser extent os.name, os.full).

This in turn means it's difficult to reliably filter for "Linux" events, for example. Currently some sources put "Linux" in one of the fields, some other sources put distro details in some of these fields (e.g. rhel / redhat) with no trace of the word Linux in sight 🙂 

**Motivation**:

We want to make it easy to filter for Windows, Linux, Unix and MacOS events. They're broadly speaking the main commercial platforms out there.

**Detailed Design**:

The proposal is to add a single new field that will have allowed values that need to be used.

Field name: `os.type`
Field type: `keyword`
Allowed values (lowercase):

* windows
* linux
* unix
* macos

Note: This issue is not about revamping the guidance on the OS field set; although this would be worth doing, in preparation for the next major release.

webmat:
See problem statement in #1110

Closes #1110


webmat:
@randomuserid Does this satisfy the need?

randomuserid:
> @randomuserid Does this satisfy the need?

It should enable us to test for platform type such as Linux, Mac, or Windows. TBH I think using something simpler and easier to remember like `host.os.name` - which almost satisfied the end today - would be a better user experience. We have a lot of host.os fields now, some of which contain the same information, and if we could simply them, and consolidate them, I think it would be easier to use and also help reduce the compute and storage costs of processing all of these fields, and would tend to be a better UX then adding an eight field to the set.

webmat:
Agreed on the need for simplification. But this will lead to breaking changes and will therefore take more time to implement.

So this new field is the only way we can solve the immediate need and meet your deadline ;-)

randomuserid:
> Agreed on the need for simplification. But this will lead to breaking changes and will therefore take more time to implement.
> 
> So this new field is the only way we can solve the immediate need and meet your deadline ;-)

Right - what is the definition of a "breaking change?" - a change to the current de facto usage of a field? Including cases where a field has not defined legal values and is being populated with different unofficial values by different teams?

randomuserid (changes requested):
I think what we need to do, in order to avoid the same problem where different models are placing different values in the field, is to adopt the design principle of normalization for fields that need to work across data streams. This would mean we explicitly define legal and expected values for the field, rather than giving examples or suggestions, and normalized fields would need to have defined legal values when they ship. Then search authors can look up the legal values when writing queries with the expectation they will work everywhere as long as legal values are being used. And data providers can be confident their events will be processed by queries using normalized fields.

docs/field-details.asciidoc
```diff
@@ -3853,6 +3853,21 @@ The OS fields contain information about the operating system.
 
 // ===============================================================
 
+| os.commercial_family
+| Categorize the operating system in one of the broad commercial families.
+
+One of these following values should be used (lowercase): linux, macos, unix, windows.
```
randomuserid, line 3859: maybe we should also make "other" a legal value..for any edge cases where someone out there has data from something we could not anticipate

webmat: Here I'd rather add specific guidance to "not populate the field if none of the options fit, but please let us know". Then we keep adding allowed values as needed.

webmat:
@randomuserid I agree with the need to tighten things up for a few of the fields in the OS field set. This is something we're tracking on the list of breaking changes under consideration for the next major release of ECS (https://github.com/elastic/ecs/issues/839). We don't have an issue for this yet. But I think we're aligned: we need to clarify how to use these fields for each of the commercial OS family. We could even come out with this in ECS 1.x as a soft guidance, and make it a hard requirement in 8.0.

> what is the definition of a "breaking change?"

The reason I'm hesitant to repurpose one of the existing fields for this mandated list of values is that we'll need to have this vetted by too many folks, and we likely wouldn't have much adoption anyway. It's not only about populating the field with these new values in the `add_host_metadata` processor. There's also potential impact on observability & security UIs, making this change in both user agent processors (LS & ES), there's likely Observability alerts that depend on OS details as well. And I'm sure I'm forgetting a bunch of things still :-)

Adding this new field, on the other hand, will mean it's either absent or adopted the right way when it is, during the 7.x line. So you would actually be able to depend on at the start of your predicate.

randomuserid (approved):
Reluctantly approved due to time pressure. I'd rather wait and use host.os.name.

 Actually, could we please get rid of the `commercial_` and simplify it too `host.os` ?

webmat:
As discussed out of band, we're changing the proposal to use a much shorter field name: `os.type`.

I like the symmetry with `event.type`, another field with a list of allowed value.

CHANGELOG.next.md
```diff
@@ -18,6 +18,7 @@ Thanks, you're awesome :-) -->
 
 * Added `event.category` "registry". #1040
 * Added `event.category` "session". #1049
+* Added `os.commercial_family`. #1111
```
ebeahan, line 21:
```suggestion
* Added `os.type`. #1111
```

schemas/os.yml
```diff
@@ -13,6 +13,20 @@
...
+      level: extended
+      type: keyword
+      short: 'Which commercial OS family (one of: linux, macos, unix or windows).'
+      description: >
+        Use the `os.type` field to categorize the operating system in one of
+        the broad commercial families.
+
+        One of these following values should be used (lowercase): linux, macos, unix, windows.
+
+        If the OS is not part of any of this list, the field should not be populated.
```
ebeahan, line 26:
```suggestion
        If the OS is not part of any family on this list, the field should not be populated.
```

Minor adjustment suggestion

schemas/os.yml
```diff
@@ -13,6 +13,20 @@
   type: group
   fields:
 
+    - name: type
+      level: extended
+      type: keyword
+      short: 'Which commercial OS family (one of: linux, macos, unix or windows).'
+      description: >
+        Use the `os.type` field to categorize the operating system in one of
```
ebeahan, line 21:
```suggestion
        Use the `os.type` field to categorize the operating system into one of
```

schemas/os.yml
```diff
@@ -13,6 +13,20 @@
...
+        the broad commercial families.
+
+        One of these following values should be used (lowercase): linux, macos, unix, windows.
+
+        If the OS is not part of any of this list, the field should not be populated.
+        Please let us know by opening an issue with ECS, to have it added to the list.
```
ebeahan, line 27:
```suggestion
        Please let us know by opening an issue with ECS, to propose adding it to the list.
```
Minor adjustment.

I think this wording better primes users to understand any additional values will need to be discussed and mutually agreed on.

ebeahan (commented):
No issues with pivoting to `os.type`.

I made notes on a few small points.

schemas/os.yml
webmat: Good idea. It's always good to know in advance one's dreams *might* be shattered 😂

schemas/os.yml
webmat: Still not a fan of the sentence here. Let's see if we can reword this some more.

docs/field-details.asciidoc
```diff
@@ -3930,6 +3930,23 @@ example: `darwin`
 
 // ===============================================================
 
+| os.type
+| Use the `os.type` field to categorize the operating system into one of the broad commercial families.
+
+One of these following values should be used (lowercase): linux, macos, unix, windows.
```
andrewstucki, line 3936: 😢 on the `unix` -- I miss my AIX and HPUX boxes

ebeahan (dismissed):
LGTM 👍

randomuserid (approved):

ebeahan (approved):

PR #1112, opened 2020-11-11
Add support for constant_keyword's 'value' parameter

### LINKED ISSUE #1052: Support the `value` parameter for `constant_keyword` fields
The [`constant_keyword` data type](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#constant-keyword-field-type) accepts the `value` parameter. On a `constant_keyword` field if the `value` parameter is specified, that value is used for all documents in the index. Otherwise, it is set based on the first document that gets indexed.

If a user wishes to use the ECS tooling to manage their ES index templates, they may also wish to control the value of `value` via their custom field definitions.

Example definition:

```yaml
  - name: acme
    title: acme
    group: 2
    short: Fields describing acme-related needs.
    description: >
      Acme-related needs
    fields:
      - name: stream
        description: stream
        level: extended
        type: constant_keyword
        value: widgets
```

webmat:
This PR adds support for setting the `value` parameter for `constant_keyword` fields.

The PR right now only lets the parameter through to the ES templates, but not the Beats field definitions. This is intentional, as ECS cannot logically set the type of some fields to `constant_keyword` on the behalf of Beats. This would have to be an override on the Beats side, according to the relevant indexing strategy.

Closes #1052

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -36,6 +36,7 @@ Thanks, you're awesome :-) -->
 * Added support for `scaled_float`'s mandatory parameter `scaling_factor`. #1042
 * Added ability for --oss flag to fall back `constant_keyword` to `keyword`. #1046
 * Added support in the generated Go source go for `wildcard`, `version`, and `constant_keyword` data types. #1050
+* Added support for `constant_keyword`'s optional parameter `value`. #1112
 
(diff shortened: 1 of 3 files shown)


ebeahan (approved):
LGTM 👍

PR #1114, opened 2020-11-12
Clarify x509 definition guidance for network events with only one cert

### LINKED ISSUE #1044: Clarify where to place x509 field set
**Description of the issue:**

The ECS docs say

> When only a single certificate is logged in an event, it should be nested under file.

This could be interpreted to apply when dealing with a TLS connection where only the server presents a certificate. The text here could be reworked to clarify when to use `file` vs `tls.server/client`.

webmat:
Closes #1044


webmat:
@dcode We had some confusion where folks faced with TLS connections with only a server cert populated `file.x509`, following initial guidance.

Does this new guidance still make sense to you? Are we missing anything?

dcode (approved):
LGTM and is consistent with the original intent.

ebeahan (approved):
LGTM 👍

PR #1115, opened 2020-11-12
Remove index parameter for wildcard fields

### LINKED ISSUE #1113: [Bug][ECS 1.7] Strip `index` attributes on `wildcard` mappings
**Description of the issue:**

The 1.7 experimental `wildcard` mappings need to have the `index` attribute stripped from them. The one that was causing problems for me is here:

https://github.com/elastic/ecs/blob/3edb6f2cf2c657f88875011625cd1801709c5d5b/experimental/generated/beats/fields.ecs.yml#L1154-L1163

Here's the error I'm getting from Elasticsearch:

```
Failed to parse mapping: unknown parameter [index] on mapper [stack_trace] of type [wildcard]
```

ebeahan:
#### Summary

Elasticsearch will reject an index template that contains a wildcard field with the `index` parameter. This change removes the `index` parameter entirely from any `wildcard` field.

#### Background

Wildcard requires that the field be indexed. If the `index` parameter is present at all, even `index: true`, Elasticsearch will reject the template as of Elasticsearch 7.10.0.

#### Questions

- [ ] With this change, the `index` attribute can still be included on a wildcard field. However, it would only serve as documentation. Is this useful, or should `index` be removed entirely if `wildcard` is the type?

Closes #1113

### DIFF
--- a/scripts/schema/cleaner.py
+++ b/scripts/schema/cleaner.py
@@ -144,6 +144,9 @@ def field_or_multi_field_datatype_defaults(field_details):
         field_details.setdefault('ignore_above', 1024)
     if field_details['type'] == 'text':
         field_details.setdefault('norms', False)
+    # wildcard needs the index param stripped
+    if field_details['type'] == 'wildcard':
+        field_details.pop('index', None)
     if 'index' in field_details and not field_details['index']:
         field_details.setdefault('doc_values', False)
 
--- a/scripts/tests/unit/test_schema_cleaner.py
+++ b/scripts/tests/unit/test_schema_cleaner.py
@@ -223,7 +223,7 @@ class TestSchemaCleaner(unittest.TestCase):
         cleaner.field_defaults({'field_details': field_details})
         self.assertEqual(field_details['doc_values'], False)
 
-        field_details = {**field_min_details, **{'type': 'wildcard'}}
+        field_details = {**field_min_details, **{'type': 'wildcard', 'index': True}}
         cleaner.field_defaults({'field_details': field_details})
         self.assertNotIn('index', field_details)
 
--- a/scripts/tests/unit/test_schema_cleaner.py
+++ b/scripts/tests/unit/test_schema_cleaner.py
@@ -223,6 +223,10 @@ class TestSchemaCleaner(unittest.TestCase):
         cleaner.field_defaults({'field_details': field_details})
         self.assertEqual(field_details['doc_values'], False)
 
+        field_details = {**field_min_details, **{'type': 'wildcard'}}
+        cleaner.field_defaults({'field_details': field_details})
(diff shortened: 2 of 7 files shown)


ebeahan:
**NOTE**: The Elasticsearch template generator already strips any instances of `index: true` in its mappings. This is why only the `experimental/generated/ecs/*` and `experimental/generated/beats/fields.ecs.yml` artifacts were modified with the changes here.

scripts/tests/unit/test_schema_cleaner.py
```diff
@@ -223,6 +223,10 @@ def test_field_defaults(self):
         cleaner.field_defaults({'field_details': field_details})
         self.assertEqual(field_details['doc_values'], False)
 
+        field_details = {**field_min_details, **{'type': 'wildcard'}}
```
webmat, line 226: `index` is not in field_min_details, you should add it in the `{'type': 'wildcard'}` dictionary, if you want to test that it properly gets removed

scripts/schema/cleaner.py
```diff
@@ -144,6 +144,9 @@ def field_or_multi_field_datatype_defaults(field_details):
         field_details.setdefault('ignore_above', 1024)
     if field_details['type'] == 'text':
         field_details.setdefault('norms', False)
+    # wildcard needs the index param stripped
+    if field_details['type'] == 'wildcard':
+        field_details.pop('index', None)
```
webmat, line 149: Yeah I think this is safe and the right approach for now.

But the more I think about this, the more this sounds wrong to have this be the behaviour of Elasticsearch. I'm sure this is simply something that was overlooked, not a hard requirement.

Let's move forward with this PR to resolve the need for 1.7, but could you also open an issue with Elasticsearch, to ask for `index` to be supported for `wildcard` (and ideally for all types)?

scripts/tests/unit/test_schema_cleaner.py
```diff
@@ -200,6 +200,7 @@ def test_field_defaults(self):
             'level': 'extended',
             'name': 'my_field',
             'type': 'unknown',
+            'index': 'true',
```
webmat, line 203: I don't think you should add this there, it could invalidate the tests on lines 219-225.

scripts/tests/unit/test_schema_cleaner.py
webmat:
```suggestion
        field_details = {**field_min_details, **{'type': 'wildcard', 'index': True}}
```

scripts/tests/unit/test_schema_cleaner.py
ebeahan: Thanks, corrected.

webmat (approved):
Alright, LGTM! :shipit:

PR #1117, opened 2020-11-12
Indicate when artifacts include experimental changes

webmat:
I'm opening this so that we give an indication on artifacts that include the experimental ECS changes.

This is the simplest implementation possible that shows up in each of the artifacts that include a mention of the version. It's very terse, however. If we go this route, we should at least include a mention of the meaning of this `+exp` in the artifact readme or elsewhere. Note that on a release tag, the version for experimental artifacts would look like `1.7.0+exp`.

I'm open to doing this differently however. We could instead include an optional comment line in the Beats YML file stating that it includes ECS experimental changes, include a second meta attribute in the ES template to that effect, and not sure how we'd handle the CSV. This approach might be more self-contained for someone looking at the artifact elsewhere.


webmat:
@ebeahan @andrewstucki Thoughts on how to message this? Do you think this terse approach is fine?

I would err towards this terse approach + a small blurb to explain the meaning of this in the repo (still to be written).

andrewstucki:
I like it, it keeps the experimental stuff as part of the semver, better than what I was thinking

webmat:
> it keeps the experimental stuff as part of the semver

Good point, nice! And to think it's my laziness that got us there 😂

scripts/generator.py
```diff
@@ -40,6 +40,10 @@ def main():
     # statements like this after any step of interest.
     # ecs_helpers.yaml_dump('ecs.yml', fields)
 
+    # Detect usage of experimental changes to tweak artifact headers
```
webmat, line 43:
```suggestion
    # Detect usage of experimental changes to tweak artifact version label
```

ebeahan (approved):
++ to the terse approach.

LGTM 👍

Issue #1116, opened 2020-11-12
[Bug][ECS 1.7] Missing `dns.answers` fields in experimental schema


andrewstucki:
**Description of the issue:**

Compare 

https://github.com/elastic/ecs/blob/6cbb5448edf921ed9b8b797bd8310003f109259a/experimental/generated/beats/fields.ecs.yml#L969-L970

With

https://github.com/elastic/ecs/blob/6cbb5448edf921ed9b8b797bd8310003f109259a/generated/beats/fields.ecs.yml#L986-L987

The first is for the `experimental` fields, the second is for the standard fields. Looks like the experimental (accidentally?) drops `dns.answers` -- not sure if there are any other fields like this -- but this one was caught by `packetbeat` integration tests.

ebeahan:
The `dns.answer` object contains `intermediate: True` when the [`dns.answer.data` object](https://github.com/elastic/ecs/blob/master/experimental/schemas/dns.yml#L6) is loaded and the deeply nested structure is generated. When working properly, it sets `intermediate: False`. I'm still assessing the best way to handle this case in the code.

This isn't unique to the experimental schema. Any custom definitions modifying `dns.answers.*` will reproduce this behavior.

A workaround is to define `answer` with type `object` in the experimental definitions:

```yaml
---
- name: dns
  fields:
    - name: question.name
      type: wildcard
    - name: answers
      type: object
    - name: answers.data
      type: wildcard
```

webmat:
I'd be fine if we did that as a temporary fix for the release. We can keep this issue opened to fix fully later, without the rush :-)

ebeahan:
Temporary workaround added in #1118

djptek:
@ebeahan working on this

PR #1118, opened 2020-11-12
Add dns.answer object into experimental schema

ebeahan:
Implements the workaround identified in #1116, which prevents the `dns.answer` object from being omitted when merging the experimental field definitions.


webmat (dismissed):
LGTM

webmat (approved):

PR #1131, opened 2020-11-18
Usage of brackets for a URL containing IPv6 address

ebeahan:
#### Summary

Detail that the `[` and `]` characters may enclose a literal IPv6 address populating `url.domain`.

#### Background

The current `url.domain` field description doesn't have guidance for this case (e.g. https://github.com/elastic/elasticsearch/pull/65150#discussion_r525635429). The format is defined in [RFC 2732](https://www.ietf.org/rfc/rfc2732.txt) and has been adopted by the major web browsers.

Example:

* Literal IPv6 address: `FEDC:BA98:7654:3210:FEDC:BA98:7654:3210`
* Formatted in a URL: `http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]/index.html`

### DIFF
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -5262,7 +5262,11 @@
       description: 'Domain of the url, such as "www.elastic.co".
 
         In some cases a URL may refer to an IP and/or port directly, without a domain
-        name. In this case, the IP address would go to the `domain` field.'
+        name. In this case, the IP address would go to the `domain` field.
+
+        If the URL contains a literal IPv6 address, `domain` may include the `[` and
+        `]` characters enclosing the address (https://www.ietf.org/rfc/rfc2732.txt[RFC
+        2732]).'
       example: www.elastic.co
     - name: extension
       level: extended
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -5264,9 +5264,9 @@
         In some cases a URL may refer to an IP and/or port directly, without a domain
         name. In this case, the IP address would go to the `domain` field.
 
-        If the URL contains a literal IPv6 address, `domain` may include the `[` and
-        `]` characters enclosing the address (https://www.ietf.org/rfc/rfc2732.txt[RFC
-        2732]).'
+        If the URL contains a literal IPv6 address enclosed by `[` and `]` (IETF RFC
+        2732), the `[` and `]` characters should also be captured in the `domain`
+        field.'
       example: www.elastic.co
     - name: extension
       level: extended
--- a/schemas/url.yml
+++ b/schemas/url.yml
@@ -58,6 +58,9 @@
 
         In some cases a URL may refer to an IP and/or port directly, without a
         domain name. In this case, the IP address would go to the `domain` field.
+
+        If the URL contains a literal IPv6 address, `domain` may include the `[` and `]`
+        characters enclosing the address (https://www.ietf.org/rfc/rfc2732.txt[RFC 2732]).
       example: www.elastic.co
 
     - name: registered_domain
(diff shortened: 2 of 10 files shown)


experimental/generated/beats/fields.ecs.yml
```diff
@@ -5262,7 +5262,11 @@
       description: 'Domain of the url, such as "www.elastic.co".
 
         In some cases a URL may refer to an IP and/or port directly, without a domain
-        name. In this case, the IP address would go to the `domain` field.'
+        name. In this case, the IP address would go to the `domain` field.
+
+        If the URL contains a literal IPv6 address, `domain` may include the `[` and
```
andrewkroh, line 5267: My preference would be for stronger wording (`shall` over `may`) to encourage consistency. Something along the lines of

> If the original URL contains a literal IPv6 address enclosed in brackets then `url.domain` shall include those brackets.

webmat: Yeah I agree with the stronger wording. We haven't been using "shall" much if at all, in ECS. Perhaps we stick with "must"?

schemas/url.yml
```diff
@@ -58,6 +58,9 @@
 
         In some cases a URL may refer to an IP and/or port directly, without a
         domain name. In this case, the IP address would go to the `domain` field.
+
+        If the URL contains a literal IPv6 address, `domain` may include the `[` and `]`
+        characters enclosing the address (https://www.ietf.org/rfc/rfc2732.txt[RFC 2732]).
```
webmat, line 63: We can link when necessary, but the asciidoc links are also clunky in our descriptions, especially when they appear in artifacts other than the documentation.

I think we could get rid of this one and just say "IETF RFC 2732". Note that I'm adding an explicit mention of "IETF", because we now also have our own RFCs :-)

WDYT @ebeahan?

experimental/generated/beats/fields.ecs.yml
sgryphon: The next strongest wording would be `should`, which indicates a strong preference, but won't, for example, start treating things as invalid if it isn't met.

schemas/url.yml
ebeahan: ++ sounds sensible.

experimental/generated/beats/fields.ecs.yml
ebeahan: I wanted to avoid inadvertently leading anyone to alter the value by adding brackets if missing in the original (albeit an unlikely scenario).

Perhaps something like:

> If the URL contains a literal IPv6 address enclosed by `[` and `]` (IETF RFC 2732), the `[` and `]` characters must also be captured in the `domain` field.

dainperkins:
Why are we populating a domain field with an IP address? If we are basing things on RFCs we should have a url.host field (name or ip) and a url.domain fields aligned with the DNS implementation shouldn't we?

I'd think the representation of the url would be dependent on what is reported (which should be []'d for ipv6 I would like to thitnk, but if the browser is not using that rfc I wouldn't want to change it...)

experimental/generated/beats/fields.ecs.yml
webmat: ++ on not leading people to alter the value, if the IPv6 is not enclosed in brackets.

I think @ebeahan's wording suggested above works well. I'm also fine with using the word "should", as suggested by @sgryphon.

webmat:
@dainperkins adding another field would be out of scope for this adjustment to the description.

The new guidance around IPv6 addresses remains in line with the previous guidance of populating this field with an IP in `.domain`, when the URL as an IP for the host. I agree this is not perfect.

But I get the sense that adding another field here leads us to the redesign of normalizing DNS names a bit more across ECS we've been talking about. It's a good idea to look at that, but I suspect it will be a big change that requires an RFC.

ebeahan:
Thanks all for the feedback! I incorporated the suggestions and would appreciate a final 👀 .

andrewkroh (approved):
LGTM

PR #1148, opened 2020-11-24
Make all fields linkable directly via an HTML ID

### LINKED ISSUE #1075: Documentation Suggestion: anchor links on ECS fields
This is really pretty simple and I'm not sure where else to properly capture it. 

The existing ECS documentation should have anchor tags on each field name in the ECS guide so you can easily deep-link directly to the relevant field. 

When collaborating, often I'll say "Look at this field(such as rule.id)". I can link the fields [category page](https://www.elastic.co/guide/en/ecs/current/ecs-rule.html), but I can't deep-link directly to the field of discussion. 

Just a small quality of life change that could speed up collaboration.

webmat:
The aim of this PR is to make all ECS fields linkable directly. This will enable collaborators to send deep links to the ECS documentation, to specific fields rather than be limited to linking to a whole field set.

I consider this a WIP for now. The IDs have been added on each field, effectively making them linkable already.  However there's no visual indicator when hovering over the field name, indicating that it's linkable (as opposed to our headings). I'll mark this as ready once I've figured out how to make this a bit more intuitive.

Closes #1075


webmat:
Got it to work with the help of @bmorelli25 ❤️ 

@jamiehynds and @MikePaquette how do you think this looks? Field names are all links now. But clicking them has the intended effect, and works otherwise like the linkable section headings :-)

webmat:
Here's the [docs preview](https://ecs_1148.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-base.html)

jamiehynds:
One small step for @webmat, one giant leap for ECS Docs 🚀 

Looks great, thanks Mat!

jamiehynds (approved):

PR #1151, opened 2020-11-25
Reinforce the exclusion of the leading dot from url.extension

webmat:
This adjustment is meant to match the improved guidance on `file.extension` we introduced in #1016.

I think we should backport this to 1.7 for it to be reflected in current docs. This guidance was already present in the description for `url.extension`, just not as obvious. It was in the long description, but not in the short description.

### DIFF
--- a/code/go/ecs/url.go
+++ b/code/go/ecs/url.go
@@ -87,11 +87,14 @@ type Url struct {
 	// differentiate between the two cases.
 	Query string `ecs:"query"`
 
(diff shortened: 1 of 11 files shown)


webmat:
[Doc preview](https://ecs_1151.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-url.html)

leehinman (approved):
Looks great.  Thank You.

webmat:
Thanks @leehinman

PR #1157, opened 2020-11-30
Correct the email RFC number, which was already in use

webmat:
The RFC no for email was assigned before the first merge. That number was already in use at the time we merged it.

This assigns RFC id 10 to the email RFC.


webmat:
@P1llus Just need a quick sanity check. RFC no 8 and 9 are already in use, so email will be no 10 :-)

ebeahan (approved):

PR #1158, opened 2020-12-01
Set the merge date on RFC 0005 stage 2

webmat:
Forgot to assign the date, when merging #1028


webmat:
@ebeahan 2 days in a row where I forget a detail when merging an RFC 🤦 

Let's see if we get the SIP adjustments ready by tomorrow: perhaps we can bring the streak to 3 in a row 😂

ebeahan (changes requested):
Can we also capture the correction PR, similar to what was done [here](https://github.com/elastic/ecs/pull/996/files#diff-7b8d76708f4fd4d90377e53f87fcb13e65aad42fa80997c76a31ea6d0de1e8f0R767)?

ebeahan (approved):
Thanks for the adjustment 👍

PR #1162, opened 2020-12-02
Add notice to the tracing field set, about not nesting field names.

webmat:
This understandable misunderstanding comes up every so often, so let's get that notice in :-)

[Doc preview of the change](https://ecs_1162.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-tracing.html)


ebeahan (approved):
👍 LGTM

PR #1164, opened 2020-12-02
Tracing fields should be at top level in Beats artifact

### LINKED ISSUE #1163: Incorrect output of the "tracing" fields in the Beats yml file
Just like the `base` fields, the `tracing` fields are not nested under the name of the field set. So it's not `base.@timestamp`, it's `@timestamp`, and it's not `tracing.trace.id`, it's `trace.id`.

In the Beats field yaml file the ECS project generates, the tracing fields are incorrectly nested under a `tracing` section, which means Beats interprets the field names incorrectly (`tracing.trace.id`).

This is a bug, these fields shouldn't be nested this way.

In order to fix this issue, we should remove this nesting in the Beats yml output. Just like `@timestamp` and other base fields are not nested under a field group.

I think this bug fix will be at minimum backported to 1.7. Thoughts welcome on this, is there a need to backport to 1.6 as well?

The Beats PR https://github.com/elastic/beats/pull/22571 to import ECS 1.7 should be adjusted with these changes, once the bug fix is ready. cc @andrewstucki

ebeahan:
The `fields.ecs.yml` artifact was erroneously adding the [tracing](https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html) fields into a `tracing` object.

Like `base` fields, the `tracing` fields should exist at the top-level and require special handling when generating the template.

Closes #1163

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -30,7 +30,7 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
-* `tracing` fields should be at top level of Beats `fields.ecs.yml` artifacts. #1164
+* `tracing` fields should be at root of Beats `fields.ecs.yml` artifacts. #1164
 
 #### Added
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -30,6 +30,8 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
+* `tracing` fields should be at top level of Beats `fields.ecs.yml` artifacts. #1164
+
 #### Added
 
 * Added ability to supply free-form usage documentation per fieldset. #988
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -13,8 +13,7 @@ def generate(ecs_nested, ecs_version, out_dir):
     allowed_fieldset_keys = ['name', 'title', 'group', 'description', 'footnote', 'type']
     # other fieldsets
     for fieldset_name in sorted(ecs_nested):
-        # skip fieldsets already added
-        if 'base' in fieldset_name:
+        if 'base' == fieldset_name:
             continue
         fieldset = ecs_nested[fieldset_name]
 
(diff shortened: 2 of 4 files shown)


scripts/generators/beats.py
```diff
@@ -10,10 +10,17 @@ def generate(ecs_nested, ecs_version, out_dir):
     # base first
     beats_fields = fieldset_field_array(ecs_nested['base']['fields'], df_whitelist, ecs_nested['base']['prefix'])
 
+    # tracing fields also need top-level handling
+    tracing_fields = fieldset_field_array(
+        ecs_nested['tracing']['fields'], df_whitelist, ecs_nested['tracing']['prefix'])
+    beats_fields.extend(tracing_fields)
+
     allowed_fieldset_keys = ['name', 'title', 'group', 'description', 'footnote', 'type']
+    fieldsets_to_skip = ['base', 'tracing']
```
webmat, line 19: The special handling for `base` here had two purposes. One of which was putting the fields at the root rather than nested, just like we're fixing for the tracing fields. But the other purpose was to literally [start the file](https://github.com/elastic/ecs/blob/master/generated/beats/fields.ecs.yml#L9-L45) with the base fields. Just like we display "Base" first in the list of field sets, in the documentation.

In order to correctly handle future field sets whose fields are also at the root, I think we should use the field set's `root` attribute. It's set to true only for field sets that shouldn't be nested.

So concretely, I would leave the special handling for `base` as it was, to keep the fields first in the file.

But in the loop we could test for `if fieldset['root']`, then either `.extend` or copy.

Note: I think we should try to avoid `root:true` field sets in general, because they cause confusion. But that doesn't mean they're not going to happen again :-)

webmat (dismissed):
LGTM, thanks for fixing this!

My comment below could either be addressed now or separately, your call.

CHANGELOG.next.md
```diff
@@ -30,6 +30,8 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
+* `tracing` fields should be at top level of Beats `fields.ecs.yml` artifacts. #1164
```
webmat, line 33: I think "root" is clearer.

```suggestion
* `tracing` fields should be at root of Beats `fields.ecs.yml` artifacts. #1164
```

scripts/generators/beats.py
ebeahan: This approach seems sensible since we already have the attribute present and set to true for these fields. 😄 

I've adjusted here.

webmat (dismissed):
LGTM

Issue #1167, opened 2020-12-03
Add fuzzy hashes (ssdeep)


andrewstucki:
**Summary**

For file/large byte data we have fields for various popular hashing algorithms: `md5`, `sha256`, etc. While I'd want to keep entity-specific hashing algorithms isolated to field sets related to them (`imphash` to PE, `ja3` to TLS, etc.), I think it would be worthwhile to add some standardized fuzzy byte hashes to `hash`. Specifically I think we could start with something like [`ssdeep`](https://ssdeep-project.github.io/ssdeep/index.html). If there are others that folks think are worthwhile, we could likely add them at the same time.

**Motivation**:

Fuzzy hashes would allow us to group by similar entities rather than identical. Helpful for things like similar file identification, etc.

**Detailed Design**:

New fields:

| Name | Type | Description |
| ---- | ---- | ----------- |
| `hash.ssdeep` | keyword | [SSDEEP](https://ssdeep-project.github.io/ssdeep/index.html) hash. |

PR #1169, opened 2020-12-04
Add ssdeep hash

andrewstucki:
Closes https://github.com/elastic/ecs/issues/1167


schemas/hash.yml
```diff
@@ -5,12 +5,16 @@
...
-    The hash fields represent different hash algorithms and their values.
+    The hash fields represent different bitwise hash algorithms and their values.
 
     Field names for common hashes (e.g. MD5, SHA1) are predefined. Add fields for other hashes
     by lowercasing the hash algorithm name and using underscore separators as appropriate
     (snake case, e.g. sha3_512).
 
+    Note that this fieldset should be used for common hashes that may be computed
+    over a range of generic bytes. Entity-specific hashes such as ja3 or imphash should be
+    placed in the fieldsets to which they relate (i.e. tls and pe).
```
ebeahan, line 16:
```suggestion
    placed in the fieldsets to which they relate (tls and pe, respectively).
```

nit: minor rewording suggestion

schemas/hash.yml
```diff
@@ -5,12 +5,16 @@
...
   description: >
-    The hash fields represent different hash algorithms and their values.
+    The hash fields represent different bitwise hash algorithms and their values.
 
     Field names for common hashes (e.g. MD5, SHA1) are predefined. Add fields for other hashes
     by lowercasing the hash algorithm name and using underscore separators as appropriate
     (snake case, e.g. sha3_512).
 
+    Note that this fieldset should be used for common hashes that may be computed
+    over a range of generic bytes. Entity-specific hashes such as ja3 or imphash should be
```
ebeahan, line 15:
```suggestion
    over a range of generic bytes. Entity-specific hashes, such as ja3 or imphash, are
```

nit: minor rewording suggestion

schemas/hash.yml
```diff
@@ -5,12 +5,16 @@
...
   short: Hashes, usually file hashes.
   description: >
-    The hash fields represent different hash algorithms and their values.
+    The hash fields represent different bitwise hash algorithms and their values.
 
     Field names for common hashes (e.g. MD5, SHA1) are predefined. Add fields for other hashes
     by lowercasing the hash algorithm name and using underscore separators as appropriate
     (snake case, e.g. sha3_512).
 
+    Note that this fieldset should be used for common hashes that may be computed
```
ebeahan, line 14:
```suggestion
    Note that this fieldset is used for common hashes that are computed
```

nit: minor rewording suggestion.

I believe we can say authoritatively all fields here are computed in this way? Or are there exceptions?

schemas/hash.yml
```diff
@@ -39,3 +43,13 @@
...
       description: SHA512 hash.
+
+    - name: ssdeep
+      level: extended
+      type: keyword
+      short: SSDEEP hash.
+      description: >
+        SSDEEP hash.
+
+        Learn more at https://ssdeep-project.github.io/ssdeep/index.html.
```
ebeahan, line 54: This link is broken due to the included `.`.

However, I suggest removing the link entirely. It's the top search hit for `ssdeep`, and removing it will make for one less link to maintain in the ECS docs.

WDYT?

ebeahan (commented):
I made a few comments concerning the descriptions, but `hash.ssdeep` seems like a useful, straight-forward field addition.

Is this related at all to the WIP VT module work from https://github.com/elastic/beats/pull/21815? If so, perhaps @dcode @peasead could also weigh in with any feedback?

andrewstucki:
@ebeahan -- it's independent of the VT module, but VT and other AV vendors do use ssdeep. I'd be more than happy for others to chime in though as it might expose some other suggested additions for generic fuzzy hashes.

schemas/hash.yml
```diff
@@ -5,12 +5,16 @@
   type: group
   short: Hashes, usually file hashes.
   description: >
-    The hash fields represent different hash algorithms and their values.
+    The hash fields represent different bitwise hash algorithms and their values.
```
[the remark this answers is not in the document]
andrewstucki: Yep, I'd be fine modifying the language to specifically call out the categories, but didn't necessarily want to preclude other widely used non-crypto hashes being added here in the future (i.e. FarmHash, MurmurHash, etc.).

At the same time, the "bitwise" was also to add clarification that we're not going to add what I refer to as "entity-specific" hashes here (i.e. fingerprinting hashes that are related to format-specific features). This field really is for hashes that can be computed over a generic chunk of bytes.

schemas/hash.yml
andrewstucki: All fields currently are computed that way, and this wording is to add that `hash` should stay that way, whereas any other "hash" that is based off of specific attributes of an object (like imphash) should go into a different set of fields.

ebeahan (approved):
LGTM 👍 

I'd still like @rw-access to weigh in with a quick review before merging.

rw-access (approved):

ebeahan:

PR #1173, opened 2020-12-04
Document considerations for OSS clusters

ebeahan:
Adds docs page covering the use of basic licensed data types in ECS and considerations for OSS-only users.

Items covered:

* Data type fallback
* Limitations for OSS users
* Fallback options for OSS Beats users (need to confirm the progress of feature)
* OSS-compatible ECS template for OSS Logstash users (also need to confirm the progress of feature)
* `--oss` support in ECS tooling with link to usage doc.

[Docs preview of this PR](https://ecs_1173.docs-preview.app.elstc.co/diff)


docs/using-data-types.asciidoc
```diff
@@ -0,0 +1,31 @@
+[[ecs-data-types]]
+=== Data Types
```
webmat, line 2: ECS is an Elasticsearch schema, and it's all about "data types" throughout. Giving this page the title "Data Types" makes it sound like it's going to be general information about data types in ECS, which is not the case.

I think this page should be named something that's clearly about OSS, so that most users know it doesn't concern them. Since it has to fit in the sidebar it has to be short, so it's hard to come up with a good one :-)

Some ideas:

* ECS in OSS clusters
* ECS with OSS clusters
* Basic datatypes in OSS clusters

docs/using-data-types.asciidoc
```diff
@@ -0,0 +1,31 @@
...
+
+IMPORTANT: The limitations noted in this section apply to the OSS-only distribution of the Elastic Stack. The default distribution with the Elastic license supports all data type used by ECS.
```
webmat, line 4: IMO the "default" wording makes more sense inside of Elastic than for the community. I would instead try to use the wording [on the subscriptions page](https://www.elastic.co/subscriptions): namely "Basic", "license", perhaps "tiers".

I would also avoid using the word "limitations". The stack is getting tons of new features, some in OSS, most in Basic for free, and some are in higher tiers only. 

So I'd recommend messaging this with points like this:

* All Elasticsearch data types are available in the free version of the Elastic stack, under the Basic license.
* The Elastic stack downloads are by default Basic license. Unless users specifically downloaded the OSS release, this page does not affect them.
* Some advanced data types are not available in the OSS release of the Elastic Stack.
* ECS now uses basic-licensed data types, and this page covers ways that users of the OSS release can get similar functionality by replacing some of the basic types with somewhat equivalent types available under OSS.

webmat: Note also that our tech writers or marketing folks could help tweak the messaging as well

docs/using-data-types.asciidoc
```diff
@@ -0,0 +1,31 @@
...
+
+Beginning with ECS 1.8.0, the ECS specification adopted additional Elasticsearch {ref-bare}/current/mapping-types.html[data types] only available in the default distribution of the Elastic Stack.
+
+OSS-only users can still benefit from normalizing data to ECS. Some licensed data types will have a fallback, but OSS-only users will not populate a field if a fallback is not available.
```
webmat, line 8: > but OSS-only users will not populate a field if a fallback is not available

I think this is potentially misleading or confusing. The shape of the JSON event will be unaffected by whatever type is inside Elasticsearch.

Some data types under the basic license have an equivalent but less capable fallback, and some types don't have an equivalent.

So far ECS only uses basic types that have a fallback. So perhaps we can omit this whole line of discussion for now.

But if you want to cover it, it's important to note that the JSON object can still be captured 100% in Elasticsearch (e.g. a structured field aimed at populating a `histogram` field can be shoved in an `object` field). It's just that OSS users won't have a roughly equivalent way to query it; and OSS indices will conflict with basic indices, for folks migrating to basic.

But to get back to my previous point, I think I'd avoid covering this part for now.

docs/using-data-types.asciidoc
```diff
@@ -0,0 +1,31 @@
...
+// ===============================================================
+
+| {ref-bare}/current/keyword.html#constant-keyword-field-type[constant_keyword]
+| {ref-bare}/current/keyword.html[keyword]
+
+|=====
+
+Some components will self-manage type fallback. OSS-only Beats will perform the fallback automatically, and OSS-only Logstash will install an OSS-compatible ECS template. The ECS tooling introduced a flag, `--oss`, for users to generate OSS-only compatible Elasticsearch index templates. The ECS tooling {ecs_github_repo_link}/USAGE.md#oss[usage guide] contains a detailed example.
+
+The use of fallback types are not a supported, official ECS variant but instead provided as a workaround.
```
webmat, line 31: I would separate the two points, and try to make each more direct.

```suggestion
The use of fallback types does not constitute an official variant of ECS. It's suggested as a workaround that brings most of the benefits of ECS to users of the purely OSS release of the Elastic stack.
```

webmat (commented):
Good work!

The challenge here will be getting the messaging right so that users aren't confused on whether that applies to them, and that the free and basic license is a good thing, not a "limitation".

So I think we need to get the wording right, but also adjust the importance we give to this.

I think putting this in the "Using ECS" section is giving this too much importance, for instance. Over 95% of our users use the default distribution, and are therefore not affected by this. I think it would be better to have this in "Additional Information".

Other wording comments below. I'm not a pro at messaging, though. Perhaps some of our marketing colleagues could help?

docs/using-data-types.asciidoc
jamiehynds: Great point about the placement of the document. Totally agree that it's appropriate for this to live under 'Additional Information' as it doesn't apply to the majority of users. 

Would field type make sense instead of data type (it's used in the subscription page).

Some ideas:
- ECS Considerations in OSS Clusters
- Licensed field types in OSS clusters 

Maybe we could add a link to the [subscriptions page ](https://www.elastic.co/subscriptions) as it shows wildcard is not available in OSS. Constant keywords isn't listed there, but I can look at getting it added.

webmat: ++ on "field type" wording

I think "ECS Considerations in OSS Clusters" works great, if it fits in the sidebar.

Not sure how multiline headings look in the sidebar, actually. If multiline works, I think "ECS Considerations in OSS Clusters" is my fav.

ebeahan: ++ good suggestions! I've made the update to `ECS Considerations in OSS Clusters`, and I think it looks acceptable in the sidebar.

docs/using-data-types.asciidoc
ebeahan: I took another pass, trying to capture these points better. I agree that we should get some additional 👀 to review the messaging after we're happy with our edits.

docs/using-data-types.asciidoc
ebeahan: Fair point - I removed this part.

docs/using-data-types.asciidoc
webmat: Happy to see this section title fits well in the sidebar.

Looking at it though, perhaps we should remove the leading "ECS"? I think it's a bit redundant. Perhaps something like "Considerations for OSS Clusters"... WDYT?

ebeahan: ++ I made this change.

webmat (approved):
LGTM

jamiehynds (approved):

ebeahan:

ebeahan:
Changes are no longer pertinent at this time.

PR #1176, opened 2020-12-08
Handle `error.stack_trace` case for ES 6.x template

ebeahan:
`error.stack_trace` needs special handling to re-add `index: false` and `doc_values: false` after the type fallback when the Elasticsearch 6.x legacy index template is generated.

Adding the logic to `generate_template_version` simplified the implementation, but to unit test would require investing additional time into mocking the file write at the end of the function. I opted not to invest that time now, but it is something I'd like to revisit.


webmat (approved):
Yeah that works 👍

PR #1180, opened 2020-12-09
[1.7] Add composable index templates artifacts (#1156)

webmat:
Backports the following commits to 1.7:

* Add composable index templates artifacts (#1156)
* Move _meta section back inside mappings, in legacy templates. (#1186)

### DIFF
--- a/experimental/generated/elasticsearch/7/template.json
+++ b/experimental/generated/elasticsearch/7/template.json
@@ -1,11 +1,11 @@
 {
+  "_meta": {
+    "version": "1.7.0+exp"
+  },
   "index_patterns": [
     "try-ecs-*"
   ],
   "mappings": {
-    "_meta": {
-      "version": "1.7.0+exp"
-    },
     "date_detection": false,
     "dynamic_templates": [
       {
--- a/experimental/generated/elasticsearch/7/template.json
+++ b/experimental/generated/elasticsearch/7/template.json
@@ -1,11 +1,11 @@
 {
-  "_meta": {
-    "version": "1.7.0+exp"
-  },
   "index_patterns": [
     "try-ecs-*"
   ],
   "mappings": {
+    "_meta": {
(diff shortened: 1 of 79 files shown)


webmat:
@ebeahan For this backport I had to undo the ES 6 fallback stuff, since it wasn't backported to 1.7.

I'll wait for your review on this one

experimental/generated/elasticsearch/7/template.json
```diff
@@ -1,11 +1,11 @@
 {
+  "_meta": {
+    "version": "1.7.0+exp"
+  },
```
webmat, line 4: @ebeahan Alright I took some liberties on where innocuous metadata was nested in this PR.

The 1.7 release (and all previous ones) incorrectly had the `_meta` fields inside the mappings. I wonder if I should modify this backport to not change this 🤔 

What do you think?

ebeahan: I tried testing out these changes to the legacy template, but `_meta` doesn't appear supported as a the top-level field in the legacy index template API. Trying to upload the template with `_meta` present resulted in this error:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "parse_exception",
        "reason": "unknown key [_meta] in the template "
      }
    ],
    "type": "parse_exception",
    "reason": "unknown key [_meta] in the template "
  },
  "status": 400
}
```

webmat: Oh thanks for testing that 👍 I should totally have done that! 

But this means I'll have to fix that in master as well.

webmat:
I've introduced the #1186 fix in this backport. So as expected, the legacy templates are no longer changing in the diff. Aka `_meta` is back where it belongs.

Ready for a final review when you're ready @ebeahan

ebeahan (approved):
LGTM 👍

PR #1182, opened 2020-12-09
[RFC] Add host metric fields to ECS (Stage 3)

kaiyan-sheng:
This PR is stage 3 of the RFC for adding host metric fields into ECS.

Link to the preview of the RFC: https://github.com/kaiyan-sheng/ecs/blob/add_host_fields/rfcs/text/0005-host-metric-fields.md


rfcs/text/0005-host-metric-fields.md
```diff
@@ -125,6 +125,13 @@ Stage 4: Document any new concerns and their resolution. The goal here is to eli
 <!--
 Stage 4: Identify at least one real-world, production-ready implementation that uses these updated field definitions. An example of this might be a GA feature in an Elastic application in Kibana.
 -->
+Our goal is to switch related fields to these new host metrics in Kibana Observability
+metrics UI. For example: right now under hosts inventory, CPU usage will only display
+metrics that are collected by Metricbeat `system` module. With using the new host
+metric fields, CPU metric from `system` module will be `host.cpu.usage`, as well
+as CPU metrics from all AWS EC2 instances or Azure compute VMs. With Kibana Metrics
+UI switching to these new proposed host metric fields, all hosts will be discoverd
```
ebeahan, line 133: Any issue(s) we should link as a reference for this upcoming switch?

ebeahan (commented):
Thanks, @kaiyan-sheng, for adding the section on the real-world implementation!

Could you add one or two additional example source documents? The [AWS](https://github.com/kaiyan-sheng/ecs/blob/add_host_fields/rfcs/text/0005/ec2.json) example format is great; it would be valuable to capture additional examples from other listed sources.

rfcs/text/0005-host-metric-fields.md
kaiyan-sheng: Hi @ebeahan, thanks for the review. I don't think we have any issues right now to track this work. I will create them and update here.

kaiyan-sheng: Hi @ebeahan, sorry for the delay. I just added the link for switching to new field names in Kibana Metrics UI.

kaiyan-sheng:
Hi @ebeahan, I just added two more real world example source documents, and please let me know what you think. Thanks!!

ebeahan (dismissed):
LGTM 

@kaiyan-sheng is there someone that can provide an o11y-side review while @exekias is away?

cc @cyrille-leclerc

kaiyan-sheng:
@ebeahan Thanks for the review! @jsoriano Could you take a look at this PR please 😬 TIA!!

jsoriano (approved):

dainperkins:
I just noticed Metricbeat is currently using host.network.in|out.bytes|packets, might make a lot of sense to stick with that instead of ingress/egress and updating beats / existing indices, etc.  
Its also pretty typical to see interface specific metrics described in a pure in/out perspective as theres no concept of e.g. final destination (azure, aws examples, snmp, etc.), whereas ingress/egress often refers specifically to an l2/l3 network concept of handling packets (e.g. router, switch, firewall netflow reporting)

PR #1185, opened 2020-12-10
Add missing component artifacts from #1098

ebeahan:
Following merging #1098, the `master` branch's build is ❌ due to component templates not being committed.

This follow-up PR should resolve.


webmat:
I think I just got bit by that in https://github.com/elastic/ecs/pull/1186. Now fixed there as well.

Perhaps we can fix two birds with one stone, and focus on #1186 instead?

webmat:
Solved by #1186

PR #1186, opened 2020-12-10
Move _meta section back inside mappings, in legacy templates.

webmat:
This fixes an issue introduced by #1156, discovered in #1180. Composable templates support `_meta` at the template's root, but legacy templates don't. So we're just putting it back inside the mappings for legacy templates.

This also fixes missing updates to the component template, after the introduction of wildcard in #1098.


ebeahan (dismissed):
LGTM 👍 

I noticed there's no changelog update, but that may be intentional since this is only reverting.

webmat:
Yeah forgot to add one. I'll just piggyback on existing changelogs, actually.

ebeahan (approved):

PR #1204, opened 2020-12-17
add cloud.service.name

graphaelli:
~`cloud.platform`~ `cloud.service.name` is intended to distinguish services running on different platforms within a provider.  For example, it is important to know if an application is running on EC2 or under Lambda though both are within AWS so that contextual information about cold starts can be displayed.  This isn't limited to serverless, for example Azure VM vs App Service vs Function enables greater insight into deployed infrastructure inventory.

### DIFF
--- a/schemas/cloud.yml
+++ b/schemas/cloud.yml
@@ -86,7 +86,9 @@
       example: lambda
       short: The cloud platform name.
       description: >
-        The cloud platform name.
+        The cloud platform name is intended to distinguish services running on
+        different platforms within a provider, eg AWS EC2 vs Lambda,
+        GCP GCE vs App Engine, Azure VM vs App Server.
 
         Examples: app engine, app service, cloud run, fargate, lambda.
 
--- a/schemas/cloud.yml
+++ b/schemas/cloud.yml
@@ -80,13 +80,13 @@
 
         Examples: AWS account name, Google Cloud ORG display name.
 
-    - name: platform
+    - name: service.name
       level: extended
       type: keyword
       example: lambda
-      short: The cloud platform name.
+      short: The cloud service name.
       description: >
-        The cloud platform name is intended to distinguish services running on
+        The cloud service name is intended to distinguish services running on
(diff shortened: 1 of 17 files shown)


ebeahan:
It makes sense to me to have a field for capturing which cloud service the event is associated with.

If I understand the intent, this field wouldn't be limited to just PaaS-type cloud services like Azure App Service or AWS Elastic Beanstalk, but may also include EC2, GCE, and other IaaS-type services?

If the field is intended for the cloud service name, would `cloud.service.name` be more appropriate?

ebeahan:
I'd appreciate it if others could weigh-in (or loop in the best person) from a data collection standpoint.

@urso @exekias @andrewkroh

schemas/cloud.yml
```diff
@@ -80,6 +80,16 @@
 
         Examples: AWS account name, Google Cloud ORG display name.
 
+    - name: platform
+      level: extended
+      type: keyword
+      example: lambda
+      short: The cloud platform name.
+      description: >
+        The cloud platform name.
```
andrewkroh, line 89: The PR's description contains extra context that I think would be useful to help users understand this field. Consider adding some of that info into the description.

graphaelli: done

graphaelli:
Is `cloud.service.name` the way forward then?

ebeahan:
@graphaelli, thanks for updating the description.

Suggesting `cloud.service.name` came from my own experiences of `service` being a common term for various cloud providers' IaaS/PaaS offering. I think this aligns with the description provided too?

I want to get some additional input from others mentioned in my last comment if there's any feedback.

ebeahan:
@graphaelli let's move forward with `cloud.service.name`.

Would you update the name in the field definitions and regenerate the artifacts?

graphaelli:
@ebeahan thanks, I've updated the definitions and artifacts

ebeahan (approved):
Thanks! LGTM 👍

Issue #1203, opened 2020-12-17
Unable to fully override composable templates


faudebert:
**Description of the problem including expected versus actual behavior**:

When running `scripts/generator.py` with `--template-setting`, `generated/elasticsearch/template.json` is partially overridden with provided custom values. That is, if specified, mappings are properly overridden but template part always remains to its default.

**Steps to reproduce**:
1. Check `template-settings.json` content (only `index_patterns` here for brevity)
```sh
$ grep index_patterns usage-example/fields/template-settings.json
  "index_patterns": ["acme-weblogs-*"],
```
2. Run `generator.py`:
```sh
$ python scripts/generator.py --ref v1.6.0 \
  --subset            usage-example/fields/subset.yml \
  --include           usage-example/fields/custom/ \
  --out               usage-example/ \
  --template-settings usage-example/fields/template-settings.json \
  --mapping-settings  usage-example/fields/mapping-settings.json
```
3. Check `template.json` content (again, only `index_patterns` for brevity):
```sh
$ grep -A 2 index_patterns usage-example/generated/elasticsearch/template.json
  "index_patterns": [
    "try-ecs-*"
  ],
```

We are expecting `"acme-weblogs-*"`.

**Any additional context:**

Reading the source, it looks defaults are hardcoded in [save_composable_template()](https://github.com/elastic/ecs/blob/master/scripts/generators/es_template.py#L23-L30) while its legacy equivalent is calling [template_settings()](https://github.com/elastic/ecs/blob/12d8f06ecc887c66a0a9eab1ea79431e53cb5483/scripts/generators/es_template.py#L189) which is reading the user-provided file.

PR #1205, opened 2020-12-18
Fix composable template settings override (#1203)

faudebert:
This is an attempt to fix composable template settings override (see #1203).

I did not reuse `default_template_settings()` methods as it look specific to legacy templates (and not used here previously anyway...).

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -15,6 +15,8 @@ Thanks, you're awesome :-) -->
 
 #### Bugfixes
 
+* Honor `--template-settings` option from `scripts/generator.py`. #1203
+
 #### Added
 
 #### Improvements
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -56,7 +56,7 @@ def main():
         exit()
 
     csv_generator.generate(flat, ecs_version, out_dir)
-    es_template.generate(nested, ecs_version, out_dir, args.mapping_settings)
+    es_template.generate(nested, ecs_version, out_dir, args.template_settings, args.mapping_settings)
     es_template.generate_legacy(flat, ecs_version, out_dir, args.template_settings, args.mapping_settings)
     beats.generate(nested, ecs_version, out_dir)
     if args.include or args.subset:
--- a/scripts/generators/es_template.py
+++ b/scripts/generators/es_template.py
@@ -11,36 +11,48 @@ from schema.oss import TYPE_FALLBACKS
 
 # Composable Template
 
-def generate(ecs_nested, ecs_version, out_dir, mapping_settings_file):
+def generate(ecs_nested, ecs_version, out_dir, template_settings_file, mapping_settings_file):
     """This generates all artifacts for the composable template approach"""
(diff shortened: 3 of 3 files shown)


faudebert:
Contributor agreement is being reviewed by our legal team. Hope we'll being able to sign it shortly :)

ebeahan:
Hi @faudebert have you been able to work with your team yet regarding the CLA?

faudebert:
Hi @ebeahan !

The process has been delayed a little bit but I'm expecting having the CLA signed in few days or so.🤞

faudebert:
Thanks @gtback!

@ebeahan Let me know if I can do know anything to help you with that PR.

ebeahan:
Thanks, @faudebert. Sorry for the delayed response.

One challenge with `--template-setting` (that we don't have with the mapping object provided with `--mapping-settings`) is the difference between the legacy `_template`and `_index_template` APIs. If we use the single `--template-settings` file to generate legacy-compatible and composable templates, one of the two will have incompatible settings.

Using this `template-settings.json` as an example:

```json
{
  "index_patterns": ["acme-weblogs-*"],
  "order": 1,
  "settings": {
    "index": {
      "codec" : "best_compression",
      "mapping": {
        "total_fields": {
          "limit": 1000
        }
      },
      "refresh_interval": "2s"
    }
  }
}
```

The legacy `_template` compatible `template.json` that's generated will work as expected:

```bash
$ curl -XPUT "http://localhost:9200/_template/try-ecs" \
   -H "Content-Type: application/json" \
   -d @generated/elasticsearch/7/template.json | jq '.'
{
  "acknowledged": true
}
```

But the composable template will return an exception.

```bash
$ curl -sSL -XPUT "http://localhost:9200/_index_template/try-ecs" \
   -H "Content-Type: application/json" \
   -d @usage-example/generated/elasticsearch/template.json | jq '.'
{
  "error": {
    "root_cause": [
      {
        "type": "x_content_parse_exception",
        "reason": "[1:410] [index_template] unknown field [order]"
      }
    ],
    "type": "x_content_parse_exception",
    "reason": "[1:410] [index_template] unknown field [order]"
  },
  "status": 400
}
```

Index templates use `priority` but the index templates instead use `order`. There's a collection of other settings that would have to be discerned between the two.

Some users may only require either the legacy _or_ the composable index template, and the provided `template-settings.json` file could be adapted for one use of the other. But, ideally, the tooling should support generating usable versions of both templates.

This makes addressing #1203 a bit more involved, I think. One possible solution would be introducing an additional flag, something like  `--template-settings-v1` for the legacy index template settings.  My preference would be to introduce a new flag for the legacy templates versus the composable template since using the current index template with the `_index_template` API is the recommendation. However, that creates a breaking change for users depending on `--template-settings` applying to their legacy index template.

I will continue brainstorming but feel free to drop in any thoughts or feedback. I'd love any insights into how you're using the tooling today and if there's a direction that would be more useful.

kgeller:
Closing this PR due to Eric's comments above and my different implementation of a solution https://github.com/elastic/ecs/pull/1737.

PR #1208, opened 2020-12-21
add http.request.id

graphaelli:
It is very common for web servers to generate a unique id for all incoming requests.  For example, [heroku](https://devcenter.heroku.com/articles/http-request-id) standardizes this in `X-Request-ID`.  Elastic APM Server itself has assigned unique request ids since v6.2.

This information is useful for analyzing per-request logs.

### DIFF
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -13,7 +13,11 @@
       type: keyword
       short: HTTP request ID.
       description: >
-        A unique identifier for each HTTP request.
+        A unique identifier for each HTTP request to correlate logs between clients
+        and servers in transactions.
+        
+        The id may be contained in a non-standard HTTP header, such as `X-Request-ID`
+        or `X-Correlation-ID`. 
 
       example: 123e4567-e89b-12d3-a456-426614174000
 
--- a/schemas/http.yml
+++ b/schemas/http.yml
@@ -8,6 +8,15 @@
   type: group
   fields:
 
+    - name: request.id
+      level: extended
+      type: keyword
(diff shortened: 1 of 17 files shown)


schemas/http.yml
```diff
@@ -8,6 +8,15 @@
   type: group
   fields:
 
+    - name: request.id
+      level: extended
+      type: keyword
+      short: HTTP request ID.
+      description: >
+        A unique identifier for each HTTP request.
```
ebeahan, line 16: Proposing we expand the description.

```suggestion
        A unique identifier for each HTTP request to correlate logs between clients
        and servers in transactions.
        
        The id may be contained in a non-standard HTTP header, such as `X-Request-ID`
        or `X-Correlation-ID`. 
```

ebeahan (commented):
Should we also consider a response field, `http.response.id`?

This does touch on a long-outstanding issue of how to best handle capturing HTTP headers in ECS (#232). ECS will eventually have better guidance on where to capture arbitrary HTTP headers, including non-standard X-prefixed headers. Regardless of how that guidance shapes up, certain headers will still be valuable captured in their individual fields.

In this case, the unique identifier could appear across multiple header names depending on the underlying app, service, or framework implementation (`X-Request-ID`, `X-Correlation-ID`, etc.). So I think normalising to `http.[request|response].id` is reasonable here.

graphaelli:
I don't have a need for `http.response.id` but up for doing the legwork if that would help move `http.response.id` along.

PR #1219, opened 2021-01-12
[RFC] Email - Stage 1 Proposal

P1llus:
Moving the Email RFC to Stage 1

[Preview of the RFC proposal](https://github.com/P1llus/ecs/blob/master/rfcs/text/0010-email.md)

Criteria for Stage 1

### DIFF
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -1,7 +1,7 @@
 # 0010: Email
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **2 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
 - Date: **2020-11-30** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 This RFC proposes a new top-level field to facilitate email use cases.
@@ -77,6 +77,43 @@ Stage 1: Provide a high-level description of example sources of data. This does 
 <!--
 Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting.
 -->
+```json
+{
+	"EndDate": "2020-11-10T22:12:34.8196921Z",
+	"FromIP": "8.8.8.8",
+	"Index": 25,
+	"MessageId": "\\u003c95689d8d5e7f429390a4e3646eef75e8-JFBVALKQOJXWILKBK4YVA7APGM3DKTLFONZWCZ3FINSW45DFOJ6EAQ2ENFTWK43UL4YTCMBYGIYHYU3NORYA====@microsoft.com\\u003e",
+	"MessageTraceId": "ff1a64a3-cafb-41b7-1efb-08d8848aedc3",
+	"Organization": "testdomain.onmicrosoft.com",
+	"Received": "2020-11-09T04:50:06.3312635",
+	"RecipientAddress": "john@testdomain.onmicrosoft.com",
+	"SenderAddress": "o365mc@microsoft.com",
+	"Size": 64329,
+	"StartDate": "2020-11-08T22:12:34.8196921Z",
+	"Status": "Delivered",
+	"Subject": "Weekly digest: Microsoft service updates",
+	"ToIP": null
+}
+
+{
+	"EndDate": "2020-11-10T22:12:34.8196921Z",
+	"FromIP": null,
+	"Index": 8,
+	"MessageId": "\\u003c72872e16-f4c2-4eef-a393-e5621748a0ff@AS8P19vMB1605.EURP191.PROD.OUTLOOK.COM\\u003e",
+	"MessageTraceId": "a4bd8c4c-3a4f-427f-8952-08d8850f9c20",
+	"Organization": "testdomain.onmicrosoft.com",
+	"Received": "2020-11-10T00:28:56.3306834",
+	"RecipientAddress": "o365mc@microsoft.com",
+	"SenderAddress": "postmaster@testdomain.onmicrosoft.com",
+	"Size": 96627,
+	"StartDate": "2020-11-08T22:12:34.8196921Z",
+	"Status": "Delivered",
+	"Subject": "Undeliverable: Message Center Major Change Update Notification",
+	"ToIP": "8.8.8.8"
+}
+```
+
+
 
 <!--
 Stage 3: Add more real world example source documents so we have at least 2 total, but ideally 3. Format as described in stage 2.
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -1,7 +1,7 @@
 # 0010: Email
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **2 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
 - Date: **2020-11-30** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 This RFC proposes a new top-level field to facilitate email use cases.
@@ -195,5 +195,6 @@ e.g.:
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
-* Stage 1: https://github.com/elastic/ecs/pull/999
+* Stage 1 (formerly proposal stage): https://github.com/elastic/ecs/pull/999
   * RFC ID correction: https://github.com/elastic/ecs/pull/1157
+* Stage 1 (draft): https://github.com/elastic/ecs/pull/1219
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -287,7 +287,7 @@ Does the initial set of `email` fields need to consider observability and email 
 
 ### Additional event categorization values
 
-Should a new event.category field (email) be created, and, if so, which `event.type` values the `email` category should be combined with.
+Should a new event.category field (email) be created, and, if so, which `event.type` values the `email` category should be combined with?
 
 ### Display names
 
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -272,7 +272,7 @@ Should attachments be considered in this initial proposal? If so, should the fie
 
 ### Spam processing details
 
-Should fields intended to capture details around spam processing like sender policy framework (SPF) or domainkeys identified mail (DKIM) be in scope for this proposal as well?
+Should fields intended to capture details around spam processing like sender policy framework (SPF), domainkeys identified mail (DKIM), or domain-based message authentication, reporting, and conformance (DMARC) be in scope for this proposal as well?
 
 <!--
 Stage 2: Document new concerns or resolutions to previously listed concerns. It's not critical that all concerns have resolutions at this point, but it would be helpful if resolutions were taking shape for the most significant concerns.
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -16,7 +16,7 @@ Stage 1: Describe at a high level how this change affects fields. Which fieldset
 
 | field | type | description |
 | --- | --- | --- |
-| `email.from` | keyword | Stores the `from` email address |
+| `email.from` | keyword | Stores the `from` email address from the RFC5322 `From:` header field. |
 | `email.origination_timestamp` | date | The date and time the email message was composed. Many email clients will fill this in automatically when the message is sent by a user. |
 | `email.delivery_timestamp` | date | The date and time the email message was received by the service or client. |
 | `email.to` | keyword (array) | The email address(es) of the message recipient(s) |
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -31,9 +31,12 @@ Stage 1: Describe at a high level how this change affects fields. Which fieldset
 | `email.x_mailer` | keyword | What application was used to draft and send the original email. |
 | `email.attachments` | nested | Nested object of attachments on the email. |
 | `email.attachments.file.mime_type` | keyword | MIME type of the attachment file. |
-| `email.attachments.file.name` | keyword | Name of the file including the extension. |
-| `email.attachments.file.extension` | keyword | File extension, excluding the leading dot. |
-| `email.attachments.file.size` | long | File size in bytes. |
+| `email.attachments.file.name` | keyword | Name of the attachment file including the extension. |
+| `email.attachments.file.extension` | keyword | Attachment file extension, excluding the leading dot. |
+| `email.attachments.file.size` | long | Attachment file size in bytes. |
+| `email.attachments.hash.md5` | keyword | MD5 hash of the file attachment. |
+| `email.attachments.hash.sha1` | keyword | SHA-1 hash of the file attachment. |
+| `email.attachments.hash.sha256` | keyword | SHA-256 hash of the file attachment. |
 
 ### Additional event categorization allowed values
 
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -18,6 +18,7 @@ Stage 1: Describe at a high level how this change affects fields. Which fieldset
 | --- | --- | --- |
 | `email.from` | keyword | Stores the `from` email address |
 | `email.origination_timestamp` | date | The date and time the email message was composed. Many email clients will fill this in automatically when the message is sent by a user. |
+| `email.delivery_timestamp` | date | The date and time the email message was received by the service or client. |
 | `email.to` | keyword (array) | The email address(es) of the message recipient(s) |
 | `email.subject` | keyword; `.text` text multi-field | A brief summary of the topic of the message |
 | `email.cc` | keyword (array) | The email address(es) of the carbon copy (CC) recipient(s) |
--- a/rfcs/text/0010-email.md
+++ b/rfcs/text/0010-email.md
@@ -169,7 +169,7 @@ Stage 2: Included a real world example source document. Ideally this example com
 #### Original log
 
 ```
-<38>1 2016-06-24T21:00:08Z - ProofpointTAP - MSGBLK [tapmsg@21139 messageTime="2016-06-24T21:18:38.000Z" messageID="20160624211145.62086.mail@evil.zz" recipient="clark.kent@pharmtech.zz, diana.prince@pharmtech.zz" sender="e99d7ed5580193f36a51f597bc2c0210@evil.zz" senderIP="192.0.2.255" phishScore="46" spamScore="4" QID="r2FNwRHF004109" GUID="c26dbea0-80d5-463b-b93c-4e8b708219ce" xmailer="Spambot v2.5"]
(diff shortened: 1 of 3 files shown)


ebeahan:

With this PR targeting the now "legacy" stage two, I'm thinking we update to target stage 1. Stage 1 (draft) fields will still be added as experimental fields as was done with legacy stage 2 fields.

rfcs/text/0010-email.md
```diff
@@ -1,7 +1,7 @@
 # 0010: Email
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **1 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **2 (proposal)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 4:
```suggestion
- Stage: **1 (draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```

As my earlier [comment](https://github.com/elastic/ecs/pull/1219#issuecomment-768588593) mentions, this is only to align with the new, condensed stages. Moving to `Draft` will still add the `email.*` fields to ECS as experimental fields.

rfcs/text/0010-email.md
```diff
@@ -77,6 +77,43 @@ Stage 1: Provide a high-level description of example sources of data. This does
 <!--
 Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting.
 -->
+```json
```
ebeahan, line 80: It looks like these are Microsoft 365 example source events for one successful and one undeliverable message, but can we add some labels making that clearer here?

ebeahan (commented):
Thanks, @P1llus, for opening this. I listed some of the areas we want to go over and capture to move forward.

### Field definitions

I suggest we replace any instance of `wildcard` with `keyword` since we've paused on introducing `wildcard` field support into ECS for now.

Can we create a directory using the RFC's number and add a YAML for the proposed set of field definitions? Here's an example: https://github.com/elastic/ecs/tree/master/rfcs/text/0009

### Example data

We've added a couple of great examples from M365, and it'd be great to capture one or two more examples from other email data sources. 

### Concerns

Are there any new concerns that have come up to be captured? Or, do any of the existing concerns have updates or resolutions we can capture?

### Feedback

Any individuals or teams we should ask for their feedback around the approach, fields, etc.?

ebeahan:
ECS team will review and determine the next steps to continue moving this forward.

@jamiehynds, are you still willing to act as a sponsor?

jamiehynds:
Thanks @ebeahan - yes, I'll continue to sponsor the RFC.

ebeahan:
The `email` RFC proposal has been updated and is ready for some feedback! 😄 🎉

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
-| --- | --- |
-| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
-| `event.start` | When the email event started
-| `event.end` | When the email event ended
-| `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
-| `network.protocol` | Type of email protocol used |
-| `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
-
-
+| `email.from` | keyword (array) | Stores the `from` email address(es) |
```
kgeller, line 19: maybe a silly question, but how can one get an email from multiple email addresses?

devonakerr, line 19: Should source be consider singular in this context? I would expect the source of a message to be potentially one-to-many if not one-to-one, which would mean an array is less ideal.

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
+| `email.timestamp` | keyword (array) | The local date and time the message was written |
```
devonakerr, line 20: I don't see a concept representing when an email was received, which is another time value useful for security purposes.

rfcs/text/0010-email.md
devonakerr: ++

rfcs/text/0010-email.md
```diff
@@ -97,37 +249,48 @@ The goal here is to research and understand the impact of these changes on users
...
+
+Should the display name be captured separately from the email address for senders and recipients. If so, how do we accomplish this in a document while keeping the 1:1 of a display name to email address.
+
+### Attachments
+
+Should attachments be considered in this initial proposal? If so, should the fields should mirror (or potentially nest) the `file.*` fields?
+
+### Spam processing details
+
+Should fields intended to capture details around spam processing like sender policy framework (SPF) or domainkeys identified mail (DKIM) be in scope for this proposal as well?
```
devonakerr, line 275: Additionally DMARC

rfcs/text/0010/email.yml
```diff
@@ -0,0 +1,118 @@
...
+      description: >
+        Stores the `to` email address(es).
+      example: "employee@example.com"
+      normalize:
+        - array
+
+    - name: from
+      level: extended
+      type: keyword
+      short: Sender address(es)
```
devonakerr, line 24: Consider whether a keyword is more appropriate than an array

rfcs/text/0010/email.yml
```diff
@@ -0,0 +1,118 @@
...
+      description: >
+        Stores the `from` email address(es).
+      example: "administrator@example.com"
+      normalize:
+        - array
+
+    - name: cc
```
devonakerr, line 31: Shouldn't cc and bcc be arrays, assuming plurality of values?

devonakerr (commented):
I think there are some questions to answer:

- will a source ever be an array of values?
- are cc and bcc fields arrays and not keywords?
- how will we account for attachments?

rfcs/text/0010-email.md
ebeahan: Good catch - this field shouldn't be an array! 😅

ebeahan: Agreed. I'll remove the array from the `.from` field.

rfcs/text/0010/email.yml
ebeahan: I have the field definition defined as an array (`normalize: array`), but I didn't use an array for the example value. Will correct.

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
-| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
-| `event.start` | When the email event started
-| `event.end` | When the email event ended
-| `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
-| `network.protocol` | Type of email protocol used |
-| `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
-
-
+| `email.from` | keyword | Stores the `from` email address |
+| `email.timestamp` | keyword (array) | The local date and time the message was written |
```
peasead, line 20:
```suggestion
| `email.timestamp.written` | keyword (array) | The local date and time the message was written |
| `email.timestamp.received` | keyword (array) | The local date and time the message was received |
```

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
+| `email.to` | keyword (array) | The email address(es) of the message recipient(s) |
+| `email.subject` | keyword; `.text` text multi-field | A brief summary of the topic of the message |
+| `email.cc` | keyword (array) | The email address(es) of the carbon copy (CC) recipient(s) |
+| `email.bcc` | keyword (array) | The email address(es) of the blind carbon copy (CC) recipient(s) |
+| `email.content_type` | keyword | Information about how the message is to be displayed. Typically a MIME type |
+| `email.message_id` | wildcard | Unique identifier for the email message that refers to a particular version of a particular message |
+| `email.reply_to` | keyword | Address that replies should be delivered to |
+| `email.direction` | keyword | Direction of the message based on the sending and receiving domains |
+| `email.x_mailer` | keyword | What application was used to draft and send the original email.
+
```
peasead, line 30: | `email.attachment.mime_type` | keyword | MIME type should identify the format of the attachment using IANA official types, where possible. When more than one type is applicable, the most specific type should be used. |

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
+| `email.timestamp` | keyword (array) | The local date and time the message was written |
+| `email.to` | keyword (array) | The email address(es) of the message recipient(s) |
+| `email.subject` | keyword; `.text` text multi-field | A brief summary of the topic of the message |
+| `email.cc` | keyword (array) | The email address(es) of the carbon copy (CC) recipient(s) |
+| `email.bcc` | keyword (array) | The email address(es) of the blind carbon copy (CC) recipient(s) |
+| `email.content_type` | keyword | Information about how the message is to be displayed. Typically a MIME type |
+| `email.message_id` | wildcard | Unique identifier for the email message that refers to a particular version of a particular message |
+| `email.reply_to` | keyword | Address that replies should be delivered to |
+| `email.direction` | keyword | Direction of the message based on the sending and receiving domains |
+| `email.x_mailer` | keyword | What application was used to draft and send the original email.
```
peasead, line 29:
```suggestion
| `email.x_mailer` | keyword | What application was used to draft and send the original email. |
```

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
+
+### Additional event categorization allowed values
```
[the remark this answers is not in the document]
ebeahan: I agree we should include fields for emails attachments eventually. Still, I'm undecided whether to have them or not in this initial round of the `email.*` fields or add attachment fields in a separate proposal later.

If folks feel there's a strong case for including them from the start, I'm open to adding.

Part of my hesitation - I haven't thought of a great option for multiple attachments in a single document without resorting to an array of objects and using a `nested` field. We introduced `nested` recently in ECS for other use cases. However, I am still cautious about `nested` since a) users still misunderstand the behavior of nested fields, b) nested fields consume additional storage, and c) lack of nested agg support in Kibana for visualizations.

rfcs/text/0010-email.md
ebeahan: This timestamp captures the [origination date](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.1) value. The description should explain that better; I'll address it.

I had considered using something like `email.sent_timestamp` and `email.received_timestamp`. I struggled with `received,` though.  A log emitted from an MTA (like Mimecast) might only have a single timestamp, but if the source is similar to an EML file with [trace fields](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.7). Open to any ideas on how to best represent these nuances.

cc @peasead

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
-
+| `email.from` | keyword | Stores the `from` email address |
+| `email.timestamp` | keyword (array) | The local date and time the message was written |
+| `email.to` | keyword (array) | The email address(es) of the message recipient(s) |
+| `email.subject` | keyword; `.text` text multi-field | A brief summary of the topic of the message |
+| `email.cc` | keyword (array) | The email address(es) of the carbon copy (CC) recipient(s) |
+| `email.bcc` | keyword (array) | The email address(es) of the blind carbon copy (CC) recipient(s) |
+| `email.content_type` | keyword | Information about how the message is to be displayed. Typically a MIME type |
+| `email.message_id` | wildcard | Unique identifier for the email message that refers to a particular version of a particular message |
+| `email.reply_to` | keyword | Address that replies should be delivered to |
```
ericwbentley, line 27: It should be clarified that this is referring to RFC5322.ReplyTo and is not the same as Return-Path or RFC5321.MailFrom

rfcs/text/0010-email.md
```diff
@@ -1,68 +1,44 @@
...
-| --- | --- |
-| `event.duration` | The duration related to the email event. Could be the total duration in Quarantine, how long the email took to send from source to destination etc |
-| `event.start` | When the email event started
-| `event.end` | When the email event ended
-| `process.name` | When the event is related to a server or client. Does not take MTA into account which is part of a ongoing discussion |
-| `network.protocol` | Type of email protocol used |
-| `tls.*` | Used for TLS related information for the connection to for example a SMTP server over TLS |
-
-
+| `email.from` | keyword | Stores the `from` email address |
```
ericwbentley, line 19: It should be clarified that this is referring to the RFC5322.From or "Header From" address specifically

There should be another field for RFC5321.MailFrom "Envelope From" and/or Return-Path as RFC5321.MailFrom and RFC5322.From often differ

rfcs/text/0010-email.md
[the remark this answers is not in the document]
ericwbentley: Can we consider adding another ID field, for example` email.local-id`? Exchange uses it's own IDs in addition to message-id for example `internal-message-id` or `network-message-id`. Other email servers will have their own session IDs as well.

rfcs/text/0010-email.md
peasead: Thanks @ericwbentley 

So, something like this:
| - | - | - |
| - | - | - |
| `email.from` | keyword | Stores the `from` email address provided in the "From" originator field |

Reference: https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.2

rfcs/text/0010-email.md
peasead: Thanks @ericwbentley 

So, something like this:
| - | - | - |
| - | - | - |
| `email.reply_to` | keyword | Stores the email address provided in the "Reply-To" originator field |

rfcs/text/0010-email.md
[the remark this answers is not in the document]
peasead: That sounds like a good idea, @ericwbentley 

Could you use the "insert suggestion" option to share how you think that'd need to be structured?

rfcs/text/0010-email.md
peasead: Good question.

It looks like trace fields are part of a different [RFC](https://datatracker.ietf.org/doc/html/rfc5321#section-4.4), maybe those trace fields would be better tracked in an SMTP fieldset?

If not, would there be a way to make the field an array of timestamps if necessary? Is there a way to connect the different trace timestamps with the other trace elements (return, received, etc.)

peasead:
What happened to this E-mail fieldset? https://github.com/elastic/ecs/pull/999

I see it was merged, but I don't see it in `master`.

ebeahan:
@peasead This picks up on the same work from #999 that @P1llus @jamiehynds started. @P1llus has been focused on other commitments, and I'm helping to move the proposal forward.

The field changes proposed in #999 aren't anywhere in the schema yet. Once we have a consensus and merge this PR, the agreed changes will be added to the [experimental](https://github.com/elastic/ecs/tree/master/experimental) schema.

rfcs/text/0010-email.md
ericwbentley: I would go as far to explicitly state the RFC in the description to eliminate any confusion. Or at least the name of the field in the headers.

```suggestion
| `email.reply_to` | keyword | Address that replies should be delivered to (RFC5322.ReplyTo) |
```

rfcs/text/0010-email.md
ebeahan: After some thought, I added an `email.delivery_timestamp` field described as `The date and time the email message was received by the service or client.`. This reserves the word `received` for potential use with the trace records (which aren't currently part of this proposal).

Thoughts?

rfcs/text/0010-email.md
ericwbentley: I'd reference the RFC in the description and add a second field for the envelope from address since they should be distinct from each other.

```suggestion
| `email.from` | keyword | Stores the header `from` email address (RFC5322.From) |
| `email.envelope_from` | keyword | Stores the envelope `from` email address (RFC5321.MailFrom) |
```

rfcs/text/0010-email.md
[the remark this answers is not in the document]
ericwbentley: Without using language too specific towards one product or another I think this covers what the field would be used for

```suggestion
### Additional event categorization allowed values
| `email.local_id` | keyword | Stores the local ID of the email from the server that created the event and not persistent across server hops |
```

rfcs/text/0010-email.md
ebeahan: @ericwbentley Makes sense to reference RFC 5322 - thanks for suggesting the distinction.

Like I put in the [`Concerns` section](https://github.com/P1llus/ecs/blob/master/rfcs/text/0010-email.md#email-messages-vs-protocols), right now, this proposal focuses on the IMF from RFC5322. However, I like the idea of also introducing an `smtp` field set later that focuses on details of the protocol (and other email protocols, if helpful).

I see `smtp.*` fields pairing well with `email.*`, but I want to avoid increasing the scope here too much. Open to feedback, though.

ebeahan:
> will a source ever be an array of values?

Corrected

> are cc and bcc fields arrays and not keywords?

The `cc` and `bcc` are arrays of `keyword` values.

> how will we account for attachments?

I added a nested `attachment` object and updated the Proofpoint TAP sample mapping to include an attachment. Let me know what you think.

rfcs/text/0010-email.md
peasead: I'm not seeing "ReplyTo" anywhere in the RFC. I do see ["Reply-To" in the originator field section](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.2). Is that what you're referring to?

So, something like this:
| - | - | - |
| - | - | - |
| `email.reply_to` | keyword | Stores the email address provided in the "Reply-To" originator field (RFC5322) |

rfcs/text/0010-email.md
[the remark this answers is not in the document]
ebeahan: @ericwbentley @peasead I had added `email.external_id`, but I like what you proposed with `email.local_id` better. 😄 

I adjust the description a bit: https://github.com/elastic/ecs/pull/1219/commits/6368af1cb3b1805d07fae332ad928b44fa2363a3. Let me know if you have any feedback.

rfcs/text/0010-email.md
ebeahan: I changed `email.timestamp` to `email.origination_timestamp` and added `email.delivery_timestamp` to capture both timestamps. Thoughts?

rfcs/text/0010-email.md
ebeahan: I've added a nested object at `email.attachments` with each object having these fields available for each attachment:

* `file.mime_type`
* `file.name`
* `file.extension`
* `file.size`
* `hash.md5`
* `hash.sha1`
* `hash.sha256`

I definitely can expand this list if there are other fields we should account for. I don't think completely nesting `file.*` at `email.attachments.*` is necessary, but I did try to match the naming to the existing `file.*` field set for consistency.

rfcs/text/0010-email.md
ebeahan: Trying to improve and clarify this point in https://github.com/elastic/ecs/pull/1219/commits/a856c0a6462d07bb5fe94d2ba7c6fbcd1eb1c87c. Let me know what you think.

ebeahan:
Thanks, all, for the comments so far! I think I've addressed all the outstanding feedback.

Would appreciate additional looks to see if this is set for stage one.

@kgeller @devonakerr @peasead @P1llus @jamiehynds

rfcs/text/0010-email.md
peasead: I like it.

peasead (approved):
I think this looks good for a Stage 1.

kgeller (approved):
LGTM 👍

wasserman:
In regards to `Display Name`, you could look into some options.

1. Allow for name/address pairs in all email fields instead of just the email address itself.
2. Permit RFC 5322 Email formats like `First Last <first.last@example.com>`.  Then a keyword with a normalizer could index just the emails while still keeping the display names intact.  Of course the names wouldn't be searchable.
3. Multi-fields with some variation of this could work and allow for some form of the `keyword` and `text` fields.

wasserman:
@JamisonWhite @peasead In email the to/cc/bcc is always a singular label regardless of having multiples.  I would vote to not to include the `_addresses` part since this would be understood and a bit redundant.  Hopefully aggregations by email addresses will not be impacted by this structure.

ebeahan:
Thanks, @wasserman @JamisonWhite, for the feedback!

I will incorporate the thinking from this discussion into the stage 2 PR #1593

PR #1229, opened 2021-01-19
add time_zone, postal_code, and continent_code

### LINKED ISSUE #1217: Maxmind vs Logstash's GeoIP Filter
**Summary**

The Logstash GeoIP filter uses the a maxminddb-formatted database (that may or may not be provided by Maxmind) to populate a number of fields to enrich an event based on an IP address. As presently implemented in the Logstash filter, all fields need to be sub-fields of a single `target` (e.g, with `target => client`, we would have `client.geo.*`, `client.as.*`, etc.).

**Motivation**:

Clearly define destinations for all GeoIP Filter fields as a sub-field of a single target, so that users can enable ECS Compatibility Mode without losing metadata that they currently rely on.

Specifically, we have six fields that do not have direct analogues in ECS: 

 - `timezone`: the IANA name of the timezone e.g., America/New_York
 - `postal_code`: a string postal code, length varies by country
 - `continent_code`: "A two character continent code like "NA" (North America) or "OC" (Oceania)"
 - `organization`: the name of the business or ISP associated with an address, reportedly available for ~40% of lookups.
 - `isp`: the name of the ISP
 - `dma_code`: us-only code representing Designated Market Area (~metro area)

> Many of the existing ECS fields underneath `geo.*` are named aligning with Geo*2 like these here.  I see some fields, such as `timezone`, being a good candidates as ECS additions, but I’m not so sure about others, such as the US-specific `dma_code`.
>
> -- @ebeahan 

**Detailed Design**:

1) propose the addition of specs to ECS for
   - `geo.timezone` (IANA name, presently up to 30 characters, e.g., `America/Argentina/Buenos_Aires`),
   - `geo.postal_code` (freeform but relatively space limited, regulated by each country; see [wikipedia](https://en.wikipedia.org/wiki/List_of_postal_codes)),
   - `geo.continent_code` ([docs](https://dev.maxmind.com/geoip/legacy/codes/country_continent/) define `AF`, `AN`, `AS`, `EU`, `NA`, `OC`, and `SA`)
2) guidance for namespacing the other fields so that they are usable and unlikely to present future conflict, bearing in mind that they _must_ be sub-fields of the singular `target` that is the parent of the related `geo` and `as` fields.
   - (a) under a `mmdb` sub-key? e.g, `${target}.mmdb.organization`
     - since the `mmdb` being used may or may not be provided by Maxmind, I'd like to avoid explicitly using "maxmind" in the key name 
     - `dma_code`, while provided in mmdb, is a US-only [Nielsen Ratings construct](https://www.nielsen.com/us/en/intl-campaigns/dma-maps/) and may not be adequately described with a `mmdb` prefix.
   - (b) some other way? 

Related: https://github.com/logstash-plugins/logstash-filter-geoip/issues/163

ebeahan:
### Summary

Add three new fields to the ECS `geo.*` fieldset:

* `geo.timezone`
* `geo.postal_code`
* `geo.continent_code`

### Motivation

These fields will allow for additional geolocation context on events. Like the other `geo.*` fields, these values can be populated by automatically by GeoIP enrichment or manually by the user.

### Other Considerations

* I originally chose `time_zone` over `timezone`. However, reviewing existing geoip implementations ([[0](https://www.elastic.co/guide/en/elasticsearch/reference/master/geoip-processor.html)], [[1](https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/src/main/java/org/logstash/filters/Fields.java)]), opted for `timezone`.

* Considered explicitly listing the seven two-letter continent codes in the description as used by other sources. However, despite other sources referencing [ISO-3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) or [UN M.49](https://unstats.un.org/unsd/methodology/m49/), I couldn't find a definitive, available standard confirming the seven two-letter continent codes as an official standard. Any opinions or feedback one way or the other?

Closes #1217

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -2942,7 +2942,7 @@ example: `Quebec`
 [[field-geo-timezone]]
 <<field-geo-timezone, geo.timezone>>
 
-| The time zone of the location.
+| The time zone of the location, such as IANA time zone name.
 
 type: keyword
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -2792,23 +2792,7 @@ example: `Montreal`
 [[field-geo-continent-code]]
 <<field-geo-continent-code, geo.continent_code>>
 
-| Two-letter code representing the continent.
-
-Possible codes are:
-
-  * AF - Africa
-
-  * AN - Antarctica
-
-  * AS - Asia
-
-  * EU - Europe
-
-  * NA - North America
-
-  * OC - Oceania
-
-  * SA - South America
+| Two-letter code representing continent's name.
 
 type: keyword
 
@@ -2908,13 +2892,15 @@ example: `boston-dc`
 [[field-geo-postal-code]]
 <<field-geo-postal-code, geo.postal_code>>
 
-| Postal code. Also known as postcode, post code, or ZIP code.
+| Postal code associated with the location.
+
+Values appropriate for this field may also be known as a postcode or ZIP code.
 
 type: keyword
 
 
 
-example: `0`
+example: `94040`
 
 | core
 
@@ -2956,7 +2942,7 @@ example: `Quebec`
 [[field-geo-time-zone]]
 <<field-geo-time-zone, geo.time_zone>>
 
-| The time zone associated with location, as specified by the IANA Time Zone Database.
+| The time zone of the location.
 
 type: keyword
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -2789,6 +2789,38 @@ example: `Montreal`
 // ===============================================================
 
 |
+[[field-geo-continent-code]]
+<<field-geo-continent-code, geo.continent_code>>
+
+| Two-letter code representing the continent.
+
+Possible codes are:
+
+  * AF - Africa
+
+  * AN - Antarctica
+
+  * AS - Asia
+
+  * EU - Europe
+
+  * NA - North America
+
+  * OC - Oceania
+
+  * SA - South America
+
+type: keyword
+
+
+
+example: `NA`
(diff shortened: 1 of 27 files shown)


grandmasterv:
This looks great to me.  My team was just thinking about proposing these extensions.  Thank you @ebeahan.

docs/field-details.asciidoc
```diff
@@ -2900,6 +2916,24 @@ example: `boston-dc`
 
 // ===============================================================
 
+|
+[[field-geo-postal-code]]
+<<field-geo-postal-code, geo.postal_code>>
+
+| Postal code associated with the location.
+
+Values appropriate for this field may also be known as a postcode or ZIP code.
```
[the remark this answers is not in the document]
ebeahan: Captured this in the updated description. WDYT?

docs/field-details.asciidoc
```diff
@@ -2932,6 +2966,22 @@ example: `Quebec`
 
 // ===============================================================
 
+|
+[[field-geo-timezone]]
+<<field-geo-timezone, geo.timezone>>
+
+| The time zone of the location.
```
yaauie, line 2973: Worth referencing IANA's timezone db?

Otherwise, PR LGTM.

ebeahan: Updated the description with a mention of the IANA time zone names.

I prefer to avoid limiting the field's value exclusively to IANA time zone names, but I know the IANA names will be the most common values here (by a huge margin 😄 ).

ebeahan:
@yaauie, mind giving this one last look?

ebeahan:
Moving forward merging per the 👍 here: https://github.com/elastic/ecs/pull/1229#discussion_r565633677

PR #1231, opened 2021-01-25
[RFC] File modifications and multiple files

andrewstucki:
This introduces an RFC for multiple files in an event. Specifically the use-cases we're wanting to model:

1. File modifications (i.e. duplicate attributes such as two `name` values or file `owner`s)
2. File copies where we want to capture both the source and destination of the copy operation.

The main outstanding question is what do do about the sheer number of fields that making `file` reusable would cause. I propose limiting the reusable subset to the top-level `file` fields rather than making additional nested fields present (i.e. no `file.target.pe`), but would definitely be open to feedback.

Since I tried to flesh this out with the majority of my thoughts and some real-world use-cases we've already run into in beats, I'm marking this as `Stage 2`, but we can go slower if you want @ebeahan.

[Preview of the markdown proposal](https://github.com/andrewstucki/ecs/blob/multiple-files/rfcs/text/0013-multiple-files.md)


rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
+# 0013: Multiple files in an event
+
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **2 (Draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 5:
```suggestion
- Stage: **1 (Draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```

Under the [new stages](https://elastic.github.io/ecs/stages.html) (with stage 4 removed), `Draft` would be considered stage 1.

rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
+# 0013: Multiple files in an event
```
ebeahan, line 1:
```suggestion
# 0000: Multiple files in an event
```

It may remain as `0013`, but just a note the RFC number will be officially allocated when this PR is merged.

rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
...
+
+The new fields discussed in this RFC are the following:
+
+- `file.target.*`
+- `file.changes.*`
+
+Note that `file` would need to be made reusable with the following entries.
+
+```YAML
+# schemas/file.yml excerpt
```
ebeahan, line 22: Thanks for defining the field definitions already 👍 

Can you also add these in a separate, standalone YAML file? Similar to what was done [here](https://github.com/elastic/ecs/blob/master/rfcs/text/0007/user.yml)?

The standalone file allows for anyone to grab it independently using `--include` for testing the schema changes, e.g.:

`$ python scripts/generator.py --include rfcs/text/00xx/file.yml`

rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+Notice that `changes` only captures the fields that changed, not the fields that remain the same between modifications.
+
+Additionally, in some cases we're not always going to have access to the original field value (i.e. running `chmod 755 ...` isn't going to contain data about the original file permissions). Users should actively fill in all data that they can for the top-level fields, but, even if there is no corresponding top-level field, `changes` may contain the fields that are assumed to be modified.
+
+### File Copies
+
+This is a bit more complex than a modification. This can happen, for example if a file is duplicated via `cp` or some other utility, if a file is remotely transfered, or if a symlink or hardlink is created that references back to the original file. Unlike the modification events, I would imagine that copying would likely require most of the fields from the top-level of the file fields to be present in `target`, as file ownership, location, extensions, devices, timestamps, and names can all change during a copy event.
```
ebeahan, line 84: For remotely transferred files, do you have any thoughts for the capturing of the remote host that'll be relevant to `file.target.*`? Or guidance on how to best capture remote copies using `file.target.*` plus other ECS fields?

rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
...
+    top_level: true
+    expected:
+      # Added for this RFC
+      - at: file
+        as: target
+      - at: file
+        as: changes
+```
+
+Currently `file` has 21 top-level fields and has a good number of fields that can be nested under it such as `code_signature`, `hash`, `pe`, and `x509`, so we may want to clarify exactly which fields we want to include (maybe just the top-level?).
```
ebeahan, line 33: Yes, I think _not_ having all of `file.*` nestings carry along to `file.target.*` and `file.changes.*` is sensible for this case.

I'll go back and see if there are any changes in the tooling necessary to support.

rfcs/text/0013-multiple-files.md
```diff
@@ -0,0 +1,220 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+- Stage 2: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 220:
```suggestion
- Stage 1: https://github.com/elastic/ecs/pull/1231
```

ebeahan (commented):
I left an initial review. Thanks for all the work you've already done here, @andrewstucki!

Do you have anyone in mind for sponsoring as the proposal advances?

rfcs/text/0013-multiple-files.md
andrewstucki: With regard to "capturing of the remote host" do you mean like hostnames/ips and such? If think that leveraging the `source`/`destination` or `client`/`server` fields (depending on protocol/modeling) would probably make sense for something like that. Do you want me to add that to the doc?

ebeahan: Yeah, please do.

djptek:
Have you considered using "type": "nested" to avoid having the same field name occurring at multiple levels in the mapping?

My concerns with the multi-level solution are:

- repeating the same properties at distinct depths in the document structure may be confusing to look at and could create problems when constructing DSL queries via the API
- might have issues when more than 2 files are involved, for example, how to express a command such as:

`cp source_file_1 source_file_2 ... source_file_N target_folder/`

Using nested also has a downside since it adds overhead, however, the File Copy example in [RFC 0013](https://github.com/andrewstucki/ecs/blob/multiple-files/rfcs/text/0013-multiple-files.md#file-copies) could be modified to use an array with nested field type and (optionally) the addition of an additional field to denote the role of each file in the operation - this would ensure that any property such as `files.name` would always occur at a consistent depth in the hierarchy

The RFC gives this example:

```
# ...0013-multiple-files.md#file-copies
POST file_copy_test_v1/_doc
{
  "file": {
    "name": "abc.txt",
    "directory": "/home/user",
    "extension": "txt",
    "target": {
      "name": "def.exe",
      "directory": "/tmp",
      "extension": "exe"
    }
  }
}
```

which could be expressed like this

```
# alternative using "type": "nested"
# create explicit mapping
PUT file_copy_test_v2
{
  "mappings": {
    "properties": {
      "files": {
        "type": "nested",
        "properties": {
          "directory": {
            "type": "keyword"
          },
          "extension": {
            "type": "keyword"
          },
          "name": {
            "type": "keyword"
          },
          "role": {
            "type": "keyword"
          }
        }
      }
    }
  }
}

# add 2 distinct documents
PUT file_copy_test_v2/_doc/abc2def
{
  "files": [
    {
      "role": "source",
      "name": "abc.txt",
      "directory": "/home/user",
      "extension": "txt"
    },
    {
      "role": "target",
      "name": "def.exe",
      "directory": "/tmp",
      "extension": "exe"
    }
  ]
}

PUT file_copy_test_v2/_doc/def2abc
{
  "files": [
    {
      "role": "source",
      "name": "def.txt",
      "directory": "/tmp",
      "extension": "txt"
    },
    {
      "role": "target",
      "name": "abc.exe",
      "directory": "/home/user",
      "extension": "exe"
    }
  ]
}

# use nested query - should only match document with "_id" : "def2abc"
GET file_copy_test_v2/_search
{
  "query": {
    "nested": {
      "path": "files",
      "query": {
        "bool": {
          "must": [
            {
              "match": {
                "files.role": "target"
              }
            },
            {
              "match": {
                "files.name": "abc.exe"
              }
            }
          ]
        }
      }
    }
  }
}

```

djptek:
Hi @andrewstucki did you have a chance to try the alternative strategy for your example as per my previous comment? Please let me know if you need any additional explanation of my question

devonakerr:
Added the `deferred` label to indicate that review may resume at a later time but is unlikely to advance in the near-term.

ebeahan:
Closing for now. Feel free to re-open when the work on this proposal resumes.

PR #1234, opened 2021-01-26
Add binary file content fields

andrewstucki:
This PR aims to fill in a couple of gaps in ECS around binary file content.

1. Currently the `http.*.body.content` fields are `keyword` with a `text` multi-field. Neither makes sense if you have a binary chunk of data that you want to include in a document. To address this I'm proposing adding a `binary`-type multi-field called `raw` to the `content` fields.
2. Ideally we'd have the same sort of ability to return file samples in a `content` field. In the Endgame product for example, you had the ability to download and analyze malware samples that might have triggered an alert on a monitored system. Having a `file.content` field to contain such information would standardize this.

The only somewhat different thing between the added fields is that for the `http` fields, the `binary` field is added as a multi-field, presumably because you're often going to have response bodys that you may be interested in sending generally in html format or something of the sort. In the case of `file` content, you could really be working with pretty much anything, so rather than add a `binary` multi-field, the `binary` field is primary and there's an additional `text` field that can be used for non-binary data added as a multi-field.

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -2406,6 +2406,28 @@ example: `["readonly", "system"]`
 // ===============================================================
 
 |
+[[field-file-content]]
+<<field-file-content, file.content>>
+
+| Content of the file, base64 encoded
+
+type: binary
+
+Multi-fields:
+
+* file.content.text (type: text)
+
+
+
+
+
+
+
+| extended
+
+// ===============================================================
+
+|
 [[field-file-created]]
 <<field-file-created, file.created>>
 
@@ -3394,6 +3416,9 @@ Multi-fields:
 * http.request.body.content.text (type: text)
 
 
+* http.request.body.content.raw (type: binary)
+
+
 
 
 
@@ -3524,6 +3549,9 @@ type: wildcard
 Multi-fields:
 
 * http.response.body.content.text (type: text)
+
+
+* http.response.body.content.raw (type: binary)
 
 
 
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -1560,6 +1560,15 @@
         Leave unpopulated if a certificate was unchecked.'
       example: 'true'
       default_field: false
+    - name: content
+      level: extended
+      type: binary
+      multi_fields:
+      - name: text
+        type: text
+        norms: false
+      description: Content of the file, base64 encoded
+      default_field: false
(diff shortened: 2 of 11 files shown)


rw-access:
Do we have other binary fields in ECS right now? And do we expect a lot of searching on these fields? If we don't need to search them, then we could be wasting a lot of index space. Another q -- how does this interact with `ignore_above`?

andrewstucki:
@rw-access we don't currently have any `binary` fields (which is why the build is failing -- the go code generator is broken for `binary` types 😞 ). I would imagine that these fields wouldn't always be set--mainly set in exceptional circumstances--doing something like dumping file content for every file on disk or every http request coming out of a host would be pretty ridiculous.

As it is, because we don't have `store: true` set on the field, we wouldn't actually be able to search on it, rather it would be primarily retrievable as an artifact from the source document. So, the predominant way I would imagine using these are searching for a bunch of metadata/surrounding context about a file, and then potentially wanting to see what the file itself contains--pretty similar to what was used for example, in malware samples based off of alert triggers in the past.

ebeahan:
> the go code generator is broken for binary types

Yeah, anytime a new type is introduced into ECS a the type has to be [added](https://github.com/elastic/ecs/blob/master/scripts/cmd/gocodegen/gocodegen.go#L276-L295) else the go-generator will fail. 😞 

> Currently the http.*.body.content fields are keyword with a text multi-field. Neither makes sense if you have a binary chunk of data that you want to include in a document. To address this I'm proposing adding a binary-type multi-field called raw to the content fields.

Multi-fields index the same value in different ways for different purposes. How do `keyword` and `text` interact with a `binary` type only accepts the binary value as a Base64 encoded string? I'm not very familiar with the `binary` type and haven't experimented yet with using `binary` as a multi-field, so I'm very likely overlooking the obvious. 😃 

> Ideally we'd have the same sort of ability to return file samples in a content field. In the Endgame product for example, you had the ability to download and analyze malware samples that might have triggered an alert on a monitored system. Having a file.content field to contain such information would standardize this.

Yeah, this definitely sounds like something we should standardize. Out of curiosity, is there a known, existing approach that Elastic Endpoint/Elastic Security here takes when collecting malicious files from an endpoint when malicious activity is detected? Just wondering out loud if there's any other existing efforts we should align with?

djptek:
Hey @andrewstucki are you still actively working on this?

PR #1236, opened 2021-01-26
[CI] Switch to GitHub actions

andrewstucki:
Pretty sure we can switch via something like this. The `remote` addition is so that CI will actually work on forks that don't have the same tags as the `elastic/ecs` repo, you might be able to remove it. Also, might want to look into adjusting the build triggers, right now it's on any push/PR.


.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Unit Tests
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.1'
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - run: git remote add upstream https://github.com/elastic/ecs
```
ebeahan, line 17: > The remote addition is so that CI will actually work on forks that don't have the same tags as the elastic/ecs repo, you might be able to remove it.

Yes, I ran into the same problem with my fork while testing.

Do you see a need for testing to work on forks? I'm leaning towards removing it; all the tests can be ran locally and rely on CI once opening a PR.

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: Tests
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
```
ebeahan, line 7:
```suggestion
    runs-on: ubuntu-20.04
```

nit: I'd prefer explicitly controlling the version over relying on `*-latest`.

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: Tests
+
+on: [push, pull_request]
```
ebeahan, line 3: > Also, might want to look into adjusting the build triggers, right now it's on any push/PR.

I don't think TravisCI is/was any more restrictive, so let's start more permissive and can revisit if necessary.

ebeahan (approved):
LGTM 👍 - Thanks @andrewstucki for putting together something so quickly!

I left some comments, but I think overall this should work well.

.github/workflows/test.yml
andrewstucki: Not particularly, I just can never recall how the branch references work in terms of PRs -- if the `git fetch --prune --unshallow --tags` gets run in the context of the repo of the PR author rather than in the context of the `elastic` org with changes overlayed on it, then it's not going to work if the user doesn't have the latest build tags.

Want me to remove it and then add it back in if things don't work?

ebeahan: Yeah, let's remove it. We can easily re-add if we run into issues.

ebeahan (approved):

PR #1239, opened 2021-01-27
[RFC] Multiple Users - Update to align with new proposal stages

ebeahan:
Update the multiple users RFC proposal's stage to align with new RFC [proposal stages](https://elastic.github.io/ecs/stages.html).

### DIFF
--- a/rfcs/text/0007-multiple-users.md
+++ b/rfcs/text/0007-multiple-users.md
@@ -1,8 +1,8 @@
 # 0007: Multiple users in an event
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **2 (candidate)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **2020-11-11** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Stage: **3 (finished)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **2021-02-17** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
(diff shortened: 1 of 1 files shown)


epixa:
For multi-user, what items remain before we can consider stage 3?

ebeahan:
The stage 3 criteria (minus the PR step):

- [x] Included multiple real-world example source documents
- [ ] No objections from the sponsor, ECS team, or subject matter experts

I feel the proposal doc itself is complete, with several examples and all captured concerns having a documented resolution. The only outstanding item I see is an agreement that the fields as defined are complete. I don't expect any objections to arise, though.

Implementation of ECS 1.8 is ongoing (e.g., https://github.com/elastic/beats/issues/23118) without any major hurdles AFAIK. The multiple user changes are being released as `beta` in 1.8, giving us some flexibility if any adjustments become needed.

ebeahan:
After more discussion out-of-band, the ECS team feels comfortable and confident considering the multiple users RFC as finished.

@adriansr @marc-gr, you both have been doing most of the ECS 1.8 implementation work in beats/integrations (https://github.com/elastic/beats/issues/23118) to adopt the multiple users' fields. Any issues or feedback you have from that work that should be captured before considering the multiple users RFC finished?

adriansr:
@ebeahan I can't think of any issue that could prevent the RFC to be finished.

Only issue that I remember is not being able to map some events that modify multiple users at the same time, understandable as related can't be an array.

And at some point we should give some thought to similar changes for `group.*` fields.

ebeahan:
From the feedback, sounds like we're all in agreement to consider this RFC finished. 🎉 

I've updated the proposal to reflect that change. Once this is merged, I'll open a separate PR to remove the `beta` attribute from these fields in the schema.

ebeahan:
@epixa Would you give this one final look?

epixa (approved):
LGTM

PR #1240, opened 2021-01-27
[RFC] Service Environment Field - Update to align with new proposal stages

ebeahan:
Update the service environment field RFC proposal's stage naming to align with new RFC [proposal stages](https://elastic.github.io/ecs/stages.html).


epixa (approved):
LGTM

PR #1241, opened 2021-01-28
add 'package.namespace' field (#1038)

### LINKED ISSUE #1038: Add package group/namespace field
**Summary**

The [`package`](https://www.elastic.co/guide/en/ecs/current/ecs-package.html) fields should include a field for the package's group/namespace, to accommodate package types like Maven and npm.

**Motivation**:

Many package types include an explicit group/namespace field, which cannot be properly represented with just `name` and `version`.

For example: Maven's [groupId](https://maven.apache.org/guides/mini/guide-naming-conventions.html), and npm's [scopes](https://docs.npmjs.com/about-scopes).

**Detailed Design**:

* Field names: `package.group` or `package.namespace`
* Example values for the fields: `org.elasticsearch.client`
* Suggested appropriate datatypes: keyword
    Downloads / installations from an internal artifact repository:
    ```
    123.456.78.9 - - [20/Oct/2020:00:00:00 -0700] "GET /repository/maven-central/org/elasticsearch/client/elasticsearch-rest-high-level-client/elasticsearch-rest-high-level-client-7.9.2.jar" HTTP/1.0" 200 - 1323 5 "Apache-Maven"
    ```

   ```json
    { 
        "package": {
            "type": "maven",
            "group": "org.elasticsearch.client",
            "name": "elasticsearch-rest-high-level-client",
            "version": "7.9.2",
        }
    }
    ```

rgmz:
Adds a field to represent package namespaces. This fixes #1038 .

### DIFF
--- a/schemas/package.yml
+++ b/schemas/package.yml
@@ -9,6 +9,12 @@
   type: group
   fields:
 
+    - name: namespace
+      level: extended
+      type: keyword
+      description: Package's namespace
+      example: org.elasticsearch.client
+
     - name: name
       level: extended
       type: keyword
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -20,6 +20,7 @@ Thanks, you're awesome :-) -->
 * Added `http.request.id`. #1208
 * Added `cloud.service.name`. #1204
 * Added `hash.ssdeep`. #1169
+* Added `package.namespace`. #1038
 
 #### Improvements
 
--- a/code/go/ecs/package.go
+++ b/code/go/ecs/package.go
@@ -27,6 +27,9 @@ import (
 // contains general information about a package, such as name, version or size.
 // It also contains installation details, such as time or location.
 type Package struct {
+	// Package's namespace
+	Namespace string `ecs:"namespace"`
+
 	// Package name
 	Name string `ecs:"name"`
 
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -4864,6 +4864,22 @@ example: `go`
 // ===============================================================
 
 |
+[[field-package-namespace]]
+<<field-package-namespace, package.namespace>>
+
+| Package's namespace
+
+type: keyword
+
+
+
+example: `org.elasticsearch.client`
+
+| extended
+
+// ===============================================================
+
+|
 [[field-package-path]]
 <<field-package-path, package.path>>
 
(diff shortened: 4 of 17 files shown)


ebeahan:

schemas/package.yml
```diff
@@ -9,6 +9,12 @@
   type: group
   fields:
 
+    - name: namespace
+      level: extended
+      type: keyword
+      description: Package's namespace
```
ebeahan, line 15: Can we expand on the description a bit? I offered a suggestion here but feel free to adjust.

```suggestion
      short: Package's namespace
      description: > 
        Package's namespace.
        
        Some software registries and repositories allow for a namespace as part of a
        package's naming, such as Maven groupId and npm scopes.
        
```

ebeahan (commented):
Thanks, @rgmz!

As mentioned previously (https://github.com/elastic/ecs/issues/1038#issuecomment-718934940), I think we should consider if OS packages have a similar concept to consider here. @andrewkroh @andrewstucki WDYT? Any insights?

andrewstucki:
@ebeahan sorry for the late reply. As far as groups of packages go at the OS level--the packaging ecosystems I'm a bit more familiar with:

1. `yum` -- has a notion of "group" installation i.e. `yum groupinstall` where individual groups can be created on a per-repo basis and managed via `yum-groups-manager`
2. `apt` -- relies on "meta" packages that define all of the desired packages as dependencies

In either case we're not really capturing this sort of information in auditbeat at the moment AFAIK. For example, on systems that use `yum` as a package manager, we're pretty much just querying the rpm cache directly, so we'll have no notion of higher level concepts like yum groups.

To be honest though, this PR does raise an issue in my mind that I think we might want to circle back on. From the docs, I didn't actually ever think about leveraging the `package` fields to model specific programming language package management (i.e. maven, npm, etc.). I'm wondering if we want to clarify in the description of the field set that this is used for both system-level OS packages as well as packages for specific programming languages. I can open an issue for it.

schemas/package.yml
djptek: @rgmz how do you feel about these suggestions from the last review?

kgeller:
This PR has been stale for a while, and we're going to close it.

Feel free to re-open this PR if you think it should stay open and is worth rebasing.

rgmz:
Hey @kgeller, apologies for letting this get stale.

Most of my time is spent on my work laptop, which doesn't allow contributions to GitHub.com. I still believe this is a meaningful change, and will rebase this soon.

PR #1247, opened 2021-01-29
[RFC] Wildcard - stage 2 proposal

ebeahan:
### Summary

Revisions to the stage one wildcard RFC for consideration of advancement for stage two.

Opening this PR with minimal changes. Feedback and concerns as the wildcard changes are adopted can be captured in this PR and incorporated into the proposal as appropriate.

- [x] Opened pull request for this candidate revising the existing draft
- [x] Subject matter experts weighed in on technical utility of field definitions in the pull request
- [x] Complete Elasticsearch benchmarking and discuss indexing and storage impacts based on results.
- [x] Identifies scope of impact of changes to ingestion mechanisms (e.g. beats/logstash), usage mechanisms (e.g. Kibana applications, detections), and the ECS project (e.g. docs, tooling)

[Markdown preview of this RFC](https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-2/rfcs/text/0001-wildcard-data-type.md)


ebeahan:
I'm capturing some general thoughts here as a comment first, and I plan to incorporate them into the proposal doc next. I'm using these insights to revisit the identified fields in this proposal and re-evaluate each one.

### Objectives

Restating the original objectives of the `wildcard` migration in ECS:

* Leverage the improved searching using `wildcard` of regular expressions and leading wildcard on high-cardinality fields. Wildcard excels when searching for something in the middle of strings when there are many unique values to evaluate. It also addresses issues with the partial matching of string values important to [security use cases](https://socprime.com/blog/elastic-for-security-analysts-part-1-searching-strings/).
* Remove security blindspots caused by `keyword` field size limits (`ignore_above`) and the Lucence hard limit (32k for each value) when dealing with large event values or messages.
* Potentially less disk usage for high-cardinality fields
* Potentially simpler search expressions

> NOTE: Case insensitivity is also crucial to fully addressing the security use cases, but outside this proposal's scope. Users can control case sensitivity for both `keyword` and `wildcard` fields using the [`case_insensitive` query flag](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html#term-field-params).

### Impact

`keyword` and `wildcard` are both members of the keyword type family but use different underlying data structures. Wildcard uses two data structures to accelerate wildcard and regexp searching:

1. n-gram index of all three character sequences in the string values
2. binary doc value store of the original values

Wildcard fields require more disk space for the additional n-gram index. This disk cost is only recovered if the full doc values that are also stored compress better than the compression of the keyword fields for the doc values. Keyword values are compressed based on common-prefixes in the values, but wildcard values use LZ4 based on blocks of 32 values. These two compression approaches vary depending on size, duplicate values, the cardinality of the data, and so on.

The ECS team has a better understanding of the underlying data structures and their characteristics. Data's been reviewed from benchmarks comparing `keyword` and `wildcard` fields on the same data sets. We will now revisit the proposed field mappings and suggest changes that avoid excessive impact indexing performance or increased storage usage.

#### Storage and Index Costs

When assembling the initial list of candidate fields to migrate to `wildcard,` we split focus between query performance improvements and removing security blind spots. However, we overlooked the storage and indexing costs when switching fields to be indexed as `wildcard.`

ECS fields will be re-evaluated now in terms of storage and indexing using the following criteria:

* Underestimating cardinality of particular fields. How many unique fields are expected for a given field? Thousands? Hundreds of Thousands? Millions?
* Disk costs for mostly identical values. Which fields are more likely to have values sharing common prefixes and better compression as `keyword`?

#### Query Performance

Query performance of `wildcard` vs. `keyword` hasn't been benchmarked extensively. ECS recognizes some fields may need `wildcard` even with an increase in storage and indexing overhead due to other query benefits.

Keyword vs. wildcard query characteristics:

* `keyword` will perform queries faster for a prefix query (`foo*`) on a low-cardinality field (< hundreds of thousands of unique values) than `wildcard.`
* `wildcard` will perform much faster than `keyword` for leading wildcard or regexp queries but only on a high-cardinality field (> millions of unique values).
* `wildcard` fields should avoid being used extensively for sorting and aggregation features

### Fields to Potentially Remove

The next sections may _not_ be good candidates for `wildcard` after more investigation.

#### File paths and names

File path values are likely to compress extremely well as `keyword` since `keyword` fields have common-prefix-based compression (`wildcard` values are blocks of 32 values compressed into a single LZ4 blob). On top of the worse doc values compression, the number of `postings` also increases significantly due to n-grams.

#### Host and Organization Naming

Hostname values are very likely duplicated from event to event. Even though an index might have thousands of different hosts, it's unlikely to see millions of unique hostname values. 

Depending on an organization's host naming convention, there's also some possibility of common prefixing (hosts named `USNHCDBRD-D001` and `USNHCW2K8-P001` both share the prefix `USNHC`).

Organization names such as `as.organization.name` and `organization.name` might be better suited to `text.` 

#### User Identifiers

User identifiers, like usernames or email addresses, are also likely duplicated across events. Common prefixing is also a potential consideration.

### Fields to Potentially Remain

#### Error Stack Traces

Large, unstructured, unique, and unfriendly with text analyzers. Users will often want to query for a substring in the stack trace's value that would be a good fit for `wildcard.`

#### HTTP Body Content

Like stack traces, the values here may be large, often unique, and wouldn't play well with default text analyzers.

#### URLs, Payloads, Commands, Processes

We need to remain mindful of a [security blindspot](https://twitter.com/cyb3rops/status/1253290280352190465) that wildcard aims to address: Accepting large values that may 1) exceed the current default of `ignore_above: 1024` or 2) exceed the max Lucene size of 32KB for a `keyword` value.

### Field Types to Potentially Add?

#### `message`

Often a large field and is mapped a `keyword` to simplify substring searches. But:

* larger field = many unique values = slow substring searches
* size limits for `keyword` will drop overly large values

Unlike a migration from `keyword` to `wildcard,` migrating from `text` to `wildcard` would be a breaking change. The initial wildcard proposal focused on non-breaking changes of existing keyword fields in ECS, but now understanding the storage/indexing characteristics of `wildcard` better and the new implementation timeline may be worth discussing more for the next ECS major.

#### `log.original`

Similar reasons as `message`: large field, many unique values, size limits many affect accurate ingestion.

epixa:
How do you feel about getting some more subject matter experts identified to help make sure we have coverage for reviews and such?

ebeahan:
> How do you feel about getting some more subject matter experts identified to help make sure we have coverage for reviews and such?

++ to adding more folks to the discussion here. Definitely want additional feedback and insights for how to proceed.

My next steps:

- [ ] Formalize my notes above by capturing them into the proposal
- [ ] Come up with a list of candidate fields to anchor discussions.
- [ ] Pull in other folks to discuss those candidate fields

ebeahan:
I've completed updating the proposal doc. The two significant changes in this revision:

* [Performance differences](https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-2/rfcs/text/0001-wildcard-data-type.md#performance-differences): Captures the concerns and potential resolutions around disk usage and indexing performance.
* [List of proposed fields to target for this initial migration](https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-2/rfcs/text/0001-wildcard-data-type.md#identified-wildcard-fields): This is a much smaller list than before. Fields that were file paths, file names, hostnames, domain names, organization names, or user identifiers were removed.

ebeahan:
@rw-access @leehinman you both weighed in during the earlier wildcard discussions. Would love any feedback you have around these latest proposed changes.

gingerwizard:
@ebeahan do you want a measurement of the performance impact of these changes re disk and indexing throughput?

ebeahan:
@gingerwizard Yes, absolutely, that would be great!

First, I'd like to collect and incorporate any feedback if we have any changes to what I've proposed. I can then follow up to arrange the testing.

There will still be storage increases for some fields, but as I've narrowed down the field list, I've focused on ones that remove the security blind spots caused by keyword value size limits (like `url.original` or `process.command_line`) as a trade-off.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -21,29 +28,10 @@ For a field to use wildcard, it will require changing the field's defined schema
...
-| [`server`](0001/server.yml) | `server.domain`<br> `server.registered_domain` |
-| [`source`](0001/source.yml) | `source.domain`<br> `source.registered_domain` |
-| [`tls`](0001/tls.yml) | `tls.client.issuer`<br> `tls.client.subject`<br> `tls.server.issuer`<br> `tls.server.subject` |
-| [`url`](0001/url.yml) | `url.full`<br> `url.original`<br> `url.path`<br> `url.domain`<br> `url.registered_domain` |
-| [`user`](0001/user.yml) | `user.name`<br> `user.full_name`<br> `user.email` |
-| [`user_agent`](0001/user_agent.yml) | `user_agent.original` |
-| [`x509`](0001/x509.yml) | `x509.issuer.distinguished_name`<br> `x509.subject.distinguished_name` |
+| [`error`](0001/error.yml) | `error.stack_trace` |
+| [`http`](0001/http.yml) | `http.request.body.content`<br> `http.response.body.content` |
+| [`process`](0001/process.yml) | `process.command_line` |
```
rw-access, line 33: If we could also keep `file.path` and `process.executable`, I think that would be valuable. I think it's okay to ignore `process.parent.executable` though.

rfcs/text/0001-wildcard-data-type.md
```diff
@@ -21,29 +28,10 @@ For a field to use wildcard, it will require changing the field's defined schema
...
-| [`file`](0001/file.yml) | `file.directory`<br> `file.path`<br> `file.target_path` |
-| [`geo`](0001/geo.yml) | `geo.name` |
-| [`host`](0001/host.yml) | `host.hostname`<br> |
-| [`http`](0001/http.yml) | `http.request.referrer`<br> `http.request.body.content`<br> `http.response.body.content` |
-| [`log`](0001/log.yml) | `log.file.path`<br> `log.logger` |
-| [`organization`](0001/organization.yml) | `organization.name` |
-| [`os`](0001/os.yml) | `os.name`<br> `os.full` |
-| [`pe`](0001/pe.yml) | `pe.original_file_name` |
-| [`process`](0001/process.yml) | `process.command_line`<br> `process.executable`<br> `process.name`<br> `process.thread.name`<br> `process.title`<br> `process.working_directory`<br> |
-| [`registry`](0001/registry.yml) | `registry.key`<br> `registry.path`<br> `registry.data.strings` |
```
rw-access, line 39: This change makes sense when I think about it, because these wildcards are often not leading wildcards, so they are less costly to do on a keyword fields.

rw-access:
Commented on a few of the fields. Thanks for the changes!

FYI @ebeahan I made a feature request to Elasticsearch for better query planning on wildcard fields. https://github.com/elastic/elasticsearch/issues/70612
In my experience writing rules, most of the time that I use wildcards, it's in combination with an `term` query on a keyword field. My hope from the ticket is that we can have better runtime performance in these cases without any additional storage or ingest cost. 🤞 

I don't think it'll be on the road map for a while but I think it could lead to a huge savings in search time, for many common scenarios.

rfcs/text/0001-wildcard-data-type.md
rw-access: `registry.data.strings` though, could be worth keeping

rfcs/text/0001-wildcard-data-type.md
ebeahan: A large area of feedback from benchmarking was file paths would likely compress very well as `keyword` since file paths share prefixes. Even fields with moderate cardinality (50,000 - 100,000 unique values) saw increases of 20-30x storage space 😞 

Doesn't mean that all fields that capture file paths are out of consideration for `wildcard`; we just need to weigh any gains against the (fairly steep) costs.

rfcs/text/0001-wildcard-data-type.md
ebeahan: Ok, I'm good with keeping `registry.data.strings`. I'll re-add.

ebeahan:
@elastic/security-external-integrations can someone take a look at this revised, shortened list of fields proposed for `wildcard`?

https://github.com/ebeahan/ecs/blob/wildcard-rfc-stage-2/rfcs/text/0001-wildcard-data-type.md#identified-wildcard-fields

gingerwizard:
@ebeahan we have a [runbook for testing](https://github.com/elastic/elasticsearch-sizing/blob/master/logging/run_books/index_compression.md) these changes using our tooling. We can assist and help enable someone on your team to test changes like this and others moving forward.

djptek:
>>help enable someone on your team to test changes like this and others moving forward.

@ebeahan I'm in

ebeahan:
> @ebeahan we have a runbook for testing these changes using our tooling. We can assist and help enable someone on your team to test changes like this and others moving forward.

Fantastic! Thanks, @gingerwizard!

ebeahan:
Latest update:

I'm in the process of running a benchmark for these `wildcard` changes. Hitting some access issues with the tooling, but I'm working through that with the ES performance team.

ebeahan:
I was able to complete the benchmarking. Full details were collected, but here is a summary of some of the results.

### Indexing performance

| Metric | Task | Baseline | Contender | Diff  | Unit |
| ------ | ----- | -------- | ----------- | ---- | ---- |
|   Mean Throughput |  bulk-index |     12131.9 |     11534.8 | -597.129 | docs/s |
|   Median Throughput |  bulk-index |     12118.8 |     11518.2 | -600.571 | docs/s |
|   Max Throughput |  bulk-index |     12430.1 |     11771.2 |  -658.91 | docs/s |

Indication of approximately 5% reduction in indexing performance.

### Disk Usage

#### Keyword

Raw Message: 91.05 GB
Index Size: 120.51 GB
Compression Ratio: 1.393

#### Wildcard

Raw message: 91.05 GB
Index Size: 125.15 GB
Compression Ratio: 1.460

Indication of approximately 5% additional disk usage in total.

### Additional notes

We continue to see a significant increase in postings. While the overall disk usage increase was smaller due to the smaller set of `wildcard` fields, 

Example field comparison from the application log set:

Field name: `error.stack_trace`

#### Keyword

total: 26,957
postings: 138
docvalues: 25,218

#### Wildcard

total: 627,878,375
postings: 68,248,307
docvalues: 559,612,154

Total: 26MB vs. 613MB

Another field comparison from nginx access logs:

Field name: `url.original`

#### Keyword

total: 661,065,465
postings: 141,514,804
docvalues: 478,645,564

#### Wildcard

total: 6,255,586,374
postings: 2,897,411,659
docvalues: 3,358,094,638

Total: 645MB vs. ~6GB

ebeahan:
Current list of candidate `wildcard` candidates:

| Field Set | Field(s) |
| --------- | -------- |
| [`error`](0001/error.yml) | `error.stack_trace` |
| [`http`](0001/http.yml) | `http.request.body.content`<br> `http.response.body.content` |
| [`process`](0001/process.yml) | `process.command_line` |
| [`registry`](0001/registry.yml) | `registry.data.strings` |
| [`url`](0001/url.yml) | `url.full`<br> `url.original`<br> `url.path` |

This list was adopted for a group of integration packages - see https://github.com/elastic/integrations/issues/1163.

ebeahan:
@elastic/ecs I believe this PR is ready for a final ECS team review before advancing the current proposal to stage 2.

djptek (approved):
LGTM!

PR #1042, opened 2020-10-20
Add support for `scaling_factor` in the generator

webmat:
Closes #1033

PR #1250, opened 2021-02-02
Add scaled_float type to go generator

ebeahan:
When a field uses `type: scaled_float` in the field definitions, `make` fails due to `scaled_float` not being a defined data type in `go` generator:

```
$ make
build/ve/bin/python scripts/generator.py --strict --include ""
build/ve/bin/python scripts/use-cases.py --stdout=true >> /dev/null
find code/go/ecs -name '*.go' -not -name 'doc.go' | xargs rm
cd scripts \
          && GO111MODULE=on go run cmd/gocodegen/gocodegen.go \
                -version=2.0.0-dev \
                -schema=../schemas \
                -out=../code/go/ecs
no translation for scaled_float (field cpu.usage)
exit status 1
make: *** [gocodegen] Error 1
```

This fix adds a type mapping for `scaled_float`.

I'll plan on backporting to 1.8 as well.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -33,6 +33,7 @@ Thanks, you're awesome :-) -->
 #### Bugfixes
 
 * Clean up `event.reference` description. #1181
+* Go code generator fails if `scaled_float` type is used. #1250
 
 #### Added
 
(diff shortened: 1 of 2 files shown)


kaiyan-sheng:
@ebeahan Thanks for fixing this! Question: do we also need to add support for `scaling_factor` here or this is already done?

ebeahan:
@kaiyan-sheng We added support for `scaling_factor` back in https://github.com/elastic/ecs/pull/1042, so I believe we'll be set once the other fix is merged.

kaiyan-sheng (approved):

PR #1254, opened 2021-02-04
Add nested type support to go code generator

ebeahan:
Introducing `type: nested` will require handling of `nested` fields be defined, else the go code generator will fail.


scripts/cmd/gocodegen/gocodegen.go
```diff
@@ -286,7 +286,7 @@ func goDataType(fieldName, elasticsearchDataType string) string {
 		return "time.Time"
 	case "boolean":
 		return "bool"
-	case "object":
+	case "object", "nested":
```
ebeahan, line 289: @andrewkroh Foresee any issues with adding this mapping for `nested`?

ebeahan:
Adjusted this PR to add both `nested` as well as `flattened` types to the generator's mapping logic.

marc-gr:
Is it possible to add some test cases un the `tests` folder for this addition?

andrewstucki:
@ebeahan @marc-gr I know that there is a plan to eventually strip out the go generators from the ECS repo, does it make sense to do so now? I believe that the only elastic project using the go code is packetbeat, and I think it's pinned to an older version. Otherwise, if we're planning on keeping these around for awhile, maybe we could add the [binary type](https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html), making it `[]byte` would allow the default json marshaler to automatically do base64 encoding.

ebeahan:
@andrewstucki @marc-gr Thanks for taking a look. These are some good discussion points, but I don't want to expand the original scope here too much.  Any objections to the proposed mapping for `nested` and `flattened`?

> Is it possible to add some test cases un the tests folder for this addition?

Ideally, we'd have something in place for testing the go generator portions of the repo, but we don't today. I don't think we should add that testing framework into the ECS repo right now since...

> I know that there is a plan to eventually strip out the go generators from the ECS repo, does it make sense to do so now?

...we definitely want to remove the go generator and eventually the other golang dependencies from this repo. 

I have no issue with pulling it out, but I'm also not familiar with where it's a dependency. A team or individual would need to take ownership of setting up the new repo; I don't see this as something the ECS team would be best to maintain (right now at least).

marc-gr:
> Ideally, we'd have something in place for testing the go generator portions of the repo, but we don't today. I don't think we should add that testing framework into the ECS repo right now

👍

scripts/cmd/gocodegen/gocodegen.go
andrewkroh: For `nested` maybe we should generate a more strict data type (one that incorporates the nested object's schema). Assuming we know the schema for the nested object then generating a type like `[]Segment` (in the case of PE segments) would be superior IMO.

That would look like:

```
type Pe struct {
    Segments []Segment
}

type Segment {
    Type string
    Sections string
}

or 

type Pe struct {
    Segments []struct{
        Type string
        Sections string
    }
}
```

ebeahan: Yeah, I understand how having the schema more strictly defined would be superior. Unfortunately, this is currently a blocker to adding `nested` and `flattened` types since not having the top defined here causes an exception when generating a new set of artifacts. 

If this is something that we definitely should do right, I'm more than happy to help make that happen. Though, do we know if anything consuming the code generated here relies on `nested`? I prefer to avoid sinking time into a larger effort if the need isn't really there right now.

andrewstucki: @ebeahan to my knowledge the only thing on the Elastic side of things that consumes this right now is Packetbeat, which doesn't use any nested fields.

ebeahan:
Will focus on solely on adding `nested` type support here. Separate PR opened to add support for `flattened`: https://github.com/elastic/ecs/pull/1302.

ebeahan:
Took a shot at the improved implementation suggested in https://github.com/elastic/ecs/pull/1254#discussion_r574843883. 

@andrewkroh @andrewstucki would appreciate a look.

andrewkroh (commented):
I think this is good, but I would find it very helpful if you could temporarily add some nested fields to demonstrate what the generated code looks like and then paste that into comment.

ebeahan:
I generated some examples using some `nested` fields from the experimental schema: https://gist.github.com/ebeahan/2c3038b0accd1786f78d7d7f0b4353da

andrewkroh:
Thanks! It looks like the tag is missing the key name. It has 

```
	// ELF object segment list.
	Segments []Segments `ecs:""`
```

where I expected

```
Segments []Segments `ecs:"segments"`
```

ebeahan:
Thanks, @andrewkroh. I believe it's now addressed.

Update the sample gist with the latest [versions](https://gist.github.com/ebeahan/2c3038b0accd1786f78d7d7f0b4353da/revisions).

andrewkroh (approved):

PR #1261, opened 2021-02-11
Stage 1 changes for RFC 0015 - add elf fieldset

ebeahan:
Adding `elf.*` fieldset from [RFC 0015](https://github.com/elastic/ecs/blob/master/rfcs/text/0015-create-file-elf.md) (https://github.com/elastic/ecs/pull/1077) to the experimental schema and artifacts.

For reviewers, the bulk of the changes are located in `experimental/schemas/elf.yml`. Most files are generated automatically based on files in `experimental/schemas`.

### DIFF
--- a/experimental/schemas/elf.yml
+++ b/experimental/schemas/elf.yml
@@ -1,0 +1,198 @@
+---
+- name: elf
+  title: ELF Header
+  group: 2
+  description: >
+    These fields contain Linux Executable Linkable Format (ELF) metadata.
+  type: group
+  reusable:
+    top_level: false
+    expected:
+      - file
+      - process
+  fields:
+    - name: creation_date
+      short: Build or compile date.
+      description: >
+        Extracted when possible from the file's metadata. Indicates when it was
+        built or compiled. It can also be faked by malware creators.
+      type: date
+      level: extended
+
+    - name: architecture
+      description: >
+        Machine architecture of the ELF file.
+      type: keyword
+      level: extended
+      example: ARM, x86-64, etc
+
+    - name: byte_order
+      description: >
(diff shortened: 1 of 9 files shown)


experimental/schemas/elf.yml
```diff
@@ -0,0 +1,198 @@
...
+        built or compiled. It can also be faked by malware creators.
+      type: date
+      level: extended
+
+    - name: architecture
+      description: >
+        Machine architecture of the ELF file.
+      type: keyword
+      level: extended
+      example: ARM, x86-64, etc
```
ebeahan, line 27:
```suggestion
      example: x86-64
```

No preference on which one, but let's pick one single value as the example here.

experimental/schemas/elf.yml
```diff
@@ -0,0 +1,198 @@
...
+      type: keyword
+      level: extended
+      example: Little Endian, Big Endian
+
+    - name: cpu_type
+      description: >
+        CPU type of the ELF file.
+      type: keyword
+      level: extended
+      example: Intel, PowerPC, RISC, etc.
```
ebeahan, line 41:
```suggestion
      example: Intel
```

Again, let's pick one example value to use.

experimental/schemas/elf.yml
devonakerr: I like the suggested edit, it will have longevity as future architectures become supported.

devonakerr (commented):
@peasead I think just two suggested changes if you don't mind taking a look.

ebeahan:
Moving forward with merging these experimental changes before the 1.9 FF.

PR #1127, opened 2020-11-16
[RFC] Threat Intel - Stage 2

shimonmodi:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)?
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)?
- If submitting code/script changes, have you verified all tests pass locally using `make test`?
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?
- Is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed.
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)?

[Preview the RFC](https://github.com/shimonmodi/ecs/blob/shimonmodi-stage-2/rfcs/text/0008-threat-intel.md)

PR #1268, opened 2021-02-17
Stage 1 experimental changes for RFC 0008 - threat.indicator fields

ebeahan:
Adding `threat.indicator.*` fields from [RFC 0008](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md) (#1127) to the experimental schema and artifacts.

For reviewers, the significant changes are located in `experimental/schemas/threat.yml`. Most files are generated automatically based on files in `experimental/schemas`.


ebeahan:
Moving forward with merging these experimental changes before the 1.9 FF.

Issue #1249, opened 2021-02-02
Add 2 fields to code signature


Trinity2019:
**Summary**

Add following fields to [code_signature](https://www.elastic.co/guide/en/ecs/current/ecs-code_signature.html):
```
code_signature.team_id
code_signature.signing_id
```

**Motivation**:

performance enhancement: Collecting the team id and signing id can be done very fast whereas the signer name can be slow so there are cases where a data shipper may not want to report the signer name and just report the team id and signing id.

Need to point out team id and signing id are *OS(including macOS) specific and important for *OS. Combined together they are enough for *OS to identify/verify the origin and authenticity of a software vendor's code. Apple takes care of the uniqueness of team id and also checks the reputation of the team/company. Signing id is also important for *OS to identify applications from a software vendor.

Other big venders uses team id and signing id(a.k.a. bundle id) for profiling/whitelisting/etc. purpose: 
https://docs.microsoft.com/en-us/mem/intune/configuration/kernel-extensions-settings-macos
https://derflounder.wordpress.com/2018/04/12/whitelisting-third-party-kernel-extensions-using-profiles/
https://www.jamf.com/jamf-nation/discussions/28183/configure-apple-team-id-in-mdm-profile

**Detailed Design**:

* example team_id: 
  `EQHXZ8M8AV` // which is a team id for Google
  `UBF8T346G9` // which is a team id for Microsoft
  suggested datatype for team id will be string
* example signing_id:
  `com.apple.xpc.launchd` // which is the signing id for `/sbin/launchd`
  `com.apple.xpc.proxy` // which is the signing id for `/usr/libexec/xpcproxy`
  suggested datatype for signing id is string

ebeahan:
Thanks for the detailed write-up, @Trinity2019! Would you be willing to open a PR for these changes?

I'm not familiar with code signing between different OSs, so the answer to the following may be a simple "no". 😅 

Are there similar values/IDs provided by different OSs that would ever be relevant values for these fields, in addition to macOS/OSX?

ebeahan:
Closed via #1269

Trinity2019:
Thank you @ebeahan for all the help! and thanks @rw-access for the PR review!

PR #1269, opened 2021-02-17
Add 2 fields to code_signature

Trinity2019:
Issue: https://github.com/elastic/ecs/issues/1249


schemas/code_signature.yml
```diff
@@ -57,3 +57,25 @@
...
+      
+    - name: team_id
+      level: extended
+      type: keyword
+      short: The team identifier used to sign the process.
+      description: >
+        The team identifier used to sign the process.
+
+        This is used to identify the team or vendor of a software product.
+        Leave unpopulated if unavailable.
```
ebeahan, line 69:
```suggestion
```

nit: This is the general guidance for ECS fields, so we can shorten the description by omitting.

ebeahan (approved):
LGTM 👍 

I had minor note on the description, but everything looks great! Thanks for adding the example values. A rebase should fix the `CHANGELOG.next` conflict.

@rw-access Would you take a look at these additions to `code_signature.*`? Any feedback?

experimental/generated/beats/fields.ecs.yml
```diff
@@ -969,6 +999,16 @@
       description: Subject name of the code signer
       example: Microsoft Corporation
       default_field: false
+    - name: code_signature.team_id
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: 'The team identifier used to sign the process.
```
rw-access, line 1006: I think for both of these, it's worth noting that it's mostly relevant to macOS or other Apple operating systems. It might not be intuitive for people trying to map this field to a Windows sensor, like Elastic Endpoint or Sysmon.

experimental/generated/beats/fields.ecs.yml
```diff
@@ -1636,6 +1676,16 @@
       description: Boolean to capture if a signature is present.
       example: 'true'
       default_field: false
+    - name: code_signature.signing_id
```
rw-access, line 1679: is this a duplicate?

schemas/code_signature.yml
```diff
@@ -57,3 +57,24 @@
         This is useful for logging cryptographic errors with the certificate validity or trust status.
         Leave unpopulated if the validity or trust of the certificate was unchecked.
       example: ERROR_UNTRUSTED_ROOT
+      
+    - name: team_id
+      level: extended
+      type: keyword
+      short: The team identifier used to sign the process.
+      description: >
```
rw-access, line 65: I think for both of these, it's worth noting that it's mostly relevant to macOS or other Apple operating systems. It might not be intuitive for people trying to map this field to a Windows sensor, like Elastic Endpoint or Sysmon.

ebeahan: Makes sense.

I had initially wondered if we could make these fields more generic for use across different OS types, but based on what @Trinity2019 [shared](https://github.com/elastic/ecs/issues/1249#issuecomment-772062111), Windows and macOS don't seem to align enough.

Trinity2019: I added suggested comment: https://github.com/elastic/ecs/pull/1269/commits/5a6d1b20e1865d265956859a996b433294154183
Will fix other `make` generated files once you guys approve this change^

ebeahan: LGTM

Trinity2019: Thanks @ebeahan !

ebeahan (approved):
LGTM 👍

PR #1283, opened 2021-03-01
Set go version to 1.15.x in tests 

ebeahan:
Current Github action config allows `go` version to update the minor, which was causing [unexpected errors](https://github.com/elastic/ecs/pull/1230#issuecomment-787019042) when CI runs `make check`.

This change will lock the `go-version` at `1.15`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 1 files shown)


djptek (approved):
nodejs convention [https://semver.npmjs.com/] implies `go-version: '~1.15.0'` however, I tried that with `python-version: '~3.0'` and the latter fails so I agree it would be better to use "Major.Minor.x" to have consistency across `test.yml`

LGTM

PR #1288, opened 2021-03-04
Add file.fork_name to ECS

### LINKED ISSUE #1285: Add Alternate Data Stream Name to File Events
**Summary**

A category of [malwares](https://blog.malwarebytes.com/101/2015/07/introduction-to-alternate-data-streams/) tend to leverage file alternate data stream content to store malicious code in an attempt to avoid AV file system scan. The file name will contain a colon ":" symbol and for rules creation or simply searches it become an expensive operation as one need to add prefix wildcards to file.name:*\:* 

Below an example of how the ADS file name looks like: 

image

We need an [ECS field](https://www.elastic.co/guide/en/ecs/current/ecs-file.html) that captures this info (e.g. file.alternate_data or file.ads)

**Motivation**:

**Detailed Design**:

* Field names
* Example values for the fields
* Suggested appropriate datatypes

rw-access:
Resolves #1285

Added `file.fork_name` to ECS.


CHANGELOG.next.md
```diff
@@ -47,6 +47,7 @@ Thanks, you're awesome :-) -->
 * Extended `pe` fields added to experimental schema. #1256
 * Add `elf` fieldset to experimental schema. #1261
 * Add `threat.indicator` fields to experimental schema. #1268
+* Added `file.fork_name` field #1288
```
ebeahan, line 50: Can this entry be moved above under `Unreleased > Schema Changes > Added` instead?

schemas/file.yml
```diff
@@ -184,3 +184,21 @@
         https://www.iana.org/assignments/media-types/media-types.xhtml[IANA official types],
         where possible. When more than one type is applicable, the most specific
         type should be used.
+
+    - name: fork_name
+      level: extended
+      type: keyword
+      description: >
+        A fork is additional data associated with a filesystem object.
```
ebeahan, line 192: Thanks for the detailed description. 👍 

This type of "real-world usage" detail here would be great to capture in a `file.*` usage doc, similar to what was done with [user](https://www.elastic.co/guide/en/ecs/current/ecs-user-usage.html).

Not necessary for this addition, though, but a note for when it's revisited. 😃

schemas/file.yml
```diff
@@ -184,3 +184,21 @@
...
+
+        On Linux, a resource fork is used to store additional data with a filesystem object. A file always has at
+        least one fork, for the data portion and additional forks may exist.
+
+        On Windows, this is analogous to an Alternate Data Stream (ADS), and the default data stream for a file is just called $DATA.
```
ebeahan, line 197:
```suggestion
        On Windows NTFS, this is analogous to an Alternate Data Stream (ADS), and the default data stream for a file is just called $DATA.
```

Should we explicitly call out NTFS?

schemas/file.yml
```diff
@@ -184,3 +184,21 @@
...
+        Zone.Identifier is commonly used by Windows to track contents downloaded from the Internet. An ADS is typically of the form:
+        `C:\path\to\filename.extension:some_fork_name`, and `some_fork_name` is the value that should populate `fork_name`.
+        `filename.extension` should populate `file.name`, and `extension` should populate `file.extension`.
+        The full path, `file.path` will include the fork name.
```
ebeahan, line 201:
```suggestion
        The full path, `file.path`, will include the fork name.
```

schemas/file.yml
rw-access: I could go either way, we could just say "On Windows, ..."

schemas/file.yml
rw-access: Good point. Yeah this info spans multiple fields so it makes sense to capture that at a higher level.

There's still work to do on the PR and some cleanup in my end, so I'm gonna do another round or two of edits

schemas/file.yml
ebeahan: Ok, I'm good either way as well. Author's choice then 😄

schemas/file.yml
djptek: @rw-access are you happy with the suggested changes from the last review?

If you run `make` locally and `commit` the changed artifact files this can often resolve the failed checks

ebeahan:
@rw-access Is this addition still something you want to move forward on?

rw-access:
Revived the PR and moving from draft to ready for review

CHANGELOG.next.md
```diff
@@ -74,6 +74,8 @@ Thanks, you're awesome :-) -->
 
 ## Unreleased
 
+* Added `file.fork_name` field #1288
```
ebeahan, line 77: Can you bump this up under `#### Added` at line 23?

ebeahan (approved):
One note about the changelog entry, but otherwise LGTM!

I merged https://github.com/elastic/ecs/pull/1509 this morning which changed the `version` in the master branch, and it's the cause of the merge conflicts.

Pulling in those changes and re-running `make` should clear it up.

CHANGELOG.next.md
rw-access: 199c64a

PR #1292, opened 2021-03-05
Stage 1 experimental changes for RFC 0012 - orchestrator fieldset

ebeahan:
Adds `orchestrator.*` fieldset from [RFC 0012](https://github.com/elastic/ecs/blob/master/rfcs/text/0012-orchestrator-field-set.md) (https://github.com/elastic/ecs/pull/1230) to the experimental schema and artifacts.

### DIFF
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -4163,7 +4163,7 @@
       level: extended
       type: keyword
       ignore_above: 1024
-      description: Organization affected by the event (for multi-tenant orchestrator
+      description: Organization affected by the event (for multi-tenant orchestrator setups).
         setups).
       example: elastic
       default_field: false
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -4120,6 +4120,74 @@
       type: keyword
       ignore_above: 1024
       description: Observer version.
+  - name: orchestrator
+    title: Orchestrator
+    group: 2
+    description: Fields that describe the resources which container orchestrators
+      manage or act upon.
+    type: group
+    fields:
+    - name: api_version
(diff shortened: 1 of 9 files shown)


experimental/schemas/orchestrator.yml
```diff
@@ -0,0 +1,69 @@
...
+      level: extended
+      type: keyword
+      description: >
+        Name of the cluster.
+
+    - name: cluster.url
+      level: extended
+      type: keyword
+      description: >
+        URL of the cluster.
```
jsoriano, line 21: Perhaps this description should be specific about being the URL of the API used to manage the cluster.
Something like this.
```suggestion
        URL of the API used to manage the cluster.
```

jsoriano (approved):
Thanks! Added a small suggestion about the docs of a field, not sure if this can be applied now after the RFC.

experimental/generated/beats/fields.ecs.yml
```diff
@@ -4120,6 +4120,74 @@
...
+      type: keyword
+      ignore_above: 1024
+      description: Namespace in which the action is taking place.
+      example: kube-system
+      default_field: false
+    - name: organization
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: Organization affected by the event (for multi-tenant orchestrator
```
kaiyan-sheng, line 4166:
```suggestion
      description: Organization affected by the event (for multi-tenant orchestrator setups).
```

kaiyan-sheng (approved):
lgtm, besides one nit :)

experimental/schemas/orchestrator.yml
ebeahan: Thanks, great suggestion!

Small adjustments to improve descriptions are fine, as long as we're not significantly changing the intent of the field.

PR #1293, opened 2021-03-08
[RFC] Create Threat Fieldset - Stage 2 Proposal

peasead:
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)? yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? NA
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? NA
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? Yes

**CC**
- @devonakerr 
- @shimonmodi 

[Preview of the markdown proposal](https://github.com/elastic/ecs/blob/create-threat-stage-2/rfcs/text/0008-threat-intel.md)


rfcs/text/0008-threat-intel.md
```diff
@@ -262,109 +262,47 @@ There are many sources of threat intelligence including open source, closed sour
...
-2020-10-29 19:16:35,190.45.24.210,80,2020-11-02,Heodo
-2020-10-29 19:16:32,109.242.153.9,80,2020-11-02,Heodo
-2020-10-29 19:16:28,169.1.39.242,80,2020-11-02,Heodo
-2020-10-29 19:14:24,201.171.244.130,80,2020-11-02,Heodo
-2020-10-29 19:14:20,64.207.182.168,8080,2020-11-02,Heodo
-2020-10-29 19:14:19,173.173.254.105,80,2020-11-02,Heodo
-2020-10-29 19:14:16,153.204.122.254,80,2020-10-30,Heodo
-2020-10-29 19:14:13,201.163.74.203,80,2020-11-02,Heodo
-```
+- Abuse.ch Malware - This dataset from Abuse.ch provides a list of malware hashes.
```
ebeahan, line 265: The additional details and examples of the intelligence sources are great! 🙌

ebeahan (commented):
Leaving observations here that I couldn't easily add into the file diff:

* I noticed in both usage examples the `threat.indicator.type` field is an array. Is that expected? If so, let's update the field definition to reflect.
* I originally proposed using `wildcard` for `threat.indicator.description`. Since we're reevaluating `wildcard` usage in ECS for now, I suggest we revert to `keyword.`
* `threat.indicator.dataset` has a typo in the `Example` column: `theatintel`
* For `threat.indicator.marking.tlp`, I realized the examples here are using all caps, but the expected values in the field's description are only capitalizing the first letter (`RED` vs. `Red`). The TLP convention appears to use all caps for TLP colors. Do we imagine most users will search TLP designators using all caps? Do we want to explicitly specify a convention for the field or leave it flexible?
* Reviewing the second example mapping, `Adding threat intelligence match/enrichment to another document which could be in a source event index or signals index.`, it looks like the `indicator.matched.*` fields could be a list. If these values could be a list, do we see these fields residing under a `nested` object?
* @peasead @dcode, you've both contributed to the proposal extensively, so feel free to also list yourselves as authors in addition to SMEs under `People.` 😃

peasead:
* I noticed in both usage examples the `threat.indicator.type` field is an array. Is that expected? If so, let's update the field definition to reflect.
  [x] Done
  - I was just providing multiple examples in the table. I've updated it to be clearer on the markdown doc and used "expected values", in the yml, to reflect all of the other options.
* I originally proposed using `wildcard` for `threat.indicator.description`. Since we're reevaluating `wildcard` usage in ECS for now, I suggest we revert to `keyword.`
  [x] Done
* `threat.indicator.dataset` has a typo in the `Example` column: `theatintel`
  [x] Done
* For `threat.indicator.marking.tlp`, I realized the examples here are using all caps, but the expected values in the field's description are only capitalizing the first letter (`RED` vs. `Red`). The TLP convention appears to use all caps for TLP colors. Do we imagine most users will search TLP designators using all caps? Do we want to explicitly specify a convention for the field or leave it flexible?
   [x] Updated all TLP examples to match
  - I think we should leave it flexible. MISP, as an example, is using all lower-case TLP. I would prefer to set the examples to align with the standard, but I wouldn't want an error or to reject based on case.
* Reviewing the second example mapping, `Adding threat intelligence match/enrichment to another document which could be in a source event index or signals index.`, it looks like the `indicator.matched.*` fields could be a list. If these values could be a list, do we see these fields residing under a `nested` object?
  - I think that was the idea, @rylnd does that sound right?
* @peasead @dcode, you've both contributed to the proposal extensively, so feel free to also list yourselves as authors in addition to SMEs under `People.` 😃 
  [x] Done

rylnd:
> Reviewing the second example mapping, `Adding threat intelligence match/enrichment to another document which could be in a source event index or signals index.`, it looks like the `indicator.matched.*` fields could be a list. If these values could be a list, do we see these fields residing under a `nested` object?
> * I think that was the idea, @rylnd does that sound right?

As I see it, there are two proposed use cases for this fieldset: the indicator data themselves, and the enrichment of indicator/match data on extant event data. Defining a common mapping for both uses has several options/tradeoffs that should be discussed:

1. `threat.indicator` is a nested object
   a. Pro: Allows enriched events to be queried by individual indicators
   b. Con: storage/performance issues with ingestion of nested documents from beats/agents
1. `threat.indicator` is a regular object
   a. Con: enriched events lose the context of the indicators that they matched: we retrieve a [flattened](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-arrays-flattening-objects) object and cannot accurately query e.g. "events which matched a domain from provider A"
   b. Con: mapping conflict with detection alerts
     * Because 1a was deemed critical functionality for CTI rules, Detections [already uses](https://github.com/elastic/kibana/pull/89899) a nested mapping for enriched alerts. Conflicts with e.g. detection rules written against filebeat-* could potentially be solved within the detection engine, but fundamentally there would be a mapping conflict between indicator indices and alerts indices.

peasead:
> > Reviewing the second example mapping, `Adding threat intelligence match/enrichment to another document which could be in a source event index or signals index.`, it looks like the `indicator.matched.*` fields could be a list. If these values could be a list, do we see these fields residing under a `nested` object?
> > 
> > * I think that was the idea, @rylnd does that sound right?
> 
> As I see it, there are two proposed use cases for this fieldset: the indicator data themselves, and the enrichment of indicator/match data on extant event data. Defining a common mapping for both uses has several options/tradeoffs that should be discussed:
> 
> 1. `threat.indicator` is a nested object
>    a. Pro: Allows enriched events to be queried by individual indicators
>    b. Con: storage/performance issues with ingestion of nested documents from beats/agents
> 2. `threat.indicator` is a regular object
>    a. Con: enriched events lose the context of the indicators that they matched: we retrieve a [flattened](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-arrays-flattening-objects) object and cannot accurately query e.g. "events which matched a domain from provider A"
>    b. Con: mapping conflict with detection alerts
>    
>    * Because 1a was deemed critical functionality for CTI rules, Detections [already uses](https://github.com/elastic/kibana/pull/89899) a nested mapping for enriched alerts. Conflicts with e.g. detection rules written against filebeat-* could potentially be solved within the detection engine, but fundamentally there would be a mapping conflict between indicator indices and alerts indices.

Thanks, @rylnd that was my understanding as well and just wanted to double-check.

@ebeahan if you were asking if `threat.indicator.matched.*` would be a nested object, it looks like we're in agreement that it should be. Does that answer your question?

ebeahan:
> @ebeahan if you were asking if threat.indicator.matched.* would be a nested object, it looks like we're in agreement that it should be. Does that answer your question?

In the second example mapping, the `matched` field is a single field, but the comment states it may be a list:

```
                /* `matched` will provide context about which of the indicators above matched on this
                    particular enrichment. If multiple matches for this indicator object, this could
                    be a list */
                "matched": "sha256",
```

The list of fields in the proposal defines `threat.indicator.matched.*` as an object with three child fields:

* `threat.indicator.matched.atomic`
* `threat.indicator.matched.field`
* `threat.indicator.matched.type`

I wanted to understand better which direction to take and make sure that's captured correctly. Right now, the field definitions don't reflect that these fields are either a list or a nested object:

```yaml
  - name: indicator.matched.atomic
    level: extended
    type: keyword
    short: Indicator atomic match
    description: >
      Identifies the atomic indicator that matched a local environment endpoint or network event.
    example: example.com

  - name: indicator.matched.field
    level: extended
    type: keyword
    short: Indicator field match
    description: >
      Identifies the field of the atomic indicator that matched a local environment endpoint or network event.
    example: file.hash.sha256

  - name: indicator.matched.type
    level: extended
    type: keyword
    short: Indicator type match
    description: >
      Identifies the type of the atomic indicator that matched a local environment endpoint or network event.
    example: domain-name
```

rfcs/text/0008/threat.yml
```diff
@@ -161,10 +161,10 @@
       Traffic Light Protocol sharing markings.
 
       Expected values are:
```
ebeahan, line 163:
```suggestion
      Recommendedded values are:
```

> I think we should leave it flexible. MISP, as an example, is using all lower-case TLP. I would prefer to set the examples to align with the standard, but I wouldn't want an error or to reject based on case.

I'm also fine with the uppercase values being a "soft" recommendation.

Let's leave the examples as all-uppercase, but instead of `Expected`, how about rephrasing to `Recommended`?

rylnd:
Briefly capturing a few notes from recent discussions around this work:

* Some of the discussion above (mostly me, sorry 😅) pertains to the dual usage of this fieldset, and the presence/absence/mappings of these fields in each situation. It's now my understanding that, while necessary, those details are orthogonal to the RFC itself, and should mainly be captured in a separate document pertaining to examples/usage.
* In order to provide more context for the enrichment use case, we'd like to prescribe adding the `event` fieldset (`threat.indicator.event.*`). @ebeahan should that be captured in this proposal, or should this sort of composition only be captured in the aforementioned examples/usage documentation?

devonakerr (approved):
I think the move from wildcard to keyword is the right one for threat.indicator.description, and acknowledge the TLP value updates. Also pleased to see the author suggestions incorporated.

ebeahan:
> Some of the discussion above (mostly me, sorry 😅) pertains to the dual usage of this fieldset and the presence/absence/mappings of these fields in each situation. It's now my understanding that, while necessary, those details are orthogonal to the RFC itself and should mainly be captured in a separate document pertaining to examples/usage.

Good context to capture to some extent in the RFC proposal doc still, I think. 

> In order to provide more context for the enrichment use case, we'd like to prescribe adding the event fieldset (threat.indicator.event.*). @ebeahan should be captured in this proposal, or should this sort of composition only be captured in the aforementioned examples/usage documentation?

Yes, we should capture that intent. I imagine it would be implemented like the other proposed field nestings ([example](https://github.com/elastic/ecs/blob/master/experimental/schemas/hash.yml))?

Can we update the current examples or add another example demonstrating how these fields are intended to be used?

peasead:
> > Some of the discussion above (mostly me, sorry 😅) pertains to the dual usage of this fieldset and the presence/absence/mappings of these fields in each situation. It's now my understanding that, while necessary, those details are orthogonal to the RFC itself and should mainly be captured in a separate document pertaining to examples/usage.
> 
> Good context to capture to some extent in the RFC proposal doc still, I think.
> 
> > In order to provide more context for the enrichment use case, we'd like to prescribe adding the event fieldset (threat.indicator.event.*). @ebeahan should be captured in this proposal, or should this sort of composition only be captured in the aforementioned examples/usage documentation?
> 
> Yes, we should capture that intent. I imagine it would be implemented like the other proposed field nestings ([example](https://github.com/elastic/ecs/blob/master/experimental/schemas/hash.yml))?
> 
> Can we update the current examples or add another example demonstrating how these fields are intended to be used?

@rylnd can I help with anything here?

rylnd:
@peasead @ebeahan as discussed, I added the `event` fieldset in fb057b0. However, we expect this nested fieldset to only be necessary in the "enrichment" use case, as one could leverage the existing root `event` fields for the indicator document/event itself.

This brings us back to the point about documentation of use cases: with the exception of the above, I believe that the existing [use case documentation](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#usage), introduced in #1127, sufficiently describe the situation. If these RFCs are meant to be the source for that type of usage data, then I think we're good here!

rylnd:
> I wanted to understand better which direction to take and make sure that's captured correctly. Right now, the field definitions don't reflect that these fields are either a list or a nested object

@ebeahan to this point, I think the issue that I'm struggling with is that the mappings could be different depending on the usage. I tried to break this down in [a previous comment](https://github.com/elastic/ecs/pull/1293#issuecomment-799783474) but I haven't seen any responses.

devonakerr:
Ryland, I think both use-cases defined are applicable - what other use-cases do you anticipate?

ebeahan:
I think the two use cases described in [Usage section](https://github.com/elastic/ecs/blob/create-threat-stage-2/rfcs/text/0008-threat-intel.md#usage) are doing a good job of describing the dual usage. 

> @ebeahan to this point, I think the issue that I'm struggling with is that the mappings could be different depending on the usage. I tried to break this down in a previous comment but I haven't seen any responses.

@rylnd With my [earlier comment](https://github.com/elastic/ecs/pull/1293#issuecomment-804336773), my intent was to point out the discrepancy between the example and how the proposed fields are defined (`indicator.matched` as a leaf field vs. an object), and I don't believe this is a difference between the enrichment vs. indicator use cases.

rylnd:
> I don't believe this is a difference between the enrichment vs. indicator use cases.

👍 agreed; in both cases `indicator.matched` would be an object and not a field. Brief aside: I don't think there's currently a purpose for `indicator.matched` in the indicator use case.

rylnd:
We had a brief synchronous discussion about the above comments and came to the following conclusions:

1. Recommended mappings are different for the two current use cases: nested fields makes sense for the enrichment use case, but for the indicators themselves this is unnecessary and would in fact have a negative impact on ingestion and storage.
2. Fields that are only relevant to a particular use case (`indicator.matched.*`) implies that we may be trying to "overfit" the schema to match two conflicting use cases; the aforementioned mapping conflict also reinforces this.
3. The most straightforward solution to both of these problems is to define _two different fields_ to cover these use cases:
    1. the existing `indicator.*` fields will continue as proposed, and represent the schema for the indicator use case.
    2. A new, TBD field (e.g. `threat.indicator_matches`?) will cover the enrichment use case, and will be defined roughly as  "an list of nested indicator objects, each of which contains the indicator fieldset and additional matched.* fields to represent the context of the match".
    
If there are no disagreements, I'm going to remove the references to the enrichment use case from this RFC, and move that to a new RFC containing the proposal for 3b above.

I think an argument could be made for either leaving `indicator.matched.*` in this proposal (albeit [unused](https://github.com/elastic/ecs/pull/1293#issuecomment-824968073) for now), or removing them here and constraining them to the match/enrichment RFC only.

Please share any opinions or recommendations on the above; in the absence of feedback I'll proceed with these changes tomorrow.

peasead:
> We had a brief synchronous discussion about the above comments and came to the following conclusions:
> 
> 1. Recommended mappings are different for the two current use cases: nested fields makes sense for the enrichment use case, but for the indicators themselves this is unnecessary and would in fact have a negative impact on ingestion and storage.
> 2. Fields that are only relevant to a particular use case (`indicator.matched.*`) implies that we may be trying to "overfit" the schema to match two conflicting use cases; the aforementioned mapping conflict also reinforces this.
> 3. The most straightforward solution to both of these problems is to define _two different fields_ to cover these use cases:
>    
>    1. the existing `indicator.*` fields will continue as proposed, and represent the schema for the indicator use case.
>    2. A new, TBD field (e.g. `threat.indicator_matches`?) will cover the enrichment use case, and will be defined roughly as  "an list of nested indicator objects, each of which contains the indicator fieldset and additional matched.* fields to represent the context of the match".
> 
> If there are no disagreements, I'm going to remove the references to the enrichment use case from this RFC, and move that to a new RFC containing the proposal for 3b above.
> 
> I think an argument could be made for either leaving `indicator.matched.*` in this proposal (albeit [unused](https://github.com/elastic/ecs/pull/1293#issuecomment-824968073) for now), or removing them here and constraining them to the match/enrichment RFC only.
> 
> Please share any opinions or recommendations on the above; in the absence of feedback I'll proceed with these changes tomorrow.

Thanks @rylnd 

I agree.

rylnd:
@peasead @ebeahan I removed the documentation about the enrichment use case from this branch, and [opened a stage 0](https://github.com/elastic/ecs/pull/1386) for the enrichment use case as discussed above. IMO this should be ready for a final pass, but I defer to the experts.

ebeahan:
Thanks, @rylnd!

@peasead based on the new direction, which `threat.indicator.*` fields should we keep and which should we remove from the proposal here? We can also now remove all the field reuses in favor of using the top-level fields instead, right?

rfcs/text/0008-threat-intel.md
```diff
@@ -28,21 +28,18 @@ Field | Type | Example | Description
...
 threat.indicator.sightings | long | 20 | Number of times this indicator was observed conducting threat activity
-threat.indicator.type | keyword | ipv4-addr, domain-name, email-addr | Type of indicator as represented by Cyber Observable in STIX 2.0
-threat.indicator.description | wildcard | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
-threat.indicator.dataset | keyword | theatintel | Identifies the name of specific dataset from the intelligence source.
-threat.indicator.module | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific module where the data is coming from.
+threat.indicator.type | keyword | ipv4-addr | Type of indicator as represented by Cyber Observable in STIX 2.0
+threat.indicator.description | keyword | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
+threat.indicator.dataset | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific dataset from the intelligence source.
```
rylnd, line 33: @peasead is this one (and some others, I think) redundant with e.g. `event.dataset`, or does this serve a separate purpose?

peasead: Removed this. Thanks.

rfcs/text/0008-threat-intel.md
```diff
@@ -28,29 +28,26 @@ Field | Type | Example | Description
...
-threat.indicator.matched.field | keyword | threat.indicator.ip | Identifies the field of the atomic indicator that matched a local environment endpoint or network event.
-threat.indicator.matched.type | keyword | ipv4-addr, domain-name, email-addr, url | Identifies the type of the atomic indicator that matched a local environment endpoint or network event.
 
 ### Proposed New Values for Event Fieldset
 
 Field | New Value | Description
 --- | --- | ---
-event.kind | enrichment | Propose adding this value to capture the type of information this event contains and how it should be used. Threat intelligence will be used to enrich source events and signals. Enrichment could also appy to other types of contextual data sources (not just threat intelligence) such as directory services, IPAM data, asset lists.
+event.kind | enrichment | Propose adding this value to capture the type of information this event contains and how it should be used. Threat intelligence will be used to enrich source events and signals. Enrichment could also apply to other types of contextual data sources (not just threat intelligence) such as directory services, IPAM data, asset lists.
```
ebeahan, line 48: Are these new proposed values for `event.*` fields still in the scope of this proposal?

rfcs/text/0008-threat-intel.md
```diff
@@ -378,7 +204,7 @@ Stage 2: Identifies scope of impact of changes. Are breaking changes required? S
  * ECS project (e.g. docs, tooling)
 The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
 -->
- * Ingestion mechanism: Primary ingestion mechanisms will be Filebeat modules and Ingest Packages. There will be no impact on ingestion mechanisms.
+ * Ingestion mechanism: Primary ingestion mechanisms will be Filebeat modules and Ingest Packages. There will be no impact on ingestion mechanisms. [Filebeat module](https://www.elastic.co/guide/en/beats/filebeat/7.12/exported-fields-threatintel.html) is scheduled to be released in `7.12`.
```
ebeahan, line 207:
```suggestion
 * Ingestion mechanism: Primary ingestion mechanisms will be Filebeat modules and Ingest Packages. There will be no impact on ingestion mechanisms. [Filebeat module](https://www.elastic.co/guide/en/beats/filebeat/7.12/exported-fields-threatintel.html) was released in `7.12`.
```

rfcs/text/0008-threat-intel.md
```diff
@@ -89,9 +86,9 @@ Stage 1: Describe at a high-level how these field changes will be used in practi
 
 The additions described above will be used to enable cyber threat intelligence capabilities in Elastic Security solution. A new rule type Indicator match will be introduced in 7.10 and the proposed ECS updates will enable a new category of detection alerts that match incoming log and event data against threat intelligence sources. Additionally in the future we will also develop enrichment flows that add context from threat intelligence to alerts and events to assist analysts in their investigative workflows.
```
ebeahan, line 87:
```suggestion
The additions described above will be used to enable cyber threat intelligence capabilities in Elastic Security solution. A new rule type Indicator match was introduced in 7.10 and the proposed ECS updates will enable a new category of detection alerts that match incoming log and event data against threat intelligence sources. Additionally in the future we will also develop enrichment flows that add context from threat intelligence to alerts and events to assist analysts in their investigative workflows.
```

rfcs/text/0008-threat-intel.md
rylnd: @ebeahan while we don't currently have a definite use case for e.g. `event.kind: enrichment` (since indicator match rules generate signals, not enrichments), I think having this new kind makes sense if one wanted an audit trail for enrichment, and similarly I don't think the other fields are still relevant as well 👍

rylnd:
> It's still unclear from what's captured in the document which new fields are being proposed.

@ebeahan @peasead my understanding is that _all_ of the fields in the table are still part of the proposal. If they're omitted from the Usage section then that's an oversight on our end.

The necessity of those additional fields (e.g. `threat.indicator.ip`) is due to the fact that ECS does not provide context-agnostic equivalents for the different values that may be contained in an indicator and so we cannot simply use existing ECS fields for those values. As an example, if an indicator represents a malicious IP address, it wouldn't make sense to populate that value in either `destination.ip` or `source.ip`, and so `threat.indicator.ip` allows us to provide a generic field to hold that category of value (IP address).

rfcs/text/0008-threat-intel.md
```diff
@@ -87,11 +84,11 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+The additions described above will be used to enable cyber threat intelligence capabilities in Elastic Security solution. A new rule type Indicator match was introduced in 7.10 and the proposed ECS updates will enable a new category of detection alerts that match incoming log and event data against threat intelligence sources. Additionally in the future we will also develop enrichment flows that add context from threat intelligence to alerts and events to assist analysts in their investigative workflows.
 
-There are two primary uses for these fields.
+While there are two primary uses for these fields, this RFC deals primarily with the first: ingestion/storage of threat intelligence.
 
-1. **Storing threat intelligence as an event document in threat index(s).**
+**Storing threat intelligence as an event document in threat index(s).**
 
     Threat intelligence data will be collected from multiple sources stored in threat indices. The ECS fields proposed here will be used to structure the documents collected from various sources.
```
rylnd, line 94: @peasead the example below declares `indicator.*` instead of `threat.indicator.*`, and as mentioned by @ebeahan is also missing many of the fields in the above table.

peasead: Yep. I'm working that.

peasead:
> It's still unclear from what's captured in the document which new fields are being proposed.
> 
> The table underneath `Fields` is listing 15 fields:
> 
> * threat.indicator.first_seen
> * threat.indicator.last_seen
> * threat.indicator.sightings
> ...

Thanks for the eagle eye, @ebeahan 

I have updated the examples to match the fields table and `.yml` file.

rfcs/text/0008-threat-intel.md
peasead: yessir. I got that buttoned up.

rfcs/text/0008-threat-intel.md
```diff
@@ -87,170 +85,136 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
     },
+
+    // Tags for context
     "tags": [
         "threatintel",
         "forwarded"
-    ],
+    ]
 }
-```
```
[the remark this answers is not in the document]
peasead: Doh. Sorry about that!

rfcs/text/0008-threat-intel.md
```diff
@@ -87,171 +85,139 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+        "type": [
+            "file"
+        ],
+        "description": "Implant used during an Angler EK campaign.",
+        "dataset": "threatintel.malwarebazaar",
+        "module": "threatintel",
+        "provider": "Abuse.ch",
+        "reference": "https://bazaar.abuse.ch/sample/f3ec9a2f2766c6bcf8c2894a9927c227649249ac146aabfe8d26b259be7d7055",
+        "confidence": "High",
+        "file": {
```
ebeahan, line 189: Apologies, I know I keep circling back to this topic, but I also keep seeing instances of it:

For the indicator use case, would this actually be the top-level `file.*` fields, or nested under `threat.indicator.*` like the example here?

@rylnd @peasead

rylnd: @ebeahan I agree, I think several of these examples should be moved up a level, e.g.:

* `threat.indicator.dataset` -> `event.dataset`
* `threat.indicator.reference` -> `event.reference`
* `threat.indicator.file.*` -> `file.*`

peasead: Thanks, @rylnd and @ebeahan 

After reviewing the [threat enrichment RFC](https://github.com/rylnd/ecs/blob/threat-enrichment-stage-1/rfcs/text/0021-threat-enrichment.md), I understand better. 

I'll make those changes.

rfcs/text/0008-threat-intel.md
```diff
@@ -28,29 +28,27 @@ Field | Type | Example | Description
...
-threat.indicator.description | wildcard | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
-threat.indicator.dataset | keyword | theatintel | Identifies the name of specific dataset from the intelligence source.
-threat.indicator.module | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific module where the data is coming from.
+threat.indicator.type | keyword | ipv4-addr | Type of indicator as represented by Cyber Observable in STIX 2.0
+threat.indicator.description | keyword | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
+threat.indicator.dataset | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific dataset from the intelligence source.
+threat.indicator.module | keyword | threatintel | Identifies the name of specific module where the data is coming from.
 threat.indicator.provider | keyword | Abuse.ch | Identifies the name of intelligence provider.
+threat.indicator.reference | keyword | https://urlhaus.abuse.ch/url/1292596 | Provides a reference for the indicator for additional information.
```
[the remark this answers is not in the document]
peasead: If we're moving all of the `event.` data up, we should move `event.reference` up as well.

Example: `event.reference:https://urlhaus.abuse.ch/url/1321083/`

rfcs/text/0008-threat-intel.md
```diff
@@ -28,29 +28,27 @@ Field | Type | Example | Description
...
 threat.indicator.sightings | long | 20 | Number of times this indicator was observed conducting threat activity
-threat.indicator.type | keyword | ipv4-addr, domain-name, email-addr | Type of indicator as represented by Cyber Observable in STIX 2.0
-threat.indicator.description | wildcard | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
-threat.indicator.dataset | keyword | theatintel | Identifies the name of specific dataset from the intelligence source.
-threat.indicator.module | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific module where the data is coming from.
+threat.indicator.type | keyword | ipv4-addr | Type of indicator as represented by Cyber Observable in STIX 2.0
+threat.indicator.description | keyword | 201.10.10.90 was seen delivering Angler EK | Describes the type of action conducted by the threat
+threat.indicator.dataset | keyword | threatintel.{abusemalware,abuseurl,misp,otx,limo} | Identifies the name of specific dataset from the intelligence source.
```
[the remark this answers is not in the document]
peasead: This should just be `event.module`. As an example, for the Filebeat TI module, it would be `event.module:threatintel`.

Typo and I'll fix it.

rfcs/text/0008-threat-intel.md
```diff
@@ -28,29 +28,27 @@ Field | Type | Example | Description
...
+threat.indicator.module | keyword | threatintel | Identifies the name of specific module where the data is coming from.
```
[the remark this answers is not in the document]
peasead: Yes, this should just be `event.dataset:threatintel.{abusemalware,abuseurl,misp,otx,limo}`.

I'll get that fixed.

rfcs/text/0008-threat-intel.md
peasead: @rylnd and @ebeahan 

Shouldn't we move `threat.indicator.domain` to the existing ECS URL fieldset (`url.domain`) as well?

peasead: Going with `.domain` being under the existing URL fieldset (`url.domain.`).

ebeahan:
Thanks, @peasead, for making the latest rounds of changes!

@rylnd @peasead there was some out-of-band discussion last week. Were there any outcomes or changes that should be reflected here? Or is the PR ready for another review?

peasead:
> Thanks, @peasead, for making the latest rounds of changes!
> 
> @rylnd @peasead there was some out-of-band discussion last week. Were there any outcomes or changes that should be reflected here? Or is the PR ready for another review?

I feel like we're ready, but I'll wait for @rylnd and @devonakerr to voice any dissent.

devonakerr:
No dissent, I'll start my review.

rylnd (approved):
All my comments have been addressed. I defer to @devonakerr for official approval, but LGTM!

rfcs/text/0008-threat-intel.md
```diff
@@ -87,171 +83,139 @@ Stage 3: Add or update all remaining field definitions. The list should now be e
...
+        "type": "indicator",
+        "reference": "https://urlhaus.abuse.ch",
+        "severity": 7,
+        "risk_score": 10,
+        "original": "2020-10-29 19:16:38"
+    },
+
+    // Metadata about the indicator data
+    "threat.indicator": {
+        "first_seen": "2020-10-01",
```
rylnd, line 113: @peasead it's probably outside the scope of this RFC, but I wanted to mention that the need for an `updated_at` field has been requested in the context of indicator indexes, in order to be able to query the most relevant indicators by sorting on said field. I think that this goes against our current idea of indicators as timeseries data, but the underlying problem can be expressed as three events, in chronological order:

1. Domain indicator A is active
2. Domain indicator B is active
3. Domain indicator A is inactive

If the analyst is only concerned with active domains, then a naive query may include both active and inactive versions of indicator A if scoping simply on `@timestamp` or `event.ingested`.

peasead: I think we should add it now.

I'll do that.

rfcs/text/0008/threat.yml
```diff
@@ -21,6 +21,15 @@
 
     example: "2020-11-05T17:25:47.000Z"
 
+  - name: indicator.updated_at
```
rylnd, line 24: `updated_at` or `modified_at`?

peasead: Seemingly...both.

Fix't

devonakerr (approved):
LGTM

rylnd:
@ebeahan: @peasead is out this week, but I think this is good to merge! I'm happy to do any of the leg work here, just let me know.

ebeahan (approved):
LGTM!

I'll set the date and merge 🎉 🚢

PR #1302, opened 2021-03-16
Add flattened type support to Go code generator

ebeahan:
Separating implementation of `flattened` type into the go code generator from #1254. I believe `flattened` should be a more straight-forward change.

Opted to use `string` since `flattened` fields do not support defining sub-fields like other object types, such as `object` or `nested`.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -30,6 +30,8 @@ Thanks, you're awesome :-) -->
 
 #### Improvements
 
+* Go code generator now supports the `flattened` data type. #1302
+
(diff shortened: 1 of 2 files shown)


andrewkroh:
> Opted to use string since flattened fields do not support defining sub-fields like other object types

Flattened is an object type with different indexing behavior. So I would add it to the `case: "object"`.

ebeahan:
> So I would add it to the case: "object".

Thanks, done.

andrewkroh (approved):

PR #1303, opened 2021-03-16
Generate ECS detection rules

leehinman:
- generates required rules
 - generates allowed_values rules
 - generates expected_values rules for event.category

## Example Required rule

```
{
  "description": "This rule checks to make sure the required field `@timestamp` is present.",
  "from": "now-1d",
  "interval": "1d",
  "name": "ECS Check: @timestamp required field",
  "query": "not @timestamp:*",
  "risk_score": 42,
  "severity": "high",
  "type": "query"
}
```

## Example allowed_values rule

```
{
  "description": "This rule checks to make sure the `event.outcome` field only contains allowed values.",
  "from": "now-1d",
  "interval": "1d",
  "name": "ECS Check: event.outcome allowed values",
  "query": "event.outcome:* and not event.outcome:failure and not event.outcome:success and not event.outcome:unknown",
  "risk_score": 42,
  "severity": "high",
  "type": "query"
}
```

## Example expected_values rule

```
{
  "description": "This rule checks to make sure that `event.type` contains expected values when `event.category` is web.",
  "from": "now-1d",
  "interval": "1d",
  "name": "ECS Check: expected `event.type` for `event.category` web",
  "query": "event.category:web and event.type:* and not event.type:access and not event.type:error and not event.type:info",
  "risk_score": 42,
  "severity": "high",
  "type": "query"
}
```

## Added Gherkin feature files.

With a feature file that looks like this:
``` gherkin
Feature: Address field should be populated if ip or domain is polulated.
  Scenario: If the source.ip field exists then so should the source.address field.
    Given an event
    And the source.ip field is present
    Then the source.address field is present
```

We get documentation that looks like:
Screen Shot 2021-04-30 at 12 35 36

And a detection rule that looks like:
``` json
{
  "description": "If the source.ip field exists then so should the source.address field.",
  "from": "now-1d",
  "interval": "1d",
  "name": "ECS Check: If the source.ip field exists then so should the source.address field.",
  "query": "source.ip:* and not source.address:*",
  "risk_score": 42,
  "severity": "high",
  "type": "query"
}
```

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -10,6 +10,7 @@ from generators import csv_generator
 from generators import es_template
 from generators import ecs_helpers
 from generators import intermediate_files
+from generators import rules_generator
 
 from schema import loader
 from schema import oss
@@ -59,6 +60,7 @@ def main():
     es_template.generate(nested, ecs_version, out_dir, args.mapping_settings)
     es_template.generate_legacy(flat, ecs_version, out_dir, args.template_settings, args.mapping_settings)
     beats.generate(nested, ecs_version, out_dir)
+    rules_generator.generate(flat, ecs_version, out_dir)
     if args.include or args.subset:
         exit()
 
--- a/scripts/generators/rules_generator.py
+++ b/scripts/generators/rules_generator.py
@@ -2,7 +2,9 @@ import json
 import os.path
 import sys
 
-from generator import ecs_helpers
+
+from generators import ecs_helpers
+from generators import bp_generator
 
 
 def generate(ecs_flat, version, out_dir):
@@ -12,7 +14,7 @@ def generate(ecs_flat, version, out_dir):
     rules.extend(gen_category_expected_types_rules(ecs_flat))
     if len(rules) > 0:
         ecs_helpers.make_dirs(os.path.join(out_dir, 'rules'))
-        save_rules(os.path.join(out_dir, 'rules/rules.json'), rules)
+        bp_generator.save_ndjson(os.path.join(out_dir, 'rules/ecs-rules.ndjson'), rules)
 
 
 def gen_required_rules(ecs_flat):
@@ -21,15 +23,10 @@ def gen_required_rules(ecs_flat):
     for field_name, field_properties in required_fields.items():
         if field_properties['required'] == False:
             next
-        rule = {"name": "ECS Check: {0} required field".format(field_name),
-                "description": "This rule checks to make sure the required field `{0}` is present.".format(field_name),
-                "risk_score": 42,
-                "severity": "high",
-                "type": "query",
-                "interval": "1d",
-                "from": "now-1d",
-                "query": "not {0}:*".format(field_name),
-                }
+        rule = bp_generator.default_detection_rule()
+        rule['name'] = "ECS Check: {0} required field".format(field_name)
+        rule['description'] = "This rule checks to make sure the required field `{0}` is present.".format(field_name)
+        rule['query'] = "not {0}:*".format(field_name)
         rules.append(rule)
     return rules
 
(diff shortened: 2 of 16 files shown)


scripts/generators/rules_generator.py
```diff
@@ -0,0 +1,81 @@
...
+        save_rules(os.path.join(out_dir, 'rules/rules.json'), rules)
+
+
+def gen_required_rules(ecs_flat):
+    rules = []
+    required_fields = dict(filter(lambda elem: 'required' in elem[1], ecs_flat.items()))
+    for field_name, field_properties in required_fields.items():
+        if field_properties['required'] == False:
+            next
+        rule = {"name": "ECS Check: {0} required field".format(field_name),
```
ebeahan, line 24: A possible change for later:

Similar how some of the generated asciidoc files use [templates](https://github.com/elastic/ecs/tree/master/scripts/templates), we could consider creating rule templates to have some of the rule configuration outside the rule generator here.

scripts/generator.py
```diff
@@ -59,6 +60,7 @@ def main():
     es_template.generate(nested, ecs_version, out_dir, args.mapping_settings)
     es_template.generate_legacy(flat, ecs_version, out_dir, args.template_settings, args.mapping_settings)
     beats.generate(nested, ecs_version, out_dir)
+    rules_generator.generate(flat, ecs_version, out_dir)
```
ebeahan, line 63: Will we want to include the also add the generated rules file? Not seeing the `generated/rules/` directory included right now.

scripts/generators/rules_generator.py
```diff
@@ -0,0 +1,81 @@
...
+                "type": "query",
+                "interval": "1d",
+                "from": "now-1d",
+                "query": "event.category:{0} and event.type:* and {1}".format(allowed_value['name'], " and ".join(event_types))
+                }
+        rules.append(rule)
+    return rules
+
+
+def save_rules(filename, rule_list):
```
ebeahan, line 75: Looks like there's also a similar function in the `es_template` generator. Might be a good candidate to move into `ecs_helpers`.

ebeahan (commented):
Thanks, @leehinman! A few observations here and below.


* I think the types and content of the rules looks great.
* I noticed `rules.json` is a collection of JSON objects, but it isn't ndjson. Were you thinking the generated rules file would be something that could be uploaded as detection rules directly, or for another use?
* If we do add `generated/rules` as a maintained generated artifact, can we add an entry in the generated artifacts [README](https://github.com/elastic/ecs/blob/master/generated/README.md)?

leehinman:
> Thanks, @leehinman! A few observations here and below.
> 
>     * I think the types and content of the rules looks great.
> 
>     * I noticed `rules.json` is a collection of JSON objects, but it isn't ndjson. Were you thinking the generated rules file would be something that could be uploaded as detection rules directly, or for another use?

Thanks, fixed that.

> 
>     * If we do add `generated/rules` as a maintained generated artifact, can we add an entry in the generated artifacts [README](https://github.com/elastic/ecs/blob/master/generated/README.md)?

Thanks, added to generated/README.md

scripts/generators/rules_generator.py
leehinman: I ended up not using ecs_helpers.py but a new file because using ecs_helper.py with behave was giving me import errors because import paths were different.  Probably worth fixing in a follow on PR.

ebeahan:
@leehinman are you ready for another review, or is this still something you're iterating on?

leehinman:
> @leehinman are you ready for another review, or is this still something you're iterating on?

ready for another review

djptek:
@leehinman there's a few conflicts have emerged & `elasticsearch-ci/docs` may need a re-run

ebeahan (commented):
Sorry for not reviewing again sooner, @leehinman.

I like the use of Gherkin to specify the best practices as feature files. I think the plain-text spec works well to describe the outcomes we look for in ECS.

I also like the idea of having a true ECS `Best Practices` section in the docs, but I'm unsure if a collection of bullet points is the best way to communicate those concepts to users. However, maybe the bullet points can be high-level, brief descriptions and more detailed examples could follow later.

It took me a couple of passes to wrap my head around where the different ECS rule sets were coming from in the generator, but I think with some documentation that could help avoid user confusion.

If this is something we still want to try and introduce into the ECS tooling, we should also add some supporting unit testing.

ebeahan:
@leehinman Is this PR still something you want to move forward on?

ebeahan:
@leehinman I'm going to close this PR out for now. Feel free to re-open in the future.

PR #1304, opened 2021-03-17
Fix rendering of restricted data_stream fields

felixbarny:
Also adding `,` and `#`, based on https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params


ebeahan:
Suggested we fold this fix into the other changes in the `data_stream` fields stage 3 PR: https://github.com/elastic/ecs/pull/1212#issuecomment-801239826

ruflin:
I cherry picked this change into https://github.com/elastic/ecs/pull/1212 Hope that is ok for you @felixbarny

felixbarny:
Sure that works, thanks!

PR #1309, opened 2021-03-18
[RFC] Version Type Field Migration - Stage 0

ebeahan:
### Summary

Stage 0 proposal for adopting the `version` type for fields containing software version values.

[`version`](https://www.elastic.co/guide/en/elasticsearch/reference/current/version.html) is specialized to support precedence rules based on [Semantic Versioning](semver.org).

Stage 0 (Strawperson) Criteria:

- [x] Discuss with domain or subject matter experts the utility of these changes
- [x] Discuss with ECS team whether these changes seem appropriate for ECS

[Preview of markdown proposal](https://github.com/ebeahan/ecs/blob/rfc/0000/adopt-version-field-type/rfcs/text/0000-version-type-field-migration.md)

### DIFF
--- a/rfcs/text/0000-version-type-field-migration.md
+++ b/rfcs/text/0000-version-type-field-migration.md
@@ -107,6 +107,10 @@ e.g.:
 
 <!-- Insert any links appropriate to this RFC in this section. -->
 
+* https://github.com/elastic/ecs/issues/887
+* https://github.com/elastic/ecs/issues/842
+* https://www.elastic.co/guide/en/elasticsearch/reference/current/version.html
+
 ### RFC Pull Requests
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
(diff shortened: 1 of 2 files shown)


ebeahan:
Discussed in another channel that engaging the owners of the [proposed version fields](https://github.com/ebeahan/ecs/blob/rfc/0000/adopt-version-field-type/rfcs/text/0000-version-type-field-migration.md#fields) and determining which of these fields are candidates for `version` are good next steps. After some thought, I think that it would be best to hold off on those broader, cross-team discussions until the stage 1 PR.

Stage 0 acceptance is an agreement that the proposed idea has merit for consideration into ECS, so I think we focus on those criteria in this PR.

kgeller (approved):
LGTM! 

I agree with your sentiment about having the broader discussion in phase 1.

PR #1298, opened 2021-03-10
RFC Remove log.original Stage 0

djptek:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
YES

- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)?
YES

- Is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed.
YES

PR #1314, opened 2021-03-24
[RFC] 0017 Remove log.original stage 1

djptek:
- If submitting code/script changes, have you verified all tests pass locally using `make test`? n/a stage 1
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? n/a stage 1
- Have you added an entry to the [CHANGELOG.next.md] (https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? n/a stage 1


kgeller (approved):

rfcs/text/0017-remove-log-original.md
```diff
@@ -16,16 +16,18 @@ The removal of `log.original` will be considered a breaking change since the fie
 
 | Field Set | Field(s) |
 | --------- | -------- |
-| [`log`](0000/log.yml) | `log.original` |
-| [`event`](0000/event.yml) | `event.original` |
+| [`log`](0017/log.yml) | `log.original` |
```
ebeahan, line 19: Including field definition files when adding or changing fields has proven useful in a couple of ways:

* The fields can be used either with the `--include` flag to test out the field changes
* Sometimes, the proposed field definitions get a little long, especially for brand new fieldsets. Separating the field definitions from the markdown keeps everything a bit tidier.

The way that custom files are merged with the current ECS fields, even if you remove a current ECS field it'll still be present in all the outputs. So it's not a problem to include these field definitions here, but unfortunately, we can't use them with the `--include` flag and get the expected result.

I do think we could consider having a consistent way to mark a field as `deprecated` in the ECS docs until we're ready to entirely remove the field. Looking at the Elastic docs repo, there is an asciidoc admonition available for `deprecated` that we might be able to leverage somehow: https://github.com/elastic/docs#old-section.

We have something similar for the `beta` attribute: https://github.com/elastic/ecs/blob/master/scripts/templates/field_details.j2#L13-L18

rfcs/text/0017-remove-log-original.md
```diff
@@ -16,16 +16,18 @@ The removal of `log.original` will be considered a breaking change since the fie
...
 
-<!--
-Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting.
--->
+- The internal description of the field `log.original` in [`log`](0017/log.yml) should be amended by addition of a notice of deprecation and subsequently removal if/when Deprecation progresses to Removal
 
-<!--
-Memo: see https://github.com/elastic/ecs/pull/1298#discussion_r591870463
--->
+- The internal description of the field `event.original` in [`event`](0017/event.yml) should be updated to reflect the revised scope
```
ebeahan, line 24: 👍

rfcs/text/0017-remove-log-original.md
```diff
@@ -16,16 +16,18 @@ The removal of `log.original` will be considered a breaking change since the fie
...
-<!--
-Memo: see https://github.com/elastic/ecs/pull/1298#discussion_r591870463
--->
+- The internal description of the field `event.original` in [`event`](0017/event.yml) should be updated to reflect the revised scope 
+
+- The [`Beats default fields inclusion list`](../../scripts/generators/beats_default_fields_whitelist.yml) list should be updated by removing `log.original` if/when Deprecation progresses to Removal
+
+- The extended description of `log.original` in the [`Log Fields documentation`](../../docs/field-details.asciidoc#field-log-original) should be amended by addition of a notice of deprecation and subsequently removal if/when Deprecation progresses to Removal
+
+- The extended description of `event.original` in the [`Event Fields documentation`](../../docs/field-details.asciidoc#field-event-original) should be amended by addition of a notice of deprecation and subsequently removal if/when Deprecation progresses to Removal
```
ebeahan, line 30: Would we have anything that we'd need to deprecate for `event.original` if we remove `log.original`?

ebeahan (commented):
Thanks for iterating on this, @djptek!

I think overall it's shaping up well. Couple of "virtual paperwork" items 😸 :

* Would you add the URL for this PR to the section towards the bottom for stage 1?
* Also would you add my handle, @ebeahan, as the sponsor under the `People` section?

rfcs/text/0017-remove-log-original.md
djptek: Thanks to @jamiehynds for spotting https://github.com/elastic/ecs/pull/1298#discussion_r604750476

djptek: `Google Cloud integration includes a var.keep_original_message (disabled by default) which populates log.original. Will need to adjust the module to populate event.original instead.`

ebeahan: Ok, good catch @jamiehynds @djptek.

I don't see a need to edit the ECS description for `event.original` if we move forward with deprecating and then removing `log.original` at the next major. When we deprecate `log.original`, we can add a note `log.original's` description suggesting to migrate to `event.original`.

Thoughts?

djptek: Agree, that was wrong, I've changed to:

- The extended description of `event.original` in the [`Event Fields documentation`](../../docs/field-details.asciidoc#field-event-original) should be amended to clarify the absorption of `log.original`

ebeahan:
Thanks, @djptek, for the updates! Everything is looking great here.

I'm going to list out the criteria for advancing to stage one. The one remaining item is capturing some of the potential implementation challenges and concerns in the proposal doc.

- [ ] Identified potential concerns and implementation challenges/complexity
    * Let's note under `Scope of impact` that Beats modules and Agent integration packages would be required to migrate if this change is adopted as proposed. No need to find or list every instance, but we should call out that those ingestion sources will be affected if this change takes place.
    * I think lines 10-13 noting the breaking changes could go under the `Scope of impact` section as well.

djptek:
Thanks for the comments @ebeahan I've updated as discussed during the call today + applied the changes requested here in `Identified potential concerns and implementation challenges/complexity`

rfcs/text/0017-remove-log-original.md
```diff
@@ -1,31 +1,28 @@
...
-Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting.
--->
+- The internal description of the field `log.original` in [`log`](0017/log.yml) should be amended by addition of a notice of deprecation and subsequently removal if/when Deprecation progresses to Removal
 
-<!--
-Memo: see https://github.com/elastic/ecs/pull/1298#discussion_r591870463
--->
+- The internal description of the field `event.original` in [`event`](0017/event.yml) should be updated to reflect the revised scope 
+
+- The [`Beats default fields inclusion list`](../../scripts/generators/beats_default_fields_whitelist.yml) list should be updated by removing `log.original` if/when Deprecation progresses to Removal
```
ebeahan, line 21:
```suggestion
- The [`Beats default fields inclusion list`](../../scripts/generators/beats_default_fields_allowlist.yml) list should be updated by removing `log.original` if/when Deprecation progresses to Removal
```

Recalled we recently updated that file name.

ebeahan (approved):
Other than the one nit item, LGTM 👍 !

I'll let you handle merging, @djptek. Before you do, please add a commit setting the `Date` at the top of the proposal doc to the date you're merging on. When you push the commit, you'll dismiss the approval ✅ , but feel free to merge without additional approval.

Typically once an RFC has advanced as stage one, another PR is opened to add new fields or changes to the experimental schema. Since the only proposed change here is updating the `log.original` field's description, I don't think we worry about adding the experimental schema changes. 

I'm also open to discussing more if we think there might be value to including those changes, though.

PR #1316, opened 2021-03-25
Move host metric fields RFC to finished

kaiyan-sheng:
This PR is to move adding host metric fields RFC from candidate to finished.


ebeahan (dismissed):
LGTM 👍 

Thanks!

PR #1323, opened 2021-03-26
lowercasing component names for es template

kgeller:
Correcting fieldset name capitalization when generating the generated es template


djptek (approved):
Looks good

ebeahan (commented):
Thanks for looking into this one, @kgeller!

A couple of items I wanted to note.

#### Testing

We don't have much testing coverage of the component and composable template portions of the Elasticsearch generator, but I think we should add test coverage for this case as we're modifying it a bit; at least some testing coverage for the uppercasing-to-lowercasing handling.

The existing tests are in `scripts/tests/test_es_template.py` for the ES template generator. The other test cases should be helpful examples, and I think since the casing change is pretty targeted, the nesting data structure can be pretty small. Something like:

```
        ecs_nested = {
            "Acme": {
                "name": "Acme",
            }
        }
```

#### Example shell script

As I was reviewing this change, I realized this [example shell script](https://github.com/elastic/ecs/tree/master/generated/elasticsearch#composable-and-component-index-templates) for the composable/component index templates would also use the uppercase filename (e.g. `Acme.json`).

Maybe we can also add lowercasing of the file names in that example shellcode snippet with a command such as `tr` or `awk`?

PR #1336, opened 2021-03-31
adding process info on rfc folder

kgeller:
Adding in mention of the RFC folder to the guidance on RFCs

### DIFF
--- a/rfcs/0000-rfc-template.md
+++ b/rfcs/0000-rfc-template.md
@@ -13,6 +13,10 @@ Feel free to remove these comments as you go along.
 Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
 -->
 
+<!--
+Stage 1: If the changes include field additions or modifications, please create a folder titled as the RFC number under rfcs/text/. This will be where proposed schema changes as standalone YAML files or extended example mappings and larger source documents will go as the RFC is iterated upon.
+-->
+
 ## Fields
 
 <!--
@@ -36,7 +40,7 @@ Stage 1: Provide a high-level description of example sources of data. This does 
 -->
 
 <!--
-Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
+Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the correspdonding RFC folder.
 -->
 
 <!--
--- a/rfcs/0000-rfc-template.md
+++ b/rfcs/0000-rfc-template.md
@@ -40,7 +40,7 @@ Stage 1: Provide a high-level description of example sources of data. This does 
 -->
 
 <!--
-Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the correspdonding RFC folder.
+Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
 -->
 
 <!--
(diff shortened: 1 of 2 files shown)


rfcs/PROCESS.md
```diff
@@ -4,6 +4,8 @@ Changes to ECS are proposed as Requests for Comments (RFC) in [rfcs/](./rfcs/) a
 
 Each RFC is represented as a markdown document following a prescribed template that gets committed to the repo. Each stage of the RFC is represented as a pull request against that document.
 
+Every RFC should also have a corresponding folder (named after the RFC number) in [rfcs/text/](./text/). The folder should contain the proposed schema changes as standalone YAML files, or extended example mappings and larger source documents.
```
ebeahan, line 7:
```suggestion
If proposing new fields or changing existing fields, the RFC should also have a corresponding folder (named after the RFC number) in [rfcs/text/](./text/). The folder should contain the proposed schema changes as standalone YAML files or extended example mappings and larger source documents.
```

I think most RFCs will be adds/changes, and having the directory will be necessary. There may also be cases, like removing fields, where it's probably not needed, though. 

Also, a minor grammar nitpick.

rfcs/0000-rfc-template.md
```diff
@@ -36,7 +36,7 @@ Stage 1: Provide a high-level description of example sources of data. This does
 -->
 
 <!--
-Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting.
+Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the correspdonding RFC folder.
```
ebeahan, line 39:
```suggestion
Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
```

rfcs/0000-rfc-template.md
```diff
@@ -16,11 +16,11 @@ Stage 0: Provide a high level summary of the premise of these changes. Briefly d
 ## Fields
 
 <!--
-Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting.
+Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
```
ebeahan, line 19: It'll be redundant, but from experience sometimes folks don't always read `rfcs/PROCESS.md` before jumping into the template. So maybe before the first mention of the folder, we reiterate that RFCs making field changes/adds should create an accompanying folder? Maybe even as a separate comment towards the top of the template?

ebeahan:
@kgeller I removed the `needs_backport` label since we don't worry about backporting the RFCs proposals or the RFC process docs. The activity can all happen on `master`.

This is one advantage of splitting the RFC proposal PRs from their implementation PRs. The proposals live on `master`, and we can backport any schema changes separately.

rfcs/0000-rfc-template.md
```diff
@@ -36,7 +40,7 @@ Stage 1: Provide a high-level description of example sources of data. This does
 -->
 
 <!--
-Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
+Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the correspdonding RFC folder.
```
ebeahan, line 43:
```suggestion
Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
```

Looks like this was fixed but then slipped back in.

ebeahan (dismissed):
Minus the one final item looks good! 🚢

rfcs/0000-rfc-template.md
kgeller: oops!

ebeahan (approved):

PR #1346, opened 2021-04-07
[RFC] Create Mach-O - Stage 1 Proposal

peasead:
[RFC Preview](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/0000-create-file-macho.md)

Resolves https://github.com/elastic/ecs/issues/1096

- If submitting code/script changes, have you verified all tests pass locally using `make test`? NA at this stage
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?  NA at this stage
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)?  NA at this stage


peasead:
Thanks for the update @dcode

ebeahan:
Since we're losing some context from the original PR #1097, was there ever a final resolution for this discussion? https://github.com/elastic/ecs/pull/1097#discussion_r588759432

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,152 @@
...
+| macho.cpu.type                 | keyword | CPU type for the file.                                          |
+| macho.headers                  | nested  | Header information for the file.                                |
+| macho.headers.commands.number  | long    | Number of load commands for the Mach-O header.                  |
+| macho.headers.commands.size    | long    | Size of load commands of the Mach-O header.                     |
+| macho.headers.commands.type    | keyword | Type of the load commands for the Mach-O header.                |
+| macho.headers.magic            | keyword | Magic field of the Mach-O header.                               |
+| macho.headers.flags            | keyword | Flags set in the Mach-O header.                                 |
+| macho.segments                 | nested  | Segment information for the file.                               |
+| macho.segments.name            | keyword | Name of this segment.                                           |
+| macho.segments.physical_offset | long    | File offset of this segment.                                    |
```
ebeahan, line 31: The `macho.segments` fields in the sample field definitions are using:

* `macho.segments.fileoff`
* `macho.segments.filesize`
* `macho.segments.flags`
* `macho.segments.name`
* `macho.segements.offset`
* `macho.segments.sections`
* `macho.segements.size`
* `macho.segments.vmaddr`
* `macho.segments.vmsize`

Looks like there's some alignment, but some fields are mismatched. I prefer the more verbose naming here in the table.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,152 @@
...
+| macho.segments.physical_size   | keyword | Amount of memory to map from the file.                          |
+| macho.segments.virtual_address | keyword | Memory address of this segment.                                 |
+| macho.segments.virtual_size    | keyword | Memory size of this segment.                                    |
+| macho.segments.sections        | keyword | Section names contained in this segment.                        |
+| macho.sections                 | nested  | Section information for the segment of the file.                |
```
ebeahan, line 36: Should this be `macho.segments.sections`? And likewise for the rest of the `macho.sections.*` fields?

dcode: No, while there is definitely a relationship between segments and sections, they are actually two different views into the binary. Keeping sections and segments at the top-level of the binary structure allows for a more consistent analysis experience across PE, ELF, and MachO

rfcs/text/0000-create-file-macho.md
dcode: Agree, the more verbose naming is better. We'll get that cleaned up.

devonakerr (approved):
LGTM - sorry for the delayed approval.

ebeahan:
Thanks, @devonakerr, for reviewing.

The one item I still have is the [fields list](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/0000-create-file-macho.md#fields) doesn't match the [definitions](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/macho/macho.yml). Once that's addressed, I believe we'll be in good shape to advance this one.

cc @peasead @dcode

peasead:
> Thanks, @devonakerr, for reviewing.
> 
> The one item I still have is the [fields list](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/0000-create-file-macho.md#fields) doesn't match the [definitions](https://github.com/peasead/ecs/blob/file.macho-create/rfcs/text/macho/macho.yml). Once that's addressed, I believe we'll be in good shape to advance this one.
> 
> cc @peasead @dcode

@ebeahan I think we're cleaned up now.

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,152 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 1: https://github.com/elastic/ecs/pull/1097
```
ebeahan, line 147:
```suggestion
* Stage 1: https://github.com/elastic/ecs/pull/1346
```

rfcs/text/0000-create-file-macho.md
```diff
@@ -0,0 +1,152 @@
...
+- Stage: **1 (draft)**
+- Date: **TBD**
+
+Create the Mach Object (Mach-O) sub-field, of the `file` or `process` top-level fieldsets. This document metadata can be used for malware research, as well as coding and other application development efforts.
+
+## Fields
+
+**Stage 0**
+
+This RFC is to create the Mach-O sub-field within the `file.` fieldset. This will include 35 sub-fields. `macho` itself is a nested
```
ebeahan, line 12: I'd prefer to leave the top-level ECS fields all as `objects`.

What about defining `macho.architectures` as `nested` and having each architecture's fields defined underneath? Borrowing from what @andrewstucki proposed in https://github.com/elastic/beats/pull/24195.

```yaml
- name: macho
   title: Mach-O file information.
   type: group
   description: These fields contain macOS Mach Object (Mach-O) metadata.
   fields:
     - name: architectures
        description: Object files contained inside this file by architecture
        type: nested
        fields:
          - name: cpu
             description: CPU architecture target for the file.
             type: keyword
             example: 64-bit

          - name: byte_order
             description: Byte order for the file.
             type: keyword
             example: little-endian

          - name: type
             description: Mach-O file type.
             type: keyword
```

peasead: So, you're suggesting changing:
- `macho.cpu.{architecture,byte_order,type,...}` to `macho.architecture.{cpu,byte_order,type,...}`?
- `macho.*` being a group instead of nested and `macho.architecture.cpu{cpu,byte_order,type,...}` being nested?

ebeahan: Yes, mostly. One difference - I'm proposing `macho.architecture` being the nested field that all the other fields would fall under for each architecture type.

dcode: The goal is that the "interface" to ELF, MachO, and PE have a similar feel. This will make building analytics and understanding of the data model. I discussed with @andrewstucki a while back about means to do that. The problem is that MachO fat binaries are literally two fully, self-sufficient executables in one big MachO wrapper. This means that _any_ field that we decide to nest as architecture specific will have to contain all the architecture-dependent data. The result of this discussion is that it probably made the most sense to make `file.macho` as a nested type, which drastically simplifies the data.

We should meet to discuss.

peasead:
Closing until ready to revisit this.

PR #1357, opened 2021-04-14
Stage 0 RFC for `rpc` fieldset

axw:
I would like to propose a new `rpc` fieldset, for capturing details of a remote procedure call such as gRPC service and method name, and status code.


ebeahan (approved):
Thanks, @axw, for submitting!

 LGTM for stage 0 👍 

I'll assign a number, set the date, and merge.

PR #1354, opened 2021-04-13
Adjust field set usage docs sidebar links

djptek:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Y
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)? Y
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)? n/a
- If submitting code/script changes, have you verified all tests pass locally using `make test`? Y
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? Y
- Is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed. Y
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? Y

Address https://github.com/elastic/ecs/issues/1072#issuecomment-818859084

PR #1366, opened 2021-04-19
Allow custom descriptions for self-nesting reuses

kgeller:
This adds the ability to override a short description when self nesting fields. Also, a small tweak to the ascii docs tables, for a third column to capture the field being nested.


scripts/schema/cleaner.py
```diff
@@ -234,3 +236,21 @@ def single_line_beta_description(schema_or_field, strict=True):
...
         else:
             ecs_helpers.strict_warning(msg)
+
+
+def single_line_short_override_description(schema_or_field, strict=True):
+    for field in schema_or_field['schema_details']['reusable']['expected']:
+        if not 'short_override' in field:
+            continue
+        short_length = len(field['short_override'])
+        if "\n" in field['short_override'] or short_length > SHORT_LIMIT:
```
ebeahan, line 246: An observation, but not a blocking issue to merging this as is.

Might be an opportunity to refactor and combine what's happening here and `single_line_short_description` since there's some duplication.

scripts/templates/field_details.j2
```diff
@@ -112,14 +112,15 @@ Note also that the `{{ fieldset['name'] }}` fields are not expected to be used d
...
-| Nested fields | Description
+| Nested fields | Location | Description
 
 // ===============================================================
 
 
 {% for entry in render_nestings_reuse_section -%}
 
 {#- Beta marker on nested fields -#}
+| `{{ entry['name'] }}`
```
ebeahan, line 123: I wonder if it makes sense for the `Nested fields` column values to now contain the link to the field set vs. the reused location?

For example, looking at the `client.*` page from the [preview](https://ecs_1366.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-client.html#_field_reuse_2), instead of `client.as.*`, `as` instead is linked to the `as.*` field set page.

ebeahan (commented):
I have a couple of notes, but overall the changes look great, @kgeller!

In addition to `process.parent.*`, I'm looking forward to using these changes in some other places. We can handle those in a separate PR, though 😄 .

scripts/templates/field_details.j2
kgeller: I think that would make a lot of sense! Will switch that over

scripts/templates/field_details.j2
```diff
@@ -4,6 +4,12 @@
 
 {{ fieldset['description']|replace("\n", "\n\n") }}
 
+{%- if usage_doc %}
```
ebeahan, line 7: I don't believe this PR's branch here includes the changes made in #1354, so this change is trying to revert it by adding this here and removing https://github.com/elastic/ecs/pull/1366/files#diff-e2eb14d7048aba757706fb174c9a191a774f83bbeaf7bd25f9148e7ed56603f7L144.

scripts/schema/cleaner.py
```diff
@@ -197,18 +199,37 @@ def field_assertions_and_warnings(field):
 SHORT_LIMIT = 120
 
 
-def single_line_short_description(schema_or_field, strict=True):
-    short_length = len(schema_or_field['field_details']['short'])
-    if "\n" in schema_or_field['field_details']['short'] or short_length > SHORT_LIMIT:
+def single_line_short_check(short_to_check, short_name):
```
ebeahan, line 202: 👍  Nice refactoring!

scripts/templates/field_details.j2
kgeller: I think I got this sorted now. Not quite sure what I did there

ebeahan (approved):
LGTM 👍 

I see some other cases of text wrapping even after switching the header to `Field Set`, but I don't think we should hold off on introducing these changes. We can iterate on it in another PR.

Screen Shot 2021-04-20 at 2 57 33 PM

PR #1382, opened 2021-04-23
Provide updated descriptions for `user.*` field reuses

ebeahan:
Using the `short_override` field, update the field reuse descriptions for these three `users.*` reuses:

* `user.changes.*`
* `user.effective.*`
* `user.target.*`

These updated descriptions will better describe the use of each.


kgeller (approved):
Excited to see some usages of this! LGTM

PR #1387, opened 2021-04-26
Add `user` to process schema

matt-moses:
image

Add `user` to process schema to capture the user that a processing is running as.

For issue: https://github.com/elastic/ecs/issues/1380

Test passed. Committed auto-generated changes.

PR #1388, opened 2021-04-29
[1.x] Add `user` to process schema (#1387)

kgeller:
 - Add `user` to process schema (#1387)


kgeller:
@ebeahan @djptek I ran into some conflicts on the generated file's version info. Is this correct (I went with 2.0.0-dev, but should it be 1.11.0-dev)?

PR #1390, opened 2021-04-30
Introduce index mappings JSON Schema generator

ebeahan:
### Summary

Introduce a new type of generator to the ECS tooling that generates a [JSON Schema](json-schema.org) document based on the targeted ECS version.

### Usage

Example:

```
$ python scripts/generator.py --ref v1.9.0 --es-mappings-json-schema
```

This generator isn't run by default; it's only called when the `--es-mappings-json-schema` flag is passed.

The JSON schema file is written to `<output_dir>/json_schema/schema.json` where `output_dir` is the value passed to the `--out` argument.

Example of the schema created: https://github.com/ebeahan/ecs-validator/blob/main/schemas/ecs-1.9.0.json


djptek:
@ebeahan I took a look at the generated JSON - The definitions use a back reference for the type definition, so e.g.:

```
"@timestamp": {
      "description": "Date/time when the event originated. This is the date/time extracted from the event, typically representing when the event was generated by the source. If the event source has no original timestamp, this value is typically populated by the first time the event was received by the pipeline. Required field for all events.",
      "properties": {
        "type": {
          "$ref": "#/definitions/types/date"
        }
      },
      "required": [
        "type"
      ],
      "type": "object"
    }
```

Could that be simplified to just show the type so as to be closer to the underlying `_mapping` e.g. 

```
"@timestamp": {
      "description": "Date/time when the event originated. This is the date/time extracted from the event, typically representing when the event was generated by the source. If the event source has no original timestamp, this value is typically populated by the first time the event was received by the pipeline. Required field for all events.",
      "properties": {
        "type": "date"
      }
    }
```

likewise, I'm not clear on what 

```
"required": [
        "type"
      ]
```

&

```
 "type": "object"
```

represent in this context - are they there for Kibana?

Also, I was looking for a multi-field e.g. `http.request.body.content` and this appears to be defined as `"#/definitions/types/keyword` . Could this be shown as a multi-field such as:

```
"properties": {
      "content": {
        "type": "keyword",
        "fields": {
          "text": { 
            "type":  "text"
          }
        }
      }
    }
```
 
or maybe that is the intention with ` "type": "object"`?

ebeahan:
I want to revisit this in the future, but I'm not planning on moving the effort forward anytime soon.

PR #1399, opened 2021-05-06
Update generator.py

djptek:
Remove lines where changes cached by VSCode ?

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -26,11 +26,9 @@ def main():
 
     # default location to save files
     out_dir = 'generated'
-    docs_dir = 'docs'
     if args.out:
         default_dirs = False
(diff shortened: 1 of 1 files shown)


djptek:
The build shows no changes for the docs diff - that's desired behavior as the code I removed was vestigial from a previous PR and shouldn't have impacted the docs

ebeahan:
All of these changes have existed for some time. Removing them would break quite a bit of functionality. You can see the last time any line was modified and the author of that change in the `git blame`:

https://github.com/elastic/ecs/blame/master/scripts/generator.py

You did modify this one line where one of the function parameters had a name change, but the entire line should not be deleted.

https://github.com/elastic/ecs/blame/master/scripts/generator.py#L67

PR #1400, opened 2021-05-06
[RFC] Threat Enrichment - Stage 1

rylnd:
Much of this was taken from what was deleted from #1293 and is in various stages of completion. Will annotate and iterate accordingly.

[RFC Preview](https://github.com/rylnd/ecs/blob/threat-enrichment-stage-1/rfcs/text/0021-threat-enrichment.md)

### Notes
I've stuck with `threat.enrichments` for this field name, as it is agnostic of any "matching" verbiage that may not make sense universally, but that brings up its own question (see below). Of note, other potential names that were discussed:
* `threat.matches`
* `threat.indicators`
* `threat.indicator_matches`

### Outstanding Questions
- [x] Is `matched.type` sufficient to cover all enrichment cases?
    * `matched.type: "manual"` or `matched.type: "analyst"` would cover the case where an analyst manually finds a matching indicator
    * Moving forward as is; see [summary below](https://github.com/elastic/ecs/pull/1400#discussion_r643295834)
    * Would it ever be valid to enrich an event without a field/value to reference? E.g. if an analyst _suspects_ a potential correlation?
- [x] How do I indicate that the `threat.indicator` fields are nested (as a nested array) under `threat.enrichments`?


rfcs/text/0021-threat-enrichment.md
```diff
@@ -39,12 +50,130 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+            "md5": "1eee2bf3f56d8abed72da2bc523e7431"
+          },
+          "size": 656896,
+          "name": "invoice.doc"
+        },
+        /* `matched` will provide context about which of the indicators above matched on this
+              particular enrichment. If multiple matches for this indicator object, this could
+              be a list */
+        "matched": {
+          "atomic": "0c415dd718e3b3728707d579cf8214f54c2942e964975a5f925e0b82fea644b4",
```
rylnd, line 110: I'm not convinced that `matched.atomic` and `matched.field` are sufficient to cover more sophisticated matching mechanisms than an exact match: if e.g. a user wanted to write an indicator match rule with a partial match, or if there were a more sophisticated indicator that itself represented a wildcard/regex value, then the value of the indicator would not be identical to the value from the event and so we may want two fields here.

Similarly, we would not be able to reproduce the exact conditions of the match with only one `field` value. While I've been told that the indicator field being matched upon should be self-evident, it seems safer to explicitly state it in another field, and I wanted to bring it up one more time before I shut up about it 😉 .

In general, I'm viewing these `matched.*` fields as the answer to both HOW and WHY a given event was enriched, so keep that in mind and/or correct me on that thinking.

rfcs/text/0021-threat-enrichment.md
```diff
@@ -39,12 +50,130 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+  // TODO I think this is now redundant with matched.type, but maybe still useful?
+  "tags": ["threat-match"],
+  // This should already exist from the original ingest pipeline of the document
+  "related": {
+    "hash": ["0c415dd718e3b3728707d579cf8214f54c2942e964975a5f925e0b82fea644b4"]
+  }
+}
+```
+
+### Proposed enrichment pipeline mechanics pseudocode
```
rylnd, line 129: @dcode @peasead I know this section is out of date, but I had some trouble grokking and could use some help on this one.

rfcs/text/0021/threat.yml
```diff
@@ -0,0 +1,38 @@
...
+  fields:
+
+  - name: enrichments
+    level: extended
+    type: nested
+    short: List of indicators enriching the event
+    description: >
+      A list of associated indicators enriching the event, and the context of that association/enrichment
+
+  - name: enrichments.matched.atomic
```
rylnd, line 13: @ebeahan I need to add the rest of the `matched` fields from above here, but I first wanted to verify how those should be represented given the `nested` questions above.

rfcs/text/0021-threat-enrichment.md
```diff
@@ -29,6 +24,22 @@ Stage 1: If the changes include field additions or modifications, please create
...
+1. The indicator's indicator fields, as defined in RFC 0018
+2. Other relevant ECS fields from the indicator (see below)
+3. Fields representing the context of the enrichment itself
+
+### Proposed new fields
+
+Field | Type | Example | Description
+--- | --- | --- | ---
+matched.atomic | keyword | 2f5207f2add28b46267dc99bc5382480 | The value that matched between the event and the indicator
+matched.id | keyword | db8fb691ffdb4432a09ef171659c8993e6ddea1ea9b21381b93269d1bf2d0bc2 | The _id of the indicator document that matched the event
```
rylnd, line 38: I think that `matched.id` and `matched.index` could arguably just be `_id` and `_index` at the indicator level, but I don't think those are standard ECS fields, so looking for input on this here as well.

I think the argument for placing them within this `matched` fieldset is that the primary use is as metadata about the match (i.e. "what was the exact document that matched"), but these may make sense in a more general field 🤷

rfcs/text/0021/threat.yml
ebeahan: Sorry for the delay in responding here @rylnd!

For a `nested` field, you'll need to define the `nested` field itself as `type: nested` and then each individual field underneath that `nested` object:

```yaml
    - name: enrichments.matched
      description: >
        Section information of the ELF file.
      type: nested
      level: extended

  - name: enrichments.matched.atomic
     level: extended
     short: Indicator atomic match
     ...
```

rylnd: `matched` is not intended to be `nested`, only `enrichments` is. So it sounds like the current format is correct, then?

ebeahan: Sorry, I misunderstood. You're correct; the format looks good to me. 👍

rfcs/text/0021-threat-enrichment.md
rylnd: Update: after some offline discussion and the knowledge that there exist no broad, general-purpose fields for elasticsearch-specific domain info like this, we're going to stick with `match.id` and `match.index` for now. When we fill out the corresponding Usage documentation about these fields, we'll want to note that their main purpose is auditing/debugging of enrichment logic.

rfcs/text/0021-threat-enrichment.md
peasead: @dcode, @rylnd and I chatted about this, but may still need some specifics if you could provide those.

rfcs/text/0021-threat-enrichment.md
peasead: > a user wanted to write an indicator match rule with a partial match, or if there were a more sophisticated indicator that itself represented a wildcard/regex value, then the value of the indicator would not be identical to the value from the event

How are you envisioning this? Are you thinking that there is a mechanical way needed to match _like_ fields. I can think of a use-case, DGAs. 

So, if you know an aggressor uses a specific domain structure for C2 (like `abc123\.12345abcdef\.xyz`), you'd want to be able to match if they use
```
abc123\.12345abcdef\.xyz
plo958\.59874qwersd\.xyz
lje456\.01258iekduh\.xyz
```
So, you'd need a way to match `url.full:/[aA-zZ]{3}[0-9]{3}\.[0-9]{5}[aA-zZ]{6}\.xyz MATCHES threat.indicator.url.full`?

Or am I off?

peasead (approved):
LGTM. Excited to see this taking shape.

rfcs/text/0021-threat-enrichment.md
```diff
@@ -39,12 +50,130 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+
+**NOTE**: There may be some optimization on which enrichments we attempt based upon the event categorization fields. For instance, we know that data that presents the netflow model or "interface" doesn't contain a sha256 hash. Since those categorization fields are lists, if data presented as both netflow and file (for whatever reason), then we'd check both network-related lookups and file-related lookups
+
 ## Source data
 
 <!--
 Stage 1: Provide a high-level description of example sources of data. This does not yet need to be a concrete example of a source document, but instead can simply describe a potential source (e.g. nginx access log). This will ultimately be fleshed out to include literal source examples in a future stage. The goal here is to identify practical sources for these fields in the real world. ~1-3 sentences or unordered list.
 -->
 
+Source data are ECS indicator documents as specified RFC 0008. At present, the best source of these documents is the filebeat threatintel module.
```
ebeahan, line 175:
```suggestion
Source data are ECS indicator documents as specified RFC 0008. At present, the best source of these documents is the [filebeat threatintel module](https://github.com/elastic/beats/tree/master/x-pack/filebeat/module/threatintel).
```

Any objection to including a link to the module here?

rfcs/text/0021-threat-enrichment.md
```diff
@@ -29,6 +24,22 @@ Stage 1: If the changes include field additions or modifications, please create
...
+
+1. The indicator's indicator fields, as defined in RFC 0018
+2. Other relevant ECS fields from the indicator (see below)
+3. Fields representing the context of the enrichment itself
+
+### Proposed new fields
+
+Field | Type | Example | Description
+--- | --- | --- | ---
+matched.atomic | keyword | 2f5207f2add28b46267dc99bc5382480 | The value that matched between the event and the indicator
```
ebeahan, line 37: Can these fields be written with their "fully qualified" field name here for clarity?

I think `matched.atomic` will be `threat.enrichments.matched.atomic`, but I'd rather that's explicitly captured.

rfcs/text/0021-threat-enrichment.md
```diff
@@ -69,6 +198,10 @@ The goal here is to research and understand the impact of these changes on users
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
 
+I believe this format should actually simplify much of the enrichment logic originally proposed in RFC 0008, since a naive implementation would simply copy the fields directly from each indicator document into `threat.enrichments`, and add the appropriate `matched` fields.
+
+While not a concern for ECS consumers at large, our existing experimental implementation within Kibana Security Solution will have to change significantly (for the better!), with accompanying data migration: https://github.com/elastic/security-team/issues/1102
```
ebeahan, line 203: Is there an accompanying public issue or a summary that we can include here instead of linking to the private issue?

ebeahan (commented):
Thanks for continuing to move this forward, @rylnd!

Do we have a sponsor for this RFC yet?

rfcs/text/0021-threat-enrichment.md
rylnd: I will create a public issue for this.

rylnd: https://github.com/elastic/kibana/issues/100510

rfcs/text/0021-threat-enrichment.md
rylnd: @peasead right, the idea is that with anything other than an exact match, the values for the LHS and RHS of the match are going to be different, and we don't have fields to account for both of those right now.

IPs is another example, where one could specify a CIDR block instead of a single IP address.

What is SOP for DGAs within threat intel, currently? Is there any attempt to generalize the pattern/algorithm generating those values?

rfcs/text/0021-threat-enrichment.md
```diff
@@ -39,12 +50,130 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+          "field": "file.hash.sha256",
+          "id": "abc123f03",
+          "index": "threat-indicators-index-000001",
+          "type": "indicator_match_rule"
+        }
+      }
+    ]
+  },
+  // Tag the enriched document to indicate the threat enrichment matched
+  // TODO I think this is now redundant with matched.type, but maybe still useful?
```
rylnd, line 120: Note to self: remove the TODO; this may be redundant, but tags often are.

rfcs/text/0021-threat-enrichment.md
peasead: @rylnd did you get the information you needed on this?

rfcs/text/0021-threat-enrichment.md
peasead: I'm not saying it's a must now.

There are some [experimental machine learning jobs](https://github.com/elastic/detection-rules/releases/tag/ML-experimental-detections-20200506-3) for DGAs.

rfcs/text/0021-threat-enrichment.md
rylnd: @peasead I think I finally grokked and updated that section appropriately. If you and @dcode can verify that what I wrote still makes sense, lemme know 👍

rylnd:
Updated all outstanding threads; I think this is ready for review!

@devonakerr has agreed to be sponsor on this one, as well.

rfcs/text/0021-threat-enrichment.md
rylnd: I think we can continue with the implicit "exact value match" semantics for this RFC. If in the future we need to support the aforementioned functionality, I think that we can do so with the addition of two new fields as discussed.

ebeahan (approved):
LGTM. Thanks, @rylnd, for the iterations here!

Can we capture @devonakerr as the sponsor under the `People` section in the document?

rfcs/text/0021-threat-enrichment.md
```diff
@@ -39,12 +50,127 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+          "last_seen": "2020-11-01",
+          "sightings": 4,
+          "type": ["sha256", "md5", "file_name", "file_size"],
+          "description": "file last associated with delivering Angler EK"
+        },
+        // event and file fields are copied from the indicator doc, if relevant
+        "event": {
+          "provider": "Abuse.ch",
+          "dataset": "threatintel.abusemalware",
+          "module": "threatintel"
```
rylnd, line 96: note: add `event.reference` here as well

devonakerr (approved):
Tidy PR, Ryland - LGTM

ebeahan:
@rylnd Am I good to set the date and merge this PR? 

You had [this note](https://github.com/elastic/ecs/pull/1400#discussion_r645062002); I wasn't sure if you wanted to make those changes here or in a later stage PR.

rylnd:
@ebeahan I'll make that change right now, and then we should be good to merge here 👍

rylnd:
@ebeahan all good on my end 👍

Issue #1340, opened 2021-04-01
Remove `--oss` flag from ECS generator script


ebeahan:
### Summary

Remove the `--oss` flag as an option in the ECS generator script, the [oss-fallback module](https://github.com/elastic/ecs/blob/master/scripts/schema/oss.py), and remove the `--oss` section in `USAGE.md`.

### Background

The `--oss` flag was added (#991) to enable users to fallback to OSS-compatible Elasticsearch types when possible.

The feature was deprecated (#1232) when (as of 7.11) an Apache 2.0 licensed distribution of Elasticsearch was no longer available with intent to remove in the next ECS major version.

PR #1404, opened 2021-05-10
removing --oss

kgeller:
https://github.com/elastic/ecs/issues/1340

draft until changelog entry


ebeahan:
Changelog entry for 8.0 breaking changes PR: https://github.com/elastic/ecs/pull/1408

scripts/generators/es_template.py
```diff
@@ -279,3 +278,10 @@ def es6_type_fallback(mappings):
                 field_or_multi_field_datatype_defaults(mappings[name])
         if 'properties' in details:
             es6_type_fallback(details['properties'])
+
+
+TYPE_FALLBACKS = {
```
ebeahan, line 283: Good catch migrating this out of the now removed `schema.oss`.

nit: I prefer defining constants like this at the top of the file since they should be module-wide. Not a strong opinion though, I can see value in keeping the definition close to the single use of it here.

ebeahan (approved):
LGTM 👍

scripts/generators/es_template.py
kgeller: That's a fair point, I'll move

Issue #1391, opened 2021-05-04
Remove use cases directory and code


ebeahan:
The `use-cases` directory has not been maintained for a long time. Both the directory and the code that maintains those files should be removed.

Previous PR: https://github.com/elastic/ecs/pull/632

PR #1405, opened 2021-05-10
removing use-cases

kgeller:
https://github.com/elastic/ecs/issues/1391


scripts/helper.py
```diff
@@ -110,10 +110,6 @@ def get_markdown_row(field, link, multi_field):
         # Add ticks around examples to not break table
         example = "`{}`".format(field["example"])
 
-    # If link is true, it link to the anchor is provided. This is used for the use-cases
```
ebeahan, line 113: I don't think `helper.py` is referenced anywhere in the repo other than `use-cases.py`.

We may be able to remove the file entirely?

kgeller: Good catch! Removed

ebeahan (approved):
🙌 

Screen Shot 2021-05-12 at 12 16 29 PM

PR #1402, opened 2021-05-10
Client port is wrongly documented

srikwit:
The client port is wrongly documented as 64734 instead of 54162.

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)?
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)?
- If submitting code/script changes, have you verified all tests pass locally using `make test`?
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?
- Is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed.
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)?

PR #1406, opened 2021-05-10
[master] Client port is wrongly documented (#1402)

ebeahan:
Forward ports the following commits to `master`:

* Client port is wrongly documented (#1402)


kgeller (approved):
LGTM

PR #1411, opened 2021-05-12
[Tooling] Add --exclude flag to Generator to support field removal testing

djptek:
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process] (https://github.com/elastic/ecs/blob/master/rfcs/README.md)? n/a
- Have you added an entry to the [CHANGELOG.next.md] WIP (https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)?


djptek:
Note: Moved to Draft while fixing regression

djptek:
Regression fixed: the files in folder `.rfcs/text/0017` have been included temporarily as I used those for testing, these will not be included in the final commit

scripts/generator.py
```diff
@@ -50,7 +50,7 @@ def main():
         oss.fallback(fields)
     cleaner.clean(fields, strict=args.strict)
     finalizer.finalize(fields)
-    fields = subset_filter.filter(fields, args.subset, out_dir)
+    fields = subset_filter.filter(fields, args.subset, args.exclude, out_dir)
```
ebeahan, line 53: What do you think about moving `exclude` into its own module instead of extending `subset_filter`?

Outside of the functions for loading the files, it looks like the two are independent of each other. Might help when it comes to writing tests to have things more isolated.

rfcs/text/0017/exclude-set.yml
```diff
@@ -0,0 +1,4 @@
+---
```
ebeahan, line 1: Looks like you have some good test case built! These will be good examples to help build out some of the unit testing.

scripts/schema/subset_filter.py
```diff
@@ -18,6 +20,45 @@ def filter(fields, subset_file_globs, out_dir):
...
+
+
+def pop_field(fields, path):
+    '''pops a field from yaml derived dict using path derived from ordered list of nodes'''
+    node_path = path.copy()
+    if node_path[0] in fields:
+        if len(node_path) == 1:
+            b4 = fields.copy()
+            fields.pop(node_path[0])
+            print("removed field:", (set(b4.keys() ^ set(fields.keys()))).pop())
```
ebeahan, line 36: I'm finding the set operation here a bit confusing to follow. Is the intent to show the field that was removed?

`.pop()` will remove the value but also returns the element if you want to store it in a variable.

scripts/generator.py
djptek: OK, I'll refactor :+1:

scripts/schema/subset_filter.py
djptek: Yeah, I didn't use the return value of pop because the field itself can be huge with subfields and comments and all and I didn't want to swamp the output. The set operation was useful to me while debugging but I agree it is OTT now so I'll find a more succinct way, maybe parse the pop return value after all :+1:

scripts/generator.py
djptek: I've moved that to `exclude_filter`. Where functions would have been exactly duplicated, I kept these as imports from `subset_filter` to keep things DRY

rfcs/text/0017/exclude-set.yml
djptek: thx!

scripts/schema/subset_filter.py
djptek: I've simplified this to take advantage of `flat_name`, test output now looks like:

```
djptek@Elastic-MacBook-Pro ecs % python scripts/generator.py --exclude="./rfcs/text/0017/exclude-set.yml"                (exclude_set)ecs
Removed field log.original
djptek@Elastic-MacBook-Pro ecs % python scripts/generator.py --exclude="./rfcs/text/0017/exclude-set-pass2.yml"          (exclude_set)ecs
Removed field log.original
Removed field log.file.path
djptek@Elastic-MacBook-Pro ecs % python scripts/generator.py --exclude="./rfcs/text/0017/exclude-set-fail*.yml"          (exclude_set)ecs
No match for exclusion: test_fail1_nonexistent_root.file.path
No match for exclusion: test_fail2_nonexistent_node.path
No match for exclusion: path.test_fail3_nonexistent_leaf
```

djptek:
moved to draft while updating unit tests

djptek:
Refactor: 
- moved common yml loading code & related tests from `subset_filter` to `loader`
- unstaged simplified tests from 0017
- added unit tests for `exclude_filter`
- converted warnings to exceptions in `exclude_filter`
- removed `print` statements from `exclude filter`

Move back to "ready for review"

scripts/generator.py
djptek: I did a more extensive refactor, see: https://github.com/elastic/ecs/pull/1411#issuecomment-841362269

scripts/schema/subset_filter.py
djptek: All print statements now removed in favour of unit tests, see: https://github.com/elastic/ecs/pull/1411#issuecomment-841362269

scripts/schema/loader.py
```diff
@@ -251,3 +251,33 @@ def merge_fields(a, b):
...
+        return yaml.safe_load(f.read())
+
+
+# You know, for silent tests
+def warn(message):
+    print(message)
+
+
+def eval_globs(globs):
+    '''Accepts an array of glob patterns or file names, returns the array of actual files'''
```
kgeller, line 267: Nit: should we use `"` here instead of `'` to be consistent with the `exclude_filter`?

kgeller (approved):

scripts/schema/exclude_filter.py
```diff
@@ -0,0 +1,62 @@
+import glob
```
ebeahan, line 1: Do `glob`, `yaml`, or `os` modules get called directly here? If not, we can clean up these imports.

ebeahan (commented):
Thanks for refactoring a bit and adding the unit tests!

scripts/schema/loader.py
djptek: Good catch @kgeller - I checked and that was also inconsistent elsewhere so I've normalized to `"""` as per [PEP-0257](https://www.python.org/dev/peps/pep-0257/)

scripts/generator.py
```diff
@@ -70,6 +72,8 @@ def argument_parser():
                                                        Note that "--include experimental/schemas" will also respect this git ref.')
     parser.add_argument('--include', nargs='+',
                         help='include user specified directory of custom field definitions')
+    parser.add_argument('--exclude', nargs='+',
```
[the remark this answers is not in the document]
djptek: Good catch, added to `USAGE.md`

scripts/schema/exclude_filter.py
djptek: nope, removed unused imports here and also `loader.py`and `subset_filter.py`

scripts/tests/unit/test_schema_exclude_filter.py
```diff
@@ -0,0 +1,97 @@
+from unittest import result
+from unittest.case import TestCase
+from schema import exclude_filter
+import mock
+import os
+import pprint
```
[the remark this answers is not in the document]
djptek: removed unused imports :+1:

PR #1418, opened 2021-05-14
Support `match_only_text` in Go code generator

ebeahan:
Running `make gocodegen` (and in turn, `make`) will fail if a type isn't explicitly handled in `scripts/cmd/gocodegen/gocodegen.go`.

Adding support in the tooling for `match_only_text` as we look to introduce the type into ECS ([RFC 0023](https://github.com/elastic/ecs/blob/master/rfcs/text/0023-match_only_text-data-type.md)).

### DIFF
--- a/scripts/cmd/gocodegen/gocodegen.go
+++ b/scripts/cmd/gocodegen/gocodegen.go
@@ -321,7 +321,7 @@ func goDataType(fieldName, elasticsearchDataType string) string {
 	}
 
 	switch elasticsearchDataType {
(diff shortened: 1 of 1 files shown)


djptek (approved):
I amended a typo in the CHANGELOG hope that's OK, LGTM

neu5ron:
is `match_only_text` lowercase like `text`?

kgeller:
@neu5ron https://www.elastic.co/guide/en/elasticsearch/reference/7.x/text.html yes it is!

Issue #779, opened 2020-03-07
Inconsistency between related.user and user.* object


vbohata:
Description of "related.user" says "All the user names seen on your event.". In this case the field related.user should be "related.user.name" as "related.user" should be object. I can imagine user.id should be also very useful in related section but it is impossible to add in the future with the related.user set to keyword.

janniten:
@vbohata I agree with you.  @webmat  what do you think?
Regards
Ana

ebeahan:
Like how `related.hosts` is intended for all hostnames or host identifiers seen on an event, `related.user` should include any usernames or user identifies seen on an event. I think the field description for `related.user` could be improved to better capture that intent.

The `related.*` fields are meant to facilitate pivoting with a trivial query. I think by turning the `related.*` fields into more complex objects, some of that simplicity would loss. Instead of simply querying `related.user:alice`, you would have to be more familiar with the object structure. For example, is it `related.user.name` or `related.user.id`?

ebeahan:
Opened #1355 to track improving the `related.user` field's description.

janniten:
Hi @ebeahan 
Today I was just about to post about this topic. I'm working with some windows events (4673,4732) where it would be useful to have the related.user.id in order to pivot data. Sometimes the user name is not present in the event and you need to pivot over the user id to find in a related event the user name. I found that case today while I was analyzing the event 4732 ( add user to a group) with not user name (MemberName) but with user id (MemberSID).

janniten:
I'll prepare some examples in order to show the importance to have the related.user.id field

janniten:
Hi @ebeahan ,
I agree with you that turning a related.user into an object will be go against the simplicity, but I found some cases where would be useful to have in the related.user the full user object
Here an example
Consider the windows event 4732 (A member was added to a security-enabled local group). From the event 4732 documentation: "For local groups this field typically has “-“ value, even if new member is a domain account. For some well-known security principals, such as LOCAL SERVICE or ANONYMOUS LOGON, the value of this field is “-”."
So is not possible to identify which member was added to the group, but...
There is a lot of events where the SID information is present as well as the user.name so if we were able to pivot with the related.user.id the final user could be easily found

![image](https://user-images.githubusercontent.com/33020901/118268737-10963f00-b4be-11eb-9173-b3fa01decbf6.png)

![image](https://user-images.githubusercontent.com/33020901/118268821-2b68b380-b4be-11eb-900d-7cd96d325db1.png)

[...]

PR #1420, opened 2021-05-17
Explicitly include user identifiers in `related.user` field description

### LINKED ISSUE #1355: Improve `related.user` field description
The `related.user` field should be populated with any username or user identifier value found on an event, but the current description could use updated to reflect this intent better (like was done for the `related.hosts` field description).

ebeahan:
Capturing user identifiers in addition to user names in `related.user` can be useful, but the description for `related.user` implies the field is limited to only "user names".

Here's one instance where pivoting around the user's SID from one event helps track that SID back to a user name in another: https://github.com/elastic/ecs/issues/779#issuecomment-841204487

Expanding the description for `related.user` also aligns with the description for `related.hosts`, specifying both "hostnames and other host identifiers".

Closes #1355


ebeahan:
@elastic/security-external-integrations Any feedback for or against expanding `related.user` description to include user identifiers beyond user names?

kgeller (approved):
Love the clarification

LGTM

PR #1422, opened 2021-05-18
Add `host.user.*` deprecation notice in field reuse description

ebeahan:
Using the recently added`short_override` option to better document that `host.user.*` is [deprecated](https://github.com/elastic/ecs/blob/master/docs/usage/user.asciidoc#deprecations).


kgeller (approved):
LGTM!

PR #1427, opened 2021-05-24
[PROPOSAL] Add Stage X to the RFC process

ebeahan:
### Overview

This PR proposes a new RFC stage:  `Stage X` - a discrete stage for abandoned RFCs.

Once abandoned, an RFC would:

* no longer advance in the RFC process
* have any candidate field definitions removed from ECS
* include a brief explanation of why it is being marked as abandoned.

Inactive or stale RFCs would be marked as abandoned after an extended period of inactivity on the judgment of the ECS committee. 

### Changes in this PR

* A new section comment with guidance for capturing why a proposal is marked as abandoned.
* The additional stage added to the ECS RFC stages page.


djptek (approved):
LGTM

Issue #514, opened 2019-08-08
Define ECS set of fields for SQL information


simitt:
Different beats modules exist that are collecting SQL information (MySQL, PostgreSQL). APM also collects SQL information when available. Aligning the information by defining a common field set in ECS would help retrieving information from the data in a unified way.

Following fields are currently defined in APM and beats and might help as a starting point for discussions:

**packetbeat**

[transaction_event fields](https://www.elastic.co/guide/en/beats/packetbeat/current/exported-fields-trans_event.html):
`transaction_event`
`event.dataset`
`type`
`method`
`query`

[postgresql](https://www.elastic.co/guide/en/beats/packetbeat/current/exported-fields-pgsql.html):
`pgsql.error_code`
`pgsql.error_message`
`pgsql.error_severity`
`pgsql.num_fields`
`pgsql.num_rows`

[mysql](https://www.elastic.co/guide/en/beats/packetbeat/current/exported-fields-mysql.html):
`mysql.affected_rows`
`mysql.insert_id`
`mysql.num_fields`
`mysql.num_rows`
`mysql.query`
`mysql.error_code`
`mysql.error_message`

**filebeat**
[ecs compliant](https://www.elastic.co/guide/en/beats/filebeat/7.2/exported-fields-ecs.html)
`service.type`
`event.module`
`event.dataset`

[postgresql](https://www.elastic.co/guide/en/beats/filebeat/7.2/exported-fields-postgresql.html)
`postgresql.log.timestamp`
`postgresql.log.core_id`
`postgresql.log.database`
`postgresql.log.query`
`postgresql.log.timezone`
`postgresql.log.thread_id`
`postgresql.log.user`
`postgresql.log.level`
`postgresql.log.message`

[mysql](https://www.elastic.co/guide/en/beats/filebeat/7.2/exported-fields-mysql.html)
`mysql.thread_id`
` mysql.error.thread_id`
`mysql.error.level`
`mysql.error.message`
`mysql.slowlog.lock_time.sec`
`mysql.slowlog.rows_sent`
`mysql.slowlog.rows_examined`
`mysql.slowlog.rows_affected`
`mysql.slowlog.bytes_sent`
`mysql.slowlog.bytes_received`
`mysql.slowlog.query`
`mysql.slowlog.id`
`mysql.slowlog.schema`
`mysql.slowlog.current_user`
`mysql.slowlog.last_errno`
`mysql.slowlog.killed`
`mysql.slowlog.query_cache_hit`
[...]

PR #1428, opened 2021-05-24
[RFC] Add new database object / fieldset - Stage 0

mbrancato:
Add new database object / fieldset to cover multiple database use cases.

Related to #514 


djptek:
Hi @mbrancato the [PR 514](https://github.com/elastic/ecs/issues/514) referenced mentions the APM `span.db.*`fields which currently include:

```
span.db.statement
span.db.type
span.db.instance
span.db.link
span.db.user
```

They are not currently in ECS, however, the APM functionality in Elastic Stack already provides a host of plug-and-play tools for capturing and analyzing database events as transactions and spans, here is a [blog showing how to monitor mysql transactions in a java app](https://www.elastic.co/blog/monitoring-java-applications-and-getting-started-with-the-elastic-apm-java-agent)

Do you think the [fields under span.db.*](https://www.elastic.co/guide/en/apm/server/7.13/exported-fields-apm-span.html) might fulfill your requirement if they were in ECS either as custom or as part of the standard?

mbrancato:
Hi @djptek - I think I first ran into this looking at Beats and then stumbled into that GH issue. As in that GitHub issue, the problem is there are already several tools doing different schemas to capture database information. For my specific use-case the top-level `span` is more confusing as they are not tracing spans. But surely, a reusable database field set could be placed at `span.db` or anywhere else.

I do think there would still be postgres, mysql, and other specific schemas, but having a common place to record this information makes querying the data easier.

At a minimum, I think there should be a single field set that can identify:
- DB type
- DB name(s) - this could be an array
- data container type (table, index, collection, etc)
- data container(s) - this should be an array
- query (raw query)
- fields (if any)
- operation (main operation name like SELECT, $match, _bulk)

I'm thinking nested statements should be split up if needed into multiple events.

My use case here is security related. it allows building rules around possibly queries to inserts into specific databases or tables without needing technology-specific fields. To me, having a top-level DB field set is just like having `source` or `destination` fieldsets. Trying to capture data under DB-specific technologies is like trying to have specific link-layer fieldsets like Ethernet, Frame Relay, DSL, etc.

djptek:
>> I'm thinking nested statements should be split up if needed into multiple events.

That would work really well with the APM tooling - even though the use case is Security it may still be useful to navigate up and down a stack of nested queries

re: `span.db.statement` or similar, you could get the best of both worlds 

- with slightly increased storage cost by copying selected data from e.g. field(s) such as you are proposing
- or with slightly increased complexity by aliasing the field(s)

rfcs/text/0000-database-object.md
```diff
@@ -0,0 +1,108 @@
...
+## References
+
+https://github.com/elastic/ecs/issues/514
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 103:
```suggestion
* Stage 0: https://github.com/elastic/ecs/pull/1428
```

rfcs/text/0000-database-object.md
```diff
@@ -0,0 +1,108 @@
+# 0000: Database Object
```
ebeahan, line 1:
```suggestion
# 0000: Database Field Set
```

Any objection (no pun intended 😂 ) to rephrasing the title?

ebeahan (approved):
Thanks, @mbrancato, for putting this together!

For a stage 0 RFC PR, the [criteria](https://elastic.github.io/ecs/stages.html) is kept pretty simple: `The premise of these changes is not obviously useless or inappropriate for ECS`. A database-focused field set has been discussed before, and I do think the concept would be an appropriate addition to ECS.

I agree, though, that as this RFC advances we'll want to consider how any proposed fields relate to the existing `span.db.` fields.

A couple of other thoughts (no need to address/comment now - we can discuss further as the proposal advances):

* `db.*` or `database.*`? ECS tries to avoid abbreviations when possible, but it also makes exceptions for well-known concepts like `ip`, `geo`, `os`. I think `db.*` is definitely a well-known shorthand, but the full `database.*` is also short. So I'm on the fence 😄 
* I'm not clear what type of use cases would need to reuse a `db.*` field set. Again, something we can discuss in a later stage PR.

mbrancato:
Thanks. I did have one other use case I identified. That is being able to log functions called. It would work along side fields in my current mockup.

- fields (if any)
- functions (if any)
  - fields (if any, for value-only functions)

ebeahan:
Thanks, @mbrancato.

@djptek, did you have anything you wanted to continue discussing before we move this proposal to stage 0?

djptek:
> Thanks, @mbrancato.
> 
> @djptek, did you have anything you wanted to continue discussing before we move this proposal to stage 0?

LGTM!

djptek:
@mbrancato please use RFC 0026

ebeahan:
Thanks, all. I'll go ahead and set the RFC number, set the date, and then merge.

PR #1443, opened 2021-05-28
Remove guidance to lowercase `http.request.method`.

### LINKED ISSUE #1363: Remove guidance to lowercase `http.request.method`
### Summary

Completely remove guidance to lowercase `http.request.method`.

### Background

The guidance to lowercase the value populating `http.request.method` was deprecated in ECS 1.6.0 #838 . It should be removed entirely from the field's description at the next major.

ebeahan:
In addition to removing the deprecation note, I included [new guidance](https://github.com/elastic/ecs/pull/1443/files#diff-535c241aff61a02effef7c3af2446d787e3c4641139ca0f9e53b507ed7225086R4238) about retaining the casing from the original event.

I also limited the `Example` attribute to a single value since this field is not expected to be an array. 

Closes #1363

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -13,6 +13,7 @@ Thanks, you're awesome :-) -->
 * Removing deprecated --oss from generator #1404
 * Removing use-cases directory #1405
 * Remove `host.user.*` field reuse. #1439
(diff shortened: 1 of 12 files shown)


djptek (approved):
LGTM

PR #1450, opened 2021-06-02
[RFC] Remove `process.ppid` - Stage 1

ebeahan:
### Summary

Revisions to the stage zero remove `process.ppid` RFC for consideration of advancing to stage one.

### Criteria

- [x] Subject matter experts identified and weighed in on the high-level utility of these changes in the pull request
- [x] ECS team weighed in on the appropriateness of these changes in the pull request

[Proposal preview](https://github.com/ebeahan/ecs/blob/rfc/0022/stage-1/rfcs/text/0022-remove-process-ppid.md)


ebeahan:
This proposal is to migrate from populating `process.ppid` to `process.parent.id` and eventually remove the `process.ppid` field from ECS as a future, breaking change.

Sharing with the mentioned group here for awareness and any feedback.

@andrewkroh @jamiehynds @jonathan-buttner @gabriellandau  @rw-access @peasead @axw @felixbarny @cyrille-leclerc @jasonrhodes @exekias

jasonrhodes:
Thanks! Do we have a process for deprecating a field in ECS and then mentioning what version it stopped being an active part of the spec, so we don't ever lose it from the ECS docs? That seems like a good idea for these kind of deprecations, I'm just curious what the overall process is.

jonathan-buttner:
@ebeahan with the removal of `ppid` would we be moving to `process.parent.id` or `process.parent.pid`? @ferullo correct me if I'm wrong but on the endpoint side we populate `process.pid` and `process.parent.pid` if we have it.

ferullo:
I think that's correct. I spot checked some Endpoint data and it at least appears to be true. In any case, it would be easy for Endpoint to stop producing `process.ppid` and always use `process.parent.pid`. I did notice that Endpoint produces `process.parent.ppid` (which is the parent's parent's pid). Would that change to `process.parent.parent.pid` or no longer be allowed in ECS?

More broadly than just considering Endpoint, I do wonder if there are any downstream affects on Detection Engine rules or EQL rules (@rw-access) to removing this field?

ebeahan:
> @ebeahan with the removal of ppid would we be moving to process.parent.id or process.parent.pid?

@jonathan-buttner `process.parent.pid` - thanks for catching! I've edited the earlier comment.

ebeahan:
> I did notice that Endpoint produces process.parent.ppid (which is the parent's parent's pid). Would that change to process.parent.parent.pid or no longer be allowed in ECS?

As proposed right now, removing `process.ppid` would also remove `process.parent.ppid`.

@ferullo by produce do you mean Endpoint is actively populating that field? Or is it just present in the field mappings?

rw-access:
We have two rules using that field as of 7.13.
I'll get a PR in to switch to `process.parent.pid` with an x-ref back.

gabriellandau:
> As proposed right now, removing process.ppid would also remove process.parent.ppid.
> 
> @ferullo by produce do you mean Endpoint is actively populating that field? Or is it just present in the field mappings?

Regarding `process.parent.ppid`, Endpoint does not set it for any regular events.  Endpoint does attempt to set it for alerts.

As to whether we need it:
* Security App - We have process lineage for Analyzer.
* Rules - It seems [no rules](https://github.com/elastic/detection-rules/search?l=TOML&q=ppid) currently use `process.parent.ppid`.
* Allowlisting - PIDs are ephemeral, not really good for static allowlist entries.  We did add the ability to have dynamic allowlist entries that compare fields within the alert to each other.  I could imagine an allowlist entry "allow process X to do Y to its grandchildren" aka `allow if process.pid == Target.process.parent.ppid`, but it's strictly a hypothetical at the moment, without any specific real-world applications.

ebeahan:
> Do we have a process for deprecating a field in ECS and then mentioning what version it stopped being an active part of the spec, so we don't ever lose it from the ECS docs?

@jasonrhodes Removing a field is proposed through the [RFC process](https://github.com/elastic/ecs/tree/master/rfcs). If there's an agreement to remove the field, the field's description is updated in the docs to indicate it's been deprecated and will be removed in the next ECS major version.

Right now, there isn't a single place in the ECS docs listing deprecated fields, but I like the idea of introducing one.

jonathan-buttner:
> * Security App - We have process lineage for Analyzer.

Within the process analyzer we don't use it to layout the graph (we use `entity_id` for that). I don't _think_ we use it anywhere else except maybe displaying the value in the side panel. @michaelolo24 or @kqualters-elastic is that accurate?

rfcs/text/0022-remove-process-ppid.md
```diff
@@ -15,50 +15,82 @@ Stage 0: Provide a high level summary of the premise of these changes. Briefly d
...
 
 ## Fields
 
-<!--
-Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting.
--->
+Removing `process.ppid` will take place in two steps:
 
-### Timeline
+1. ECS `1.x`: Indicate that `process.ppid` is deprecated in the fields description in an upcoming ECS minor release. Producers and consumers of `process.ppid` should transition to using `process.parent.id` instead.
```
jonathan-buttner, line 24: @ebeahan just a heads up there's a couple places in the RFC that we'll want to switch to `process.parent.pid` instead of `process.parent.id` 👍

michaelolo24: > > * Security App - We have process lineage for Analyzer.
> 
> Within the process analyzer we don't use it to layout the graph (we use `entity_id` for that). I don't _think_ we use it anywhere else except maybe displaying the value in the side panel. @michaelolo24 or @kqualters-elastic is that accurate?

Yea, I don't believe it's used anywhere, just displayed.

kgeller (approved):
LGTM!

PR #1452, opened 2021-06-03
Improve cloud region and AZ field descriptions

### LINKED ISSUE #1287: Improve `cloud.region` field description
Today, `cloud.region` uses this description: `Region in which this host is running.`

The `this host` portion could be interpreted to limit the field's usage to IaaS-type services (EC2, GCE, Azure VMs), but there are many other cloud services where the deployed region of the service/resource may need capturing.

I suggest we clarify this by updating the field's description.

ebeahan:
Expand the field definitions for `cloud.region` and `cloud.availability_zone` to include services or resources. The current definition implies these fields are limited to hosts/VMs running in an IaaS service like EC2 or GCE.

Closes #1287


kgeller (approved):
Looks great!

PR #1454, opened 2021-06-07
Add field event.agent_id_status

andrewkroh:
This adds a field that can be used to reflect the status of the agent.id verification performed by the receiving system or data pipeline. If the receiving system checks that the sender is authorized for a given agent.id value then the outcome can be added to this field.

For example you might implement mTLS for authenticating agents sending data to Logstash. You could add the agent's ID to the agent's client cert subject and then validate incoming events in your Logstash pipeline to ensure the data has the agent ID.

Or with Elasticsearch you could provide each of your agents with an API key that has the agent.id associated to the API key metadata. Then using an Ingest Node pipeline you can validate the agent.id against the client's API key metadata.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -24,6 +24,7 @@ Thanks, you're awesome :-) -->
 * `elf.*` field set added as beta. #1410
 * Remove `beta` from `orchestrator` field set. #1417
 * Extend `threat.*` field set beta. #1438
+* Added `event.agent_id_status` field.
 
 #### Improvements
 
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -24,7 +24,7 @@ Thanks, you're awesome :-) -->
 * `elf.*` field set added as beta. #1410
 * Remove `beta` from `orchestrator` field set. #1417
 * Extend `threat.*` field set beta. #1438
-* Added `event.agent_id_status` field.
+* Added `event.agent_id_status` field. #1454 
 
 #### Improvements
 
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ generate: generator codegen
 
 # Run the new generator
 .PHONY: generator
-generator:
+generator: ve
 	$(PYTHON) scripts/generator.py --strict --include "${INCLUDE}"
 
 # Generate Go code from the schema.
--- a/schemas/event.yml
+++ b/schemas/event.yml
@@ -762,20 +762,17 @@
         the event, otherwise one of the other allowed values should be used.
 
         If no validation is performed then the field should be omitted.
+
+        The allowed values are:
+
+        `verified` - The `agent.id` field value matches expected value obtained
+        from auth metadata.
+
+        `mismatch` - The `agent.id` field value does not match the expected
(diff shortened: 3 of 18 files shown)


schemas/event.yml
```diff
@@ -744,3 +744,38 @@
...
+
+        For example if the agent's connection is authenticated with mTLS and
+        the client cert contains the ID of the agent to which the cert was issued
+        then the `agent.id` value in events can be checked against the certificate.
+        If the values match then `event.agent_id_status: verified` is added to
+        the event, otherwise one of the other allowed values should be used.
+
+        If no validation is performed then the field should be omitted.
+      example: verified
+      allowed_values:
```
ebeahan, line 766: Setting `allowed_values` was introduced for the event category fields and, unfortunately, only works with those four fields right now. The attribute isn't very well documented right now, which should be fixed. It's also why we see the docs CI failure right now.

Can we move the allowed values into the `agent_id_status` description field and delete the `allowed_values` section?

I do like the idea of having a defined attribute to specify the allowed values outside of the event categorization fields, but implementing it is outside the scope here. 😄

CHANGELOG.next.md
```diff
@@ -24,6 +24,7 @@ Thanks, you're awesome :-) -->
 * `elf.*` field set added as beta. #1410
 * Remove `beta` from `orchestrator` field set. #1417
 * Extend `threat.*` field set beta. #1438
+* Added `event.agent_id_status` field.
```
ebeahan, line 27:
```suggestion
* Added `event.agent_id_status` field. #1454 
```

Makefile
```diff
@@ -67,7 +67,7 @@ generate: generator codegen
 
 # Run the new generator
 .PHONY: generator
-generator:
+generator: ve
```
ebeahan, line 70: 👍

ebeahan (commented):
Two thoughts about naming and the allowed values:

1. What about dropping the `agent_id_` prefix and simply use `mismatch` and `missing? The prefix feels a bit redundant with the field name specifying `agent_id` already.
2. I don't have an issue with `event.agent_id_status` and understand shorter field names are often preferred, but I do think something more detailed, like `event.agent_id_validation_status`, describes the field's purpose better.

schemas/event.yml
andrewkroh: Done

andrewkroh:
> What about dropping the agent_id_ prefix and simply use mismatch and missing? The prefix feels a bit redundant with the field name specifying agent_id` already.

Good idea. That change is reflected in the changes I pushed.

> I don't have an issue with event.agent_id_status and understand shorter field names are often preferred, but I do think something more detailed, like event.agent_id_validation_status, describes the field's purpose better.

I had similar thoughts. Initially I had `event.agent_id_validation_status`, but I struggled with the overall length. So that how I got to the shorter name. But if more folks weigh in and want the more descriptive value then I can be swayed.

andrewkroh:
Doc preview:

Screen Shot 2021-06-09 at 5 19 04 PM

ebeahan (approved):
Thanks, the doc preview for the field looks great 👍

PR #1468, opened 2021-06-23
[RFC] 0008: Correct expected indicator.type value for X509 Certificates

adriansr:
This is a minor change to a field's description in RFC 0008 (Threat Intel).

The documentation for the `indicator.type` field lists `x-509-certificate` as an expected value. However, the correct STIX 2.0 Cyber Observable type name for X509 Certificates is `x509-certificate`, according to [the STIX 2.0 docs.](http://docs.oasis-open.org/cti/stix/v2.0/cs01/part4-cyber-observable-objects/stix-v2.0-cs01-part4-cyber-observable-objects.html#_Toc496716295)

### DIFF
--- a/rfcs/text/0008/threat.yml
+++ b/rfcs/text/0008/threat.yml
@@ -53,7 +53,7 @@
         * url
         * user-account
         * windows-registry-key
-        * x-509-certificate
+        * x509-certificate
 
     example: ipv4-addr
 


ebeahan:
Great catch, @adriansr!

Would you also update it [here](https://github.com/elastic/ecs/blob/master/experimental/schemas/threat.yml#L103) in the experimental schema definitions?

I'm also adding @peasead @rylnd as reviewers.

ebeahan:
Actually, let's handle this differently since we have an active PR for this RFC.

I cherry-picked your change @adriansr and incorporate it into #1293: https://github.com/elastic/ecs/pull/1293/commits/31616fa2fa8b4f6b692da1cf73e4981f10df396b

PR #1488, opened 2021-06-29
Ensure `doc_values` parameter is passed into Beats field defintions

### LINKED ISSUE #1482: `doc_values` parameter not set in Beats artifact
Certain fields have `index: false` and `doc_values: false` in their ECS definition, like `event.original`:

https://github.com/elastic/ecs/blob/master/schemas/event.yml#L577-L599

When `doc_values: false` is defined in the field definition, it's not being added to the maintained Beats fields YAML artifact:

https://github.com/elastic/ecs/blob/master/generated/beats/fields.ecs.yml#L1737-L1750

ebeahan:
`doc_values` should be set in the Beat field definition artifacts if the parameter is listed on a field's definition. 

This change adds `doc_values` into the Beats generator allow list (`allowed_keys`).

Also adds basic unit testing for portions of the Beats generator module.

Closes #1482


djptek (approved):
LGTM

ebeahan:
Planning to backport to `1.11` in addition to `1.x`

PR #1502, opened 2021-07-01
CTI 1.11 Addendum

rylnd:
@ebeahan I think this is ready for review, just a few questions:

1. How/where should the RFC docs be updated to reflect these changes (for posterity, not the definitions themselves)
2. A critical omission from the enrichments RFC is that the `indicator` fields are not present under `threat.enrichments`! How do I go about defining `threat.enrichments.indicator.*` as coming from `threat.indicator.*` ?


ebeahan (approved):
LGTM.

Thanks, @rylnd @peasead!

devonakerr (approved):
Apologies for the delay in reviewing, LGTM.

PR #1508, opened 2021-07-09
RFC 13 address stage 0 feedback

leehinman:
- switch to field set that can be nested under `source` and `destination`

### DIFF
--- a/rfcs/text/0013-network-headers.md
+++ b/rfcs/text/0013-network-headers.md
@@ -1,16 +1,11 @@
 # 0013: network headers
-<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
 - Date: **2021-02-05** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 <!--
 As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
 Feel free to remove these comments as you go along.
--->
-
-<!--
-Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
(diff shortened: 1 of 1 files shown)


ebeahan:
Adding this entry listing the stage 1 criteria to help us track outstanding items.

Stage 1 Criteria:

- [ ] Opened pull request for this proposal revising the existing strawperson
- [ ] Identified "sponsor" at Elastic who will participate in RFC process and take ownership of the change after the process completes
- [ ] Outlined initial field definitions
- [ ] High-level description of examples of usage
- [ ] High-level description of example sources of data
- [ ] Identified potential concerns and implementation challenges/complexity
- [ ] Subject matter experts identified and weighed in on the high level utility of these changes in the pull request
- [ ] ECS team weighed in on appropriateness of these changes in the pull request

ebeahan:
@leehinman When you have some time to come back to this, would you be able to create field definitions of these additions? These definitions can be organized into a directory named `0013` after the RFC like we've done [elsewhere](https://github.com/elastic/ecs/tree/master/rfcs/text/0001).

I think accounting for capturing the different headers for directional flows is a good improvement. I worry about using `header` as the field set name for all flag data because values like `.ip` and `.port` are also in the connection headers, and I could see it becoming confusing quickly.

legoguy1000:
I'll add my comment that I think that the header fields should stay under `network` as all the tools the supply that data, don't distinguish. For all traffic the header will always be from the source. 🤷‍♂️

legoguy1000:
@any updates on this??

legoguy1000:
un-stale :)

rwaight:
If the current plan is to put `network` as a field set _under_ `source` and `destination`, then it should also be placed _under_ `client` and `server` as this is relevant to data coming from Zeek.

legoguy1000:
don't be stale :)

rwaight:
> - [ ] Identified "sponsor" at Elastic who will participate in RFC process and take ownership of the change after the process completes

If this still needs an internal "sponsor" at Elastic, I am willing to sponsor this RFC.  As mentioned in https://github.com/elastic/ecs/pull/1508#issuecomment-1112650070. I still agree that the header fields should stay under `network.` as mentioned in https://github.com/elastic/ecs/pull/1508#issuecomment-965757358.

If we have an internal sponsor, what are the next steps to move this forward, @ebeahan and @leehinman ?

legoguy1000:
Bump

PR #1504, opened 2021-07-06
Threat intelligence RFCs stage 2 addenda

ebeahan:
Incorporate changes from the threat intelligence addenda for stage 2: #1502 

[Docs preview](https://ecs_1504.docs-preview.app.elstc.co/diff)

PR #1478, opened 2021-06-24
Stage 2 changes for RFC 0021 - Threat enrichments

ebeahan:
Apply stage 2 beta changes proposed in the threat enrichments RFC [0021](https://github.com/elastic/ecs/blob/master/rfcs/text/0021-threat-enrichment.md): #1460

PR #1471, opened 2021-06-23
Stage 2 changes for RFC 0008 - threat indicator fields

ebeahan:
Apply stage 2 beta changes proposed in the threat indicator fields RFC [0008](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md): #1293 

[Docs preview of these changes](https://ecs_1478.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-threat.html)

PR #1510, opened 2021-07-12
New event categorization values to support threat intel use cases

ebeahan:
***NOTE***: Opening PR as _draft_ until the functionality is added into the ECS tooling to include the `beta` tooltip for event categorization values.

### Summary

This PR introduces three categorization field allowed values.

These fields were initially proposed in [RFC 008 - threat intelligence fields](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#proposed-new-values-for-event-fieldset) and will be added as `beta` values to align with RFC 0008 currently being at stage 2.

#### `event.kind: enrichment`

Introduced here to support threat enrichment, but it will not be limited to that use case. Available for any ECS event that will be used to enrich others.

#### `event.category: threat`

Categorize events that contain details about threat actors' tactics, techniques, and procedures.

#### `event.type: indicator`

Sub-bucket threat enrichment indicators of compromise.

### Questions for reviewers

* Do these descriptions capture the intent of each new value accurately?
* Particularly with `event.catgory:threat`, I limited the scope to `cybersecurity threats` in the description. Is this too narrow?
* Should the `threat` category include additional expected event types beyond `indicator`?

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -2576,7 +2576,7 @@ Note: this field should contain an array of values.
 
 *Important*: The field value must be one of the following:
 
-authentication, configuration, database, driver, file, host, iam, intrusion_detection, malware, network, package, process, registry, session, web
+authentication, configuration, database, driver, file, host, iam, intrusion_detection, malware, network, package, process, registry, session, threat, web
 
 To learn more about when to use which value, visit the page
 <<ecs-allowed-values-event-category,allowed values for event.category>>
@@ -2749,7 +2749,7 @@ type: keyword
 
 *Important*: The field value must be one of the following:
 
-alert, event, metric, state, pipeline_error, signal
+alert, event, metric, state, pipeline_error, signal, enrichment
 
 To learn more about when to use which value, visit the page
 <<ecs-allowed-values-event-kind,allowed values for event.kind>>
@@ -3006,7 +3006,7 @@ Note: this field should contain an array of values.
 
 *Important*: The field value must be one of the following:
 
-access, admin, allowed, change, connection, creation, deletion, denied, end, error, group, info, installation, protocol, start, user
+access, admin, allowed, change, connection, creation, deletion, denied, end, error, group, indicator, info, installation, protocol, start, user
 
 To learn more about when to use which value, visit the page
 <<ecs-allowed-values-event-type,allowed values for event.type>>
--- a/docs/field-values.asciidoc
+++ b/docs/field-values.asciidoc
@@ -46,6 +46,7 @@ The value of this field can be used to inform how these kinds of events should b
 * <<ecs-event-kind-state,state>>
 * <<ecs-event-kind-pipeline_error,pipeline_error>>
 * <<ecs-event-kind-signal,signal>>
+* <<ecs-event-kind-enrichment,enrichment>>
 
 [float]
 [[ecs-event-kind-alert]]
@@ -111,6 +112,16 @@ Usage of this value is reserved, and pipelines should not populate `event.kind` 
 
 
 
+[float]
+[[ecs-event-kind-enrichment]]
+==== enrichment
+
+The `enrichment` value indicates an event collected to provide additional context, often to other events.
+
+An example is collecting indicators of compromise (IOCs) from a threat intelligence provider with the intent to use those values to enrich other events. The IOC events from the intelligence provider should be categorized as `event.kind:enrichment`.
+
+
+
 [[ecs-allowed-values-event-category]]
 === ECS Categorization Field: event.category
 
@@ -136,6 +147,7 @@ This field is an array. This will allow proper categorization of some events tha
 * <<ecs-event-category-process,process>>
 * <<ecs-event-category-registry,registry>>
 * <<ecs-event-category-session,session>>
+* <<ecs-event-category-threat,threat>>
 * <<ecs-event-category-web,web>>
 
 [float]
@@ -315,6 +327,18 @@ start, end, info
 
 
 [float]
+[[ecs-event-category-threat]]
+==== threat
+
+Relating to cybersecurity threats. Use this category to visualize and analyze events describing threat actors' targets, motives, or behaviors.
+
+
+*Expected event types for category threat:*
+
+indicator
+
+
+[float]
 [[ecs-event-category-web]]
 ==== web
 
@@ -348,6 +372,7 @@ This field is an array. This will allow proper categorization of some events tha
 * <<ecs-event-type-end,end>>
 * <<ecs-event-type-error,error>>
 * <<ecs-event-type-group,group>>
+* <<ecs-event-type-indicator,indicator>>
 * <<ecs-event-type-info,info>>
 * <<ecs-event-type-installation,installation>>
 * <<ecs-event-type-protocol,protocol>>
@@ -439,6 +464,16 @@ The error event type is used for the subset of events within a category that ind
 ==== group
 
 The group event type is used for the subset of events within a category that are related to group objects. Common example: `event.category:iam AND event.type:creation AND event.type:group`. You can further distinguish group operations using the ECS `event.action` field.
+
+
+
+[float]
+[[ecs-event-type-indicator]]
+==== indicator
+
+The indicator event type is used for the subset of events within a category that contain details about indicators of compromise (IOCs).
+
+A common example is `event.category:threat AND event.type:indicator`.
 
 
 
--- a/experimental/generated/ecs/ecs_flat.yml
+++ b/experimental/generated/ecs/ecs_flat.yml
@@ -2407,6 +2407,11 @@ event.category:
     - end
     - info
     name: session
+  - description: Relating to cybersecurity threats. Use this category to visualize
+      and analyze events describing threat actors' targets, motives, or behaviors.
+    expected_event_types:
+    - indicator
+    name: threat
   - description: 'Relating to web server access. Use this category to create a dashboard
       of web server/proxy activity from apache, IIS, nginx web servers, etc. Note:
       events from network observers such as Zeek http log may also be included in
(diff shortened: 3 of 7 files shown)


docs/field-values.asciidoc
```diff
@@ -314,6 +326,18 @@ The session category is applied to events and metrics regarding logical persiste
 start, end, info
 
 
+[float]
+[[ecs-event-category-threat]]
+==== threat
+
+Relating to cybersecurity threats. Use this category to visualize and analyze events describing threat actors' targets, motives, or behaviors.
```
peasead, line 333:
```suggestion
Use this category to visualize and analyze events describing threat actors' targets, motives, or behaviors.
```

peasead: I think that `threat` could be used to describe more than just cybersecurity. Especially if this is expanded at some point to include physical threat uses - such as for traditional kinetic threats or human threats like you could observe in law enforcement, terrorism, or economic espionage.

experimental/generated/ecs/ecs_flat.yml
```diff
@@ -2407,6 +2407,11 @@ event.category:
     - end
     - info
     name: session
+  - description: Relating to cybersecurity threats. Use this category to visualize
```
peasead, line 2410:
```suggestion
  - description: Use this category to visualize
```

peasead: Reference: https://github.com/elastic/ecs/pull/1510/files#r676902191

generated/ecs/ecs_flat.yml
```diff
@@ -2057,6 +2057,11 @@ event.category:
     - end
     - info
     name: session
+  - description: Relating to cybersecurity threats. Use this category to visualize
```
peasead, line 2060:
```suggestion
  - description: Use this category to visualize
```

peasead: Reference: https://github.com/elastic/ecs/pull/1510/files#r676902191

generated/ecs/ecs_nested.yml
```diff
@@ -2835,6 +2835,11 @@ event:
         - end
         - info
         name: session
+      - description: Relating to cybersecurity threats. Use this category to visualize
```
peasead, line 2838:
```suggestion
      - description: Use this category to visualize
```

schemas/event.yml
```diff
@@ -296,6 +305,12 @@
             - start
             - end
             - info
+        - name: threat
+          description: >
+            Relating to cybersecurity threats. Use this category to visualize and analyze events
```
peasead, line 310:
```suggestion
            Use this category to visualize and analyze events
```

experimental/generated/ecs/ecs_nested.yml
```diff
@@ -3185,6 +3185,11 @@ event:
         - end
         - info
         name: session
+      - description: Relating to cybersecurity threats. Use this category to visualize
```
peasead, line 3188:
```suggestion
      - description: Use this category to visualize
```

rylnd (commented):
I had a similar thought to @peasead about generalizing the definition of `threat`, so I'm glad that's covered here 👍 .

The only other note I had here, broadly, was about the use of `event.kind: enrichment`. Coupled with the other sub-categories introduced here, it seems potentially redundant to define a new top-level `event.kind` solely for an intended use case. 

I think the argument could be made that the data generated by filebeat's threatintel modules is more general than enrichment: it's intelligence; something (potentially) noteworthy that was observed and recorded. That might be a broad enough definition to simply reuse `event`, or maybe we generalize to `intelligence` or even `observable`, but IMO `enrichment` describes a use case as opposed to a category of data.

As an example, the dashboards that ship with filebeat threatintel modules describe the population of ingested data, and do not touch on enrichment at all, because that's a downstream process (brief aside: I _could__ see `event.kind: enrichment` being used to capture an actual enrichment event, e.g. "event A was enriched with indicator B")

peasead:
> I had a similar thought to @peasead about generalizing the definition of `threat`, so I'm glad that's covered here 👍 .
> 
> The only other note I had here, broadly, was about the use of `event.kind: enrichment`. Coupled with the other sub-categories introduced here, it seems potentially redundant to define a new top-level `event.kind` solely for an intended use case.
> 
> I think the argument could be made that the data generated by filebeat's threatintel modules is more general than enrichment: it's intelligence; something (potentially) noteworthy that was observed and recorded. That might be a broad enough definition to simply reuse `event`, or maybe we generalize to `intelligence` or even `observable`, but IMO `enrichment` describes a use case as opposed to a category of data.
> 
> As an example, the dashboards that ship with filebeat threatintel modules describe the population of ingested data, and do not touch on enrichment at all, because that's a downstream process (brief aside: I _could__ see `event.kind: enrichment` being used to capture an actual enrichment event, e.g. "event A was enriched with indicator B")

To be honest, I never quite understood the `event.kind: enrichment` for the data provided in a threat feed. I always thought `event.kind: enrichment` should be intelligence coupled with an observable or, at a minimum, multiple data points combined together...ideally all three.

I assumed it was for some schema requirement I had missed. I continually forgot to ask about it in the right audience.

I think that `event.kind: event` makes the most sense. I don't think `observable` is right because nothing has been observed and it's not generally "intelligence" until it is contextually relevant to the local environment. 

You mentioned it's not in the dashboards, so that would be a pretty simple and non-impactful change (enrichments -> event), right?

docs/field-values.asciidoc
ebeahan: Great points - will make this adjustment.

ebeahan:
@rylnd @peasead No objections from my end if we pause on introducing `event.kind: enrichment`.

However, the `threatintel` Beats modules are already actively using `event.kind: enrichment`: https://github.com/elastic/beats/blob/v7.13.4/x-pack/filebeat/module/threatintel/abusemalware/test/abusechmalware.ndjson.log-expected.json#L5.

Ideally, those modules will use valid ECS categorization values. If there's a change in direction, the Beats modules will also need to reflect that change, I believe.

rylnd:
> Ideally, those modules will use valid ECS categorization values. If there's a change in direction, the Beats modules will also need to reflect that change, I believe.

Yep, that's a big consideration that I should have mentioned. If we need _some_ new value for `event.kind` I think that `enrichment` is acceptable, but I wanted to ensure that a) we do need a new `event.kind` value and that b) `enrichment` isn't a significant departure from the other values (usage vs categorization as I mentioned above).

@P1llus @dcode would you care to weigh in?

dcode:
I think `event.kind: enrichment` makes a lot of sense for a general use case. The documents do not generally describe something that happened, but rather some additional knowledge we can combine with something that happened. So, it doesn't make sense to call the threatintel documents "events". 

At the same time, there's plenty of use cases that it could open up: named geopoints/shapes, organizational department data for a given username or IP, CVE info (!)... I think once these mechanics are worked out, there's so much value that could be added.

I also agree with @peasead that observable and intelligence doesn't make sense.

MikePaquette:
Hi @rylnd Thanks for moving this PR forward, and for sanity-checking with all of us on the `event.kind` value.

I am +1 for moving forward with the proposed `event.kind:enrichment` and its proposed definition.

>  a) I wanted to ensure that we do need a new event.kind value

We do.  As @dcode mentions above, none of the existing [allowed values](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-kind.html#ecs-allowed-values-event-kind) is appropriate for additional knowledge we can combine with something that happened (e.g., events, metrics) to provide beneficial information and context to the analyst.

>  b) [I wanted to ensure that] enrichment isn't a significant departure from the other values (usage vs categorization as I mentioned above)

I don't think it's inconsistent, as it seems to "fit" quite well in this example, threat intel, and some other envisioned for the future, such as, again suggested by @dcode, named geopoints/shapes, organizational department data for a given username or IP, CVE info.  

Brainstorming _**hypothetical**_ future categorization triples to demonstrate consistent usage:

- `event.kind:enrichment`, `event.category:threat`, `event.type:indicator` (this example)
- `event.kind:enrichment`, `event.category:geo`, `event.type:point` (future geo points)
- `event.kind:enrichment`, `event.category:geo`, `event.type:shape` (future geo shapes)
- `event.kind:enrichment`, `event.category:user`, `event.type:???` (future user data)
- `event.kind:enrichment`, `event.category:vulnerability`, `event.type:cve` (future CVE lists)
- etc.

peasead:
Great feedback. TL;DR were good with `event.kind:enrichment`. 

I always understood indicators would be combined with other events to make enrichments and intelligence, I just wasn't aware  "evil.com" would be considered an `event.kind:enrichment` when it was data provided by a feed and not enriched. Just data. I assumed once "evil.com" + geo (in the examples) happens, then the document would be the `event.kind:enrichment`. 

This is why I didn't bring it up before as I assumed it was a schema implementation I wasn't thinking about properly. I assumed `event.kind` was for what data _was_ not what it could be. 

That's why I held my nose and suggested `event.kind:event`, because I wasn't sure that raw data provided by a threat feed met the bar for a new `event.kind`.

ebeahan:
Thanks, all, for the discussion. Looks like we have a consensus to move forward with these additions.

 I also propose these changes are backported to the `1.11` release branch even though it's frozen:

1. We're already using these values in the `threatintel` module, and I think it's better to have them formally established in ECS sooner than wait for 1.12.
2. The accompanying `threat.*` changes (#1471, #1478, #1504) are already included in `1.11`
3. There are no field changes made with this PR. It's only adding the new allowed values to the appropriate sections of the ECS docs.

Not expecting objections, but I did want to raise and open the opportunity for any feedback on including these changes for `1.11`.

ebeahan:
I'd appreciate a final 👀 and ✅ if there's no additional feedback.

docs/field-details.asciidoc
```diff
@@ -2749,7 +2749,7 @@ type: keyword
 
 *Important*: The field value must be one of the following:
 
-alert, event, metric, state, pipeline_error, signal
+alert, event, metric, state, pipeline_error, signal, enrichment
```
kgeller, line 2752: for consistency with the others, should we alphabetize this?

```suggestion
alert, enrichment, event, metric, pipeline_error, signal, state
```

kgeller (approved):

PR #1511, opened 2021-07-12
Present `beta` tooltip for categorization values

ebeahan:
### Summary

Allow for a `beta` attribute to be defined on an ECS categorization allowed value and that `beta` attribute's value to be passed into the ECS docs as a `beta` tooltip.

This expands the `beta` label supported initially added in #1051.

### Description

An allowed value defined underneath one of the ECS categorization fields will now support the `beta` attribute.

It's defined on the desired value:

```yaml
        - name: example
          description: >
            an example description
          beta: Use of this allowed value is currently beta
```

Will render as part of the allowed values description in the docs:

Screen Shot 2021-07-12 at 5 00 15 PM

And will act as a tooltip displaying a `beta` notice when moused over:

Screen Shot 2021-07-12 at 5 02 11 PM


kgeller (approved):
This is awesome! LGTM

PR #1512, opened 2021-07-13
[RFC] Host and Hostname fields - Stage 0

hadadata59:
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/master/CONTRIBUTING.md)? yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/master/rfcs/README.md)? yes

### DIFF
--- a/rfcs/text/0000-host-and-hostname-fields.md
+++ b/rfcs/text/0000-host-and-hostname-fields.md
@@ -104,7 +104,7 @@ e.g.:
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
-* Stage 0: https://github.com/elastic/ecs/pull/NNN
+* Stage 0: https://github.com/elastic/ecs/pull/1512
 
 <!--
 * Stage 1: https://github.com/elastic/ecs/pull/NNN


rfcs/text/0000/agent.yml
```diff
@@ -0,0 +1,7 @@
+- name: agent
+  fields:
+    - name: hostname
```
ebeahan, line 3: The `agent.*` fields are meant to describe the software entity collecting events on a host or observer. As a software entity, the `agent.hostname` field has been left out intentionally since the hostname is instead an attribute of a `host.*` or an `observer.*`.

rfcs/text/0000/destination.yml
```diff
@@ -0,0 +1,7 @@
+- name: destination
+  fields:
+    - name: hostname
```
ebeahan, line 3: With the `source`/`destination`/`client`/`server` field sets, the address value should populate the `.address` field and be duplicated to the appropriate field based on the value:`.ip` for IP addresses, `.domain` for FQDNs or hostnames.

I believe the `.domain` field serves the same function you're proposing here. Or do you have different motivations for proposing this addition?

rfcs/text/0000/host.yml
```diff
@@ -0,0 +1,25 @@
+- name: host
+  fields:
+    - name: model
```
ebeahan, line 3: Capturing these types of inventory attributes has come up in past ECS discussions. One pitfall to avoid would be limiting them to certain field sets that wouldn't allow them to describe a broader range of assets someone might have in their inventory or CMDB.

Examples could be power supplies, generators, or server racks. These items would still have `model`, `manufacturer`, or `serial_number` attributes to capture but wouldn't necessarily still be considered hosts in the ECS sense of a host.

In past brainstorming, the idea of creating an `inventory.*` or `asset.*` field set has been suggested, but I think that idea would best be discussed as its own RFC.

rfcs/text/0000-host-and-hostname-fields.md
```diff
@@ -0,0 +1,112 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 107:
```suggestion
* Stage 0: https://github.com/elastic/ecs/pull/1512
```

rfcs/text/0000-host-and-hostname-fields.md
```diff
@@ -0,0 +1,112 @@
...
+<!--
+As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
+Feel free to remove these comments as you go along.
+-->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+Include host information (serial number, manufacturer, and model), bring hostname across other applicable fieldsets.
```
ebeahan, line 16: This summary is sparse.

Can you describe the motivation, purpose, and/or impact of these proposed changes a bit more?

ebeahan (commented):
Thanks for submitting this stage 0 RFC, @hadadata59!

I left some initial feedback.

rfcs/text/0000/agent.yml
hadadata59: We have seen evidence of records (observer.*) which report on a host (host.*) and regarding the agent (agent.*) where the hostnames of each (observer, host, and agent) are unique.

rfcs/text/0000/destination.yml
hadadata59: Using the example of _somehost.example.com_ as a fully qualified domain name: 
where some _somehost_ is the hostname 
and _example.com_ is the domain name
See https://docs.microsoft.com/en-us/windows/win32/dns/naming-conventions
There is general confusion when the FQDN can be both .domain as well as host.name.  
The point here is to isolate the hostname (I.E. _somehost_), as well as the domain (._domain_) for a more accurate and reliable representation of the data and for ease of user search.

rfcs/text/0000/host.yml
hadadata59: I am not sure I understand the distinction you're trying to make here.  We have other lower level objects that are disimmilar (.name).  Not sure why using a .model for host would preclude using a .model with a different description for another object would be problematic?  The inventory. or asset. concept is still talking about an entity (a host), so it would be nice for context but it would also make my search problematic (E.G. if I needed to see a SW inventory on a host).  In that case the fact that the scan came from tenable or my EDR host module would be the indicator that it was a point in time inventory vice an event with an associated host.

rfcs/text/0000-host-and-hostname-fields.md
hadadata59: These fields represent fields not currently represented in the ECS standard that we see in our data sources Tenable SC, Tanium, etc.) that we believe are important enough to create an ECS field to capture for our ELK Siem users.

rfcs/text/0000-host-and-hostname-fields.md
hadadata59: Got it.  Thank you!

rfcs/text/0000-host-and-hostname-fields.md
hadadata59: These fields represent fields not currently represented in the ECS standard that we see in our data sources Tenable SC, Tanium, etc.) that we believe are important enough to create an ECS field to capture for our ELK Siem users.

ebeahan: Thanks for the context. Can you capture that in the proposal doc?

rfcs/text/0000/agent.yml
ebeahan: Do you have an example you'd be willing to share for this discussion?

hadadata59: @ebeahan 
Here is an anonymized representation of the event output from the agent (represented with 123) reporting on the HOST and the other agent deployed on the host (321). 
|application.reporting_agent.assetid | GUID123_HOSTGUID|
|--- | ---|
|application.reporting_agent.endpoint.log.level | INFO|
|application.reporting_agent.endpoint.product.version | 2.7|
|host.broker_guid | BROKERGUID|
|host.domain | HOSTNAME.DOMAIN.COM|
|host.resident_agent.name | HOSTNAME123|
|host.hostname | HOSTNAME|
|host.id | HOSTGUID|
|host.ip | HOSTIP|
|reporting_agent.server | HOSTNAME.DOMAIN.COM|
|reporting_agentguid | GUID123|
|event.resident_agent.version | 10.4|
|host.resident_agent_server.guid | GUID321|
|host.resident_agent_server.name | "HOSTNAMESVR321"|
|reporting_agent.server.name | "HOSTNAMESVR123.DOMAIN.COM"|

rfcs/text/0000-host-and-hostname-fields.md
hadadata59: Done

rfcs/text/0000/agent.yml
kgeller: Hi @hadadata59! Just catching up here and want to verify I understand correctly. 

From your example, you have agent 123 and agent 321 and both run on the same host that they are monitoring?

rfcs/text/0000/host.yml
kgeller: I think what @ebeahan was trying to say was that in ECS, we describe host broadly as a 'general computing instance' meaning it can be anything from hardware to virtual machine to docker container, etc. The intention with a `inventory.*` or `asset.*` would be specifically a physical item we want to keep track of.

rfcs/text/0000/destination.yml
kgeller: I think this addition can make sense here.

jamiehynds:
@melissaburpo this RFC may be relevant for mappings osquery host data. Is there someone on your team that could provide feedback as to whether the fields could be leveraged by osquery?

melissaburpo:
Hi @jamiehynds - I'll take a look, but @aleksmaus or @james-elastic from our team may have some input as well. Thanks for the ping!

melissaburpo:
I think these three proposed fields in particular map well to values that are retrievable via osquery from a query to the [system_info](https://osquery.io/schema/5.0.1#system_info) table:

| Proposed ECS field  | Osquery possible mapping | Example value |
| ------------- | ------------- | ------------- |
| host.model | system_info.hardware_model | MacBookPro16,2 |
| host.manufacturer | system_info.hardware_vendor | Apple Inc. |
| host.serial_number | system_info.hardware_serial | C02F3... |

Regarding `hostname`: we currently populate `host.hostname` for osquery results, so I'm not sure if the proposed hostname fields -- `agent.hostname, destination.hostname, source.hostname` -- would be specifically needed for mapping osquery host data, but it is possible these could be used for mapping values from other host fields; it depends on the use case. There are around 12 osquery tables that include hostname/host fields (per a quick search through the [schema](https://osquery.io/schema/5.0.1)).

rfcs/text/0000/agent.yml
hadadata59: @kgeller yes.  two agents, one host.

jamiehynds:
`host.os.codename` and `host.containerized` were suggested in [this issue](https://github.com/elastic/ecs/issues/75). Worth adding to this RFC?

rfcs/text/0000/agent.yml
ebeahan: What are thoughts on leveraging the existing `agent.name` to capture the hostname?

For example, Beats does this as the default, unless overridden in the configuration: https://github.com/elastic/beats/pull/18000

rfcs/text/0000/host.yml
ebeahan: As mentioned in https://github.com/elastic/ecs/pull/1512#issuecomment-952973774, I like the symmetry of using `product` over `model` to match the existing `observer.product` field naming.

ebeahan:
I've added comments in-line to the open conversation threads, but I wanted to summarize the discussion in a single place.

### Additional `host.*` fields

Regardless of specific field naming and placement, there seems to be agreement around adding fields to capture a host's serial, manufacturer/vendor, and model/product.

We can continue to discuss naming and placement in the PR for stage one.

### Adding `.hostname` fields

I've shared past conversations and my own hesitation about adding (or re-adding) `.hostname` fields across other applicable fieldsets (`source`, `destination`, `client`, `server`, `agent`).

However, we can note this under the `Concerns` section and make sure we revisit in stage 1.

### Next steps

These are the steps I've identified before merging:

- [ ] Verify there's agreement around adding fields to capture a host's serial, manufacturer/vendor, and model/product.
- [ ] If the `.hostname` additions remain in the proposal, note past discussion in the `Concerns` section of the document.

rfcs/text/0000/agent.yml
hadadata59: This is a single log record (not sets of logs) where the box (host.hostname), the non-reporting 'resident' agent (?) and the reporting agent (agent.hostname) all provide unique hostnames in the record.

rfcs/text/0000/host.yml
```diff
@@ -0,0 +1,25 @@
...
+      type: keyword
+      level: extended
+      short: Model of the host.
+      example: "Latitude 5580"
+      description: >
+        The model associated with the host.
+
+    - name: manufacturer
```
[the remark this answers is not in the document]
hadadata59: we have been using vendor for software and manufacturer for hardware.  However, I do not see a reason to object.

djptek:
Perhaps we need to also update the docs, with an elaboration, see mention of `host.hostname`

https://github.com/elastic/ecs/blob/main/docs/using-guidelines.asciidoc#guidelines-for-field-names

ebeahan:
Hi @hadadata59, did you see the summary of the next steps in https://github.com/elastic/ecs/pull/1512#issuecomment-955053444 to move this proposal forward?

Issue #899, opened 2020-07-27
Align input options for `--include` and `--subset` arguments


ebeahan:
#### Summary

There is inconsistency between the expected inputs for the `--include` and `--subset` arguments:

`--include`: Expects a single directory containing the YAML schema files
`--subset`: Expects a single subset definition YAML file. Will accept a directory but execution fails with an exception.

#### Proposal

The documentation (#893) has been improved to clarify, but the difference does cause confusions for users (#843). Propose adjusting each feature to accept both individual files or directories as valid inputs.

djptek:
@ebeahan can I take this one?

ebeahan:
@djptek Absolutely!

djptek:
PR submitted https://github.com/elastic/ecs/pull/1519

ebeahan:
Resolved via #1519

PR #1519, opened 2021-07-15
Align input options for --include and --subset arguments

djptek:
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? Y
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/master/CHANGELOG.next.md)? Y

Addresses https://github.com/elastic/ecs/issues/899


djptek:
In Draft as need to add tests

djptek:
low level tests passing, working with include, exclude, subset. next step add higher level tests

CHANGELOG.next.md
```diff
@@ -10,6 +10,8 @@ Thanks, you're awesome :-) -->
 
 ### Schema Changes
 
+* Align input options for --include and --subset arguments #1519
```
kgeller, line 13: This is more of a tooling addition than a schema change, right?

kgeller (approved):
Looks great Dominic!

CHANGELOG.next.md
djptek: Indeed, thanks for spotting @kgeller

ebeahan:
@djptek, with this change aligning the input behavior of these two arguments, can it be implemented independently from the @timestamp conversation happening in #1646?

djptek:
@ebeahan I think I need to fix the `@timestamp` first in order otherwise this PR will permit non-conformant ECS with no `@timestamp`. I checked the tests and they are valid, it's just a question of order of execution

ebeahan:
> I think I need to fix the @timestamp first in order otherwise this PR will permit non-conformant ECS with no @timestamp

I believe that's already true in the current tooling and not something introduced by these changes?

If this change is ready, we can introduce your improvements here and address #1646 later.

ebeahan:
@djptek should this one be backported to `8.0`?

PR #1521, opened 2021-07-15
Correct RFC stages in the `experimental` README

ebeahan:
The `experimental` README has not been updated since ECS RFC stage 4 was [removed](https://github.com/elastic/ecs/pull/1194).

The ECS experimental schema should contain RFC changes that have made it to stage 1.


djptek (approved):
Thanks for addressing this so rapidly, LGTM!

PR #1533, opened 2021-07-22
Update `log.origin.file.line` from `integer` to `long`

ebeahan:
Migrate `log.origin.file.line` from an `integer` type to a `long` type.

This is considered a breaking change in ECS, but may not affect many users. The index templates generated by Beats and Agent already use `long` for `log.origin.file.line`.


ebeahan:

kgeller (approved):

Issue #1534, opened 2021-07-26
Add service.address field to ECS


mtojek:
**Summary**

* while performing cleanup of ECS fields in Integrations/Beats we noticed that `service.address` is not present there.

**Motivation**:

* The field is a relatively common across different Integrations ([search results](https://github.com/elastic/integrations/search?q=service.address)). Do you think we should include it in the ECS repository?

**Related**:

https://github.com/elastic/integrations/issues/1021
https://github.com/elastic/integrations/pull/1340

cc @djptek

mtojek:
Thanks, @kgeller, for driving this. Would you mind describing what are the next steps to get it versioned and released?

kgeller:
Absolutely @mtojek ! I am backporting it right now to ECS 1.12 , which will be FF in a couple weeks and then released a couple weeks after that.

kgeller:
Implemented for ECS 1.12 , closing this issue

PR #1537, opened 2021-07-27
Adding service.address

kgeller:
https://github.com/elastic/ecs/issues/1534


schemas/service.yml
```diff
@@ -105,3 +105,13 @@
 
         This id normally changes across restarts, but `service.id` does not.
       example: 8a4f500f
+
+    - name: address
+      level: extended
+      type: keyword
+      short: Address of this service.
+      description: > 
+        Address where this service is being run from.
```
andrewkroh, line 114: Maybe instead of "being run from", either "listening" or "hosted". Not sure...

schemas/service.yml
```diff
@@ -105,3 +105,13 @@
...
+
+        This could be a URL or a filepath.
```
andrewkroh, line 116: Under what circumstances would this be a file path? Is it for a service listening on a unix socket or named pipe?

schemas/service.yml
```diff
@@ -105,3 +105,13 @@
...
+      example: localhost:8000
```
andrewkroh, line 117: With my [rfc1738](https://datatracker.ietf.org/doc/html/rfc1738#section-5) glasses 👓 on this example doesn't appear to be fully formed URL. Perhaps the definition needs clarified like the value must be a URI, absolute file path, or network address (ipv4:port or [ipv6]:port).

andrewkroh (commented):
I think the definition would benefit from some additional clarification.

schemas/service.yml
mtojek: exposed, available?

schemas/service.yml
mtojek: Kylie is right here: [kvm](https://github.com/elastic/beats/blob/23e4403ae093fcc8f7905345cad2c7ad256976d8/metricbeat/module/kvm/status/_meta/data.json#L20)

schemas/service.yml
mtojek: Agree with Andrew, it would be safer to mention a network address or resource path (sockets).

schemas/service.yml
kgeller: updated!

schemas/service.yml
kgeller: I think the description should be better now after updating it per the comment below. If you think there's still more to explain, lmk

schemas/service.yml
```diff
@@ -105,3 +105,13 @@
...
       example: 8a4f500f
+
+    - name: address
+      level: extended
+      type: keyword
+      short: Address of this service.
+      description: > 
+        Address where this service is hosted.
+
+        This could be a URI, network address (ipv4:port or [ipv6]:port) or a resource path (sockets).
```
andrewkroh, line 116:
```suggestion
        This should be a URI, network address (ipv4:port or [ipv6]:port) or a resource path (sockets).
```

I'd go with "should" since I'm a fan of the definitions in https://datatracker.ietf.org/doc/html/rfc2119#section-3.

schemas/service.yml
ebeahan: What about something along the lines of `Address where data about this service was collected from`?

If the intent of `service.address` is to capture the address of the service's monitoring endpoint or path, I think we want to avoid confusion with a service's client-facing ports that service functions other than monitoring and reporting.

schemas/service.yml
ebeahan: ++

schemas/service.yml
kgeller: That sounds perfect, updating

andrewkroh (approved):

mtojek (approved):
Thanks!

CHANGELOG.next.md
```diff
@@ -9,6 +9,11 @@ Thanks, you're awesome :-) -->
 ## Unreleased
 
 ### Schema Changes
+
+#### Added
```
ebeahan, line 13: I hadn't noticed this section header was missing, and I've been putting my schema changes under `Tooling`! 😅

Thanks for catching it! We'll make sure the other sections are cleaned up.

ebeahan (approved):
LGTM 👍

jsoriano (approved):
Nice, thanks!

PR #1548, opened 2021-08-03
Update event.kind allowed value descriptions to clarify usage of "alert" and "signal"

MikePaquette:
We have received questions from community members about when to use `event.kind:alert` vs. `event.kind:signal`

These allowed values can cause confusion, since documents that contain `event.kind:signal` are often referred to as "alerts."

As background, the overloading of the term "alerts" is a longstanding issue in security information and event management, where there was frequently confusion between external alerts generated by some external system and alerts generated by the SIEM itself.  Any usage of the unmodified "alerts" was thus ambiguous.

The proposed language attempts to provide clear guidance:
- If the alert is created by a system external to the Elastic Stack,  use `event.kind:alert`
- If the alert is created _*by*_ the Elastic Stack/solution, use `event.kind:signal`

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -59,6 +59,7 @@ Thanks, you're awesome :-) -->
 * Add --exclude flag to Generator to support field removal testing #1411
 * Explicitly include user identifiers in `relater.user` description. #1420
 * Improve descriptions for `cloud.region` and `cloud.availability` fields. #1452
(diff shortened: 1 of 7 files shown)


ebeahan (approved):
LGTM 👍 

Thanks for updating these descriptions, @MikePaquette!

Since these are minor doc updates, let's include these changes into 1.11. I'll handle the backporting.

PR #1557, opened 2021-08-04
Add `code_signature.digest_algorithm` and `code_signature.timestamp` fields

### LINKED ISSUE #1524: Add digest + timestamp to `code_signature` fields
**Summary**
There are more fields we can add to `code_signature`, so we can tell signatures apart and capture more relevant information.

**Motivation**:
Related issue https://github.com/elastic/kibana/issues/106059#issuecomment-882682354

It was noted in https://github.com/elastic/kibana/issues/106059#issuecomment-882682354 that a file can be signed multiple times by the same signer but different digest algorithms. There's no way in ECS to express that, so we could add more fields to `code_signature` to differentiate better between signatures. It's also helpful information for an analyst to know.

**Detailed Design**:
Add two new fields to code_signature

- Digest algorithm, which is a hashing algorithm (md5, sha1, sha256, etc.)
- Timestamp: when the signature was generated and signed

ebeahan:
Introduces two new fields to the `code_signature.*` field set:

* `code_signature.digest_algorithm` - Hashing algorithm used to sign the process.
* `code_signature.timestamp` - When the signature was generated and signed.:w

Closes #1524

### DIFF
--- a/schemas/code_signature.yml
+++ b/schemas/code_signature.yml
@@ -57,7 +57,7 @@
         This is useful for logging cryptographic errors with the certificate validity or trust status.
         Leave unpopulated if the validity or trust of the certificate was unchecked.
       example: ERROR_UNTRUSTED_ROOT
-      
+
     - name: team_id
       level: extended
       type: keyword
@@ -68,7 +68,7 @@
         This is used to identify the team or vendor of a software product.
         The field is relevant to Apple *OS only.
       example: EQHXZ8M8AV
-      
+
     - name: signing_id
       level: extended
       type: keyword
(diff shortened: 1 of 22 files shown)


schemas/code_signature.yml
```diff
@@ -79,3 +79,22 @@
...
+        This value can distinguish signatures when a file is signed multiple times
+        by the same signer but with a different digest algorithm.
+      example: sha256
+
+    - name: timestamp
+      level: extended
+      type: date
+      short: When the signature was generated and signed.
+      description: >
+        Timestamp of when the code signature was generated and signed.
```
rw-access, line 99: "Timestamp of when" feels a little off to me, may be this?
```suggestion
        Date and time when the code signature was generated and signed.
```

rw-access: This sound right @gabriellandau?

gabriellandau: Sure 👍

ebeahan: 👍

kgeller (approved):

rw-access (approved):

Issue #1503, opened 2021-07-06
[Docs] Further explain `user.id` vs. `user.name` on main page


a03nikki:
**Summary**

Users (aka human end users) of the ECS sometimes get confused between `user.id` and `user.name`. 

**Motivation**:

Often in spoken American English the phrases "What is your user ID?" and "What is your user name?" are often used interchangeably in conversation. Then when reviewing the documentation for the [User Fields](https://www.elastic.co/guide/en/ecs/1.10/ecs-user.html) it is not as clear as it could be.  The new [User Fields Usage and Examples](https://www.elastic.co/guide/en/ecs/1.10/ecs-user-usage.html) helps but the initial page could use additional details.

**Detailed Design**:

Current (v1.10) version

> Field | Description | Level
> -- | -- | --
> user.id | Unique identifier of the user. <br/> type: keyword | core
> user.name | Short name or login of the user.<br/> type: keyword<br/> Multi-fields:<br> * user.name.text (type: text) <br/> example: albert | core

It brought confusion because `albert` is also the first name of `Albert Einstein`.  Additionally there is no example of an `user.id`.  The description of `unique identifier` could also mean a `user.name` as well because it uniquely identifies an individual within an organization too.

Clarification of the values and how they should be populated across data sources is important for large organizations who have multiple platforms (Windows, Linux, Public Cloud, Active Directory (AD), etc.) and are using the ECS for data standardization for security analytics (such as standardizing searches across platforms and user entity analytics).

An updated version of the documentation such as this may be helpful for them

> Field | Description | Level
> -- | -- | --
> user.id | Unique identifier of the user. <br/> type: keyword <br/> **example: `1001` or `S-1-5-21-202424912787-2692429404-2351956786-1000`** | core
> user.name | Short name or login of the user.<br/> type: keyword<br/> Multi-fields:<br> * user.name.text (type: text) <br/> example: **`a.einstein`** | core

- The Windows ID was copied from <https://github.com/elastic/ecs/blob/master/rfcs/text/0007-multiple-users.md#windows-privilege-escalation>.
- The `user.name` I updated to a value that is not the first name but rather a derived value still related to the users name.
- I thought about adding "in the context of a system" to `user.id` based on #117 but that may not clarify the usage of user.id any further.

ebeahan:
Thanks for the detailed issue, @a03nikki. This is great feedback, and we'll certainly use this context to improve the descriptions and examples.

[...]

PR #1566, opened 2021-08-16
Update user name & id docs for clarity

kgeller:
https://github.com/elastic/ecs/issues/1503

Adding/updating the user.id and user.name examples to try and provide clarity


ebeahan (approved):
Thanks!

PR #1583, opened 2021-08-20
Remove dangling character from `url.full` field value

ebeahan:


ebeahan:
Fixed by https://github.com/elastic/ecs/pull/1584

PR #1593, opened 2021-08-25
[RFC] Introduce `email` field set - stage 2

ebeahan:
### Summary

Continuing onto Stage 2 with this proposal to introduce the `email.*` field set to the schema.

Stage 2 (Candidate) Criteria:

- [x] Opened pull request for this draft revising the existing proposal
- [x] Included a real-world example source document
- [x] Identifies scope of impact of changes to ingestion mechanisms (e.g., beats/logstash), usage mechanisms (e.g., Kibana applications, detections), and the ECS project (e.g., docs, tooling)
- [x] Subject matter experts weighed in on the technical utility of field definitions in the pull request

[Preview of markdown proposal](https://github.com/ebeahan/ecs/blob/rfc/0010/stage-2/rfcs/text/0010-email.md)


wasserman:
In regards to `Display Name`, you could consider some of these options:
1. Allow for name/address pairs in all email fields instead of just the email address itself.
2. Permit RFC 5322 Email formats like `First Last <first.last@example.com>`.  Then a keyword with a normalizer could index just the emails while still keeping the display names intact.  Of course the names wouldn't be searchable.
3. Multi-fields with some variation of this could work and allow for some form of the `keyword` and `text` fields. 

I personally ran with option 2 for myself.  I was hoping to leverage the `uax_url_email` tokenizer, but I had to settle for a simple regex.
```
"normalizer": {
    "email_normalizer": {
      "type": "custom",
      "char_filter": [
        "email_filter"
      ]
    }
  },
  "char_filter": {
    "email_filter": {
      "type": "pattern_replace",
      "pattern": ".*?<([^>]+)>",
      "replacement": "$1"
    }
  }
```

ebeahan:
Proposed fields now include arrays of objects with both the email address and display name for the `to`, `cc`, and `bcc` recipients. The `display_name` and `address` fields have been added under `email.from.`

| Field name | Data type |
| ---- | ----- |
| `email.from.address` | keyword |
| `email.from.display_name` | keyword |
| `email.to` | nested |
| `email.to.address` | keyword |
| `email.to.display_name` |  keyword
| `email.subject` | keyword |
| `email.cc` | nested |
| `email.cc.address` | keyword |
| `email.cc.display_name` | keyword | 
| `email.bcc` | nested |
| `email.bcc.address` | keyword |
| `email.bcc.display_name` | keyword |

djptek (approved):
LGTM!

As an aside, when this is merged I'd like to use it as an example to update the docs around `nested` type, as the difference between `email.from` and `email.to` illustrates this perfectly

rfcs/text/0010-email.md
```diff
@@ -1,28 +1,30 @@
...
-Stage 1: Describe at a high level how this change affects fields. Which fieldsets will be impacted? How many fields overall? Are we primarily adding fields, removing fields, or changing existing fields? The goal here is to understand the fundamental technical implications and likely extent of these changes. ~2-5 sentences.
--->
-
 ### Email specific fields
 
 | field | type | description |
 | --- | --- | --- |
-| `email.from` | keyword | Stores the `from` email address from the RFC5322 `From:` header field. |
+| `email.from.address` | keyword | Stores the `from` email address from the RFC5322 `From:` header field. |
+| `email.from.display_name` | keyword | Stores the display name of the `from` address. |
```
peasead, line 15: Does this not need a nested field because there is only 1 sender?
```
| `email.from` | nested | Nested object with the message sender |
| `email.from.address` | keyword | Stores the `from` email address from the RFC5322 `From:` header field. |
| `email.from.display_name` | keyword | Stores the display name of the `from` address. |
```

ebeahan: There's been some back and forth discussion about allowing for multiple `From:` addresses.

Looking back at https://www.rfc-editor.org/rfc/rfc5322#section-3.6.2, IMF _does_ permit multiple addresses in the `From:` field. If multiple addresses appear in the `From:` field, the `Sender:` field is also required but will have a single address.

I see a few possibilities:

1. We keep the `email.from.address` and `email.from.display_name` expecting single values, and then the field description for the `email.from.*` fields provides specific guidance: "if the original event includes multiple `From:` addresses, use the value of the `Sender` field instead."
2. Prefix the field names with `email.sender.*` instead of `email.from`. Like no. 1, the description guides users to use the value of `Sender:` if multiple `From:` values are present.
3. Make `email.from.*` expect a list of objects using `nested` to match `to`, `cc`, and `bcc`. Optionally, discuss if adding `email.sender.*` fields is necessary to discern the `From:` field addresses from the `Sender` field address.

djptek: > 3. Make email.from.* expect a list of objects using nested to match to, cc, and bcc

I think we should use 3, as it most closely tracks the standard. From the upstream implementation point of view it doesn't add much as it will use the same logic as `to`.

Possibility 1 is incomplete since `more than one` could be > 2, therefore adding a sender field will not cover all cases.

Possibility 2 is confusing to me

> 3. adding email.sender.* fields is necessary

Yes, though these _would_ be single value. Necessary because I have worked on this use case in Legal Search, so I know it exists

djptek: Opened a PR with these suggestions here: https://github.com/ebeahan/ecs/pull/3

peasead: > Make email.from.* expect a list of objects using nested to match to, cc, and bcc
> I think we should use 3, as it most closely tracks the standard. From the upstream implementation point of view it doesn't add much as it will use the same logic as to.

I agree with this, I actually had a `suggestion` queued up for this exactly, but wanted to check first.

ebeahan: Great, thanks all!

I cherry-picked the changes to include here: https://github.com/elastic/ecs/pull/1593/commits/0133a74574df3c2b306e8744aaf343dcb45f7664

ebeahan:
The limitations around building visualizations using type `nested` fields make me question if using `nested` for the various email sender/recipient fields is the best direction.

I'm going to think this over a bit more and iterate on the proposed fields.

djptek:
@ebeahan re `nested` type - for the legal use case `nested` type is not necessary as that would center on `*.address`

Regarding visualisation of emails, we'd probably need a Chord or Sankey diagram - a quick look in Kibana Issues does not have those on the radar at this point in time though there is a Sankey example using Vega on the Blog - again this would probably center on `*.address`. so not `nested`.

There might also be value in running address and display name through ML, that would probably be incompatible with `nested` type

The only use case off the top of my head where we might want `nested` might be Spoof detection, however, given the cardinality this would probably be best done using ML for the heavy lifting and then manual inspection of anomalies, so you could work around that

@peasead Do you have a specific use case/query/aggs in mind where we'd need to leverage `nested` type?

ebeahan:
@jamiehynds as the sponsor, can you take a look at how the `email.*` fields proposal is shaping up?

peasead:
> @peasead Do you have a specific use case/query/aggs in mind where we'd need to leverage `nested` type?

Thanks for your patience.

@djptek I don't have anything specific. I wasn't sure if there'd be a use case to query nested objects independently, but thinking more, I'm not sure that'd be needed.

wasserman:
FYI, nested was just thought to be a useful way to preserve the relationships between display names and emails.  Multi-fields or a normalizer could work too.  Ultimately any smart way to lot lose the display names in the process since it could be valuable just to be able to see, if nothing else.

rfcs/text/0010-email.md
```diff
@@ -94,12 +82,16 @@ Stage 2: Included a real world example source document. Ideally this example com
...
   "email": {
     "timestamp": "2020-11-08T22:12:34.8196921Z",
-        "from": [
-  	    "o365mc@microsoft.com"
-  	],
-    "to": [
-      "john@testdomain.onmicrosoft.com"
-    ],
+     "from": {
+  	    "address": [
```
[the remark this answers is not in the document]
djptek: pushed a fix

djptek (approved):
LGTM!

djptek:
Thanks @wasserman 

> nested was just thought to be a useful way to preserve the relationships between display names and emails

Looking at one use case that leverages this relationship, e.g. checking for spoofing of a known `address`, we'd need to involve additional data - a table/index defining the a priori relationship between a defined and countable set of `address`(es) and the legitimate `display_name`(s) for these `address`(es). This would require additional logic, including a join against that table/index. Elasticsearch joins are generally best implemented at ingest time rather than query time, so this use case could perhaps be addressed by building this into the ingest pipeline, or by reindexing a subset of data related to specific `address`(es) of interest.

Conversely, where the relationship between `display_name` and `address` is not explicitly defined a priori, there is no upper limit to the number of `address`(es) in the related use cases so it may be preferable to avoid `nested_type` to ensure the most performant solution.

ebeahan:
While implementing the `email.*` field set into the schema, reusing `hash.*` at `email.attachments.file.hash.*` felt like a better approach that aligns with the existing `file.hash.*` fields and adds a few additional hash fields for any attachment: https://github.com/elastic/ecs/pull/1688#discussion_r769004415.

I'll capture these details fully in the proposal doc during stage 3.

PR #1602, opened 2021-09-07
Threat indicators schema fix

kgeller:
https://elastic.slack.com/archives/C8L4TMJGJ/p1631024860024500

Two things:
1) @P1llus realized we were missing `hash.*` under `threat.indicator.file` and `threat.enrichments.indicator.file' . I realized this was due to an ordering issue on our field reuses. We solve this issue the same way [here](https://github.com/elastic/ecs/blob/master/schemas/group.yml#L12) for 'group' at 'user'.

2) In attempting to debug ^, I noticed we were adding `hash` directly under `threat.indicator` and `threat.enrichments.indicator' when it is intended to be used via the nesting at `file', so this removes that incorrect nesting.


kgeller:
Split into https://github.com/elastic/ecs/pull/1603 and https://github.com/elastic/ecs/pull/1604

PR #1603, opened 2021-09-08
Hash ordering fix

kgeller:
@P1llus realized we were missing `hash.*` under threat.indicator.file and `threat.enrichments.indicator.file' like the [threat RFC](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#using-existing-ecs-fields-to-store-indicator-information) describes. I realized this was due to an ordering issue on our field reuses. We solve this issue the same way here for 'group' at 'user'.


schemas/hash.yml
```diff
@@ -17,6 +17,7 @@
 
   reusable:
     top_level: false
+    order: 1
```
rylnd, line 20: Ooh; is this a tie-breaking mechanism?

rylnd: Duh, should've read the PR description...

epixa (approved):
LGTM

PR #1604, opened 2021-09-08
Threat implementation reuse correction

kgeller:
In the debugging of another issue, I noticed a discrepancy in the implementation of an RFC. 

[Here](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#using-existing-ecs-fields-to-store-indicator-information) under Hash, you'll see we intend to reuse under `file` and `process`, but we implemented as reuse under `threat.indicator`.


rylnd:
@kgeller I think the deletions make sense, but I was also expecting to see the addition of hash fields to `threat.indicator.file` and `threat.enrichments.indicator.file`; is that going to be done here or in a separate PR?

kgeller:
> @kgeller I think the deletions make sense, but I was also expecting to see the addition of hash fields to `threat.indicator.file` and `threat.enrichments.indicator.file`; is that going to be done here or in a separate PR?

Separate PR, just to keep it a little more organized. See https://github.com/elastic/ecs/pull/1603

rylnd (approved):
Deletions here LGTM! We have https://github.com/elastic/ecs/pull/1603 for the necessary additions, as well.

One related issue I noticed: we may have the same problem with the `pe` fields, where those are typically nested under `indicator.file` and not as top-level `indicator.pe` fields, as the schema currently suggests. We can pick that up in another PR as needed, though.

:shipit: !

epixa (approved):
LGTM

PR #1605, opened 2021-09-09
Pe ordering fix

kgeller:
Just like in https://github.com/elastic/ecs/pull/1603, we realized we were missing `pe.*` under threat.indicator.file and `threat.enrichments.indicator.file' like the [threat RFC](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#using-existing-ecs-fields-to-store-indicator-information) describes. I realized this was due to an ordering issue on our field reuses. We solve this issue the same way here for 'group' at 'user'.

Sibling PR to remove the incorrect nestings https://github.com/elastic/ecs/pull/1606


generated/ecs/ecs_nested.yml
```diff
@@ -14448,6 +14624,13 @@ threat:
   - threat.indicator.x509
   prefix: threat.
   reused_here:
+  - full: threat.indicator.pe
```
rylnd, line 14627: I think we want these gone, right? `pe` will be nested under `file` or `process`, but not as a sibling

rylnd (changes requested):
I think we should delete `pe` as a top-level field on `indicator`; otherwise this looks good to me.

rylnd (approved):
Disregard the last review, it looks like you already made my proposed changes in #1604. :shipit: 

I appreciate the logically separated PRs; I do not envy merging them back together, though!

epixa (approved):
LGTM

PR #1606, opened 2021-09-09
Threat implementation reuse correction part 2

kgeller:
Just like https://github.com/elastic/ecs/pull/1604 , we are also correcting pe. [Here](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md#using-existing-ecs-fields-to-store-indicator-information) under Pe, you'll see we intend to reuse under `file` and `process`, but we implemented as reuse under `threat.indicator`. 

Sibling PR to fix the ordering and get pe nested under threat ... file https://github.com/elastic/ecs/pull/1605


epixa (approved):
LGTM

Issue #1620, opened 2021-10-05
ECS - Missing fields, possible nesting issue


dstepanic17:
**Summary**

There are some missing ECS fields related to the threat intelligence features in Elastic Stack.  Is it possible to add these into ECS? I believe it might be related to a previous issue around grouping and ordering that was recently fixed.  Since the packages map fields to the generated ECS, the CI components are failing.  

```
threat.indicator.file.x509.*

threat.indicator.file.hash.sha384
threat.indicator.file.hash.tlsh
threat.indicator.file.hash.pehash
```

kgeller:
@dstepanic17 thanks so much for adding this! 

I will work on getting the `x509` field set corrected. In regards to `threat.indicator.file.hash.sha384` and `threat.indicator.file.hash.tlsh` , I don't see those two fields in the hash field set, are you sure they are formally in ECS? [schema for hash](https://github.com/elastic/ecs/blob/master/schemas/hash.yml)

dstepanic17:
Thanks, I guess those two fields are not currently in ECS.  Is it possible to add those two fields in? Should I create a PR to do that?  They would mimic the other hash [entires](https://github.com/elastic/ecs/blob/master/schemas/hash.yml), but with their own respective names.

kgeller:
Can absolutely get those added.

Normally for schema changes, we recommend going through the [RFC process](https://github.com/elastic/ecs/tree/master/rfcs) as it allows us to fully vet the proposal. However these fields appear straightforward, so I am inclined to say we can simply add them via a PR. What do you think?

dstepanic17:
Thanks, yeh I was thinking the same thing. The additions seem clear-cut so if it's possible to skip the RFC process that would be great.

P1llus:
threat.indicator.file.hash.pehash also seems to be missing @dstepanic17

kgeller:
Awesome @dstepanic17 , I think we can move forward with those 3 new field additions (counting in the additional from @P1llus ). 

I will separately correct the nesting for `x509`.

rylnd:
Just updating with notes from today's CTI Initiative sync: if `pehash`, `tlsh` and `sha384` are being populated by threatintel modules, we shouldn't add mappings for `threat.indicator` until these fields are part of ECS. If the aforementioned PR isn't merged when those modules ship, we should either:

* populate `threat.indicator.file.hash.{pehash,tlsh,sha384}` without a corresponding mapping
[...]

PR #1621, opened 2021-10-06
Nesting fix x509

kgeller:
https://github.com/elastic/ecs/issues/1620

It was noticed that we were missing some reuse nestings due to an ordering issue.


rylnd (approved):
LGTM 👍

PR #1627, opened 2021-10-12
Stage one of cgroup RFC

fearful-symmetry:
This is the Stage 1 component of the RFC process. Sorry for the delay between the stages, I got a bit swept up with other work for a bit.

_Hopefully_ the proposed/example `cgroup.yml` file is correct, I wasn't completely sure how things should be formatted.


kgeller:
> @fearful-symmetry maybe you should add us to the "People" section of the RFC.

That would be awesome! Otherwise I can add that in when getting everything ready for merge.

> I think that's an ECS philosophy question, and I'm not sure of the answer. My thinking was to include fields that are the most likely to be used for monitoring and alerting, and not to cram in everything that I technically could.

I think this is perfect. Adding in those most-likely-to-be-used fields is exactly what we want. I just had hesitation that we were only going to be adding the fields convenient from v1 & v2 overlap, but that is not the case, so yay!

---

@fearful-symmetry @jsoriano thank you both for the insight into where these fields should sit. I totally see where you're coming from and can see `cgroups` getting their own top level field set. If @kaiyan-sheng has other inventory ideas from the work with other metrics, that would be cool to get insight on as well.

kaiyan-sheng:
@jsoriano @fearful-symmetry @kgeller We currently have these basic monitoring metrics (like CPU, disk...) added under `host`, and we are also working on adding them under `container`. Will it be good to have the common fields defined as separate field sets and reuse them if it's applicable?

fearful-symmetry:
Made a couple of the suggested changes. Apologies for the delay, I got sucked into bug squashing again.

As far as merging/combining this with other fields and metrics, my concern is the weird way that cgroups connect to other things. A cgroup can be linked to a single process, a group of processes, a container, etc. We risk "over linking" the data if we start connecting things to certain cgroups without being 100% sure that that connection makes sense.

fearful-symmetry:
@kgeller anything else we need here?

kgeller:
@fearful-symmetry I think we are all set from an ECS perspective, and are good to merge as long as your sponsor(s) / SME(s) are on board!

jsoriano (approved):
All good from my side, thanks!

PR #1633, opened 2021-10-20
Beats: Add default_field: true to fieldsets

adriansr:
For 8.0, we are changing Beats to have an implicit `default_field: false` for all fields.

This updates the generator to add `default_field: true` to ECS fieldsets so that they continue to be added to the default field setting in the index template.

Relates to https://github.com/elastic/beats/issues/28215

### DIFF
--- a/scripts/generators/beats.py
+++ b/scripts/generators/beats.py
@@ -23,7 +23,8 @@ def generate(ecs_nested, ecs_version, out_dir):
             continue
 
         beats_field = ecs_helpers.dict_copy_keys_ordered(fieldset, allowed_fieldset_keys)
-        beats_field['default_field'] = True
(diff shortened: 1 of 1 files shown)


experimental/generated/beats/fields.ecs.yml
```diff
@@ -427,6 +429,7 @@
       description: Array of user roles at the time of the event.
       example: '["kibana_admin", "reporting_user"]'
       default_field: false
+    default_field: true
```
kgeller, line 432: There's quite a few of places where `default_field: true` gets inserted directly below a `default_field: false`. I assume this isn't what you're intending?

kgeller: Oooh I see, that is clearer. Thanks for that explanation!

PR #1655, opened 2021-11-09
Remove `yamllint` dependency

ebeahan:
The `yamllint` dependency is only used for `make yamllint`, and `make yammlint` is not enforced in any build step (e.g., `make check`). Also, I suspect most folks rely on text editors for equivalent functionality, and it's unnecessary to include it as a dep.

Also adding comments noting the associated licenses for each existing dependency in the Pipfile.


djptek (commented):
I ran `make yamllint` and am not convinced that contributors are submitting well-formatted YAML, which is a concern.

I opened an issue with options on how we might address that https://github.com/elastic/ecs/issues/1662 

I also submitted a PR with fixes to the Yaml, and an alternative configuration for Yamllint
https://github.com/elastic/ecs/pull/1661

djptek:
Thanks @ebeahan for walking me through the rationale, this is a great idea, LGTM!

djptek (approved):
LGTM!

ebeahan:
Going to propose a different direction in a new PR.

PR #1660, opened 2021-11-10
RFC 0028 Stage 1 - Add `cgroup.*` fields to experimental schema

ebeahan:
Add the proposed fields from [RFC 0028](https://github.com/elastic/ecs/blob/main/rfcs/text/0028-cgroups.md) stage 1 to the experimental ECS schema. (#1627 )


djptek (approved):
LGTM

matschaffer:
Just a heads up that these might need revision per https://github.com/elastic/beats/issues/31765

I don't know all the history but it looks like cgroup memory limit/usage can be larger than the bounds of `long`.

PR #1661, opened 2021-11-11
Tidy up trailing spaces and Newlines in Schemas

djptek:
I ran `make yamllint` after reading @ebeahan 's comments in https://github.com/elastic/ecs/pull/1655

Unfortunately, the Yaml files in `schema/*.yml` had a lot of issues:
- trailing spaces
- incorrect indentation
- excessively long lines

I fixed the trailing spaces and indentation & modified the `yamllint` config to tolerate the excessively long lines, as to fix those (or not) would require an RFC 

I'm most concerned about the indentation as that could potentially break a 3rd party parser & therefore be a blocker to ECS adoption

@ebeahan @kgeller 
see PR: https://github.com/elastic/ecs/pull/1655
and issue:  https://github.com/elastic/ecs/issues/1662


djptek:
After discussion I agree it is best to follow the approach in #1655 so I have rolled back the change to the config to facilitate this

PR #1670, opened 2021-11-18
RFC 0016 extend `pe.*` field set - Set to Stage X

ebeahan:
The original authors and sponsors aren't pursuing this effort at this time.

I am also removing the additional `pe.*` fields from the `experimental` schema.


djptek (approved):
LGTM!

PR #1672, opened 2021-11-19
Separate packages required for testing/linting

ebeahan:
Most uses do not need to install the dependencies required for testing and lining, and this change splits `requirements.txt` into `requirements.txt` and `requirements-dev.txt`.

`make ve` has been updated to include the additional file, and I updated the repo docs as well. I thought it prudent to simplify our install instructions in `USAGE.md`, but I'm open to discussion if anyone feels differently.


kgeller (approved):
LGTM!

PR #1673, opened 2021-11-19
Expand `[source|destination|client|server].domain` field descriptions

### LINKED ISSUE #1663: ECS does not describe source.domain and destination.domain
In https://github.com/elastic/ecs/issues/84, there was a good conversation about what `source.domain` and `destination.domain` mean. However, today, the documentation for `destination.domain` reads:

> Destination domain.

This field (and `source.domain`) is not documented sufficiently. If someone can explain to me what it is supposed to be (is it just FQDN?), I can add an example to the documentation.

ebeahan:
Revisit the description for the `.domain` fields for `source.*`, `destination.*`, `client.*`, and `server.*` field sets.

Also, add `example` and `short` attributes for each field.

Fixes #1663

### DIFF
--- a/docs/field-details.asciidoc
+++ b/docs/field-details.asciidoc
@@ -355,7 +355,7 @@ example: `184`
 
 | The domain name of the client system.
 
-This value may be a host name, a fully qualified domain name, or other host naming format. The value may from the original event or added from enrichment.
+This value may be a host name, a fully qualified domain name, or other host naming format. The value may derive from the original event or be added from enrichment.
 
 type: keyword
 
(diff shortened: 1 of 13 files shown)


docs/field-details.asciidoc
```diff
@@ -353,13 +353,15 @@ example: `184`
 [[field-client-domain]]
 <<field-client-domain, client.domain>>
 
-| Client domain.
+| The domain name of the client system.
 
-type: keyword
+This value may be a host name, a fully qualified domain name, or other host naming format. The value may from the original event or added from enrichment.
```
djptek, line 358: `The value may from the original event ` might be better phrased as 

The value may derive from the original event `

ebeahan:

djptek (approved):
LGTM!

PR #1674, opened 2021-11-24
[RFC] enforce single-valued fields

markharwood:
Strawman RFC to discuss proposed elasticsearch feature that rejects documents with arrays when single values required.


djptek:
Looks a great start @markharwood I'll merge so we can get the ball rolling

djptek (approved):
LGTM!

markharwood:
@djptek Do you need to assign me an RFC number so I can rename the file before merging?

PR #1678, opened 2021-12-01
Added new hash fields

dstepanic17:
- If submitting code/script changes, have you verified all tests pass locally using `make test`? no script changes, only schema
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? yes

Reference: https://github.com/elastic/ecs/pull/1678

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -57,6 +57,7 @@ Thanks, you're awesome :-) -->
 #### Added
 
 * Added `faas.*` field set as beta. #1628
+* Added three new fields to hash schema. #1678
 
 #### Improvements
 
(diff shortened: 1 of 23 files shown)


kgeller (approved):
LGTM!

ebeahan:
For the `hash.*` field set, ECS lists this guidance:

> Note that this fieldset is used for common hashes that may be computed over a range of generic bytes. Entity-specific hashes such as ja3 or imphash are placed in the fieldsets to which they relate (tls and pe, respectively).

I'm not very familiar with either `pehash` or `tlsh`; are either of those entity-specific hash types?

dstepanic17:
Yeh, so `TLSH` is a fuzzy matching hash used for similarity comparisons and is similar to `ssdeep` which is currently in the `hash.*` field set. But for `pehash`, it sounds to me like it fits more of the entity-specific hash type.  I will go and move that under the `pe` schema.

PR #1688, opened 2021-12-13
RFC 0010 `email.*` field set - stage 2 changes

ebeahan:
Adding `email.*` as a `beta` field set with [RFC 0010](https://github.com/elastic/ecs/blob/master/rfcs/text/0008-threat-intel.md) now advanced to [stage 2](https://github.com/elastic/ecs/pull/1593).


schemas/hash.yml
```diff
@@ -38,6 +38,7 @@
       - file
       - process
       - dll
+      - email.attachments.file
```
ebeahan, line 41: While preparing these changes, I reflected more on what was proposed in https://github.com/elastic/ecs/pull/1593 for the `email.attachments.hash` fields.

I think it makes better sense to follow the existing convention of `file.hash.*` and nest `hash.*` at `email.attachments.file.hash`. This adds a few additional hash types and keeps better symmetry with other places in the schema.

I still think fully nesting `file.*` at `email.attachments.*` is unnecessary since many of the `file.*` fields (access time, path, inode information, fork_name) likely won't apply to an email attachment.

If this change moves forward, I'll include it into the stage 3 RFC PR.

djptek (approved):
Looks OK to me

Issue #1693, opened 2021-12-16
Mark `process.ppid` as deprecated in 1.12


ebeahan:
The `process.ppid` should be marked as deprecated in the ECS 1.12 docs in favor of `process.parent.pid`.

`process.ppid` is being removed in ECS 8.0, and we should make users aware of this in the ECS 1.x docs for the field.

ebeahan:
Resolved in https://github.com/elastic/ecs/pull/1699

kgeller:
Due to issues described #1701 , revering the change in #1699 with #1702

PR #1699, opened 2021-12-20
Ppid deprecation notice

kgeller:
Resolves https://github.com/elastic/ecs/issues/1693

`process.ppid` is marked as deprecated in ECS 1.12 in favor of `process.parent.pid`


schemas/process.yml
```diff
@@ -61,7 +61,10 @@
       format: string
       level: extended
       type: long
+      short: Deprecated parent process' pid.
       description: >
+        Deprecated for removal in next major version release. This field is superseded by `process.parent.pid`.
```
ebeahan, line 66:
```suggestion
        Deprecated for removal in the next major version. This field is superseded by `process.parent.pid`.
```

I think we can drop the word `release` and keep the same meaning.

kgeller: Works for me!

ebeahan (approved):
LGTM

PR #1711, opened 2022-01-05
Beats: fix default_field setting for root fields

adriansr:
The generator was adding the `default_field` setting to non-root fieldsets, causing top-level fields (like `message`) to be excluded from the `default_field` index template configuration.

Relates https://github.com/elastic/beats/issues/29633

### DIFF
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -20,6 +20,7 @@
 
       Required field for all events.'
     example: '2016-05-23T08:05:34.853Z'
+    default_field: true
   - name: labels
     level: core
     type: object
@@ -31,6 +32,7 @@
 
       Example: `docker` and `k8s` labels.'
     example: '{"application": "foo-bar", "env": "production"}'
+    default_field: true
   - name: message
     level: core
     type: match_only_text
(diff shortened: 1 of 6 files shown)


generated/beats/fields.ecs.yml
```diff
@@ -10167,7 +10171,6 @@
       A span represents an operation within a transaction, such as a request to another
       service, or a database query.'
     example: 3ff9a8981b7ccd5a
-    default_field: false
```
ebeahan, line 10170: Does it matter if `default_field` remains defined on a root field when it's `false`?

ebeahan (commented):
> The generator was adding the default_field setting to non-root fieldsets, causing top-level fields (like message) to be excluded from the default_field index template configuration.

With these changes, top-level fields, like `message`, will now have `default_field: true` explicitly defined if that top-level field is in the allow list?

generated/beats/fields.ecs.yml
adriansr: It can be explicitly defined as false, but is unnecessary as Filebeat defaults to false, that's why the generator doesn't include it.

adriansr:
> With these changes, top-level fields, like message, will now have default_field: true explicitly defined if that top-level field is in the allow list?

Yes, exactly.

ebeahan (approved):
LGTM

Issue #1339, opened 2021-04-01
Use higher compression as default in composable index template settings


ebeahan:
The built-in index templates settings in Elasticsearch (`logs-settings`, `metrics-settings`, `synthetics-settings`) enable a higher compression ratio (`DEFLATE` vs `LZ4`) as the default now with the new indexing strategy:

`index.codec: best_compression`

Perhaps we should adopt this setting as one of the default added to the generated [sample composable template](https://github.com/elastic/ecs/blob/main/generated/elasticsearch/composable/template.json#L65).

PR #1712, opened 2022-01-05
Higher default compression for composable templates

kgeller:
Addresses https://github.com/elastic/ecs/issues/1339

The built-in index templates settings in Elasticsearch (logs-settings, metrics-settings, synthetics-settings) enable a higher compression ratio (DEFLATE vs LZ4) as the default now with the new indexing strategy: `index.codec: best_compression`


ebeahan (approved):
LGTM

ebeahan:
@kgeller, I think this one still needs to be backported?

kgeller:
@ebeahan good catch! 

Looks like they made it in already, probably from another backport, but adding them just in case

Issue #1264, opened 2021-02-16
Add analyzer multi-field capability


sektorcap:
**Summary**
Add support for the [analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/specify-analyzer.html#specify-search-field-analyzer) in the ECS field definitions.

**Example**:
```
- name: url
  title: URL
  description: URL fields provide support for complete or partial URLs, and supports
    the breaking down into scheme, domain, path, and so on.
  type: group
  fields:
  - name: domain
    level: extended
    type: keyword
    ignore_above: 1024
    multi_fields:
    - name: text
      type: text
      analyzer: domain_analyzer
    description: 'Domain of the url, such as "www.elastic.co".
```

Example ES index template snippet:
```
      "url": {
        "properties": {
          "domain": {
            "fields": {
              "text": {
                "analyzer": "domain_analyzer",
                "norms": false,
                "type": "text"
              }
            },
            "ignore_above": 1024,
            "type": "keyword"
          },
...
```

Of course, in the `template-setting` you should define the analyzer.
```
{
  "order": 1,
  "index_patterns": [
    "your-ecs-*"
  ],
  "settings": {
    
    "index": {
      "mapping": {
        "total_fields": {
          "limit": "10000"
        }
      },
      "analysis": {
        "analyzer": {
          "domain_analyzer": {
            "type": "pattern",
            "pattern": "\\."
          }
        }
      },
      "refresh_interval": "5s"
    }
  }
}

```

BenB196:
I would like to add another use-case here. Currently url.path is only handled as a keyword, but if you have a long and complex path, and you want to search for paths containing on a specific word(s), you need to resort to wildcard usage. You could use another analyzer like pattern: "/" to parse the paths to be more search friendly.

PR #1737, opened 2022-01-24
Adding support in for custom analyzers for text and overriding the default template settings for index templates

kgeller:
Resolves https://github.com/elastic/ecs/issues/1264
Resolves https://github.com/elastic/ecs/issues/1203

This PR encompasses two smaller tooling improvements: 

1) Adds support for the [analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/specify-analyzer.html#specify-search-field-analyzer) in the ECS field definitions. 
2) Now allows for users to override the default composable template settings. 

Change 2 led to a change in the arguments for the generator. Instead of just `--template-settings`, we will now have `--template-settings-legacy` and `--template-settings`. This will be a breaking change since `--template-settings` will now refer to v2 index templates.


kgeller:
@ebeahan @djptek I added support for users to add in their search `analyzer` for text fields. In doing so, I was trying to decide how to handle the settings part of it.  I was back and forth on whether to 1) programmatically add an empty definition, 2) add no definition, or 3) adding an empty definition to the [settings starter file](https://github.com/elastic/ecs/blob/main/usage-example/fields/template-settings.json). 

I went with choice 1, but it is definitely the most complex. Do you both agree with that approach, or is there something else I should do / am not thinking of?

ebeahan:
>  In doing so, I was trying to decide how to handle the settings part of it. I was back and forth on whether to 1) programmatically add an empty definition, 2) add no definition, or 3) adding an empty definition to the settings starter file.

Would you provide some more details? I'm uncertain what exactly needs handling here.

If the user wants to declare a custom analyzer on a field, the index settings specify an analyzer. But if the user doesn't define an analyzer, that settings would be omitted, right?

kgeller:
@ebeahan 

> Would you provide some more details? I'm uncertain what exactly needs handling here.

I just wasn't sure how much 'control' ECS wanted to exert here. They 'key' part IMO was passing the analyzer along with the field, which we do. It's just a matter of how we handle the settings piece of the analyzer and if we want to add it for them or not or whatever. 

> If the user wants to declare a custom analyzer on a field, the index settings specify an analyzer. But if the user doesn't define an analyzer, that settings would be omitted, right?

Correct. Essentially I was trying to figure out how we handle that definition of the analyzer in the settings, because we have a handful of directions we can go in. I went with adding an empty definition so that the user can fill it in with whatever definition they see fit. But of course if there are no analyzers, nothing will be added. On the other end of the spectrum, we could add nothing into the settings and let the user populate it themselves. 

Does that help?

ebeahan:
>  we could add nothing into the settings and let the user populate it themselves.

I think I better understand. IMO, our tooling should support passing through whatever settings the user provides. I'll walk through a hypothetical scenario using this feature. Share if you see this coming together differently. 

Using the `domain_analzyer` example from #1264, the file passed in using `--template-settings` would contain its definition (along with a few other index template setting):

```json
{
  "order": 1,
  "index_patterns": [
    "your-ecs-*"
  ],
  "settings": {
    "index": {
      "mapping": {
        "total_fields": {
          "limit": "10000"
        }
      },
      "analysis": {
        "analyzer": {
          "domain_analyzer": {
            "type": "pattern",
            "pattern": "\\."
          }
        }
      },
      "refresh_interval": "5s"
    }
  }
}
```

The user also declares the analyzer on their field definitions. Since we've moved all text fields to `match_only_text`, these would either be custom fields or someone overriding ECS default types:

```yml
- name: acme
  title: ACME
  description: >
    Acme Inc. custom fields
  type: group
  fields:
    - name: account.domain
      type: text
      level: custom
      analzyer: domain_analyzer
      description: >
        Customer account description for this activity.
```

If a user wants a built-in analyzer, they can skip the first step and define it on the field:

```yml
  fields:
    - name: account.domain
      type: text
      level: custom
      analyzer: whitespace
      description: >
        Customer account description for this activity.
```

User runs the generator, and the (v1) template is created:

```json
{
  "index_patterns": [
    "your-ecs-*"
  ],
  "mappings": {
    "date_detection": false,
    "dynamic_templates": [
      {
        "strings_as_keyword": {
          "mapping": {
            "ignore_above": 1024,
            "type": "keyword"
          },
          "match_mapping_type": "string"
        }
      }
    ],
    "properties": {
      "@timestamp": {
        "type": "date"
      },
      "acme": {
        "properties": {
          "account": {
            "properties": {
              "domain": {
                "analyzer": "domain_analyzer",
                "type": "text"
              },
              "id": {
                "ignore_above": 1024,
                "type": "keyword"
              }
            }
          }
        }
      }
    }
  },
  "order": 1,
  "settings": {
    "index": {
      "analysis": {
        "analyzer": {
          "domain_analyzer": {
            "pattern": "\\.",
            "type": "pattern"
          }
        }
      },
      "mapping": {
        "total_fields": {
          "limit": "10000"
        }
      },
      "refresh_interval": "5s"
    }
  }
}
```


If the resulting template has a config error, I think it's okay to rely on ES to throw an exception when it's uploaded. If I typoed `type` in my analyzer config:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "analyzer [domain_analyzer] must specify either an analyzer type, or a tokenizer"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "analyzer [domain_analyzer] must specify either an analyzer type, or a tokenizer"
  },
  "status": 400
}
```

scripts/generators/es_template.py
```diff
@@ -125,12 +105,12 @@ def generate_legacy(ecs_flat, ecs_version, out_dir, template_settings_file, mapp
     mappings_section = mapping_settings(mapping_settings_file)
     mappings_section['properties'] = field_mappings
 
-    generate_legacy_template_version(7, ecs_version, mappings_section, out_dir, template_settings_file)
```
ebeahan, line 128: 👍

scripts/generators/es_template.py
```diff
@@ -204,23 +184,39 @@ def mapping_settings(mapping_settings_file):
     return mappings
 
 
-def template_settings(es_version, ecs_version, mappings_section, template_settings_file):
+def template_settings(ecs_version, template_version, mappings_section, template_settings_file, component_names=None):
```
ebeahan, line 187: nit: instead of `template_version`, you could use a `bool` kwarg like `legacy=False`. Avoids having function calls with `1` and `2` magic numbers.

Simplifies the later v1/v2 check:

```python
if legacy:
    template = default_v1_template_settings(ecs_version)
else:
    template = default_v2_template_settings(ecs_version)
```

ebeahan (commented):
Generator implementation looks good. Do you see any opportunities to cover the different template generations in the unit tests?

I had some thoughts about the proposed naming.

scripts/generator.py
```diff
@@ -92,13 +98,17 @@ def argument_parser():
...
     parser.add_argument('--template-settings', action='store',
-                        help='index template settings to use when generating elasticsearch template')
+                        help=argparse.SUPPRESS)
     parser.add_argument('--mapping-settings', action='store',
                         help='mapping settings to use when generating elasticsearch template')
     parser.add_argument('--strict', action='store_true',
                         help='enforce strict checking at schema cleanup')
     parser.add_argument('--intermediate-only', action='store_true',
                         help='generate intermediary files only')
+    parser.add_argument('--template-settings-v1', action='store',
```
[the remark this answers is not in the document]
kgeller: I like switching to `--template-settings-legacy`. I was really hesitant about introducing a breaking change into 8+. So if I can still merge into 8.0, I am super happy to make a breaking change. 

I'll update that!

scripts/generators/es_template.py
kgeller: Great point, updating

kgeller:
> Generator implementation looks good. Do you see any opportunities to cover the different template generations in the unit tests?
> 
> I had some thoughts about the proposed naming.

I love your suggestions for updated naming, I just updated all of that.

As far as unit tests, I did not. But I will take a closer look.

kgeller:
@ebeahan tests added!

ebeahan (approved):
Other than the `USAGE.md` suggestion, everything else is a nit or something we can handle outside this PR.

scripts/generators/es_template.py
```diff
@@ -186,7 +166,7 @@ def entry_for(field):
                 if mf_type == 'keyword':
                     ecs_helpers.dict_copy_existing_keys(mf, mf_entry, ['normalizer', 'ignore_above'])
                 elif mf_type == 'text':
-                    ecs_helpers.dict_copy_existing_keys(mf, mf_entry, ['norms'])
+                    ecs_helpers.dict_copy_existing_keys(mf, mf_entry, ['norms', 'analyzer'])
```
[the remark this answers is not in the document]
kgeller: Makes sense to me

PR #1739, opened 2022-01-25
broken link fix

kgeller:
Fixes the first link (`rfcs/`) [here](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md) see [fix](https://github.com/kgeller/ecs/blob/broken-link-fix/rfcs/PROCESS.md)


ebeahan (approved):

ebeahan:
I wouldn't worry about backporting this change. The release branches will contain versions of `./rfc` from when it was cut, but the up-to-date RFC proposals and process docs live in `main`.

At some point in the future, it may be tidier to split the RFCs into their own repo, maybe `ecs-spec`.

PR #1740, opened 2022-01-25
[RFC] Add a reusable `risk` field set

ajosh0504:
This is a proposal to add a reusable `risk` field set, to support two new features i.e. Host Risk Score and User Risk Score being added to the Security Solution. This new field set will help correlate information present in indices corresponding to these features, with other security-related indices.


kgeller (approved):
LGTM! 

I'll get the number and the date added in shortly, and get this merged.

PR #1744, opened 2022-01-27
[RFC 0031] Risk Fields- Stage 1

ajosh0504:
This PR adds more details to [RFC 0031](https://github.com/elastic/ecs/blob/main/rfcs/text/0031-risk-fields.md) which deals with adding `risk` fields to ECS. These fields are intended to be used as reusable fields to express risk associated with entities such as hosts, users etc.


rfcs/text/0031-risk-fields.md
```diff
@@ -21,6 +21,18 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
 <!--
 Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
 -->
+The `risk` fields being proposed are as follows:
+
+Field | Type | Example | Description | Use Case
+-- | -- | -- | -- | -- 
+risk.score | float | 88.73 | A score in the range 0-100 that quantifies the amount of risk associated with an entity. The higher the score, the higher the risk. | Can be used to indicate the risk associated with a particular host
+risk.level | keyword | High | A risk level corresponding to the risk score | Can be used to indicate the risk associated with a particular host
+risk.info.*| group | - | Information about what contributed to the risk | Rules that made a host high-risk
```
kgeller, line 30: Can you elaborate more here on what is to be expected? Were you thinking something with the `[rule.*](https://www.elastic.co/guide/en/ecs/current/ecs-rule.html)` field set?

rfcs/text/0031-risk-fields.md
```diff
@@ -31,13 +43,74 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+  },
+  "ingest_timestamp": "2022-01-27T15:49:00.335519179Z",
+  "risk": "High"
+}
+```
+
+With the introduction of ECS `risk` fields, fields in the above document would map as follows:
+* `risk_stats.rule_risks` -> `host.risk.info.rule_risks`
+* `risk_stats.risk_score` -> `host.risk.score`
+* `risk_stats.risk_multipliers` -> `host.risk.info.risk_multipliers`
```
kgeller, line 112: Could you add `risk_multipliers` to the field additions section above?

kgeller:
@ajosh0504 Have you identified a sponsor?

Per the [docs](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md): 

```
The sponsor at Elastic:

- can be the same person as the contributor if they're someone at Elastic that can take ownership of this change through membership on the ECS committee
- is involved at least from stage 1 onward if a different person than the contributor
- signs off on each stage if a different person than the contributor
- takes or coordinates ownership of the addition in terms of support and maintenance after the RFC process is completed

```

rfcs/text/0031-risk-fields.md
```diff
@@ -21,6 +21,18 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
 <!--
 Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
 -->
+The `risk` fields being proposed are as follows:
+
+Field | Type | Example | Description | Use Case
+-- | -- | -- | -- | -- 
+risk.score | float | 88.73 | A score in the range 0-100 that quantifies the amount of risk associated with an entity. The higher the score, the higher the risk. | Can be used to indicate the risk associated with a particular host
```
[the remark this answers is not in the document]
ajosh0504: I suppose `risk.score` means the same as `risk_score`. I also noticed a mention of `risk_score` in the context of rules. My hope with these `risk.*` fields was to create a reusable field set that expands on the existing concept of risk score, to include things like risk level and explainability as well, and which can then be used by other assets/entities such as rules, events, hosts, users etc.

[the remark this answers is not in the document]
ebeahan: > I also noticed a mention of risk_score in the context of rules.

Yes, `signal.rule.risk_score` is populated by the detection engine and is specified in each detection rule. While this risk scoring feature is in the early stages, it might be a good time to discuss aligning how to capture `risk` values across the Security solution.

I noted while `event.risk_score` and `signal.rule.risk_score` are mapped as `float,` the app actually won't let users define a float value:

```json
{
  "name": "Error",
  "body": {
    "statusCode": 400,
    "error": "Bad Request",
    "message": "[request body]: Invalid value \"21.5\" supplied to \"risk_score\""
  }
}
```

I don't have a strong opinion around `float` vs. `long,` but we should be consistent. I'll open an issue for Kibana asking about the `signal.rule.risk_score` behavior.

rfcs/text/0031-risk-fields.md
ajosh0504: I was thinking a reusable field set that can be used with any asset/entity. Having a separate reusable field set for `risk` also helps us standardize the definition of risk across assets. So these fields would be used with field sets such as (but not limited to) `rule.*`, `host.*`, `user.*`, `event.*`. This might mean updated to the existing risk-related fields currently present in some field sets, for example `event.risk_score` would become `event.risk.score` etc.

rfcs/text/0031-risk-fields.md
```diff
@@ -31,13 +43,74 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
...
+* "Show me Critical and high-risk users on Host X"
+
 
 ## Source data
 
 <!--
 Stage 1: Provide a high-level description of example sources of data. This does not yet need to be a concrete example of a source document, but instead can simply describe a potential source (e.g. nginx access log). This will ultimately be fleshed out to include literal source examples in a future stage. The goal here is to identify practical sources for these fields in the real world. ~1-3 sentences or unordered list.
 -->
-
+The Host and User Risk Score views in the Security solution are/will be backed by indices produced as a result of running [transforms](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html) on a variety of data sources, not limited to the alerts data streams. An example document produced by the Host Risk Score transform in the absence of ECS `risk` fields looks as follows:
```
ebeahan, line 62: Instead of nesting, could `risk.*` be a top-level field and have a field that specifies the `risk.type` as `host`, `user`, etc?

Unless both `host.risk.*` and `user.risk.*` appear together in an event, avoiding nesting might simplify the schema and queries (e.g., I only query `risk.score` instead of `host.risk.score` AND `user.risk.score`).

ebeahan (commented):
I had a couple of thoughts after reviewing.

@ajosh0504 would you also update the `Stage` at the top of the proposal to say `Stage: 1 (draft)`?

rfcs/text/0031-risk-fields.md
ajosh0504: I had a question about `risk_multipliers` and `rule_risks` actually. So these fields might not apply to all types of assets that have a risk associated with them. For example, `rule_risks` is supposed to provide information about the rules that contributed to the risk of a particular host. This might not apply to say events. Is there a better naming terminology for these?

ajosh0504:
> @ajosh0504 Have you identified a sponsor?
> 
> Per the [docs](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md):
> 
> ```
> The sponsor at Elastic:
> 
> - can be the same person as the contributor if they're someone at Elastic that can take ownership of this change through membership on the ECS committee
> - is involved at least from stage 1 onward if a different person than the contributor
> - signs off on each stage if a different person than the contributor
> - takes or coordinates ownership of the addition in terms of support and maintenance after the RFC process is completed
> ```

@kgeller Hah I had not. I thought sponsors were required only for external folks. I admittedly did not fully understand the responsibilities for a sponsor, mainly the support, maintenance and owning these changes. I don't think I'm qualified enough to do that. How do I go about choosing a sponsor within Elastic? Would someone from the ECS team be willing to help?

rfcs/text/0031-risk-fields.md
ajosh0504: @ebeahan Interesting callout about `risk_score` not accepting a float value in other places. Totally agree about getting consensus on capturing `risk` in the Security solution. I'll create a Discuss issue for this.

re: nesting, it is possible that `host.risk.*` and `user.risk.*` might appear in the same event. And the risk score for the two will likely be different.

kgeller:
@ajosh0504 We can absolutely help identify a sponsor. I have added a future agenda item for the next ECS committee meeting, but just in case, @ebeahan anyone come to mind for you now?

rfcs/text/0031-risk-fields.md
kgeller: Gotcha. For the field `risk.info.*`, what exactly did you have in mind? I see it listed as type `group`, but I don't think that is an actual ES type, and I see it ends in .* leading me to believe it's an object/field set.

rfcs/text/0031-risk-fields.md
kgeller: Better name in the sense that they are 'optional' since they don't apply to certain things?

rfcs/text/0031-risk-fields.md
ajosh0504: @kgeller I noticed [another RFC](https://github.com/elastic/ecs/blob/risk-fields-stage-1/rfcs/text/0027-faas-fields.md) that I was using as reference, which had a type `group`. Looks like they've updated that though. But yes, the intent was for the `risk.info.*` fields to be an object. This field should contain explainability about the calculated risk score for a particular asset, but these explainability metrics could look different, for different assets. I wasn't sure how to capture that- any inputs on this would be super helpful!

rfcs/text/0031-risk-fields.md
ajosh0504: Exactly!

kgeller: I think keeping their names as you have described here makes sense. And then in the description you note the 'optional-ness' alongside maybe some helpful examples.

rfcs/text/0031-risk-fields.md
kgeller: Will that object be something new, or were you planning on re-using an existing field set such as [`rule.*`](elastic.co/guide/en/ecs/current/ecs-rule.html)?

rfcs/text/0031-risk-fields.md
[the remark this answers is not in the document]
ajosh0504: I kicked off a [discussion](https://github.com/elastic/security-team/issues/2895) with the broader Security team regarding the concept of risk across the Security solution. My main takeaway:

> Let's focus on adding the new embeddable risk.* fields, then we can work with ECS team to determine where they apply. For example we could disallow the use of this new risk.* field set under event.* and rule.* so as to not cause confusion with event.risk_score, etc. We can worry about that subsequently.

rfcs/text/0031-risk-fields.md
ajosh0504: This will be new. I'm going to update this a bit.

rfcs/text/0031-risk-fields.md
ajosh0504: @kgeller Done! (Hopefully)

rfcs/text/0031-risk-fields.md
```diff
@@ -21,6 +21,22 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
...
+-- | -- | -- | -- | -- 
+risk.calculated_score_norm | float | 88.73 | A risk classifications core (0-100) calculates by an internal system as part of entity analytics and entity risk scoring | Can be used to indicate the risk associated with a particular host
+risk.calculated_level | keyword | High | A risk classification level calculated by an internal system as part of entity analytics and entity risk scoring | Can be used to indicate the risk associated with a particular host
+risk.static_score_norm | float | 83.0 | A risk classification score (0-100) obtained from outside the system, such as from some external Threat Intelligence Platform | Can be used to indicate the projected risk of a particular host based on a trusted third party intelligence feed 
+risk.static_level | keyword | High | A risk classification level obtained from outside the system, such as from some external Threat Intelligence Platform | Can be used to indicate the projected risk of a particular host based on a trusted third party intelligence feed
+risk.factors| object | See Source data section | Factors that contributed to the risk | Explainability about what contributed to the risk
```
kgeller, line 32: I am concerned here about not having a defined structure for these objects. Do you think that's something we could add?

ajosh0504: @kgeller Wdym by not having a defined structure for the objects? I have also added definitions for `risk.factors.alerts` and `risk.factors.others`. So essentially, the `risk.factors` object can only further have the fields `alerts` and `others` . Furthermore, I have also tried to specify how these fields should look like. For example `risk.factors.alerts` should be an array of objects. If there's a preferred way to elaborate on these to maybe specify the structure of the objects themselves, I am happy to do so- I just wasn't sure where and how to add that.

rfcs/text/0031-risk-fields.md
```diff
@@ -21,6 +21,26 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
...
+risk.factors.alerts | array of objects (Optional) | See Source data section | Alerts that contributed to the risk | Explainability about what contributed to the risk
+risk.factors.alerts.rule_id | keyword | 99cb72b7-ab55-4375-8bbe-696e3e1b8fe2 | Rule ID associated with the alert | Metadata about alerts that contributed to the risk
+risk.factors.alerts.rule_name | keyword | Unusual Windows Path Activity | Rule name associated with the alert | Metadata about alerts that contributed to the risk
+risk.factors.alerts.rule_risk | float | 2.41 | Risk associated with the alert | Metadata about alerts that contributed to the risk
+risk.factors.others | array of strings (Optional) | See Source data section | Factors apart from alerts that contributed to the risk | Explainability about what contributed to the risk
```
kgeller, line 37: If 'risk.factors.others' has both a 'type' and a 'description', should 'others' be type object array instead of string array?

rfcs/text/0031-risk-fields.md
kgeller: Looks awesome! Thanks for getting those added in

rfcs/text/0031-risk-fields.md
```diff
@@ -21,6 +21,26 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
...
+risk.calculated_level | keyword | High | A risk classification level calculated by an internal system as part of entity analytics and entity risk scoring | Can be used to indicate the risk associated with a particular host
+risk.static_score_norm | float | 83.0 | A risk classification score (0-100) obtained from outside the system, such as from some external Threat Intelligence Platform | Can be used to indicate the projected risk of a particular host based on a trusted third party intelligence feed 
+risk.static_level | keyword | High | A risk classification level obtained from outside the system, such as from some external Threat Intelligence Platform | Can be used to indicate the projected risk of a particular host based on a trusted third party intelligence feed
+risk.factors| object | See Source data section | Factors that contributed to the risk | Explainability about what contributed to the risk
+risk.factors.alerts | array of objects (Optional) | See Source data section | Alerts that contributed to the risk | Explainability about what contributed to the risk
```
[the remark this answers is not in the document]
ajosh0504: Good catch. Resolved in [5470920](https://github.com/elastic/ecs/pull/1744/commits/5470920ccfcafc2f49488d84d3c35434c920cabe).

kgeller:
### Stage 1 Criteria

- [x] Opened pull request for this proposal
- [ ] Subject matter experts identified and weighed in on the high level utility of these changes in the pull request

@ajosh0504 I think this is looking really great. I think the only thing remaining would just be an approval from an SME

ajosh0504:
@MikePaquette @paulewing This should now be ready for y'all's review.

MikePaquette (changes requested):
After a good deal of thought and discussion, I would like to request the following changes:

1. Remove all  `risk.factors.*` objects/fields.  Rationale: these fields and objects are complex, using nested documents, and they arguably do not meet the "C" in ECS standard - that is, they are not likely to be used widely outside of the Elastic Security solution, and thus are not "common."   Let's omit them from this ECS RFC, go about our implementation without the burden of a defined ECS field set, and then re-visit this again in the future when we've gotten some more experience, and when they might become more common.
2. Add new field `risk.calculated_score`. Rationale: to be consistent with current ECS usage of `event.risk_score` and  `event.risk_score_norm` which allows for the native score (which might not be 0-100) and the normalized score, to co-exist.
3. Add new field `risk.static_score` Rationale: same as above.

This should result in this RFC containing just 6 new fields, which seems like a nice, small, and clean addition to ECS.

ajosh0504:
@MikePaquette's I have made the changes per your comments. Could you please give a 👍 so we can merge?

Also, although we aren't adding the `risk.factors.*` fields, we can still keep them in the documents and use them to populate the Top Contributors view in the Security App right?

MikePaquette:
Hi @ajosh0504 

> Also, although we aren't adding the risk.factors.* fields, we can still keep them

They should not be in the ECS documents, but yes, we can use in them in the app and can document them in the security docs as appropriate.

kgeller:
> I am assuming that further review of field definitions will take place when the Pull request for the fields is created.

@MikePaquette are you expecting things to significantly change still, or is this in a good state for merge and implementation into ECS?

ajosh0504:
@MikePaquette Could you please confirm @kgeller 's query above? If all looks good otherwise, I can merge.

PR #1787, opened 2022-02-22
[8.2] Higher default compression for composable templates (#1712)

kgeller:
Backports the following commits to 8.2:
 - Higher default compression for composable templates (#1712)


kgeller:
@elasticmachine run

kgeller:
@elasticmachine run test

PR #1793, opened 2022-02-23
Add additional log.syslog fields

taylor-swanson:
- Add additional fields for RFC 5424 messages (log.syslog.version,
log.syslog.msgid, log.syslog.data)
- Add log.syslog.hostname, log.syslog.appname, and log.syslog.procid
for hostname, process name, and process ID fields present in syslog
messages, respectively. These fields are added since it is not always
known that user wants these values copied to the more general ECS fields
(host.hostname, process.name, process.pid).

### DIFF
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -156,3 +156,50 @@
 
         According to RFCs 5424 and 3164, the priority is 8 * facility + severity.
         This number is therefore expected to contain a value between 0 and 191.
+
+    - name: syslog.version
+      level: extended
+      type: long
+      example: 1
+      short: Syslog protocol version.
+      description: >
+        The version of the Syslog protocol specification. Only applicable for RFC 5424 messages.
+
+    - name: syslog.hostname
+      level: extended
+      type: keyword
+      example: example-host
+      short: The host that originated the Syslog message.
+      description: >
+        The hostname, FQDN, or IP of the machine that originally sent the Syslog message.
+
+    - name: syslog.appname
+      level: extended
+      type: keyword
+      example: sshd
+      short: The device or application that originated the Syslog message.
+      description: >
+        The device or application that originated the Syslog message, if available.
+
+    - name: syslog.procid
+      level: extended
+      type: keyword
+      example: 12345
+      short: The process name or id that originated the Syslog message.
+      description: >
+        The process name or id that originated the Syslog message, if available.
+
+    - name: syslog.msgid
+      level: extended
+      type: keyword
+      example: ID47
+      short: An identifier for the type of Syslog message.
+      description: >
+        An identifier for the type of Syslog message, if available. Only applicable for RFC 5424 messages.
+
+    - name: syslog.data
+      level: extended
+      type: flattened
+      short: Structured data expressed in RFC 5424 messages.
+      description: >
+        Structured data expressed in RFC 5424 messages, if available.
--- a/schemas/log.yml
+++ b/schemas/log.yml
@@ -159,7 +159,7 @@
 
     - name: syslog.version
       level: extended
-      type: long
+      type: keyword
       example: 1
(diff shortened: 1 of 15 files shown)


schemas/log.yml
```diff
@@ -156,3 +156,50 @@
...
+      example: sshd
+      short: The device or application that originated the Syslog message.
+      description: >
+        The device or application that originated the Syslog message, if available.
+
+    - name: syslog.procid
+      level: extended
+      type: keyword
+      example: 12345
+      short: The process name or id that originated the Syslog message.
```
andrewkroh, line 188: `s/id/ID/g` or `s/id/identifier/g`

schemas/log.yml
```diff
@@ -156,3 +156,50 @@
 
         According to RFCs 5424 and 3164, the priority is 8 * facility + severity.
         This number is therefore expected to contain a value between 0 and 191.
+
+    - name: syslog.version
+      level: extended
+      type: long
```
ebeahan, line 162:
```suggestion
      type: keyword
```

Versions fields in ECS are usually `keyword` since their values are identifiers and are not expected to be used for a numeric operation or `range` query. So I'd suggest `keyword` instead here.

Also future proofs against different variations of versioning (1.1, 2.0.0-rc1, etc.)

schemas/log.yml
```diff
@@ -156,3 +156,50 @@
...
+      short: An identifier for the type of Syslog message.
+      description: >
+        An identifier for the type of Syslog message, if available. Only applicable for RFC 5424 messages.
+
+    - name: syslog.data
+      level: extended
+      type: flattened
+      short: Structured data expressed in RFC 5424 messages.
+      description: >
+        Structured data expressed in RFC 5424 messages, if available.
```
ebeahan, line 205: How does the expected `flattened` object here look? Key-value pairs from the structure data?

I think it makes sense to add more guidance about which values will build out the `flattened` object captured here.

schemas/log.yml
```diff
@@ -156,3 +156,50 @@
...
+
+    - name: syslog.msgid
+      level: extended
+      type: keyword
+      example: ID47
+      short: An identifier for the type of Syslog message.
+      description: >
+        An identifier for the type of Syslog message, if available. Only applicable for RFC 5424 messages.
+
+    - name: syslog.data
```
ebeahan, line 200: Can we rename to `syslog.structured_data`?

I worry some users might confuse `data` with the free-form message portion of syslog events.

schemas/log.yml
```diff
@@ -156,3 +156,50 @@
...
+      description: >
+        The version of the Syslog protocol specification. Only applicable for RFC 5424 messages.
+
+    - name: syslog.hostname
+      level: extended
+      type: keyword
+      example: example-host
+      short: The host that originated the Syslog message.
+      description: >
+        The hostname, FQDN, or IP of the machine that originally sent the Syslog message.
```
ebeahan, line 174: Nit: I see value in adding a sentence or two here describing how this is the `hostname` taken from the syslog header, and it often can be different a different value from the host that the event took place on.

Won't be an issue for something programmatically parsing like the syslog processor, but it might help clarify confusion for anyone parsing syslog headers themselves.

schemas/log.yml
taylor-swanson: Yes, it will be key-value pairs from the structured data portion. I made a mention to the RFC section as well.

ebeahan (approved):
LGTM! Thanks for making the adjustments.

Should these changes go into 8.2?

taylor-swanson:
> LGTM! Thanks for making the adjustments.
> 
> Should these changes go into 8.2?

If we could do that, that would be great! The associated syslog parser/processor is also targeting 8.2.

PR #1796, opened 2022-02-24
Added `id`, `name` and `version` fields for `faas`

AlexanderWert:
We learned from customer feedback that in some scenarios it's reasonable to map multiple serverless functions (e.g. AWS Lambda functions) to one single service entity within Elastic. Therefore, we cannot rely on the `service.name`, `service.version` and `service.id` fields to represent the serverless function name and serverless function id.

Thus, additional `faas` fields are needed to explicitly capture capture the function id and name.

This PR adds the `faas.id`, `faas.name` and `faas.version` fields to ECS `faas` fields group.


kgeller (changes requested):
@AlexanderWert I think these changes make sense. Could you update the RFC as well to reflect these additions since faas.* is still stage 2 / beta?

AlexanderWert:
Thanks @kgeller ! I updated the RFC.

kgeller (approved):
LGTM!

PR #1798, opened 2022-02-24
Adding support for re-use as an array

### LINKED ISSUE #1736: Mark which field set reuses expect arrays
### Purpose

Indicate in the field definitions when a field set reuse expects an array.

### Background

Reuses allowing arrays may benefit specific limited use. However, there's currently no way to declare this reuse type in the ECS tooling.

```json
{
  "process": {
    "previous" : [
      {
        "name": "abc.exe"
      },
      {
        "name": "xyz.exe"
      }
    ]
  }
}
```

The above basic example presents an ECS mapped event using field set array reuse. The `previous` object is `process.*` field set reused at `process.previous`.

The `normalize: [ 'array' ]` setting controls when a particular field expects an array of values. There's no dedicated `array` type in Elasticsearch, and any field can contain zero or more values of the same type. So when the `array` attribute is specified, a note is added by the generator script to that field's description in the ECS docs. 

Field def setting:

```yaml
    - name: example
      normalize:
        - array
```

Results in this text added to field's description:

Screen Shot 2022-01-20 at 1 06 36 PM

### Proposed design

Implement a similar mechanism as leaf fields:

* specify an attribute on a field reuse
* have the asciidoc generator handle including that content in the field reuse description

### Open question(s)

* Do we reuse the existing `normalize` attribute convention? Or a simple `bool` parameter (e.g., `is_array`)?

kgeller:
Closes #1736 

By adding a normalize as array attribute a re-use definition, users will now be able to re-use a field set as an array.

Example: 
```
- at: process
  as: previous
  short_override: An array of all previous executions for the process, including the initial fork details.
  normalize:
    - array
```
-------->

The generated artifacts will now include a nested object of the re-use type as seen [here](https://github.com/kgeller/ecs/blob/a1a9571feb0fab8271fd313d42ebaf08234637b6/generated/csv/fields.csv#L618) and will have an updated definition in the docs.
Screen Shot 2022-02-28 at 4 04 54 PM


ebeahan (changes requested):
The approach here isn't making the expected changes to some generated artifacts. I left notes here, but also I'm happy to sync to discuss more.

We should also add test coverage for this new functionality where reasonable.

kgeller:
Closing and pivoting to https://github.com/elastic/ecs/issues/1838 and https://github.com/elastic/ecs/issues/1839

PR #1826, opened 2022-03-07
[RFC] Stage-3 Linux event model

mitodrummer:
- If submitting code/script changes, have you verified all tests pass locally using `make test`? n/a
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? not yet


kgeller:
@mitodrummer Upon beta implementation of this proposal, the ECS team hit a couple of snags.

1) Are the file_description.socket* fields being utilized yet? As far as I could tell from the examples in the RFC document, they are not. In that case, I'd say we hold off on those for now. It's much easier to add later, than try to remove/alter.

2) ECS re-use nesting is a complex thing when it comes to chaining re-uses. Example: if we nest `as cat at dog` and `as fish at cat`, then we will end up with `dog.cat.*` and `cat.fish.*` **and** `dog.cat.fish.*`. So when trying to implement file, fifo and pipe at file description, we will bring along everything nested at the file fieldset (hash, elf, etc) underneath file description as well. And same goes for when we nest file_description and tty at process. I didn't see any usages for the file fieldset in the examples in the RFC, but is this something worth considering: adding fields to the file_description field set itself?

3) In one of the examples in the RFC, you note `To keep backwards compat and avoid data duplication. We keep user/group info for top level process at the top level`. What is the difference then between user & process.user and group & process.user.group?

While I hope this makes sense, but in the event that it doesn't, let me know and we can zoom to discuss.

rfcs/text/0030/group.yml
```diff
@@ -19,18 +19,8 @@
   reusable:
     top_level: true # top level group relates to top level process
     expected:
-      - at: process.entry_leader
```
mitodrummer, line 22: Because of how the tooling works (an issue of ordering, nestings not being created in the right sequence), we have to simplify how we nest user and group... The fact that we won't populate process.user / process.group (in endpoint) because it's already set at the top level will need to be expressed as an implementation detail elsewhere.

rfcs/text/0030/process.yml
```diff
@@ -127,3 +127,26 @@
         Entry point information for a session.
 
         Remote client information such as ip, port and geo location.
+
+    - name: tty
```
mitodrummer, line 131: process.tty is needed in v1 of session viewer, so it has been changed to a reduced set from file_description (which actually works better in the implementation)

rfcs/text/0030/user.yml
```diff
@@ -19,18 +19,9 @@
   reusable:
     top_level: true # the top level user relates to the top level process
     expected:
-      - at: process.entry_leader
+      - at: process
```
mitodrummer, line 22: same issue as with group.

rfcs/text/0030/process.yml
```diff
@@ -21,10 +21,10 @@
     expected:
       - at: process
         as: entry_leader
-        short_override: The session leader process associated with initial terminal or remote access via SSH, SSM and other remote access protocols. Entry sessions are also used to represent a service directly started by the init process. In many cases this is the same as the session_leader.
+        short_override: First process from terminal or remote access via SSH, SSM, etc OR a service directly started by the init process.
```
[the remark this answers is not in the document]
m-sample: LGTM, but I wrote both :)

rfcs/text/0030/group.yml
m-sample: It is still valuable to capture the *users* (real, effective, saved) and groups for each process since they may differ from the 'main' process.  Ie it's valuable for the process, process.entry_leader process.session_leader, process.group_leader...

rfcs/text/0030/process.yml
```diff
@@ -21,10 +21,10 @@
     expected:
       - at: process
         as: entry_leader
-        short_override: Information about the entry session leader process.
+        short_override: First process from terminal or remote access via SSH, SSM, etc OR a service directly started by the init process.
       - at: process
         as: session_leader
-        short_override: Information about the session leader. In some cases this may be the same information as entry_leader.
+        short_override: Often the same as entry_leader. When it differs, it represents a session started within another session. e.g. using tmux
       - at: process
```
m-sample, line 28: 👍

rfcs/text/0030/process.yml
m-sample: Yes, this is cleaner - the file descriptor information was unnecessary so re-using the FileDescription struct was confusing.  A TTY may be attached to several open FDs (often stdin, stdout & stderr).

rfcs/text/0030/process.yml
```diff
@@ -127,3 +127,26 @@
...
+      description: >
+        Information about the controlling TTY device. If set, the process belongs to an interactive session.
+
+    - name: tty.char_device.major
+      level: extended
+      type: long
+      short: The TTY character device's major number.
+      description: >
+        The major number identifies the driver associated with the device.
+      example: 1
```
m-sample, line 144: it might be worth adding this sentence:  
```the character device's major and minor numbers can be algorithmically combined to produce more the familiar terminal identifiers such as "ttyS0" and "pts/0.  For more details see https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html```

rfcs/text/0030/user.yml
m-sample: We not going to track the real, effective & saved uids & usernames of all processes in the event (leaders etc), just the .process one now?

mitodrummer: we do by the fact that user can be nested under process. As far as I'm aware, this means the leader processes will have this too? @kgeller let me know if that isn't correct.

rfcs/text/0030/group.yml
kgeller: The way it works now, the user.real, user.effective and user.saved re-uses will not come through to process, process.entry_leader process.session_leader, process.group_leader re-uses, only user itself will come through. 

If this is an issue, we can definitely chat and work through it.

rfcs/text/0030/user.yml
kgeller: Yes, the leader processes will get the user and user.group fieldsets. 

The caveat to that, like I mention above, is, as of right now, it won't get the user.effective, user.saved and user.real.

mitodrummer: hmm, ok so the double nestings won't make the trip.. which means we can't have   process.entry_leader.user.real.name  etc...  cc @m-sample... process.entry_leader.user and process.entry_leader.group makes the trip, but due to the tooling, the self nestings are evaluated after the foreign ones (explained by @kgeller).. 

Do you think this is an issue Mike? We will have the rich nestings for user/group on the top level process, but perhaps fine if they don't get set for leaders?

m-sample: The Linux process model has real, effective, saved (and file which we leave out) *per* process so tracking them for each process referenced in an event is valuable (e.g. see if privileges have increased via a setuid exec or dropped by setuid, etc system calls.). Note for future: we should consider adding the capability set of a process so we can see when privileges change

For now it seems we don't have query a language that can do cross field comparison (e.g. process.session_leader.effective.uid != process.effective.uid).   We have resorted to computed synthetic fields like same_as_process to make some of these comparisons queriable.  Do we need a synthetic field such as "same_users_as_process" for these too?

mitodrummer: after a few discussions we are removing the self nesting for user and group and breaking out into individual bespoke nestings under process.

rfcs/text/0030/group.yml
mitodrummer: update commiitted to remove self nesting.

PR #1834, opened 2022-03-08
Add optional `pattern` attribute

### LINKED ISSUE #1741: Allow regex pattern to be specified in schema to improve data consistency
**Summary**

Add information to the schema to support automated data validation. I propose allowing a regular expression field that tools can use to validate a value.

For example `source.mac` would give a pattern that checks for RFC 7042 format in uppercase.

```yaml
- name: source.mac
  pattern: '^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$'
```

[JSON schema](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.8) has a similar concept and supports many more validation types, but I think have a regular expression would be a good place to start.

**Motivation**:

https://github.com/elastic/elastic-package/issues/615

I'm trying to improve the consistency of values contained in ECS fields through automated tests.

**Detailed Design**:

It would be good to go through the schema and see how many places this could be applied.

ebeahan:
### Overview

Introduces a new optional field definition attribute: `pattern`. The pattern attribute holds a regular expression (regex) which expresses the expected constraint on a string field's value.

```
pattern: '[.*]'
```

The ECS tooling checks the regex itself to ensure validity. However, The ECS tooling cannot evaluate the regex against provided field values.

### Motivation

Allow automated testing to use the regex found in the `pattern` fields against values populating ECS fields.

#### Regex validation

If specified, the regex value from `pattern` goes through a check using `re.compile` if the `--strict` flag is present. The generator will only generate a warning if `--strict` isn't specified.

#### Generated artifacts

The `pattern` field appears in the following generated artifacts:

* `generated/beats/fields.ecs.yml`
* `generated/ecs/ecs_flat.yml`
* `generated/ecs/ecs_nested.yml`

#### Misc refactoring

I took the opportunity to dedup how `--strict` checks are handled in `schema.cleaner`.

### Example usage

```yaml
    - name: mac
      level: core
      type: keyword
      short: MAC address of the source.
      example: 00-00-5E-00-53-23
      pattern: '^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$'
      description: >
        MAC address of the source.

        The notation format from RFC 7042 is suggested: Each octet (that is, 8-bit byte) is
        represented by two [uppercase] hexadecimal digits giving the value of
        the octet as an unsigned integer. Successive octets are separated by a
        hyphen.
```

Resolves #1741

### DIFF
--- a/scripts/schema/cleaner.py
+++ b/scripts/schema/cleaner.py
@@ -16,6 +16,7 @@
 # under the License.
 
 import copy
+import re
 
 from generators import ecs_helpers
 from schema import visitor
@@ -205,6 +206,8 @@ def field_assertions_and_warnings(field):
         check_example_value(field, strict=strict_mode)
         if 'beta' in field['field_details']:
             single_line_beta_description(field, strict=strict_mode)
+        if 'pattern' in field['field_details']:
+            validate_pattern_regex(field['field_details'], strict=strict_mode)
         if field['field_details']['level'] not in ACCEPTABLE_FIELD_LEVELS:
             msg = "Invalid level for field '{}'.\nValue: {}\nAcceptable values: {}".format(
                 field['field_details']['name'], field['field_details']['level'],
@@ -229,13 +232,18 @@ def single_line_short_check(short_to_check, short_name):
     return None
 
 
+def strict_warning_handler(message, strict):
+    """Handles warnings based on --strict mode"""
+    if strict:
+        raise ValueError(message)
+    else:
+        ecs_helpers.strict_warning(message)
+
+
 def single_line_short_description(schema_or_field, strict=True):
     error = single_line_short_check(schema_or_field['field_details']['short'], schema_or_field['field_details']['name'])
     if error:
-        if strict:
-            raise ValueError(error)
-        else:
-            ecs_helpers.strict_warning(error)
+        strict_warning_handler(error, strict)
 
 
 def single_line_short_override_description(schema_or_field, strict=True):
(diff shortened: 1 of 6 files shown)


ebeahan:
@andrewkroh Does the proposed implementation match up with your expectations?

I added regex validation ECS-side, but I know the support of expressions can vary between different languages/engines/implementations. I'm open to removing the validation step if it's best to remove any sort of regexp validation ECS-side.

scripts/schema/cleaner.py
```diff
@@ -244,10 +251,7 @@ def single_line_short_override_description(schema_or_field, strict=True):
...
         error = single_line_short_check(field['short_override'], field['full'])
         if error:
-            if strict:
-                raise ValueError(error)
-            else:
-                ecs_helpers.strict_warning(error)
+            strict_warning_handler(error, strict)
 
 
 def check_example_value(field, strict=True):
```
andrewkroh, line 257: Maybe `check_example_value` should verify the example are valid per the `pattern`.

andrewkroh (commented):
LGTM. Thanks for adding this to the schema.

scripts/schema/cleaner.py
ebeahan: Good idea! Added in https://github.com/elastic/ecs/pull/1834/commits/14bf0d151f5c9ff1c340b7fbde93bd4e693169a9

ebeahan: Good idea! Added in https://github.com/elastic/ecs/pull/1834/commits/14bf0d151f5c9ff1c340b7fbde93bd4e693169a9

andrewkroh (approved):

PR #1838, opened 2022-03-14
Reuse as array definition

### LINKED ISSUE #1736: Mark which field set reuses expect arrays

kgeller:
Closes #1736 

By adding a normalize as array attribute a re-use definition, users will now be able to re-use a field set as an array.

Example: 
```
- at: process
  as: previous
  short_override: An array of all previous executions for the process, including the initial fork details.
  normalize:
    - array
```
-------->

This makes it so we have a note in the docs that we intend an array as seen here:
Screen Shot 2022-02-28 at 4 04 54 PM

### DIFF
--- a/schemas/README.md
+++ b/schemas/README.md
@@ -53,7 +53,7 @@ multiple places, like for example `geo`, which can appear under `source`, `desti
 }
 ```
 
-The `reusable` attribute is composed of `top_level`, `expected`, and `short_override` sub-attributes:
+The `reusable` attribute is composed of a few sub-attributes:
 
 - top\_level (optional, default true): Is this field set expected at the root of
   events or is it only expected in the nested locations?
@@ -61,6 +61,8 @@ The `reusable` attribute is composed of `top_level`, `expected`, and `short_over
   There are two valid notations to list expected locations.
 - short_override (optional, default null): Sets the short description for the
   nested field, overriding the default top-level short description.
+- normalize: Normalization steps that should be applied at ingestion time. Supported values:
+  - array: the content of the field should be an array (even when there's only one value).
 
 The "flat" (or dotted) notation to represent where the fields are nested:
 
--- a/scripts/templates/field_details.j2
+++ b/scripts/templates/field_details.j2
@@ -136,6 +136,12 @@ Note also that the `{{ fieldset['name'] }}` fields are not expected to be used d
 | {{ entry['short'] }}
 {%- endif %}
 
+{% if entry['normalize'] and 'array' in entry['normalize'] -%}
+
+Note: this reuse should contain an array of <fieldset> objects.
+
+
+{% endif %}
 // ===============================================================
 
 
--- a/scripts/templates/field_details.j2
+++ b/scripts/templates/field_details.j2
@@ -140,7 +140,6 @@ Note also that the `{{ fieldset['name'] }}` fields are not expected to be used d
 
 Note: this reuse should contain an array of <fieldset> objects.
 
-
 {% endif %}
 // ===============================================================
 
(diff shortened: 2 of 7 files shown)


scripts/templates/field_details.j2
```diff
@@ -136,6 +136,11 @@ Note also that the `{{ fieldset['name'] }}` fields are not expected to be used d
 | {{ entry['short'] }}
 {%- endif %}
 
+{% if entry['normalize'] and 'array' in entry['normalize'] -%}
+
+Note: this reuse should contain an array of <fieldset> objects.
+
+{% endif %}
```
ebeahan, line 143:
```suggestion
{% endif -%}
```

Nit: Jinja2 whitespace can be obnoxious. If you want to drop the extra blank lines now showing up in `field-details.asciidoc`, you can adjust the whitespace trimming in the template.

It's also OK with the extra lines, though; it shouldn't change how it renders.

schemas/README.md
```diff
@@ -53,14 +53,16 @@ multiple places, like for example `geo`, which can appear under `source`, `desti
...
-The `reusable` attribute is composed of `top_level`, `expected`, and `short_override` sub-attributes:
+The `reusable` attribute is composed of a few sub-attributes:
 
 - top\_level (optional, default true): Is this field set expected at the root of
   events or is it only expected in the nested locations?
 - expected (default []): list of places the field set's fields are expected.
   There are two valid notations to list expected locations.
 - short_override (optional, default null): Sets the short description for the
   nested field, overriding the default top-level short description.
+- normalize: Normalization steps that should be applied at ingestion time. Supported values:
```
ebeahan, line 64: Thank you for documenting!

scripts/templates/field_details.j2
```diff
@@ -136,6 +136,11 @@ Note also that the `{{ fieldset['name'] }}` fields are not expected to be used d
 | {{ entry['short'] }}
 {%- endif %}
 
+{% if entry['normalize'] and 'array' in entry['normalize'] -%}
+
+Note: this reuse should contain an array of <fieldset> objects.
```
ebeahan, line 141: Is `<fieldset>` a placeholder? Right now this will generate including the brackets:

Screen Shot 2022-03-14 at 5 27 15 PM

If you want, you can have the template render the field set name:

```
Note: this reuse should contain an array of {{ entry['name'] }} field set objects.
```

scripts/templates/field_details.j2
kgeller: Thank you for this! I was trying to figure this out for so long and couldn't

ebeahan (approved):
Please add a changelog entry, but otherwise, LGTM! 🚢

PR #1841, opened 2022-03-15
Remove brackets from `email.message.id` example

### LINKED ISSUE #1836: Example data in `email.message_id`
The value shown as an example for `email.message_id` contains angle brackets which is a part of the formatting, but not the actual message-id.

[RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.4) 3.6.4 Identification Fields:
> Semantically, the angle bracket characters are not part of the msg-id; the msg-id is what is contained between the two angle bracket characters.

Currently, all examples for `email.message_id` use the value `<81ce15$8r2j59@mail01.example.com>`. This should be changed to not include the angle brackets.

In my opinion, when viewing the email header as a whole it's easy to tell we only want the values inside the brackets. This is also how all the other ECS examples are currently written (from/reply_to/to/cc).
```
From: Elastic Machine <notifications@github.com>
Reply-To: "elastic/elasticsearch" <noreply@github.com>
To: "elastic/elasticsearch" <elasticsearch@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Message-ID: <elastic/elasticsearch/releases/61306590@github.com>
```

ebeahan:
Per [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.4), the angle brackets (`<` and `>`) enclose the `Message-ID` field but are not part of its value.

Resolves #1836


ebeahan:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.2|[](https://github.com/elastic/ecs/pull/1849)|


### Questions ?

PR #1842, opened 2022-03-17
Implement proposed changes from RFC 00300 - linux event model fields

kgeller:
Implement the schema changes for linux event model proposed in [RFC 0030](https://github.com/mitodrummer/ecs/blob/stage-3-linux-event-model/rfcs/text/0030-linux-event-model.md) and updated in https://github.com/elastic/ecs/pull/1826


schemas/group.yml
```diff
@@ -29,6 +29,24 @@
...
+        beta: Reusing the `group` fields in this location is currently considered beta.
+      - at: process
+        as: real_group
+        short_override: The real group (rgid).
+        beta: Reusing the `group` fields in this location is currently considered beta.
+      - at: process
+        as: saved_group
+        short_override: The saved group (sgid).
+        beta: Reusing the `group` fields in this location is currently considered beta.
+      - at: group
```
mitodrummer, line 44: should be: 

at: process
as: supplemental_groups

schemas/process.yml
```diff
@@ -32,6 +32,44 @@
       - at: process
```
mitodrummer, line 32: As part of changes to process.yml, we were planning on dropping the pgid property, but since my process.yml is just deltas i snuck that detail in via a comment.

mitodrummer: process.group_leader.pid   === process.pgid

process.parent.group_leader.pid === process.parent.pgid

schemas/process.yml
```diff
@@ -197,3 +235,94 @@
...
+    - name: entry_meta.type
+      level: extended
+      type: keyword
+      beta: This field is beta and subject to change.
+      short: The entry type for the entry session leader.
+      description: >
+        The entry type for the entry session leader.
+        Values include: init(e.g systemd), sshd, ssm, kubelet, teleport, terminal, console
+
+    - name: entry_meta.source
```
mitodrummer, line 295: Just realized we have this defined here as well as in source.yml. I would guess we should omit from process.yml?

mitodrummer: do we need a -name: entry_meta to replace it? so source.yml has something to attach to?  kinda like below with the 'tty' field?

schemas/source.yml
```diff
@@ -28,6 +28,13 @@
     and destination details from a network transaction. If the event also contains identification of the
     client and server roles, then the client and server fields should also be populated.
   type: group
+  reusable:
+    top_level: true
+    expected:
+      - at: process.entry_meta
+        as: source
```
mitodrummer, line 35: perhaps we should add a short_description here? e.g
"Remote client information such as ip, port and geo location."

mitodrummer:
Just a few notes, but overall looking good! Did you want me to merge my PR instead of making changes to rfcs/text/0030/* stuff? I guess it's not a big deal to resolve those conflicts though.

schemas/process.yml
kgeller: Good catch, I will remove it from process

We can define it if you want to be explicit, but it isn't necessary

schemas/process.yml
kgeller: I've added a deprecation notice, with guidance on the updated location

kgeller:
> Just a few notes, but overall looking good! Did you want me to merge my PR instead of making changes to rfcs/text/0030/* stuff? I guess it's not a big deal to resolve those conflicts though.

So these changes are to the actual schema, so we can get your fields out there as beta and officially into ECS.

I think we hold off on your PR for now just since stage 3 signifies the addition is finished. Unless, you wanted to change it to a stage 2 update and we hold that Stage 3 label for a little longer.

mitodrummer:
> 

oh yea no worries, it can wait.

schemas/process.yml
```diff
@@ -197,3 +239,94 @@
...
       description: >
         The time the process ended.
+    
+    - name: interactive
+      level: extended
+      type: boolean
+      example: true
+      beta: This field is beta and subject to change.
+      short: Whether the process is connected to an interactive shell.
+      description: >
```
ebeahan, line 249: Also, I suggest some additional line breaks here for readability.

schemas/process.yml
```diff
@@ -197,3 +239,94 @@
...
+        process.entry_leader.same_as_process: true
+        Note: This field is only set on process.entry_leader, process.session_leader and process.group_leader.
+
+    - name: env_vars
+      level: extended
+      type: object
+      beta: This field is beta and subject to change.
+      short: Environment variables set at the time of the event.
+      description: >
+        Environment variables set at the time of the event.
```
ebeahan, line 281: Nit: Env vars is self-evident, but maybe include the shorthand in the description to make it explicit. 

```suggestion
        Environment variables (`env_vars`) set at the time of the event.
```

schemas/process.yml
```diff
@@ -197,3 +239,94 @@
...
+      description: >
+        Information about the controlling TTY device. If set, the process belongs to an interactive session.
+        
+    - name: tty.char_device.major
+      level: extended
+      type: long
+      beta: This field is beta and subject to change.
+      short: The TTY character device's major number.
+      description: >
+        The major number identifies the driver associated with the device. The character device's major and minor numbers can be algorithmically combined to produce the more familiar terminal identifiers such as "ttyS0" and "pts/0. For more details see https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html
```
ebeahan, line 322: Nit: Can we drop the reference URL in favor of something generic: `For more details, refer to the Linux kernel documentation`?

Saves ourselves from having to remove a broken link if the URL changes in the future.

ebeahan (commented):
With all the introduced fields and reuses, this PR more than doubles the existing total ECS field count. However, this is because the current ECS tooling is fully populating all the reuses to build its artifacts.

From the ECS side, we're planning on submitting a follow-up PR shortly to trim this total back using filtering for some of the `process.*` reuese. But that's not a blocker for moving forward with these changes.

FYI @mitodrummer - ECS will base its `process.*`  filtering on your work in the [endpoint-package](https://github.com/elastic/endpoint-package/blob/49d9a9ac02762afa3ad158febe5edc164883d715/custom_subsets/elastic_endpoint/process/linux_event_model_event.yaml#L27)

schemas/process.yml
```diff
@@ -197,3 +239,94 @@
...
+        Process interactivity is inferred from the processes file descriptors. If the character device for the controlling tty is the same as stdin and stderr for the process, the process is considered interactive.
+        Note: A non-interactive process can belong to an interactive session and is simply one that does not have open file descriptors reading the controlling TTY on FD 0 (stdin) or writing to the controlling TTY on FD 2 (stderr). A backgrounded process is still considered interactive if stdin and stderr are connected to the controlling TTY.
+   
+    - name: same_as_process
+      level: extended
+      type: boolean
+      example: true
+      beta: This field is beta and subject to change.
+      short: This boolean is used to identify if a leader process is the same as the top level process.
+      description: >
```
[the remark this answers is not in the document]
kgeller: [Better](https://ecs_1842.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-process.html#field-process-same-as-process)?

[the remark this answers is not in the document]
ebeahan: Yes! Thank you.

ebeahan (approved):
LGTM

mitodrummer:
> ing for some of the

Cool, yea @kqualters-elastic opened this PR which has a lot of nested mappings on the process fieldsets, so glad to hear you are going to filter some of those out, as I imagine this adds quite a large overhead to indexes making use of these new mappings. https://github.com/elastic/kibana/pull/128286/files

PR #1850, opened 2022-03-22
Follow up on the process mapping filter PR.

mitodrummer:
Removed a few unnecessary mappings from process. cc @ebeahan @kqualters-elastic

### DIFF
--- a/schemas/subsets/main.yml
+++ b/schemas/subsets/main.yml
@@ -139,15 +139,6 @@ fields:
             fields:
               id: {}
               name: {}
-          pe:
-            fields: "*"
-          pgid: {}
-          thread:
-            fields:
-              id: {}
-              name: {}
-          title: {}
-          uptime: {}
       env_vars: {}
       executable: {}
       exit_code: {}
@@ -213,7 +204,6 @@ fields:
             fields: "*"
           end: {}
           entity_id: {}
-          env_vars: {}
           executable: {}
           exit_code: {}
           group_leader:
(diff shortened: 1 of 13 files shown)


schemas/subsets/main.yml
```diff
@@ -366,10 +356,6 @@ fields:
       tty:
         fields: "*"
       uptime: {}
-      user:
```
ebeahan, line 369: Even if these won't be used right under `process.*`, with the way the re-uses render in the ECS docs, not having `process.user.*` exist at all could create confusion.

@kgeller WDYT?

mitodrummer: Yea, there is a bit of contention on whether we include process.user since endpoint has always used the top level user to track the user for the top level process, and wanted to keep it as such. I do agree this could create confusion. cc @ferullo @rickrobbins. Considering user is only an id and name field in this nesting, are we ok with populating both?

ebeahan: I agree to populate the top-level `user.*` over `process.user.*`.

Leaving it in ECS just leaves the reuse to "exist" so when `process.*` is reused, it makes sense how that's chained together -> `process.*` contains `process.user.*` which is reused as `process.entry_leader.user.*`

kgeller: I agree with Eric, I think we keep its existence, even if there are no plans for endpoint to utilize it.

mitodrummer: Ok, i'll add this back in.

ebeahan (approved):
LGTM

PR #1871, opened 2022-04-11
Add pattern attribute to `.mac` fields

### LINKED ISSUE #1846: Set pattern field for `mac` fields
Implement a regex pattern to validate values using the `pattern` attribute from #1834 

Proposed regex: `^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$`

ebeahan:
### Overview

This PR sets the `pattern` attribute for the `.mac` address fields. The regex value is based on the format suggested in [IETF RFC 7042](https://datatracker.ietf.org/doc/html/rfc7042#section-1.1).

### Other changes

* The pattern checking added in #1834 didn't handle examples with array values. If `array` is set in the `normalized` attribute, the `example` value is now treated as an array.
* An example value not matching the `pattern` regex will cause an exception if using `--strict.` This is now documented in `USAGE.md.`

Resolves #1846

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -408,6 +408,7 @@ Strict mode requires the following conditions, else the script exits on an excep
 
 * Short descriptions must be less than or equal to 120 characters.
 * Example values containing arrays or objects must be quoted to avoid unexpected YAML interpretation when the schema files or artifacts are relied on downstream.
+* If a regex `pattern` is defined, the example values will be checked against it.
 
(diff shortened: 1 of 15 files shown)


kgeller (approved):

ebeahan:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.3|[](https://github.com/elastic/ecs/pull/1872)|


### Questions ?

PR #1889, opened 2022-04-20
Add orchestrator (resource.parent_type and resource.ip) and container (container.image.hash) fields

daniel-almeida:
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?yes

### DIFF
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -1092,6 +1092,22 @@ type: keyword
 // ===============================================================
 
 |
+[[field-container-image-hash]]
+<<field-container-image-hash, container.image.hash>>
+
+| Hash (content-addressable identifier) of the image the container was built on.
+
+type: keyword
+
+
+
+
+
+| extended
+
+// ===============================================================
+
+|
 [[field-container-image-name]]
 <<field-container-image-name, container.image.name>>
 
@@ -6419,6 +6435,22 @@ type: keyword
 // ===============================================================
 
 |
+[[field-orchestrator-resource-ip]]
+<<field-orchestrator-resource-ip, orchestrator.resource.ip>>
+
+| IP address of the resource being acted upon.
+
+type: ip
+
+
+
+
+
+| extended
+
+// ===============================================================
+
+|
 [[field-orchestrator-resource-name]]
 <<field-orchestrator-resource-name, orchestrator.resource.name>>
 
@@ -6429,6 +6461,22 @@ type: keyword
 
 
 example: `test-pod-cdcws`
+
+| extended
+
+// ===============================================================
+
+|
+[[field-orchestrator-resource-parent-type]]
+<<field-orchestrator-resource-parent-type, orchestrator.resource.parent_type>>
+
+| Type or kind of the parent resource of the resource being acted upon.
+
+type: keyword
+
+
+
+example: `DaemonSet`
 
 | extended
 
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -897,6 +897,13 @@
       type: keyword
       ignore_above: 1024
       description: Unique container id.
+    - name: image.hash
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: Hash (content-addressable identifier) of the image the container
+        was built on.
+      default_field: false
     - name: image.name
       level: extended
       type: keyword
@@ -4512,12 +4519,25 @@
       ignore_above: 1024
       description: Unique ID of the resource being acted upon.
       default_field: false
+    - name: resource.ip
+      level: extended
+      type: ip
+      description: IP address of the resource being acted upon.
+      default_field: false
     - name: resource.name
       level: extended
       type: keyword
       ignore_above: 1024
       description: Name of the resource being acted upon.
       example: test-pod-cdcws
+      default_field: false
+    - name: resource.parent_type
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: Type or kind of the parent resource of the resource being acted
+        upon.
+      example: DaemonSet
       default_field: false
     - name: resource.type
       level: extended
--- a/generated/beats/fields.ecs.yml
+++ b/generated/beats/fields.ecs.yml
@@ -847,6 +847,13 @@
       type: keyword
       ignore_above: 1024
       description: Unique container id.
+    - name: image.hash
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: Hash (content-addressable identifier) of the image the container
+        was built on.
+      default_field: false
     - name: image.name
       level: extended
       type: keyword
@@ -4462,12 +4469,25 @@
       ignore_above: 1024
       description: Unique ID of the resource being acted upon.
       default_field: false
+    - name: resource.ip
+      level: extended
+      type: ip
+      description: IP address of the resource being acted upon.
+      default_field: false
     - name: resource.name
       level: extended
       type: keyword
       ignore_above: 1024
       description: Name of the resource being acted upon.
       example: test-pod-cdcws
+      default_field: false
+    - name: resource.parent_type
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: Type or kind of the parent resource of the resource being acted
+        upon.
+      example: DaemonSet
       default_field: false
     - name: resource.type
       level: extended
--- a/schemas/container.yml
+++ b/schemas/container.yml
@@ -75,6 +75,12 @@
       normalize:
         - array
 
+    - name: image.hash
+      level: extended
+      type: keyword
+      description: >
+        Hash (content-addressable identifier) of the image the container was built on.
+
     - name: labels
       level: extended
       type: object
--- a/schemas/orchestrator.yml
+++ b/schemas/orchestrator.yml
@@ -83,6 +83,19 @@
       description: >
         Type of resource being acted upon.
 
+    - name: resource.parent_type
+      level: extended
+      type: keyword
+      example: DaemonSet
+      description: >
+        Type or kind of the parent resource of the resource being acted upon.
+
+    - name: resource.ip
+      level: extended
+      type: ip
+      description: >
+        IP address of the resource being acted upon.
+
     - name: resource.id
       level: extended
       type: keyword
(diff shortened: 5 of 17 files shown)


schemas/container.yml
```diff
@@ -75,6 +75,12 @@
       normalize:
         - array
 
+    - name: image.hash
```
ebeahan, line 78: For container images, is there one hash type used consistently, like `sha-256`?

If so, I wonder if we should include the type in the name. It's a convention used elsewhere in ECS (for example, [`tls.client.hash.sha256`](https://www.elastic.co/guide/en/ecs/current/ecs-tls.html#field-tls-client-hash-sha256), and this approach allows for adding other hash types later, if necessary.

schemas/orchestrator.yml
```diff
@@ -83,6 +83,19 @@
       description: >
         Type of resource being acted upon.
 
+    - name: resource.parent_type
```
ebeahan, line 86: Maybe `resource.parent.type`? It's nice to have flexibility in the future with a `resource.parent.*` namespace.

schemas/orchestrator.yml
```diff
@@ -83,6 +83,19 @@
...
+      type: keyword
+      example: DaemonSet
+      description: >
+        Type or kind of the parent resource of the resource being acted upon.
+
+    - name: resource.ip
+      level: extended
+      type: ip
+      description: >
+        IP address of the resource being acted upon.
```
ebeahan, line 97:
```suggestion
        IP address of the resource being acted upon.
     normalize:
       - array
```

I suggest we specify this field to expect an array of values.

I'm not that familiar with k8s networking, but I imagine there are dual-stack deployments where a pod could have IPv4 and IPv6 addresses? Or perhaps primary and secondary IPs are assigned to a single pod or deployment?

ebeahan, line 97: > IP address of the resource being acted upon.

Can this description be expanded? Not overly detailed, but an additional sentence or two describing why an event would populate `orchestrator.resource.ip` in addition to `host.ip`?

ebeahan (commented):
Changes look good, @daniel-almeida. I just had a few thoughts about the naming.

Thanks for adding the changelog.next entry!

schemas/container.yml
daniel-almeida: For sure; it's usually sha-256. Maybe that was after your comments, but I changed this last week and included an example and description.  This value would look like this `sha256:f8fefc80e3273dc756f288a63945820d6476ad64883892c771b5e2ece6bf1b26`.

That said, I don't see any reason to deviate from the ECS convention. I'll change it.

schemas/orchestrator.yml
daniel-almeida: I believe you're correct. k8s does support dual-stack networking.

daniel-almeida: I added a short description that [I think] is better than what I had there before and a longer description with an example in the context of k8s. How does that look now?

ebeahan (approved):
LGTM!

docs/fields/field-details.asciidoc
```diff
@@ -6414,6 +6430,25 @@ type: keyword
...
+| extended
+
+// ===============================================================
+
+|
+[[field-orchestrator-resource-ip]]
+<<field-orchestrator-resource-ip, orchestrator.resource.ip>>
+
+| IP address assigned to the resource being acted upon by the orchestrator. In Kubernetes, for instance, this could be the IP of a Pod (as opposed to the Node on which the Pod is running).
+
```
m-sample, line 6442: minor edit suggestion:   "being acted upon by the orchestrator" -> "associated with the event being observed"

Why: events/activities in a Pod may from more than the just orchestrator - e.g. I'm kubectl exec'ing a shell in there and every time I run a command that creates some execution events.

docs/fields/field-details.asciidoc
```diff
@@ -6414,6 +6430,25 @@ type: keyword
...
+type: ip
+
+
+Note: this field should contain an array of values.
+
```
m-sample, line 6447: should be put a caveat comment for Pods. E.g.   "In the case of Kubernetes Pods, there will be one IP address in this array"

docs/fields/field-details.asciidoc
```diff
@@ -6434,6 +6469,22 @@ example: `test-pod-cdcws`
 
 // ===============================================================
 
+|
+[[field-orchestrator-resource-parent-type]]
+<<field-orchestrator-resource-parent-type, orchestrator.resource.parent.type>>
+
+| Type or kind of the parent resource of the resource being acted upon.
+
```
m-sample, line 6477: related to above comment, this would be clearer if "being acted upon" -> "associated with the event"

docs/fields/field-details.asciidoc
```diff
@@ -6434,6 +6469,22 @@ example: `test-pod-cdcws`
...
+type: keyword
+
+
+
+example: `DaemonSet`
+
```
m-sample, line 6483: ideally we would enumerate the values we currently expect along with whether they are case-sensitive etc. This will help sensor writers. Perhaps we could reference a Kube docs page for the list of 'nouns' and standard spelling & capitalization.

experimental/generated/beats/fields.ecs.yml
```diff
@@ -897,6 +897,13 @@
       type: keyword
       ignore_above: 1024
       description: Unique container id.
+    - name: image.hash.sha256
+      level: extended
```
m-sample, line 901: by having a separate field for each hash type (sha256, ....) we will need to update ECS for every new hash alg in use (e.g. "blake3" becomes popular next year).   There may be an expectation that all of them are populated (ie we compute them vs just scooping the one from the Kube API).   

Pardon the last minute observation on this but this approach seems very maintenance heavy (us) vs just having   "<hashalg>:<b64 of hash>" as the value's structure for image.hash field.   This approach keeps it to one field to look in and automatically supports new hash algs, and parsing isn't that hard - most folks will be looking for an equality match, I suspect so they could type "sha256:ab1234ace..." for their search.

ebeahan: I understand that perspective, and I'm fine if we decide on `image.hash`.

My thinking around `container.image.hash.sha256`:

* Follow the naming convention used by the [`hash.*` field set](https://www.elastic.co/guide/en/ecs/current/ecs-hash.html) and its reuses, like `file.hash.sha256`.
* If the `related.*` fields are being populated, [`related.hash`](https://www.elastic.co/guide/en/ecs/current/ecs-related.html#field-related-hash) helps simplify searching across the different hash fields.
* Avoid establishing `image.hash` as a leaf field

m-sample: interesting.   related.hash as an array of all the hash values doesn't seem to help as there would only be 1 entry in the array and it wouldn't be self-describing (could be sha256, sha512, ?).

I agree we should not overload "hash" to be a the existing field set or this new primitive valued one :head-explode:.

How does this sound:   Add a field to the "hash" field set, something like hash.all that holds self-describing hash values like <hash-alg>:<b64-hash-bytes>.   Sort of related.hash but co-located with hash field set and self-describing.  This would accommodate the current case of 1 hash value read from the kube API, as well as possible future use cases where we populate the different hash algs values too.  Allowed values for the hash-alg part would be the existing field names in the hash fieldset?  Thankfully they seem to align with Kubernetes use of "sha256" (same as ECS field name).

docs/fields/field-details.asciidoc
daniel-almeida: This note was auto-generated. I updated the long description to cover that as well.

experimental/generated/beats/fields.ecs.yml
daniel-almeida: Mike, just to clarify: you're suggesting both fields (hash.sha256 _and_ hash.all) should be set by producers, right?

m-sample: Just hash.all to start

m-sample: > > Add a field to the "hash" field set, something like hash.all that holds an array of self-describing hash values like :
> 
> I think adding `hash.all` under the `hash.*` field set would confuse users over when to populate `user.all` vs the existing fields. If capturing the hash as `<hash-alg>:<b64-hash-bytes>` is a convention used by Kubernetes, maybe `orchestrator.image.hash.all`?

`orchestrator.image.hash.all`  is exactly what I was suggesting. If we want overlay that `all` field just there (vs adding `all` to the `hash` field set in general), that works for me.  Using something other than `all` would be fine too; I'm chasing agreement on the structure concept vs naming.

m-sample: > The `related.*` fields don't help describe the values but exist to help search. Here's an example I put together to better describe their intended usage: [example](https://gist.github.com/ebeahan/8cb6cd4e679114c09556449213f6ec96).

Thanks for the example.  My concern with `related` is that used in conjunction with the associated hash.sha256 and other field values we're duplicating data. My sense is that size does matter (the smaller the better).  If we just populated `related.hash` we'd lose information (hash alg).

daniel-almeida: Are we happy with `orchestrator.image.hash.all = ["sha256: f7988fb6c02e0ce69257d9bd9cf37ae20a60f1df7563c3a2a6abe24160306b8d"]` then?

m-sample: That works for me - no losing information (hash alg), and no duplicates in the same event.  Any concern's @ebeahan ?  Perhaps different naming?

ebeahan: Works for me 👍

daniel-almeida: I think we meant to say `container.image...` instead of `orchestrator.image...`. Just pushed that change; let me know if it's good to go.

generated/beats/fields.ecs.yml
```diff
@@ -847,6 +847,14 @@
       type: keyword
       ignore_above: 1024
       description: Unique container id.
+    - name: image.hash.all
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: 'An array of digests of the image the container was built on. Each
+        digest consists of the hash algorithm and value in this format: `algorithm:value`.'
```
m-sample, line 855: I would just add "Algorithm names in this value should align with the field names in the ECS hash field set" if that makes sense.

m-sample (approved):
LGTM

experimental/generated/beats/fields.ecs.yml
daniel-almeida: @ebeahan , I noticed that hash values in the TLS field set you referenced are capitalized for consistency. Should we do the same here?

ebeahan: @daniel-almeida, ECS field names are always lowercase. Is that what you mean?

ebeahan (approved):
LGTM

ebeahan:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.3|[](https://github.com/elastic/ecs/pull/1896)|


### Questions ?

PR #1909, opened 2022-05-10
Introduce "docs_only" attribute for subsets

### LINKED ISSUE #1907: Documentation-only field descriptions
### Overview

Support declaring field descriptions only appearing in the documentation.

### Background

Currently, only fields defined in a top-level field set have their fields described in the [field reference](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html).

This proposed feature adds a field's description in the docs. However, no other artifacts (index mappings, Beats, `ecs.*`, etc.) include the field. The field is removed from other artifacts to reduce incorrect usage.

### Examples

* `same_as_process` exists at `process.entry_leader`, `process.session_leader`, and `process.group_leader` but not `process.*`
* `entry_meta.type` exists `process.entry_leader`, `process.session_leader`, and `process.group_leader` but also not at `process.*`

Fields using this option will state their intended usage in the fields' descriptions. For example, `This field is only populated on the process.entry_leader, process.session_leader, and process.group_leader reuses.`

ebeahan:
### Overview

Introduce a new attribute for subset file definitions, `docs_only.` Setting `docs_only: True` only filters the field to include it in the ECS docs.

### Example usage

```yaml
name: event
fields:
  base:
    fields:
      "@timestamp": {}
  agent:
    fields: "*"
  dll:
    fields: "*"
  ecs:
    fields: "*"
  process:
    fields:
      same_as_process:
        docs_only: True
```

### Use case

Support for situations like the `process.same_as_parent` field, which is only populated in `process` reuses. In the current documentation, a field still must be defined and described at the top level. The field's description includes the guidance of which reuses contains the field.

### Design

The changes to support `docs_only` were added to `schema.subset_filter`. If `docs_only` is defined on any subset fields, these extra steps take place:

* A second subset with only `docs_only` fields is created
* Filter a list of `docs_only` fields using this second subset
* Remove any `docs_only` fields from the main fields list. Removing the fields prevents them from appearing in any other artifacts.
* Pass the filtered list into the `ascii_doc` generator alongside the main fields.

Resolves #1907


ebeahan:
I've also added the two initial entries mentioned as examples in #1907 to the docs-only defs: `process.same_as_process` and `process.entry_meta.type`: [ECS docs preview](https://ecs_1909.docs-preview.app.elstc.co/guide/en/ecs/master/ecs-process.html#field-process-same-as-process)

scripts/generators/assets/docs_only_fields.yml
```diff
@@ -0,0 +1,45 @@
+process:
+  fields:
+    process.entry_meta.type:
```
andrewkroh, line 3: Is there a way that this can be implemented without duplicating the definitions from schemas/process.yml? I think this will be error prone in that someone makes a change to one and forgets to update the other.

Are there use case for this feature for fields that are not part of the schema? Perhaps there could be an attribute in `ecs/schemas/subset/main.yml` that could indicate a field should only be added to docs.

```
process:
    fields:
      same_as_process:
        docs_only: true
```

kgeller (approved):

andrewkroh (commented):
LGTM.

andrewkroh (approved):

ebeahan:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.2|[](https://github.com/elastic/ecs/pull/1914)|


### Questions ?

Issue #1915, opened 2022-05-17
agent.name documentation might be wrong.


philippkahr:
**Description of the issue:**

The documentation https://www.elastic.co/guide/en/ecs/current/ecs-agent.html says:

> Custom name of the agent.
> This is a name that can be given to an agent. This can be helpful if for example two Filebeat instances are running on the same host but a human readable separation is needed on which Filebeat instance data is coming from.
> If no name is given, the name is often left empty.

Agent.name can of course be anything that it want to be. Shouldn't we mention that our Beats and Elastic Agent per default use the Hostname as a value? Because the last line `If no name is given, the name is often left empty` is irritating.

ebeahan:
> Because the last line If no name is given, the name is often left empty is irritating.

I agree this sentence is unnecessary.

> Shouldn't we mention that our Beats and Elastic Agent per default use the Hostname as a value?

I vote against adding a mention for two reasons:

1. Users should query/alert/filter using `host.*` or `observer.*` fields for host name values and not `agent.*`. ECS [intentionally avoided ](https://github.com/elastic/ecs/issues/178#issuecomment-512586064) defining an `agent.hostname` field. Beats/Agents default to use the hostname, but other sources, like APM, use different conventions.
2. To avoid stating Beats/Agent defaults in ECS. The Beats or Agents docs best describe their defaults and config options and are more likely up-to-date should defaults change.

philippkahr:
Makes sense to me!

>Users should query/alert/filter using host.* or observer.* fields for host name values and not agent.*. ECS https://github.com/elastic/ecs/issues/178#issuecomment-512586064defining an agent.hostname field. Beats/Agents default to use the hostname, but other sources, like APM, use different conventions.

Maybe we should have a blog post that explains what fields someone should use starting with 8.x?

ebeahan:
Removed the unneeded sentence in #1919.

PR #1919, opened 2022-05-18
Remove unnecessary guidance from `agent.name` description

ebeahan:
Removing unnecessary guidance to address concerns discussed in #1915.


ebeahan:
Since we're planning on an `8.2.1` release, I'm planning to backport these changes into `8.2`.

kgeller (approved):

PR #1926, opened 2022-05-23
Advancing linux event model fields to GA

kgeller:
The [RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0030-linux-event-model.md) has advanced to stage 3, and we want to advance the fields to GA to officially finish the addition.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -54,6 +54,7 @@ Thanks, you're awesome :-) -->
 * Add `orchestrator.resource.parent.type` #1889
 * Add `orchestrator.resource.ip` #1889
 * Add `container.image.hash.all` #1889
+* Advanced linux event model fields and re-uses to GA. #1926
 
 #### Improvements
 
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -7248,9 +7248,7 @@ example: `c2c455d9f99375d`
 [[field-process-env-vars]]
 <<field-process-env-vars, process.env_vars>>
 
-| beta:[ This field is beta and subject to change. ]
-
-Environment variables (`env_vars`) set at the time of the event. May be filtered to protect sensitive information.
+| Environment variables (`env_vars`) set at the time of the event. May be filtered to protect sensitive information.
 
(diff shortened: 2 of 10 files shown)


ebeahan (approved):
LGTM

m-sample:
There is an opportunity for reducing the event size that we may want explore before locking this in.  

Here are some common scenarios:  the group_leader has same_as_process:true, ie the fields are the same as in .process since they represent the same process. To save data volume, one could leave group_leader empty except for `group_leader.same_as_process:true`.    Rules and threat hunting queries would have to be aware of this -  e.g.   `(process.group_leader.name: sudo AND group_leader.same_as_process: false) OR (process.name: sudo AND process.group_leader.same_as_process:true)`.     Somewhat clunky but reduces data volume.

This same approach could be applied when `process.session_leader` and `process.entry_leader` are the same process (e.g. by adding a same_as_entry_leader bool to session_leader)

For some workloads, process, session_leader, entry_leader and group_leader may all be the same (e.g. a multi-threaded/non-forking web server launched by systemd in its own session).

What are everyone's thoughts on this approach? too clever/confusing?

mitodrummer:
Despite those optimizations, I would guess we wouldn't change anything ECS related, except for maybe some descriptions to reflect the above (once the cost savings strategy is baked). So unless we think we will be removing any of these fields down the road, it's probably safe to move to GA.

mitodrummer (approved):

kgeller:
> So unless we think we will be removing any of these fields down the road, it's probably safe to move to GA.

I agree ^

However, if we are unsure at all, it is absolutely fine to hold off on this for now. There is no rush, I was just trying to wrap everything up since the RFC has been stage 3 for a little while.

kgeller:
Alright will do!

I'm going to close this PR for now, and we will revisit

Thanks @mitodrummer and @m-sample

m-sample:
> Despite those optimizations, I would guess we wouldn't change anything ECS related, except for maybe some descriptions to reflect the above (once the cost savings strategy is baked). So unless we think we will be removing any of these fields down the road, it's probably safe to move to GA.

We'd just want to get this cooked before people build up a large rule sets assuming the all *_leader.*  fields where there.

PR #1956, opened 2022-06-13
[Stage 0/1] Initial draft for IO fieldset RFC

mitodrummer:
- If submitting code/script changes, have you verified all tests pass locally using `make test`? n/a
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? no


rfcs/text/0033/io.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: data.time
+      level: extended
+      type: date
+      short: Timestamp for the line of IO
+      description: >
+        Timestamp for the line of IO
+
+        Timestamps are tracked per line, so that they can be replayed in realtime using libraries like xtermjs.org.
+
+    - name: data.value
```
qcorporation, line 41: Would it be better to be more explicit in the 'value' field.
Would there be any need in the future to separate this out from '.raw' to '.something_else'?

mitodrummer: Is this in regards to potentially storing the same info as plain text (w/o escape codes)? It's still not apparent how much of an issue this might be (formatting between words etc..), but if so, one could imagine a schema that has both:
process.io.data.value (or process.io.data.raw)
and process.io.data.text

rfcs/text/0033-tty-output.md
```diff
@@ -0,0 +1,159 @@
...
+-->
+
+## Fields
+
+<!--
+Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+process.io (nested IO fieldset)
+process.io.fd (number)
```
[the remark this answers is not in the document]
mitodrummer: cc @mattnite I think last I heard you guys said it wouldn't be possible to distinguish between stdout and stderr.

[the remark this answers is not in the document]
tabell: I don't think it will be possible to get the process-internal fd number no. It might be feasible to get the inode number of the final destination of the bytes - if that would be usefil @mmat11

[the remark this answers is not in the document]
mitodrummer: if it could be used to determine stdout vs stderr then great, but if not i'll just drop fd from the rfc.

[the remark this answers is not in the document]
mitodrummer: cc @nick-alayil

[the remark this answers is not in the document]
mmat11: I confirm that it isn't trivial to get that data. As far as I understood stdout,stderr are a symlink to a /dev/pts/X pseudo terminal and all file related data structures at our current hook (`tty_write`) refer to the latter. I also had a look at points in which our hook is called without success so I believe the "translation" happens in a different execution context

rfcs/text/0033-tty-output.md
```diff
@@ -0,0 +1,159 @@
...
+-->
+
+## Concerns
+
+<!--
+Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
+-->
+
+1. Data exfiltration. TTY output is a sensitive surface area to expose. It definitely should be featured gated, and an opt in for customers. At least until we implement some robust "regex scrubbing" mechanisms.
+2. Per event batch size should be considered. If batch size is too big, alerting on IO data becomes fuzzy, as rules are evaluated on the document level, not on an indiviual line of output. This could make it a challenge to figure out what part of the message triggered the alert. Also, some performance benchmarking should be done to see how rule evaluation performs on process.io.data.value when data.length > X
```
mmat11, line 112:
```suggestion
2. Per event batch size should be considered. If batch size is too big, alerting on IO data becomes fuzzy, as rules are evaluated on the document level, not on an individual line of output. This could make it a challenge to figure out what part of the message triggered the alert. Also, some performance benchmarking should be done to see how rule evaluation performs on process.io.data.value when data.length > X
```
typo

rfcs/text/0033-tty-output.md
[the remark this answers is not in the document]
mitodrummer: k, will remove.

rfcs/text/0033/io.yml
```diff
@@ -0,0 +1,43 @@
+- name: io
+  title: IO
+  group: 2
+  short: Fields describing a batch of input or output data produced by a process.
+  description: >
+    Fields describing a batch of input or output (IO) data produced by a process.
+    These fields should only appear on the top level process object when event.action is 'io'
+  type: group
+  reusable:
```
ebeahan, line 9: If the only foreseeable use for these fields is at `process.*`, what about adding directly under `process.*` instead of as a reuse?

rfcs/text/0033/io.yml
```diff
@@ -0,0 +1,43 @@
...
+      type: date
+      short: Timestamp for the line of IO
+      description: >
+        Timestamp for the line of IO
+
+        Timestamps are tracked per line, so that they can be replayed in realtime using libraries like xtermjs.org.
+
+    - name: data.value
+      level: extended
+      type: text
```
ebeahan, line 38: If using `text` types, I recommend evaluating [`match_only_text`](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html#match-only-text-field-type) over `text` for the better space efficiency. 

However, using `text` or `match_only_text` will use the `standard` analyzer, which indexes using grammar-based tokenization and is case-insensitive. Depending on how users interact with these fields, the search results may be unexpected.

The standard analyzer will remove the special characters when indexing the value `#!/bin/bash`, leaving only `bin` and `bash` tokens:

```
POST /_analyze
{
  "text": "#!/bin/bash"
}

{
  "tokens" : [
    {
      "token" : "bin",
      "start_offset" : 3,
      "end_offset" : 6,
      "type" : "<ALPHANUM>",
      "position" : 0
    },
    {
      "token" : "bash",
      "start_offset" : 7,
      "end_offset" : 11,
      "type" : "<ALPHANUM>",
      "position" : 1
    }
  ]
}
```

The [`wildcard` type](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#wildcard-field-type) is an alternative used a bit in ECS that won't strip special characters when indexing and is also case-sensitive.

rfcs/text/0033-tty-output.md
```diff
@@ -0,0 +1,157 @@
...
+Stage X: Provide a brief explanation of why the proposal is being marked as abandoned. This is useful context for anyone revisiting this proposal or considering similar changes later on.
+-->
+
+## Fields
+
+<!--
+Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+process.io (nested IO fieldset)
```
ebeahan, line 34: nit: Markdown's acting silly and rendering all the fields in one line.

`process.io (nested IO fieldset) process.io.data (array of objects) process.io.data.time (date) process.io.data.value (text)`

rfcs/text/0033/io.yml
mitodrummer: Ah good catch. I think it would be strange if we couldn't match on certain special characters, especially forward and back slashes. Though there are ascii control characters in this IO data which might benefit from being excluded in these searches, as they can appear between words. (an issue @m-sample and I have chatted about). 

It probably makes sense to use wildcard type. Any thoughts @m-sample ?

rfcs/text/0033-tty-output.md
mitodrummer: ah yea, i'll do a review in preview mode. thanks.

rfcs/text/0033/io.yml
m-sample: Back from vacation now, looking at this today.

mitodrummer: updated to use keyword type.

mitodrummer: Not sure why I updated to keyword type, I think I had it in my head that was what was mentioned above. Anyways this came up again today. Took a peek at https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#wildcard-field-type

It seems reasonable we'd want to use wildcard type, as I'm sure we'll have events with > 32kb of io data. And output data will most always be searched with a wildcard. Though, since process.io.data is an array of objects, each having a value that is a 'wildcard' type. Each line would need to exceed 32kb for it to be worth it? Sounds like wildcard type comes at a cost to storage and speed though.

rfcs/text/0033/process.yml
```diff
@@ -0,0 +1,61 @@
...
+- name: process
+  fields:
+
+    # minor typo fix in description for this field. "pts/0 -> "pts/0"
+    - name: tty.char_device.major
+      description: >
+        The major number identifies the driver associated with the device. The character device's major and minor numbers can be algorithmically combined to produce the more familiar terminal identifiers such as "ttyS0" and "pts/0". For more details, please refer to the Linux kernel documentation.
+
+    # new 'io' properties for tracking tty output
+    - name: io
```
lrishi, line 27: @mitodrummer given that we want this to be extensible for different io sources, and my assumption is that the event type will be "io", does it make more sense for this to be an array of objects? so that different io.sources can piggyback on the same process event.

mitodrummer: Might be cleaner to keep different sources in separate events? Perhaps we can use event.action as the source differentiator instead of process.io.source ? e.g event.action: 'tty_io|file_io|network_io'.

lrishi: That'll work too, but then 'io.source' becomes redundant? I'm also ok with this as is, it looks good, just wanted to bring this up wrt searchability and UI.

mitodrummer: yea, if we use event.action to differentiate "sources" we can drop process.io.source.

mitodrummer: Though not applicable to ECS, there is also the question of which "index" we want to write to? At the moment I'm assuming it will all go into logs-endpoint.events.process. Whether that is a good idea or not needs answering (perhaps from people on the endpoint side). Do we want to consider a new data stream? e.g  logs-endpoint.events.tty ?

m-sample: if we generalized io to be for any fd {file/socket/..} and for both reads and writes (input & output - IO), we'd still want it to align with the process.* fields that indicate which process did the writing and reading to those/that fd.   We'd likely need to keep timestamps for reads and writes to make sense of the stream too.   We'd probably have fine-grained controls on which programs or file/sockets/...  we actually track read and maybe writes for for data volume mgmt reasons.

Ideally this stream of events is serialized or has timing information for reads and writes (e.g. I could see that this read from a TCP socket was after this write of the request)

Off the top of my head, the simplest way to render these to events would be have each IO event represent  (Reads xor Writes) (never both), on a single tty/file/socket/... by a single process.   This approach would allow the event.action to be "input" or "output" and also only need one source per event (only be the controlling tty supported for now but eventually files, sockets, etc could fit into this model).   If we flush these events (ie stop aggregating read xor written data to a specific tty/file/socket) when the operation changes from read to write or write to read on the same tty/file/socket we could get an event storm from some types of program but likely rare?)

mitodrummer: "Ideally this stream of events is serialized or has timing information for reads and writes". Each entry in the data array could include this timing info (like suggested to support quasi realtime playback). Could the write mode be set at that level too?
e.g
process.io.data = [
  { time: <timestamp>, mode: 'w', raw: 'the raw data written' },
  { time: <timestamp>, mode: 'r', raw: 'the raw data read' }
];

this would solve the issue of frequent operation changes perhaps. Of course we run into the limitation elastic has on arrays of objects, so would be harder to query for just input or output events.

m-sample: I forgot to mention the search-ability should be precise (for rules generating alerts via queries on ES and threat hunting) e.g. - matching "output to this tty/file/socket containing this text" which is problem with the r/w mode in the array :(. So a separate event (r xor w) seems easiest - or a separate array. 

General purpose io support would also require b64 or something similar for binary data that cant be converted to utf8 or scoping these io events to text only capture (ie never send up binary io).

mitodrummer: to support binary data maybe we have ability to set either a "text" property or "binary/bytes" field.
e.g
For tty io
```
process.io.data = [
  { text: 'utf8 encoded text' },
  { text: 'utf8 encoded text' }
];
```
For binary data
```
process.io.data = [
  { bytes: <base64encodedvalue> },
  { bytes: <base64encodedvalue> }
];
```
Field types would be:
process.io.data.text = keyword|wildcard type
process.io.data.bytes = binary type (optimized type for storing base64 encoded bytes) <- i imagine this type skips any sort of tokenization or indexing for search.

**event.action** could then inform applications which data their dealing with. e.g tty_input, tty_output, file_input, file_output etc..

m-sample: The text entries in the array represent lines, ideally.  No such constraint possible for binary so likely don't need an array for it?  Also for event action, I'm not sure the source/dest type needs to be there? e.g event.action = {binary_input, binary_output, text_input, text_output} and the src/dst would be detailed in the event in other fields.

so 

```
event.action = {binary_input, binary_output, text_input, text_output} 

process.io.bytes = one base64 string the of read/input written/output binary data. Popluated for binary_input and binary_output event.action values.

process.io.data.[].txt = each array entry is a string, with best efforts to line-ify the data, including a trailing newline to indicate a complete line. Populated for text_input and text_output event.action values.

File: rfcs/text/0033/process.yml [hunk 8]
```diff
@@ -51,6 +51,44 @@
...
+      level: extended
+      type: number
+      description: >
+        The total number of bytes captured in this event.
+
+    - name: io.total_bytes_skipped
+      level: extended
+      type: number
+      description: >
+        The total number of bytes skipped in this event.
```
m-sample, line 64: Perhaps add clarity to help implementors:

"The total number of bytes of output that were not captured due to implementation restrictions such as buffer size limits.  Implementors should strive to ensure this value is always zero"

rfcs/text/0033/process.yml
```diff
@@ -51,6 +51,44 @@
...
+
+    - name: io.max_kilobytes_per_process_exceeded
```
m-sample, line 66: "max_bytes_per_process_exceeded" is more consistent with other field names in this RFC (even if the Integration config option measures it in kiB).

rfcs/text/0033/process.yml
```diff
@@ -51,6 +51,44 @@
...
+      level: extended
+      type: boolean
+      description: >
+        If true, the process producing the output has exceeded the max_kilobytes_per_process configuration setting.
+
+    - name: io.bytes_skipped
+      level: extended
+      type: object
+      description: >
+        An array of byte offsets and lengths denoting where IO data may have been skipped.
```
m-sample, line 76: I think this is clearer (no maybe about it):
"An array of byte offsets and lengths denoting where IO data has been skipped."

rfcs/text/0033/process.yml
```diff
@@ -51,6 +51,44 @@
...
+
+      normalize: array
+
+    - name: io.bytes_skipped.offset
+      level: extended
+      type: number
+      description: >
+        The byte offset showing when bytes were skipped.
```
m-sample, line 84: The following would be more precise:

The byte offset into this event's `io.text` (or `io.bytes` in the future) where `length` bytes were skipped.

rfcs/text/0033-tty-output.md
```diff
@@ -62,7 +67,7 @@ Stage 1: Provide a high-level description of example sources of data. This does
   {
     event: {
       kind: 'event',
-      action: 'text_output' (for now the only action type, though once could imagine values like: text_input, binary_output, binary_input?)
+      action: 'text_output' (for now the only action type, though one could imagine values like: text_input, binary_output, binary_input?)
```
m-sample, line 70: no question mark nec on binary_input (?) - e.g. in the future it could track binary data read from a file.

m-sample (approved):
Some suggested tightening of the descriptive text.   Otherwise this looks great and reflects exactly what we discussed.

m-sample (approved):
Looks great @mitodrummer.  Nice work.

mitodrummer:
@ebeahan This PR should be ready to move to the next stage.

lrishi (approved):

kgeller:
@mitodrummer this looks great.

Before merge, could you update your folder to `0035` since that's the RFC number?

mitodrummer:
> @mitodrummer this looks great.
> 
> Before merge, could you update your folder to `0035` since that's the RFC number?

Done! 

What stage do we go to next? I don't think there will be any further updates afaik. Let me know if I need to open up another PR. TY!

kgeller:
@mitodrummer what do you think about putting it this to stage 2? Then the fields can be added as beta, just like the linux event model RFC. And then down the line when you're sure it's not going to change, we can move to stage 3 and remove the beta tags.

mitodrummer:
> @mitodrummer what do you think about putting it this to stage 2? Then the fields can be added as beta, just like the linux event model RFC. And then down the line when you're sure it's not going to change, we can move to stage 3 and remove the beta tags.

Sounds good to me. Did you need me to update the stage/date in the md file before I merge?

kgeller (approved):
LGTM

PR #1957, opened 2022-06-15
[RFC] Stage 0 - New Fields For Login

mr1716:
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)?Yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process] (https://github.com/elastic/ecs/blob/main/rfcs/README.md)? Yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? Yes
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? Yes
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. Yes

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -18,6 +18,8 @@ Thanks, you're awesome :-) -->
 
 #### Improvements
 
+* Advances `threat.enrichments.indicator` to GA. #1928
+
 #### Deprecated
 
 ### Tooling and Artifact Changes
@@ -56,6 +58,8 @@ Thanks, you're awesome :-) -->
 #### Breaking changes
 
 #### Bugfixes
+
+* Added Deprecation Warning for `misspell` task #1993
 
 #### Added
 
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ all: generate experimental
 # Check verifies that all of the committed files that are generated are
 # up-to-date.
 .PHONY: check
-check: generate experimental test fmt misspell makelint
+check: generate experimental test fmt misspell_warn makelint
 	# Check if diff is empty.
 	git diff | cat
 	git update-index --refresh
@@ -85,6 +85,11 @@ misspell:
 		./build/misspell/install-misspell.sh -b ./build/misspell/bin >> /dev/null 2>&1 ; \
 	fi
 	./build/misspell/bin/misspell -error README.md CONTRIBUTING.md schemas/* docs/* experimental/schemas/*
+
+# Warn re misspell removal     
+.PHONY: misspell_warn
+misspell_warn:
+	@echo "Warning: due to lack of cross-platform support, misspell is no longer included in this task and may be deprecated in future\n"
 
 .PHONY: reload_docs
 reload_docs: generator docs
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -9089,9 +9089,7 @@ These fields are for users to classify alerts from all of their sources (e.g. ID
 [[field-threat-enrichments]]
 <<field-threat-enrichments, threat.enrichments>>
 
-a| beta:[ This field is beta and subject to change. ]
-
-A list of associated indicators objects enriching the event, and the context of that association/enrichment.
+a| A list of associated indicators objects enriching the event, and the context of that association/enrichment.
 
 type: nested
 
@@ -9110,9 +9108,7 @@ Note: this field should contain an array of values.
 [[field-threat-enrichments-indicator]]
 <<field-threat-enrichments-indicator, threat.enrichments.indicator>>
 
-a| beta:[ This field is beta and subject to change. ]
-
-Object containing associated indicators enriching the event.
+a| Object containing associated indicators enriching the event.
 
 type: object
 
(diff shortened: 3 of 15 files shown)


djptek:
Thanks @mr1716 

You can leave the RFC number as 0000 for now until the ECS Team accept the stage 0 RFC and allocate this number as per the RFC process

I think initially it would be helpful if you could focus on one top level field, so in this case that would probably be 

```
login.*
``` 

and you'd perhaps propose:

```
cloud.category
event.flagged
user.type
```

in complementary RFCs or alternatively use other ECS fields that might be suitable

mr1716:
@djptek Thanks for the feedback. I think all of these fields would be useful, and if I were to submit 2 RFCs, the 2 RFCs would be very similar, with minor changes to the proposed fields. At that point, the argument would be, why wasnt this made into 1 RFC, as making this 1 RFC would save some time reviewing this.

mr1716:
@djptek thanks for the feedback. I have updated this as to follow your suggestions. I removed the numbering and changed it to 0000, not 0033, and removed all non login.* fields. I will submit another RFC with the other fields at a later date.

djptek:
Hi @mr1716 can you update to the title & RFC content to remove references to `cloud` and better reflect this RFC's focus on `login`?

Thanks

mr1716:
@djptek thank you very much for the feedback. The title has been updated accordingly. Please review and let me know the next step

djptek:
Thanks @mr1716 

I think that to progress it would be helpful to refactor the RFC to focus solely on `login` fields.

The RFC still makes a lot of reference to e.g. `cloud`

```
|Proposed Field Name|Type|Value|Azure Field|
| --- | --- | --- | --- |
|cloud.category|keyword|Device|azure.auditlogs.properties.category|
```

and those fields would be better dealt with in a separate `cloud` rfc

djptek:
Hi @mr1716 that's great, do you think you might rename the RFC to `login` rather than `cloud` and maybe add a note somewhere in the description to capture your intention to use this in conjunction with a potential future `cloud` fieldset?

mr1716:
@djptek Would making a note in the RFC about future RFCs potentially set a bad precedent? I updated this to be a self contained login fields and that makes a great start.

djptek:
Thanks @mr1716 the changes to the text look great. Please could you rename the file to e.g. something about login rather than cloud and perhaps remove the reference to a specific provider so we can use this for other providers in a possible future? Mentioning concrete examples in the content is fine

mr1716:
@djptek I have removed some additional items.

djptek:
@mr1716 Please can you rename the md file to e.g. `0000-login-fields.md` or similar

djptek:
Hi @mr1716 Please could you move the RFC to Stage 0, Stage 1 would come later after the Stage 0 has been submitted. Also, anything you might do to help this be more generic e.g. looking at authorisation/authentication and then considering login within that context would be helpful

mr1716:
@djptek What would I do to move this to stage 0? This is my first time doing this.

rfcs/text/0000-login-fields.md
```diff
@@ -0,0 +1,152 @@
+# 0000: Login Fields RFC
+
+Stage: 1
```
djptek, line 3: @mr1716 please could you change the text `Stage: 1` to `Stage: 0`

We need to start the RFC process with a `Stage: 0` RFC

djptek:
I've added a note at the relevant line regarding beginning from Stage 0. 

The RFC will also benefit from a more generic focus e.g. looking at authorisation/authentication and then considering login within that context would be helpful, rather than orienting the RFC directly around login

djptek:
Hi @mr1716 

>> The RFC will also benefit from a more generic focus e.g. looking at authorisation/authentication and then considering login within that context would be helpful, rather than orienting the RFC directly around login

Do you think you might consider refocussing the RFC around authorisation/authentication and then considering login within that context?

mr1716:
@djptek I do agree that authorization/authentication are important and that login is part of that, but the request is for new login fields, which makes sense to title it the way it is. I could add in a description to make it mention authentication and authorization.

kgeller:
I agree with @djptek . I think that a specific `login.*` fieldset is too narrow in scope for the ECS project. 

One of our principles is to [maximize operability and reuse](https://www.elastic.co/guide/en/ecs/current/ecs-principles-design.html#_a_common_schema). With that in mind, I would say that adding fields around auth (authentication and authorization) would be a better fit, and then we could, of course, incorporate login concepts within that context.

mr1716:
@kgeller @djptek so then it would be auth.login.* then instead of login.*?

kgeller:
@mr1716 I am by no means an expert on auth, so I won't attempt specific suggestions. But, I think we'd want to rework the fields as a whole, so that they are more focused around authentication/authorization. And then we'd have some subset of fields like context or something that would describe that it's for login. 

I think having something like `auth.login.*` is still too specific.

mr1716:
@kgeller Then what would you recommend to accurately capture this detail in an ECS field? Or is it not worth capturing? I'm not going to disagree about the potential for being too specific, but then what to do to address this gap in ECS?

Because there is the need to capture information about whether some sort of 2FA and login information available.

kgeller:
@mr1716 I definitely think auth is worth capturing in ECS.

From some quick research on auth fields, I came across [this](https://docs.paloaltonetworks.com/pan-os/9-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/authentication-log-fields) that I think does a decent job at capturing big picture details. For example, some fields like type, object, authentication policy, authentication id, vendor, factor number, etc.

mr1716:
@kgeller So the example you provided is 1 more reason to add those fields in. But where would that data get added into the existing ECS fields? I'm trying to understand where to put this information that would meet the requirements of being broad.

kgeller:
@mr1716 I think we'd create a new `authentication` fieldset that would sit at the top level.

An example of an RFC's [stage 0](https://github.com/elastic/ecs/pull/1610) and [stage 1](https://github.com/elastic/ecs/pull/1627) PR's adding in a brand new fieldset.

Does that help?

mr1716:
@kgeller It does help provide some idea of what the next steps would be. If I took what I have for proposed fields and replaced the word login with authentication and added in what the Palo Alto logs show, would that be acceptable or too narrow?

Issue #1965, opened 2022-06-23
`service.node.role` should be `service.node.roles` and support multiple keywords


lukeelmers:
I'm working on adding a `node.roles` configuration to Kibana, and discovered some discrepancies in the spec that was merged in https://github.com/elastic/ecs/pull/1916 and the use case we had envisioned for this in Kibana.

In our configuration, we want to allow nodes to have _multiple_ roles. So for example, the following would be a valid value for the field:
```
node.roles: ['ui', 'background-tasks']
```

The PR that was merged only allows for a single item (not normalized as an array) and is also named `service.node.role` (singular).

Looking for advice on what to do here, since this was merged for 8.3. It is almost certainly too late to do anything about it, but unfortunately it means we won't be able to adhere to this spec in Kibana based on the technical needs of our project.

Should we leave the field as-is and just not use it? Deprecate it? Something else?

FWIW, we won't be actually introducing usage of this field in Kibana until at least 8.4.

cc @kobelb @matschaffer @ebeahan

matschaffer:
🤦🏻 if only we had an easy way to upgrade fields from old to new versions (string to list should be one of the easy ones).

Deprecation of the old field and introduction of a new field make sense to me.

kgeller:
@lukeelmers @matschaffer I'd say that since 8.3 hasn't been released, we can just go ahead and fix this now (`role` -> `roles` and single value -> array).

@ebeahan does that seem reasonable to you?

ebeahan:
Integrations (and potentially elsewhere) have already adopted ECS 8.3. I did a brief review and not seeing any active use of `service.node.role`, but I agree with @lukeelmers that it's too late to adjust for 8.3.

Not ideal, we could keep the name `node.role` and update from a single value to expecting an array. An event's `_source` may not match, but Elasticsearch indexing wouldn't be affected .

lukeelmers:
> we could keep the name `node.role` and update from a single value to expecting an array

Is that a breaking change though? If so, I'd just as soon rename the field as well for clarity: `service.node.roles`

ebeahan:
ECS 8.3 hasn't released yet, so changing either the field name or single/array expectation is still non-breaking in the spec itself. But, stack/solutions/integrations already adopted 8.3 changes. Renaming or removing a field may create issues downstream.

[...]

PR #1981, opened 2022-07-05
Adding service.node.roles

kgeller:
Adding the _plural_ `service.node.roles` field per https://github.com/elastic/ecs/issues/1965

@lukeelmers / @matschaffer let me know if you want any tweaks to the description. I mostly kept what you had for the singlular.

### DIFF
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -8661,6 +8661,33 @@ example: `background-tasks`
 // ===============================================================
 
 |
+[[field-service-node-roles]]
+<<field-service-node-roles, service.node.roles>>
+
+a| Roles of a service node.
+
+This allows for distinction between different running roles of the same service.
+
+In the case of Kibana, the `service.node.role` could be `ui` or `background-tasks` or both.
+
+In the case of Elasticsearch, the `service.node.role` could be `master` or `data or both`.
+
+Other services could use this to distinguish between a `web` and `worker` role running as part of the service.
+
+type: keyword
+
+
+Note: this field should contain an array of values.
+
+
+
+example: `["ui", "background-tasks"]`
+
+| extended
+
+// ===============================================================
+
+|
 [[field-service-state]]
 <<field-service-state, service.state>>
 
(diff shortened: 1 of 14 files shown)


docs/fields/field-details.asciidoc
```diff
@@ -8660,6 +8660,33 @@ example: `background-tasks`
...
+[[field-service-node-roles]]
+<<field-service-node-roles, service.node.roles>>
+
+a| Roles of a service node.
+
+This allows for distinction between different running roles of the same service.
+
+In the case of Kibana, the `service.node.role` could be `ui` or `background-tasks` or both.
+
+In the case of Elasticsearch, the `service.node.role` could be `master` or `data or both`.
```
lukeelmers, line 8673:
```suggestion
In the case of Elasticsearch, the `service.node.role` could be `master` or `data` or both.
```

(repeated a few other places in the PR)

docs/fields/field-details.asciidoc
```diff
@@ -8660,6 +8660,33 @@ example: `background-tasks`
...
 
+|
+[[field-service-node-roles]]
+<<field-service-node-roles, service.node.roles>>
+
+a| Roles of a service node.
+
+This allows for distinction between different running roles of the same service.
+
+In the case of Kibana, the `service.node.role` could be `ui` or `background-tasks` or both.
```
lukeelmers, line 8671: We ended up changing from `background-tasks` to `background_tasks` in the Kibana PR for consistency with ES: https://github.com/elastic/kibana/pull/135272#discussion_r908530524

So maybe we should change the examples in this PR to match

lukeelmers (commented):
Thanks for updating this! Noticed one typo & one thing we changed in the Kibana PR, otherwise LGTM.

cc @kobelb

docs/fields/field-details.asciidoc
matschaffer: Agreed!

matschaffer:
The origin/target part is unfamiliar to me. Are we adding that just to maintain parity with where `service.node.name` is already present in ECS?

djptek (approved):
LGTM

docs/fields/field-details.asciidoc
djptek: suggestion applied

PR #1993, opened 2022-07-08
Misspell macos fix

djptek:
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)? Y
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)? n/a
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. Y
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? TBC


Makefile
```diff
@@ -20,7 +20,7 @@ all: generate experimental
 # Check verifies that all of the committed files that are generated are
 # up-to-date.
 .PHONY: check
-check: generate experimental test fmt misspell makelint
+check: generate experimental test fmt misspell_warn makelint
```
kgeller, line 23: Since we are changing this, should we not also delete the original `misspell` task down below?

djptek: I put the new task in under a different name so I could leave the original task as a standalone in case people already use that in their processes and want to continue doing so on platforms where it still works.

kgeller: Ah ok, that makes sense

PR #1994, opened 2022-07-13
Adding `risk.*` fields as experimental

kgeller:
Adding per https://github.com/elastic/ecs/pull/1744

### DIFF
--- a/experimental/generated/beats/fields.ecs.yml
+++ b/experimental/generated/beats/fields.ecs.yml
@@ -3696,7 +3696,7 @@
       description: A risk classification score obtained from outside the system, such
         as from some external Threat Intelligence Platform, and normalized to a range
         of 0 to 100.
-      scaling_factor: 83.0
+      example: 83.0
       default_field: false
     - name: type
       level: core
@@ -7259,7 +7259,7 @@
       description: A risk classification score obtained from outside the system, such
         as from some external Threat Intelligence Platform, and normalized to a range
         of 0 to 100.
-      scaling_factor: 83.0
+      example: 83.0
       default_field: false
   - name: rule
     title: Rule
@@ -12130,7 +12130,7 @@
       description: A risk classification score obtained from outside the system, such
         as from some external Threat Intelligence Platform, and normalized to a range
         of 0 to 100.
-      scaling_factor: 83.0
+      example: 83.0
       default_field: false
     - name: changes.roles
       level: extended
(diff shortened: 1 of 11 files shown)


experimental/schemas/risk.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: risk
+  title: Risk score information
+  group: 2
+  short: Fields for describing the risk score and level.
```
kgeller, line 5: @ajosh0504 I stubbed this short and description since it wasn't explicit in the RFC document. Everything else should be verbatim.

experimental/generated/beats/fields.ecs.yml
```diff
@@ -3652,6 +3652,52 @@
         system (nsfs). Unsigned int inum in include/linux/ns_common.h.
       example: 256383
       default_field: false
+    - name: risk.calculated_level
+      level: extended
```
ajosh0504, line 3656: What's "level" here?

experimental/generated/beats/fields.ecs.yml
```diff
@@ -7162,6 +7208,59 @@
       ignore_above: 1024
       description: All the user names or other user identifiers seen on the event.
       default_field: false
+  - name: risk
```
ajosh0504, line 7211: What entity is this set associated with? I didn't see them in the `fields.csv` file!?

experimental/schemas/risk.yml
ajosh0504: That's fine!

experimental/schemas/risk.yml
```diff
@@ -0,0 +1,58 @@
...
+      level: extended
+      type: float
+      example: 830.0
+      description: >
+        A risk classification score obtained from outside the system, such as
+        from some external Threat Intelligence Platform.
+    - name: static_score_norm
+      level: extended
+      type: float
+      scaling_factor: 83.0
```
ajosh0504, line 40: Should this be `scaling_factor` or `example`?

experimental/generated/beats/fields.ecs.yml
kgeller: It essentially just means either core ECS fields or the more use-case specific https://www.elastic.co/guide/en/ecs/current/ecs-guidelines.html#_ecs_field_levels

ajosh0504 (approved):

ajosh0504:
Looking good! 🚢

PR #1999, opened 2022-07-26
Added `ios` and `android` as valid values for `os.type`

### LINKED ISSUE #1995: Mapping OS fields for mobile devices (e.g. Android, iOS)
**Summary**
For the APM Mobile agents we would like to properly populate the Operating System Fields. 
What is not clear to me is what we would set for `os.type`  and `os.platform` in case of iOS and Android. 

Let's take the example of iOS. According to [the description of ](https://www.elastic.co/guide/en/ecs/current/ecs-os.html#field-os-type)`os.type` there is no allowed value that would be valid (because `Macos` is not `iOS`and `ios` is not allowed as a value). Also, in this case, `os.platform` would have the value `darwin` (since iOS is based on Darwin). But `darwin` could also mean that it's a MacOS device (not mobile). So, there is no field where we could put the value `ios` in to properly categorize the operating system in this case.

**Motivation**:
Capture and categorize the operating system of mobile devices properly.

**Detailed Design**:
Does it make sense to add the following values as valid values for `os.type`?:
* `ios`
* `android`

Or are there other ideas how we could handle the scenario described above?

AlexanderWert:
Closes #1995 

Adds `ios` and `android` as valid values for the `os.type` field to support mobile use cases (E.g. APM / distributed tracing in mobile devices).

/cc @simitt, @felixbarny

### DIFF
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -6682,6 +6682,8 @@ Expected values for this field:
 * `macos`
 * `unix`
 * `windows`
+* `ios`
+* `android`
 
 type: keyword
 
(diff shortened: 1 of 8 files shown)


djptek:
@AlexanderWert Please can you run `make check` locally to generate the artefacts & then check those in to advance the unit tests

djptek (approved):
LGTM

PR #2010, opened 2022-07-26
Fixing risk short fields

kgeller:
Follow up of #1994 ; correcting linter warnings due to too long descriptions without shorts


djptek (approved):
LGTM

PR #2019, opened 2022-07-29
Feature: make ecs more flexible

maihde:
When using the ECS tooling for custom projects, there are a few minor changes that make would make the tooling more flexible, especially in cases where the ECS schema itself might be minimally used.  I appreciate that the primary use case for ECS tooling is the ECS schema, but given that changes are minor and their is value of increased utility for the tooling I hope they could be considered for inclusion in the baseline.

I welcome discussion on the proposed changes.


scripts/generators/es_template.py
```diff
@@ -38,13 +38,14 @@
 def generate(
     ecs_nested: Dict[str, FieldNestedEntry],
     ecs_version: str,
+    ecs_component_name_prefix: str,
     out_dir: str,
     mapping_settings_file: str,
     template_settings_file: str
 ) -> None:
```
IanLee1521, line 45: Not really having context for this tool, adding a new argument like this in the middle seems like it might not be the most compatible way of doing this.. Usually I'd expect new arguments to show up in the `kwargs` section of the list and then have a reasonable default, which in this case looks like it should be `ecs`:

```suggestion
def generate(
    ecs_nested: Dict[str, FieldNestedEntry],
    ecs_version: str,
    out_dir: str,
    mapping_settings_file: str,
    template_settings_file: str,
    ecs_component_name_prefix: str = "ecs",
) -> None:
```

IanLee1521 (commented):
LGTM, though I'm not a user of this tool, so I'm possibly missing something.

scripts/generators/es_template.py
desean1625: Moved to kwarg

IanLee1521: LGTM

ralphm (changes requested):
It would be useful to detail what improvements you made from a user's perspective, as this is not detailed on the PR description. The individual commit messages do give a short indication, but also per the [contribution guide](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md), could you please add to `CHANGELOG.next.md`?

desean1625:
This pull request allows users to 
1. Define their own template documentations for ASCII DOCS
2. Run ECS from a different directory. This allows ECS to be included in a project as a submodule
3. Set a component name that isn't ecs_

This allows ECS tooling to be utilized by more organizations for generating their component templates, and documentation.

desean1625:
👍

ralphm:
👍

scripts/generators/asciidoc_fields.py
```diff
@@ -152,8 +152,14 @@ def save_asciidoc(f, text):
 # jinja2 setup
 
 
+cur_dir = path.abspath(path.curdir)
 local_dir = path.dirname(path.abspath(__file__))
-TEMPLATE_DIR = path.join(local_dir, '../templates')
+CUR_TEMPLATE_DIR = path.join(cur_dir, 'templates')
+LOCAL_TEMPLATE_DIR = path.join(local_dir, '../templates')
+if path.exists(CUR_TEMPLATE_DIR):
```
mjwolf, line 159: If these conditions both evaluate to false, this will exit with a TypeError. It would be better to catch this and print a friendly error message

scripts/generator.py
```diff
@@ -118,6 +123,10 @@ def argument_parser() -> argparse.Namespace:
                         help='enforce strict checking at schema cleanup')
     parser.add_argument('--intermediate-only', action='store_true',
                         help='generate intermediary files only')
+    parser.add_argument('--no-ecs', action='store_true',
```
mjwolf, line 126: Could add more documentation to `help` for the new arguments, to explain more about how these are intended to be used, and to mention they are for custom use cases. It's fine to use multiline strings here, so the help message can be longer.

scripts/generators/beats.py
```diff
@@ -29,15 +29,21 @@
     FieldNestedEntry,
 )
 
+BEATS_DEFAULT_FIELDS = join(dirname(ecs_helpers.__file__), "beats_default_fields_allowlist.yml")
```
mjwolf, line 32: If this file is intended to be customized, I think it would be better to add a generator.py argument that can be used to specify the file location used

CHANGELOG.next.md
```diff
@@ -17,6 +17,10 @@ Thanks, you're awesome :-) -->
 #### Added
 
 #### Improvements
+* Allow ECS from any directory #2019
```
mjwolf, line 20: Could you move these into the "Tooling and Artifact Changes" -> Improvements section?

Villegas12:
There is still a need for this!  :)
Thanks

Marcial Villegas

Principal Solutions Architect

+1 202.930.3812

<https://www.elastic.co/>

jkb-axios:
👍

PR #2050, opened 2022-09-13
Add attested user and groups to process

daniel-almeida:
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? yes


schemas/group.yml
```diff
@@ -47,6 +47,10 @@
         beta: Reusing the `group` fields in this location is currently considered beta.
         normalize:
           - array
+      - at: process
+        as: attested_group
+        short_override: The attested group (agid).
```
mitodrummer, line 52: I guess agid is a new abbreviation? do we need to specify this, as it's not really a linux convention?

schemas/user.yml
```diff
@@ -54,6 +54,10 @@
         as: real_user
         short_override: The real user (ruid). Identifies the real owner of the process.
         beta: Reusing the `user` fields in this location is currently considered beta.
+      - at: process
+        as: attested_user
+        short_override: The attested user (auid). Identifies the attested used associated with the process.
```
mitodrummer, line 59: same here. auid isn't something that exists in linux, so maybe not needed, unless this is a name we're using elsewhere.

schemas/group.yml
daniel-almeida: I didn't think all of those were a linux convention. heh TIL. We could also not abbreviate it, I suppose.

schemas/user.yml
mitodrummer, line 59: Also, i know there isn't a ton of space for descriptions on nested fields. "short only", but I wonder if we can explain what attested means. e.g via 2factor mechanisms etc...

mitodrummer:
This works good as a nested field. When this gets merged and we update endpoint-package, we'll need to decide what nesting this attested_user lives at (mapping wise). Assuming it will only be attested once at the entry_leader process, it probably makes sense to only add a mapping for process.entry_leader.attested_user and process.entry_leader.attested_group. Not only where it's nested, but which fields in the user fieldset will be mapped. e.g email, username etc...

schemas/group.yml
mitodrummer: yea, see **man 7 credentials**

though looking i'm not really seeing a ruid (real user id), as I think that might just be uid, so we might have invented that, not sure. cc @m-sample

schemas/user.yml
andrewkroh: fwiw: `auid` is used in the context of Linux kernel auditing.

https://github.com/linux-audit/audit-documentation/blob/b5a76394e165b104c95c94127ac3d29eed68f927/specs/fields/field-dictionary.csv#L16

m-sample: "externally" attested is the key aspect ie they have done something to prove to an external source (e.g. pw & 2nd factor) that they are who claim to be and authorized to obtain the role/identity stated here.

schemas/group.yml
m-sample: I would avoid any suggestion that this related a Linux uids or gids at all;  Group names are all we get from k8s IIRC, not ids.

Should this be "attested_groups" - it is an array

schemas/user.yml
m-sample: I too would remove  "auid" - confusing.  Are we planning to populate the user.id field for the externally attested user? if not it would nice to say somewhere that when attested_user represents a k8s user, how we map the field values received by the admission controller to the user fields.

mitodrummer (approved):
LGTM!

m-sample (approved):
LGTM. Nice work & thanks @daniel-almeida.   

We can review if we need to support more than one externally attested identity per event before this leaves beta.

kgeller:
@daniel-almeida I think these changes look good, I just have one question. Can we get these added to the subset file, or is there is a reason that was omitted?

example: [process.supplemental_groups](https://github.com/elastic/ecs/blob/main/schemas/subsets/main.yml#L140)

daniel-almeida:
That was probably oversight on my part, @kgeller . I'll look into it.

kgeller (approved):
LGTM

PR #2058, opened 2022-09-22
Refining the addition of the `risk.*` fields locations

kgeller:
Cleaning up the nestings of the` risk.*` fields, so that they are only at `host.*` and `user.*`

Removing the following:
- `client.user.risk`
- `destination.user.risk`
- `server.user.risk`
- `source.user.risk`
- `user.changes.risk`
- `user.effective.risk`
- `user.target.risk`


ajosh0504 (commented):
LGTM. Thanks, Kylie!

ajosh0504 (approved):

Issue #2073, opened 2022-09-30
Update to Threat ECS Fieldset (TLP)


peasead:
**Summary**

The Traffic Light Protocol (TLP) is a system of markings that communicates information sharing permissions for threat indicators.

In August 2022, FIRST (the governing body for TLP) [released TLP v2.0](https://www.first.org/tlp/). This has an update to an existing TLP designation (`TLP:WHITE` -> `TLP:CLEAR`) and the addition of an additional TLP designation (`TLP:AMBER+STRICT`).

We should update the TLP fields for the Threat ECS fieldset to include the following:
- add a `TLP:CLEAR` field
- add a `TLP:AMBER+STRICT` field
- add a TLP version field

**Motivation**:

This will keep the ECS fieldset updated with the governing body for TLP. TLP v1.0 was deprecated in August 2022. 

**Detailed Design**:

Provide additional details around the design of the proposed changes.

* Field names:
  * [`threat.indicator.marking.tlp`](https://www.elastic.co/guide/en/ecs/current/ecs-threat.html#field-threat-indicator-marking-tlp)
  * [`threat.enrichments.indicator.marking.tlp`](https://www.elastic.co/guide/en/ecs/current/ecs-threat.html#field-threat-enrichments-indicator-marking-tlp)
  * `threat.indicator.marking.tlp.version`
  * `threat.enrichments.indicator.marking.tlp.version`
* Example values for the fields
  * `threat.indicator.marking.tlp : CLEAR, WHITE, AMBER, AMBER+STRICT, RED`
  * `threat.enrichments.indicator.marking.tlp : CLEAR, WHITE, AMBER, AMBER+STRICT, RED`
  * `threat.indicator.marking.tlp.version : 1.0, 2.0`
  * `threat.enrichments.indicator.marking.tlp.version : 1.0, 2.0`

kgeller:
Hi @peasead ! The additions of `CLEAR` and `AMBER+STRICT` have already been completed via a community [PR](https://github.com/elastic/ecs/pull/2034) and will be released in ECS 8.6.

peasead:
🚀

PR #2034, opened 2022-08-23
feature/tlp_2.0

mdavis332:
- [x] Signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- [x] Followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)?
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)?
  - Non-substantial change (IMO)
- [ ] If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?
  - Was unable to run `make` locally due to the following:
```
Collecting gitpython==3.1.27 (from -r scripts/requirements.txt (line 5))
  Could not find a version that satisfies the requirement gitpython==3.1.27 (from -r scripts/requirements.txt (line 5))
```
- [x] Pull request against main?
- [x] Added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)?

Addresses https://github.com/elastic/ecs/issues/2022.

PR #2074, opened 2022-09-30
Update to TLP

kgeller:
Continuation of https://github.com/elastic/ecs/pull/2034

Addresses https://github.com/elastic/ecs/issues/2073

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -44,10 +44,11 @@ Thanks, you're awesome :-) -->
 
 * Adding `vulnerability` option for `event.catgeory`. #2029
 * Added `device.*` field set as beta. #2030
+* Added `tlp.version` to threat #2074
 
 #### Improvements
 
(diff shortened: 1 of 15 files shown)


djptek (approved):
LGTM

vpiserchia:
Dear all, I see discrepancies between every other ECS fields related to versioning information (e.g. http.version, agent.version, etc). All of them are of ```keyword``` type while this one is a ```float```. Provided that I do no see a use case for this to be a number (ordering/sorting) I would recomment to keep the same coherence for these fields:
- threat.threat.indicator.marking.tlp.version
- threat.indicator.marking.tlp.version

kgeller:
> Dear all, I see discrepancies between every other ECS fields related to versioning information (e.g. http.version, agent.version, etc). All of them are of `keyword` type while this one is a `float`. Provided that I do no see a use case for this to be a number (ordering/sorting) I would recomment to keep the same coherence for these fields:

@vpiserchia great catch, I didn't notice that discrepancy. I have updated the new `version` fields here to keyword for parity.

PR #2075, opened 2022-09-30
Archiving ECS new issue actions

kgeller:
In the attempt to avoid confusion, archiving the actions on issues being opened/reopened


ebeahan:
Instead of editing the job, you can disable the workflow in the repo's `Actions`: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow.

Otherwise, we can just delete the file entirely. It'll be in the commit history should we ever want to re-use in the future.

kgeller:
> Instead of editing the job, you can disable the workflow in the repo's `Actions`: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow.

Oh! Yes I can. I assumed that because it came from the file, it had to be disabled there.

I have disabled and will close this out.

PR #2031, opened 2022-08-17
[RFC Stage-2] Process IO events

mitodrummer:
Updates the schemas/process.yml with the new process.io fields.

There is also two new fields that weren't in stage 1 added to track terminal window size.
-process.tty.rows
-process.tty.columns

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? yes
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)? yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)? yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? yes
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? yes
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? yes

PR #2078, opened 2022-10-04
Process io subset fix

kgeller:
Followup to https://github.com/elastic/ecs/pull/2031 

Adding `io.*` fields to the subset file at top-level process

cc @mitodrummer


mitodrummer (approved):
LGTM!

PR #2083, opened 2022-10-26
Add fields for executable object format metadata for ELF, Mach-O and PE

efd6:
Fields are added to ELF for malware detection signatures and fields in ELF for this are reflected in PE. A new Mac OS Mach-O field group is added. Not all possible fields are added to Mach-O.

Specific Go executable fields are added because the import structure of Go executables is not described in standard dynamic library imports and Go is a reasonably popular language both generally, and significantly, in malware.

See discussion at https://github.com/elastic/beats/pull/28802#discussion_r990525302.

Optimistically putting in 8.6, but happy to bump to 8.7 if this is too tight. If this needs an RFC will do that.

ELF:

- [x] `file.elf.go_import_hash`
- [x] `file.elf.go_imports`
- [x] `file.elf.go_imports_names_entropy`
- [x] `file.elf.go_imports_names_var_entropy`
- [x] `file.elf.go_stripped`
- [x] `file.elf.import_hash` — alias for ELF equivalent of imphash to simplify cross-platform searches
- [x] `file.elf.imports_names_entropy`
- [x] `file.elf.imports_names_var_entropy`
- [x] `file.elf.sections.var_entropy`

PE:

- [x] `file.pe.go_import_hash`
- [x] `file.pe.go_imports`
- [x] `file.pe.go_imports_names_entropy`
- [x] `file.pe.go_imports_names_var_entropy`
- [x] `file.pe.go_stripped`
- [x] `file.pe.imports_names_entropy`
- [x] `file.pe.imports_names_var_entropy`
- [x] `file.pe.sections.var_entropy`

The following are added to PE as analogous to the existing ELF fields. All bar virtual_size are strictly informative (`virtual_size` is always equal to `physical_size` since only ELF has compressed sections)

- [x] `file.pe.import_hash` — alias for imphash to simplify cross-platform searches
- [x] `file.pe.imports`
- [x] `file.pe.sections`
- [x] `file.pe.sections.entropy`
- [x] `file.pe.sections.name`
- [x] `file.pe.sections.physical_size`
- [x] `file.pe.sections.virtual_size` — always the same as `physical_size` on PE (meaningful on ELF as a potential marker for evasive ELF)

ECS doesn't have any fields for Mach-O so all the following are new:

- [x] `file.macho.go_import_hash`
- [x] `file.macho.go_imports_names_entropy`
- [x] `file.macho.go_imports_names_var_entropy`
- [x] `file.macho.go_imports`
- [x] `file.macho.go_stripped`
- [x] `file.macho.import_hash` — alias for Mach-O equivalent of imphash
- [x] `file.macho.imports`
- [x] `file.macho.imports_names_entropy`
- [x] `file.macho.imports_names_var_entropy`
- [x] `file.macho.sections`
- [x] `file.macho.sections.entropy`
- [x] `file.macho.sections.name`
- [x] `file.macho.sections.physical_size`
- [x] `file.macho.sections.var_entropy`
- [x] `file.macho.sections.virtual_size` — always the same as `physical_size` on Mach-O (meaningful on ELF as a potential marker for evasive ELF)
- [x] `file.macho.symhash` — alias for Mach-O equivalent of for imphash

For elastic/beats#28802

Please take a look.

### DIFF
--- a/schemas/elf.yml
+++ b/schemas/elf.yml
@@ -65,9 +65,12 @@
     - name: go_import_hash
       short: A hash of the Go language imports in an ELF file.
       description: >
-        A hash of the Go language imports in an ELF file. An import hash can be used to
-        fingerprint binaries even after recompilation or other code-level
-        transformations have occurred, which would change more traditional hash values.
+        A hash of the Go language imports in an ELF file excluding standard library imports.
+        An import hash can be used to fingerprint binaries even after recompilation or other
+        code-level transformations have occurred, which would change more traditional hash values.
+
+        The algorithm used to calculate the Go symbol hash and a reference implementation
+        are available [here](https://github.com/elastic/toutoumomoma).
       example: 10bddcb4cee42080f76c88d9ff964491
       type: keyword
       level: extended
--- a/schemas/elf.yml
+++ b/schemas/elf.yml
@@ -62,6 +62,43 @@
       level: extended
       example: Intel
 
+    - name: go_import_hash
+      short: A hash of the Go language imports in an ELF file.
+      description: >
+        A hash of the Go language imports in an ELF file. An import hash can be used to
+        fingerprint binaries even after recompilation or other code-level
+        transformations have occurred, which would change more traditional hash values.
+      example: 10bddcb4cee42080f76c88d9ff964491
+      type: keyword
+      level: extended
+
+    - name: go_imports_names_entropy
+      description: >
+        Shannon entropy calculation from the list of Go imports.
+      type: long
+      format: number
+      level: extended
+
+    - name: go_imports_names_var_entropy
+      description: >
+        Variance for Shannon entropy calculation from the list of Go imports.
+      type: long
+      format: number
+      level: extended
+
+    - name: go_imports
+      description: >
+        List of imported Go language element names and types.
+      type: flattened
+      level: extended
+
+    - name: go_stripped
+      short: Whether the file is a stripped or obfuscated Go executable.
+      description: >
+        Set to true if the file is a Go executable that has had its symbols stripped or obfuscated and false if an unobfuscated Go executable.
+      type: boolean
+      level: extended
+
     - name: header.class
       description: >
         Header class of the ELF file.
@@ -111,6 +148,32 @@
       description: >
         "0x1" for original ELF files.
 
+    - name: import_hash
+      short: A hash of the imports in an ELF file.
+      description: >
+        A hash of the imports in an ELF file. An import hash can be used to
+        fingerprint binaries even after recompilation or other code-level
+        transformations have occurred, which would change more traditional hash values.
+
+        This is an ELF implementation of the Windows PE imphash.
+      example: d41d8cd98f00b204e9800998ecf8427e
+      type: keyword
+      level: extended
+
+    - name: imports_names_entropy
+      description: >
+        Shannon entropy calculation from the list of imported element names and types.
+      format: number
+      type: long
+      level: extended
+
+    - name: imports_names_var_entropy
+      description: >
+        Variance for Shannon entropy calculation from the list of imported element names and types.
+      format: number
+      type: long
+      level: extended
+
     - name: sections
       short: Section information of the ELF file.
       description: >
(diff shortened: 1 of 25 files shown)


efd6:
The failure appears to be due to the absence of the `[ecs-macho]` section in field_details.
```
INFO:build_docs:The Asciidoctor migration is complete! --asciidoctor will emit this message
INFO:build_docs:forever in honor of our success but otherwise doesn't do anything.
INFO:build_docs:
INFO:build_docs:Building HTML from /doc/ecs/docs/index.asciidoc
INFO:build_docs:Guessed toplevel=[/doc/ecs] remote=[git@github.com:elastic/ecs] branch=[master] repo=[ecs]
INFO:build_docs:Browserslist: caniuse-lite is outdated. Please run next command `npm update`
INFO:build_docs:
INFO:build_docs:asciidoctor: WARNING: invalid reference: ecs-macho
make: *** [docs] Error 255
```
I don't see how to generate this.

peasead (commented):
Though I don't approve ECS merges, this LGTM.

schemas/elf.yml
```diff
@@ -62,6 +62,43 @@
       level: extended
       example: Intel
 
+    - name: go_import_hash
+      short: A hash of the Go language imports in an ELF file.
+      description: >
+        A hash of the Go language imports in an ELF file. An import hash can be used to
```
andrewkroh, line 68: Is this a well-known hash standard? If so can we link to its documentation. If not, can we write a document somewhere the describes how this is calculated.

andrewkroh (commented):
My hope is that ECS contains enough detail that there is a high probably that all implementations produce the same values for these fields. With that in mind, when you read the descriptions do you think there's enough detail?

schemas/elf.yml
efd6: The Go import hash is imphash/symhash but using the static Go imports rather than the dynamic imports. There is no standard. Where would you like a document put?

The parallels can be seen here https://github.com/elastic/toutoumomoma/blob/594ef30cb64046757baa988d4451b8f547dcb695/toutoumomoma.go#L149-L231

andrewkroh: I don't have a strong preference. How about in the https://github.com/elastic/toutoumomoma repo? You could point to the code as the reference implementation.

efd6: Will do.

- [ ] Enhance readme in toutoumomoma.
- [ ] Pull description from `*File.ImportHash` out to godoc for both it and `*File.GoSymbolHash`
- [ ] Link to relevant locations in the ECS descriptions

schemas/elf.yml
```diff
@@ -65,9 +65,12 @@
...
       short: A hash of the Go language imports in an ELF file.
       description: >
-        A hash of the Go language imports in an ELF file. An import hash can be used to
-        fingerprint binaries even after recompilation or other code-level
-        transformations have occurred, which would change more traditional hash values.
+        A hash of the Go language imports in an ELF file excluding standard library imports.
+        An import hash can be used to fingerprint binaries even after recompilation or other
+        code-level transformations have occurred, which would change more traditional hash values.
+
+        The algorithm used to calculate the Go symbol hash and a reference implementation
```
andrewkroh, line 72: Does ECS expect this calculated with or without the inclusion of the stdlib imports?

andrewkroh (approved):

schemas/elf.yml
efd6: I guess that depends on what we decide here since it's a new symbol. I excluded stdlib packages because it reduces the size of indexed documents. However, I can see _some_ merit in including stdlib given that it is entirely feasible to write malicious code using only stdlib packages. The argument against that is that the packages that you might use to do so are entirely valid for benign applications.

We have to make a decision and this seems like the better of the two.

andrewkroh: > excluding standard library import

Oh, it is specified. I came up with the question after reading the changes in toutoumomoma and didn't re-read this part.

By including stdlib we would have some signal to use when the binary does not utilize external packages. I assume that for any binary that only uses stdlib that the result is `MD5("")`.

efd6: > I assume that for any binary that only uses stdlib that the result is `MD5("")`.

Yes, this is correct `d41d8cd98f00b204e9800998ecf8427e`

PR #2084, opened 2022-10-31
Mapping parameters

ralphm:
When attempting to create a custom mapping for a `date` (multi-)field, I found that the ECS tooling doesn't currently support providing the `ignore_malformed` and `format` mapping parameters. Even though there is some limited support for certain parameters, these are not consistent between the main field mapping and the multi-field mapping. E.g. the `normalizer` parameter is only available for multi-fields.

Instead of implementing specific parameters for my use case, this PR adds a new property `parameters` to provide any mapping parameter, both for the main field and for multi-fields, along with a minimal test for the multi-field generation code. The existing code can remain as-is, with the minimal sanity checking that is in place, for backwards compatibility.

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -30,6 +30,8 @@ Thanks, you're awesome :-) -->
 
 #### Added
 
+* Add `parameters` property for field definitions, to provide any mapping parameter. #2084
+
 #### Improvements
 
 * Bump `gitpython` from `3.1.27` to `3.1.30` in `/scripts`. #2139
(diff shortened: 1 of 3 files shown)


ralphm:
Any chance this can get a review? The merge conflict is trivial, and I'll be happy to resolve that when the time comes.

kgeller (changes requested):
@ralphm this looks like a great add. Once the merge conflict is resolved, I will approve and merge.

kgeller (approved):

ralphm:
Thanks, @kgeller!

PR #2095, opened 2022-11-02
Fix typos in macho and pe schema files

efd6:
No changelog entry since there has not been a release including these fields.

Please take a look.


djptek (approved):
LGTM

PR #2129, opened 2023-01-05
[RFC] host metrics

ruflin:
This RFC adds a proposal to bring host metrics to ECS. These metrics should build the foundation to deliver a minimal set of metrics related to a host. The current list looks as following:

* host.cpu.system.norm.pct
* host.cpu.user.norm.pct
* host.fsstats.total_size.used (in bytes)
* host.fsstats.total_size.total (in bytes)
* host.fsstats.total_size.used.pct
* host.load.norm.1
* host.load.norm.5
* host.load.norm.15
* host.memory.actual.used.bytes
* host.memory.actual.used.pct
* host.memory.total
* host.network.egress.bytes
* host.network.ingress.bytes

One of the main challenges around this RFC is if we should prefix with `host.*` or `system.*`. See some more details in the RFC itself. It would be great to hear opinions around it.


ruflin:
Converted from draft to review to start conversations.

ebeahan:
What's the relationship between this new proposal and [RFC 0005 - host metric fields](https://github.com/elastic/ecs/blob/main/rfcs/text/0005-host-metric-fields.md)? 

cc @kaiyan-sheng who authored RFC 0005.

> One of the main challenges around this RFC is if we should prefix with host.* or system.*. See some more details in the RFC itself. It would be great to hear opinions around it.

RFC 0005 established a small group of metric fields under `host.*`. With those fields already added to the spec, I'd propose continue to use `host.*`.

ruflin:
Thanks for pointing out https://github.com/elastic/ecs/blob/main/rfcs/text/0005-host-metric-fields.md @ebeahan It definitively points in the direction of using `host.*` for the fields. For now I will to with the assumption we keep the `host.*` prefix until someone objects.

There is an overlap with RFC 5 here and it expands on it. I was aware of the network and disk fields but missed the cpu field. @kaiyan-sheng Can you comment on where these fields are used today? How does it compare to the fields proposed here?

ruflin:
Can I get some reviews on this PR to get things moving?

neptunian:
> Can you comment on where these fields are used today? How does it compare to the fields proposed here?

I found [this issue](https://github.com/elastic/kibana/issues/97106) created to switch over to using some of the new ECS host fields.  It looks like `host.cpu.usage` wasn't done.  Currently it's calculated [here](https://github.com/elastic/kibana/blob/main/x-pack/plugins/infra/common/inventory_models/host/metrics/snapshot/cpu.ts) as average of `system.cpu.user.pct` plus `average of system.cpu.system.pct` divided by `system.cpu.cores`.  Comparing the result of this to `host.cpu.usage` is very close and would be exact if we were using the normalized system values instead.  We should remove this calculation and switch to using `host.cpu.usage`.

So, from the list we can remove:
- host.cpu.system.norm.pct
- host.cpu.user.norm.pct
- host.network.egress.bytes
- host.network.ingress.bytes

rfcs/text/0037-host-metrics.md
```diff
@@ -0,0 +1,162 @@
...
+Stage 2: Identifies scope of impact of changes. Are breaking changes required? Should deprecation strategies be adopted? Will significant refactoring be involved? Break the impact down into:
+ * Ingestion mechanisms (e.g. beats/logstash)
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
+
+Currently Elastic Agent and metricbeat ship data host/system metrics under the `system.*` prefix. This would change it to `host.*`. One of the reasons for this is that some metrics for network already exist under this prefix in ECS. Another advantage is that some of these fields might use newer field types like `gauge` and `counter` delivered by TSDB in Elasticsearch which is possible without a breaking change. One of the big advantages is it needs to be figured out how to migrate to it with the existing shippers.
```
neptunian, line 105: > Currently Elastic Agent and metricbeat ship data host/system metrics under the `system.*` prefix. This would change it to `host.*`. One of the reasons for this is that some metrics for network already exist under this prefix in ECS. Another advantage is that some of these fields might use newer field types like `gauge` and `counter` delivered by TSDB in Elasticsearch which is possible without a breaking change

Should this be under "Scope of impact"?

> One of the big advantages is it needs to be figured out how to migrate to it with the existing shippers.
I'm confused by this sentence.  How is this a big advantage?  Seems more like a concern.

rfcs/text/0037/host.yml
```diff
@@ -0,0 +1,193 @@
...
+# * Memory used (in %, used, total)
+# * Disk usage (in %)  and io -> summary
+# * Network (traffic in / out)
+#
+# This translates to the detailed metric list below.
+
+
+### CPU ###
+
+# The CPU metrics must indicate under how much load the system is.
```
ebeahan, line 34: Would we need any guidance how these new proposed `host.cpu.*` fields are related to the existing [`host.cpu.usage`](https://github.com/elastic/ecs/blob/main/schemas/host.yml#L122)?

ebeahan (approved):
I made some comments and nits, but overall no objections to the premise of the proposal and merging the RFC as [stage 0](https://elastic.github.io/ecs/stages.html).

rfcs/text/0037/host.yml
```diff
@@ -0,0 +1,193 @@
...
+
+# The file system metrics indicate how full the disks are.
+# 
+# If there are multiple file systems, this is a summary of all the file systems.
+
+  # TODO: Discuss if it should be `fsstat` or `fsstats` (plural)
+- name: host.fsstats.total_size.used
+  type: long
+  format: bytes
+  time_series_metric: gauge
```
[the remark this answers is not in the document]
ruflin: Does ECS have a spec for the fields.yml file where this can be added?

ruflin:
```
So, from the list we can remove:

host.cpu.system.norm.pct
host.cpu.user.norm.pct
host.network.egress.bytes
host.network.ingress.bytes
```

@neptunian Would you remove these from the initial proposal because they are not used or do you think we should keep them as they will be useful?

rfcs/text/0037-host-metrics.md
ruflin: The `s/advantage/concern` was a pretty big typo. Fixed it now.

The other part I moved under scope of impact. It somehow is a mix between both.

rfcs/text/0037/host.yml
ruflin: Yes. I added this to the list of concerns.

neptunian:
> ```
> So, from the list we can remove:
> 
> host.cpu.system.norm.pct
> host.cpu.user.norm.pct
> host.network.egress.bytes
> host.network.ingress.bytes
> ```
> 
> @neptunian Would you remove these from the initial proposal because they are not used or do you think we should keep them as they will be useful?

Maybe I'm missing something.  `host.network.egress.bytes` and `host.network.ingress.bytes` are already ECS fields.  
https://www.elastic.co/guide/en/beats/metricbeat/current/exported-fields-ecs.html#_host.  For the CPU usage it looks like we could use `host.cpu.usage` which is already an ECS field in place of `host.cpu.system.norm.pct` and `host.cpu.user.norm.pct`.

ruflin:
The egress fields I only put in for completness: https://github.com/elastic/ecs/pull/2129/files#diff-eb1c37e580b8c563a0079437420aa4214f94d05d5b683b743b34ca09b0695446R175 If it is confusing, can also remove these.

For the cpu usage, I think it is important for users to see what cpu usage of the kernel vs user space is. If we decide we don't need this, the `host.cpu.usage` value can be used. The part that is not clear to me, what exactly do we show for `host.cpu.usage`?

neptunian:
> The egress fields I only put in for completness: https://github.com/elastic/ecs/pull/2129/files#diff-eb1c37e580b8c563a0079437420aa4214f94d05d5b683b743b34ca09b0695446R175 If it is confusing, can also remove these.

@ruflin  I see, thanks.

> For the cpu usage, I think it is important for users to see what cpu usage of the kernel vs user space is. If we decide we don't need this, the `host.cpu.usage` value can be used. The part that is not clear to me, what exactly do we show for `host.cpu.usage`? Unfortunately the definition does not contain any details around it (which should be fixed).

Sounds good to me.  If it's helpful, `host.cpu.usage` looks to be the equivalent of `system.cpu.total.pct` / `system.cpu.cores`.  

Screenshot 2023-02-28 at 10 37 53 AM

rfcs/text/0037/host.yml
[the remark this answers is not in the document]
ebeahan: There's no one source of truth for the ECS spec. I filed https://github.com/elastic/ecs/issues/2176 capturing the requirements needed to support.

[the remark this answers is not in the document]
ruflin: Thanks for the issue. For the schema, maybe some parts can be copied over from the package-spec where the same fields.yml structure exists: https://github.com/elastic/package-spec

ebeahan:
Will move forward merging as stage 0 with the two approvers. We'll continuing refining the details and addressing any outstanding concerns in subsequent stages.

ruflin:
Thanks @ebeahan for getting this in.

PR #2133, opened 2023-01-10
Update version file to 8.8.0-dev

marc-gr:
Update version


kgeller:
@marc-gr after manually updating the version file, you'll want to run `make generate` to re-generate all the artifacts with the new version

kgeller (approved):

PR #2156, opened 2023-02-02
Tlp patch

kgeller:
In https://github.com/elastic/ecs/pull/2074 , with the addition of `indicator.marking.tlp.version` we inadvertently removed `indicator.marking.tlp` since a field cannot be both it's own leaf field and a parent of another object.

In this PR, we move from `tlp.version` to `.tlp_version`

Relates [to](https://github.com/elastic/ecs/issues/2073)


taylor-swanson (approved):
LGTM

PR #2172, opened 2023-02-14
[RFC] Stage 0 - TSDB Dimensions

agithomas:
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process] (https://github.com/elastic/ecs/blob/main/rfcs/README.md)? yes

### DIFF
--- a/rfcs/text/0039-tsdb-dimensions.md
+++ b/rfcs/text/0039-tsdb-dimensions.md
@@ -2,7 +2,7 @@
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
 - Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **2023-04-06** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Date: **2023-04-11** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 <!--
 As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.


rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+Stage 1: If the changes include field additions or modifications, please create a folder titled as the RFC number under rfcs/text/. This will be where proposed schema changes as standalone YAML files or extended example mappings and larger source documents will go as the RFC is iterated upon.
+-->
+
+<!--
+Stage X: Provide a brief explanation of why the proposal is being marked as abandoned. This is useful context for anyone revisiting this proposal or considering similar changes later on.
+-->
+
+## Fields
+
+This RFC proposes the annotating of certain ecs fields as `dimension`. This change is proposed to take the advantage of using `TSDB` offerred by the elasticsearch without impacting the data injetion.
```
ebeahan, line 26:
```suggestion
This RFC proposes the annotating of certain ecs fields as `dimension`. This change is proposed to take the advantage of using `TSDB` offered by the elasticsearch without impacting the data injection.  
```

nits

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+
+<!--
+Stage X: Provide a brief explanation of why the proposal is being marked as abandoned. This is useful context for anyone revisiting this proposal or considering similar changes later on.
+-->
+
+## Fields
+
+This RFC proposes the annotating of certain ecs fields as `dimension`. This change is proposed to take the advantage of using `TSDB` offerred by the elasticsearch without impacting the data injetion.  
+
+Annotating field as `dimension` is one of the important step in the process of TSDB adoption. Failing to annotate adequate number of fields as `dimension` when `TSDB` is enabled may lead to data loss. A large majority of fields that must be annotated as `dimension` fields are ecs fields. Presently, the Integration (Service Integration, Cloud Native, etc ) developers are expected to annotate ecs fields as `dimensions` in integration configuration. To avoid the duplicatation in configuration, minimize data loss probability, the RFC is proposed. `dimension` field takes two values - `true` and `false`.
```
ebeahan, line 28: Is whether or not a field is annotated as a `dimension` something we should be defining at all within ECS? My initial opinion is leaning towards no, but I don't know very much about TSDB and the necessary configuration.

Different implementations should decide when to annotate or not based on the use cases involved. The way a field is defined in ECS is the way it should be implemented across all uses of the field. Is that the case for TSDB and `dimensions`? Should fields still be annotated for non-TSDB use cases?

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+
+Integration package development is the key beneficiary of this change. The fields of the document that are received from an integration receives a field mapping. If and when TSDB benefits are to be utilised, along with the field mapping with a metric type, atleast one of the fields  must recieve `dimension: true` annotation.
```
ebeahan, line 152:
```suggestion
Integration package development is the key beneficiary of this change. The fields of the document that are received from an integration receives a field mapping. If and when TSDB benefits are to be utilised, along with the field mapping with a metric type, at least one of the fields  must receive `dimension: true` annotation. 
```

nits

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+Stage 2: Identifies scope of impact of changes. Are breaking changes required? Should deprecation strategies be adopted? Will significant refactoring be involved? Break the impact down into:
+ * Ingestion mechanisms (e.g. beats/logstash)
+ * Usage mechanisms (e.g. Kibana applications, detections)
+ * ECS project (e.g. docs, tooling)
+The goal here is to research and understand the impact of these changes on users in the community and development teams across Elastic. 2-5 sentences each.
+-->
+
+## Concerns
+
+No concerns are known as of now. Presence of the `dimension:true` does not impact functionality. Elastic Stack version 8.7 is essential for this.
```
ebeahan, line 190: If 8.7+ is required to support `dimension` and the proposed changes here were added to ECS, would any integration package using the version of ECS including `dimension` need to explicitly support 8.7+ going forward?

rfcs/0000-tsdb-dimensions.md
agithomas: 1. The set of fields that are proposed here (as part of RFC) are those fields which must be annotated across all use cases (all datastreams in all packages). 

In addition to these fields, datastreams  may or may not have additional fields that may need `dimension:true` annotation. An integration package developer is expected to annotate ONLY the additional fields , if any, after this proposed change is accepted and included in ecs. 

2. With `dimension: true` annotation  and the user choose not to enable TSDB, it does not alter or affect the behaviour of the integration package.

rfcs/0000-tsdb-dimensions.md
agithomas: No, any package using the proposed ECS version can continue to serve Non-TSDB and TSDB settings because `dimension: true` supports both Non-TSDB and TSDB settings .

 A package can upgrade to use the proposed ECS version (with` dimension: true`) and continue to use a version prior to 8.7.

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+
+Integration package development is the key beneficiary of this change. The fields of the document that are received from an integration receives a field mapping. If and when TSDB benefits are to be utilised, along with the field mapping with a metric type, at least one of the fields  must receive `dimension: true` annotation.
```
ruflin, line 152: I expect this eventually to become the default mappings for any `metrics-*-*` data stream so it will benefit all metrics cases: https://github.com/elastic/elasticsearch/issues/85692

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+Stage 3: Document resolutions for all existing concerns. Any new concerns should be documented along with their resolution. The goal here is to eliminate risk of churn and instability by ensuring all concerns have been addressed.
+-->
+
+## People
+
+The following are the people that consulted on the contents of this RFC.
+
+* @agithomas | author
+* @ruflin | subject matter expert
+* @lalit-satapathy | reviewer
```
ruflin, line 209: Can we had here @martijnvg to have sign off from the TSDB team too?

agithomas: Sure, Done

rfcs/0000-tsdb-dimensions.md
P1llus: @agithomas Are you saying that unless we require that the package minimum version is 8.7, we cannot update our ECS versions?
At SEI we upgrade ECS versions for all our integrations every release, however we do not bump the minimum version for our packages..

P1llus:
https://github.com/elastic/ecs/pull/2172#discussion_r1117870796

This is quite concerning to me, you are asking us to bump the minimum version of all our packages to 8.7, that means no users of integrations below that version that ever receieve any hotfixes anymore..

rfcs/0000-tsdb-dimensions.md
agithomas: @P1llus , If only the package has to be TSDB enabled, proposed ECS (having `dimension : true`) is to be used. 

If the package do not want to have TSDB enablement, it can continue with old version of ECS.

P1llus: That is the problem though @agithomas . That means any package that wants to have other ECS changes other than `dimension: true` has to upgrade their minimum version.

Since we always update our ECS version of packages, we would lock users out of hotfixes for older versions.

While we had planned to bump the minimum version of packages at some point, this change would force that at this point.

ruflin: > That is the problem though @agithomas . That means any package that wants to have other ECS changes other than dimension: true has to upgrade their minimum version.

No. Even if you run Elasticsearch 8.3 and have `dimension: true` set, it will work as the field is just ignored. You just keep having the old behaviour.

agithomas:
> This is quite concerning to me, you are asking us to bump the minimum version of all our packages to 8.7, that means no users of integrations below that version can ever receieve any hotfixes anymore..

This RFC does not introduce any such limitations.  All packages can continue to choose their preferred version of ECS and ES version. 

**Only if a package needs TSDB**, it must satisfy two requirements (1. have the proposed ECS version , 2. ES Stack version 8.7 + ).

P1llus:
> > This is quite concerning to me, you are asking us to bump the minimum version of all our packages to 8.7, that means no users of integrations below that version can ever receieve any hotfixes anymore..
> 
> This RFC does not introduce any such limitations. All packages can continue to choose their preferred version of ECS and ES version.
> 
> **Only if a package needs TSDB**, it must satisfy two requirements (1. have the proposed ECS version , 2. ES Stack version 8.7 + ).

Oh okay, I thought that the package installation would error out, please ignore my comment then!

agithomas:
@ruflin , @lalit-satapathy , @ebeahan , @P1llus , @martijnvg , can you take a final look at the RFC ? If there are no further question / concerns, please approve the RFC.

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,240 @@
...
+  short: Fields describing the relevant computing instance.
+  description: >
+    A host is defined as a general computing instance.
+
+    ECS host.* fields should be populated with details about the host on which
+    the event happened, or from which the measurement was taken.
+    Host types include hardware, virtual machines, Docker containers, and Kubernetes nodes.
+  type: group
+  fields:
+    - name: hostname
```
ruflin, line 96: Only spotted this now. Should we select `host.hostname` or `host.name` ?

ruflin (commented):
It seems the fields listed in this PR is not aligned with https://github.com/elastic/integrations/issues/5193#issuecomment-1438212045 ? Can you share some background on why field were left out and others were selected?

rfcs/0000-tsdb-dimensions.md
agithomas: A related thread i found is here : https://github.com/elastic/observability-dev/issues/1056

Based on the inputs in the thread, `host.name` may be good option. I will change accordingly.

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,241 @@
...
+      type: keyword
+      dimension: true
+      short: Address of this service.
+      description: >
+        Address where data about this service was collected from.
+
+        This should be a URI, network address (ipv4:port or [ipv6]:port) or a resource path (sockets).
+      example: 172.26.0.2:5432
+
+```
```
lalit-satapathy, line 79: Is the list of dimensions given in the RFC below will be same as what is earlier documented [here](https://github.com/elastic/integrations/issues/5193#issuecomment-1438212045)?

agithomas:
@lalit-satapathy, @ruflin  ,can you please help by approving , if there are no further queries or changes?

P1llus (approved):

P1llus:
Would need another approval from @ebeahan as well at least.

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,332 @@
...
+    - name: id
+      level: core
+      type: keyword
+      short: Unique identifier of this agent.
+      description: >
+        Unique identifier of this agent (if one exists).
+
+        Example: For Beats this would be beat.id.
+      example: 8a4f500d
+      dimension: true
```
lalit-satapathy, line 137: Nit: Can we have a consistent ordering where dimension fields are added in this document. Some times it is added before and some time it is after the description field.

lalit-satapathy:
@agithomas,

What will happen to the already manually added dimension fields (example [oracle](https://github.com/elastic/integrations/blob/main/packages/oracle/data_stream/performance/fields/ecs.yml)) once the ECS dimensions updated centrally?  Do these locally mappings need to be removed OR the central ECS mapping takes a precedence?

ebeahan (approved):
LGTM.

@martijnvg's review was requested, so we'll wait for their review. Let us know if we should move forward without it.

martijnvg (approved):
From an Elasticsearch perspective the changes in this PR look good to me. 
Note that I'm not an ecs expert here, so my approval should not be weighted heavily.

agithomas:
@lalit-satapathy , if the Stage -0 proposal looks good to you, can I have your approval?

lalit-satapathy (approved):

lalit-satapathy:
> @lalit-satapathy , if the Stage -0 proposal looks good to you, can I have your approval?
done..

agithomas:
@ebeahan , 

Are any more approvals needed before going to Stage -1 of RFC ? 

From the RFC steps, i understand that the next step in the process is

`ECS committee and/or team member merges RFC, formally accepting the strawperson`

Is this something that you can help me with ?

rfcs/0000-tsdb-dimensions.md
```diff
@@ -0,0 +1,333 @@
...
+  short: Fields describing the relevant computing instance.
+  description: >
+    A host is defined as a general computing instance.
+
+    ECS host.* fields should be populated with details about the host on which
+    the event happened, or from which the measurement was taken.
+    Host types include hardware, virtual machines, Docker containers, and Kubernetes nodes.
+  type: group
+  fields:
+    - name: name
```
tetianakravchenko, line 96: any reason why not to consider `host.id`? https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-id
as mentioned in description of `host.name`:
```
It can contain what hostname returns on Unix systems, the fully qualified domain name (FQDN), or a name specified by the user. The recommended value is the lowercase FQDN of the host.
```
since it can be specified by user - it might be not unique.

in combination with other fields dimension _should_ be unique I guess though

ebeahan: The proposal as-is will be merged as Stage 0, but this concern is worth discussing in Stage 1+ and capturing. I'll make a note this item is brought up in the next stage's PR.

cc @agithomas

ebeahan:
> Is this something that you can help me with ?

Yes I'll merge.

rfcs/0000-tsdb-dimensions.md
lalit-satapathy: > any reason why not to consider `host.id`? https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-id as mentioned in description of `host.name`:
> 
> ```
> It can contain what hostname returns on Unix systems, the fully qualified domain name (FQDN), or a name specified by the user. The recommended value is the lowercase FQDN of the host.
> ```
> 
> since it can be specified by user - it might be not unique.
> 
> in combination with other fields dimension _should_ be unique I guess though

@agithomas ?

agithomas: > any reason why not to consider `host.id`? https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-id as mentioned in description of `host.name`:
> 
> ```
> It can contain what hostname returns on Unix systems, the fully qualified domain name (FQDN), or a name specified by the user. The recommended value is the lowercase FQDN of the host.
> ```
> 
> since it can be specified by user - it might be not unique.
> 
> in combination with other fields dimension _should_ be unique I guess though

Based on my discussion with ElasticSearch team, `host.name` is preferred over `host.ip`. host.ip values change often and hence it is not a preferred dimension field. 

`host.name`, in combination with other dimension fields provides a unique identifier of the asset / resource getting monitored.

agithomas: Please ignore my previous comment, i mis-read it as `host.ip`. 
I shall go through the `host.id`, its values & cardinality and its qualification of it becoming dimension field during RFC-1 phase

jsoriano:
Hey,

Sorry, I forgot to look for this PR after we talked about this last week.

Are these dimensions intended to be imported automatically in packages? If that's the case, I have a concern. Changing the set of dimensions for a time series will create a new time series. For example, if `kubernetes.pod.id` is enough to define a time series, adding `host.id` later will create a different one, even if it refers to the same time series. This may cause things as duplication of time series in visualizations. Doing it automatically may be overlooked by package developers.

In general I think that we shouldn't automatically include dimensions in packages, given that they are quite specific to each use case. As mentioned in the RFC itself, it is important to select an adequate number of fields as dimensions. In my opinion, package developers should take this decision for each case.

It can be great though to document in ECS what fields are good candidates to be considered dimensions.

lalit-satapathy:
@jsoriano,

Thanks for the feedbacks, can discuss this further (offline if needed?). Do you see an overlap  among the 7 defined [common](https://github.com/elastic/integrations/blob/main/docs/developer_tsdb_migration_guidelines.md) ECS fields defined by @agithomas (its now _host.name_ instead of _host.ip)_

ruflin:
> Are these dimensions intended to be imported automatically in packages? If that's the case, I have a concern. 

I think we should. A package dev should not have to think about the complexity of dimensions for standard cases. For example, if someone builds a package to run on its host and adds host.id as dimension, now suddenly it's also run in k8s and it might not work because container.id is missing. There are of course cases like "disk metrics" where additional dimensions are needed and the user has to think about it. @felixbarny You might have an opinion here too, as I think you are thinking in the direction of "most keyword fields" are a dimension.


> Changing the set of dimensions for a time series will create a new time series. For example, if kubernetes.pod.id is enough to define a time series, adding host.id later will create a different one, even if it refers to the same time series.

I had a similar discussion in the past with @martijnvg and if I remember correctly, he wasn't too concern. If we already start with this basic set of dimension, the above should not happen. It only happens when the data changes and then I guess it could be argued, it might be a new time series? But I agree, something we should also discuss with Lens team in more details to not have unexpected results.

jsoriano:
> For example, if someone builds a package to run on its host and adds host.id as dimension, now suddenly it's also run in k8s and it might not work because container.id is missing.

Yes, this is a good point to have something like this, along with the automatic inclusion of full ECS with dynamic mappings (https://github.com/elastic/elastic-package/pull/1073). To support metadata added by autodiscover or processors.

I am more concerned about the other way around, data streams that already have enough dimensions, but after upgrading ECS, new time series are created. This would be for example the case of container metrics that can have `container.id` as single dimension, but after adding `host.id` as dimension, they have redundant dimensions, and new time series. This will also happen on most services that have randomized identifiers or uuids.

> If we already start with this basic set of dimension, the above should not happen. It only happens when the data changes and then I guess it could be argued, it might be a new time series? 

It will happen every time a new dimension is added, even if the data doesn't change. This is what I find misleading. Package developers may not be aware that by updating ECS they are creating different time series.
But well, this is mitigated if we add the set of dimensions on this RFC by now, and we are very strict on what we add later. Perhaps we can also mitigate this with tooling (with something like https://github.com/elastic/elastic-package/issues/579).

> But I agree, something we should also discuss with Lens team in more details to not have unexpected results.

This wouldn't be an issue only with Lens, it will affect any feature that makes use of dimensions (query languages, downsampling...).

ruflin:
> It will happen every time a new dimension is added, even if the data doesn't change. This is what I find misleading. Package developers may not be aware that by updating ECS they are creating different time series.

I missed the opt-out problem. Very good point. A user should be able to add ECS but opt out of the magical dimensions.

jsoriano:
> > It will happen every time a new dimension is added, even if the data doesn't change. This is what I find misleading. Package developers may not be aware that by updating ECS they are creating different time series.
> 
> I missed the opt-out problem. Very good point. A user should be able to add ECS but opt out of the magical dimensions.

I have done a quick check with elastic-package and it won't import dimensions from ECS as it is now. So as we have to do some specific implementation for this we can take this concern into account there. We can maybe have it disabled by default, but enabled in new packages.

I am going to open a quick PR to add this concern to the list of concerns, mentioning that it needs to be taken into account by implementations consuming ECS. And regarding the implementation for Elastic Packages we can continue with the discussion on elastic-package/package-spec.

Thanks!

felixbarny:
> It will happen every time a new dimension is added, even if the data doesn't change. This is what I find misleading. Package developers may not be aware that by updating ECS they are creating different time series.

I don't think that creating a new time series would be breaking or confusing in any way. The actual _tsid is an internal implementation detail and changes to it shouldn't leak to to user. If that statement is true, we don't need to be too strict about dimensions we're adding after the fact. The only limiting factor is the dimension limit, which we will hopefully get rid of soon:
* https://github.com/elastic/elasticsearch/issues/93564

> @felixbarny You might have an opinion here too, as I think you are thinking in the direction of "most keyword fields" are a dimension.

I guess for integrations we can be specific about which fields we want to map as dimensions. But for custom metrics (`metrics-*-*`), I'd like to get to a place where any non-metric field is a dimension by default. If we don't do that, we require users to first create an explicit mapping before they can ingest their custom metrics. I think that's just way too burdensome. After initial onboarding, users can still tune their mappings and remove some dimension fields. Again, we can do that for our integrations if we think that's time well spent. But if there's no dimension limit and no significant penalty of using too many dimensions, I don't see a good reason to agonize about the dimensions fields.

felixbarny:
> If that's the case, then I agree that we shouldn't care about having too many dimensions, but I guess we are not on this situation yet?

Yes, that's just aspirational.

> This is actually also a burden for integrations development. It is time well spent in the sense that in theory this is time saved to users. Anything we can do to avoid needing to define mappings in TSDB use cases will be welcome.

Sure, the integration mappings need to work out-of-the-box for TSDB with no user configuration. That's a given. What I meant was when we're in this aspirational world where we don't have limits on dimensions, we can choose to not spend too many brain cycles agonizing about what the minimum set of dimensions is. Or we could opt-out instead of opt-in.

jsoriano:
Ok, let's forget about my concern then, and think about opt-in or opt-out approaches on the specific implementations handling these dimensions.

Thanks!

PR #2178, opened 2023-03-02
[RFC] Extend event.kind allowed values with asset

taylor-swanson:
This RFC proposes extending `event.kind` with an additional allowed value, `asset`. For the Entity Analytics project, we are looking to identify user, host, and other types of entity assets.

### DIFF
--- a/rfcs/text/0038-event-kind-asset.md
+++ b/rfcs/text/0038-event-kind-asset.md
@@ -35,11 +35,11 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
 
 ## Usage
 
-For the Entity Analytics project, we are looking to identify user, host, and other types of entity assets. The `event.kind` field with `asset` value will be used to identity these type of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
+For the Entity Analytics project, we are looking to identify user, host, and other types of entity assets. The `event.kind` field with `asset` value will be used to identity these types of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
 
 * `event.kind: asset`
 * `event.category: iam`
-* `event.type: info`
+* `event.type: user,info`
 
 <!--
 Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
@@ -150,7 +150,7 @@ e.g.:
 
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
-* Stage 0: https://github.com/elastic/ecs/pull/NNN
+* Stage 0: https://github.com/elastic/ecs/pull/2178
 
 <!--
 * Stage 1: https://github.com/elastic/ecs/pull/NNN
--- a/rfcs/text/0038-event-kind-asset.md
+++ b/rfcs/text/0038-event-kind-asset.md
@@ -35,11 +35,11 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
 
 ## Usage
 
-For the Entity Analytics project, we are looking to identify user, host, and other types of entity assets. The `event.kind` field with `asset` value will be used to identity these types of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
+We want to represent a catalog of assets where each document represents some type of user or device entity. The `event.kind` field with `asset` value will be used to identity these types of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
 
 * `event.kind: asset`
-* `event.category: iam`
-* `event.type: user,info`
+* `event.category: [iam]`
+* `event.type: [user, info]`
 
 <!--
 Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
@@ -61,7 +61,9 @@ An example of a user entity document, with `event.kind: asset`:
                 "provider": "Azure AD",
                 "kind": "asset",
                 "action": "user-discovered",
-                "category": "iam",
+                "category": [
+                    "iam"
+                ],
                 "type": [
                     "user",
                     "info"
--- a/rfcs/text/0038-event-kind-asset.md
+++ b/rfcs/text/0038-event-kind-asset.md
@@ -1,0 +1,158 @@
+# 0038: Extend event.kind allowed values with asset
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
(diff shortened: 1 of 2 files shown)


taylor-swanson:
@SourinPaul or @MikePaquette, feel free to add any links or other notes that should be added to the RFC. (If you want, just comment and I'll add it to the markdown document).

rfcs/text/0038-event-kind-asset.md
```diff
@@ -0,0 +1,158 @@
...
+-->
+
+<!--
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+For the Entity Analytics project, we are looking to identify user, host, and other types of entity assets. The `event.kind` field with `asset` value will be used to identity these types of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
```
andrewkroh, line 38: This kind of assumes the audience knows what the "Entity Analytics project" is. The ECS community might not have this knowledge. I would probably omit that reference since the rest of your usage statement can stand on its own.

I was thinking "We want to represent a catalog of assets using ECS where each document represents some type or user or device entity."

rfcs/text/0038-event-kind-asset.md
```diff
@@ -0,0 +1,158 @@
...
+
+* `event.kind: asset`
+* `event.category: iam`
+* `event.type: user,info`
```
andrewkroh, line 42: Should we represent these values as arrays?

```suggestion
* `event.category: [iam]`
* `event.type: [user, info]`
```

rfcs/text/0038-event-kind-asset.md
```diff
@@ -0,0 +1,158 @@
...
+                "type": "entity-analytics"
+            },
+            "@timestamp": "2023-02-23T20:03:17.489Z",
+            "event": {
+                "agent_id_status": "verified",
+                "ingested": "2023-02-23T20:03:18Z",
+                "provider": "Azure AD",
+                "kind": "asset",
+                "action": "user-discovered",
+                "category": "iam",
```
andrewkroh, line 64: This should be an array.

rfcs/text/0038/event.yml
```diff
@@ -0,0 +1,53 @@
...
+            Examples include lists of user identities or accounts ingested from directory
+            services such as Active Directory (AD), inventory of hosts pulled from
+            configuration management databases (CMDB), and lists of cloud storage buckets
+            pulled from cloud provider API's.
+
+            `event.kind: asset` is not used for normal system events or logs that are coming
+            from an asset/entity, nor is it used for system events or logs coming from a
+            directory or CMDB system.
+
+            By contrast, `event.kind: enrichment`` is used to indicate events whose primary
```
andrewkroh, line 50: Is this left over from something you copied?

rfcs/text/0038-event-kind-asset.md
```diff
@@ -0,0 +1,160 @@
+# 0038: Extend event.kind allowed values with asset
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 4: Since the scope of this change is small and the proposal is already well detailed, I propose we target stage 2 initially instead of stage 0.

By moving to stage 2, `event.kind: asset` will be added into ECS but as a `beta` allowed value. We'll still have some flexibility to make adjustments if necessary.

@taylor-swanson if you agree with targeting to stage 2, can you capture any necessary detail in the `Scope of impact` and `Concerns`?

rfcs/text/0038/event.yml
```diff
@@ -0,0 +1,48 @@
...
+      type: keyword
+      short: The kind of the event. The highest categorization field in the hierarchy.
+      description: >
+        This is one of four ECS ... # Omitted for brevity
+      example: alert
+      allowed_values:
+        - name: asset
+          description: >
+            This value indicates events whose primary purpose is to store an inventory of
+            assets/entities and their attributes. Assets/entities are objects that are
```
ebeahan, line 38: I think it should be clearer here about what `objects` fall under our definition of `asset`. Users and devices are implied from the examples later, but I think it's best our definition be explicit.

We can take it even further and clarify this value is reserved and to be used by features of the Elastic stack, like [`signal`](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-kind.html#ecs-event-kind-signal).

rfcs/text/0038-event-kind-asset.md
```diff
@@ -0,0 +1,160 @@
...
+-->
+
+<!--
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+We want to represent a catalog of assets where each document represents some type of user or device entity. The `event.kind` field with `asset` value will be used to identity these types of documents. It has also been proposed that we combine this with other fields to have a more exact definition. For example, a user document may have these event fields specified:
```
ebeahan, line 38: In the final (stage 3) version, it may be nice to have examples for other entities included here for completeness, like groups and devices, if those definitions are finalized at that point.

rfcs/text/0038-event-kind-asset.md
taylor-swanson: @ebeahan, That seems to be reasonable to me.

@SourinPaul, do we have anything specific to call out here for Scope of Impact or Concerns? I can add them to the PR once we have them.

taylor-swanson: cc @MikePaquette in case you have any additional input here (see my comment for Sourin immediately above).

SourinPaul: **Scope of Impact:**
We expect to use the `asset` value within the security solution to consistently across asset management, external asset integration, and persisting related metadata. This field may be further leveraged in o11y topology work as they also are exploring ways to extend the asset.* fields within ECS.

**Concerns:**
There are no breaking changes identified. I do not have any concerns to note at this time.

@taylor-swanson, please let me know if I can elaborate anything further. Thanks for driving this.

taylor-swanson: Thanks, @SourinPaul, I think this looks good.

I think this could wait to Stage 3 as Eric mentioned, but to add to his comment below, should we include verbiage in description for `asset` to say something on the lines of:

> This value is used by Elastic Security for asset management solutions.
>
> Usage of this value is reserved, and data ingestion pipelines must not populate event.kind with the value "asset".

ebeahan: > data ingestion pipelines must not populate event.kind with the value "asset".

For `event:kind: asset` events, data pipelines (like the Azure AD user entity integration) _will be_ populating this field, right?

ebeahan (approved):
LGTM

PR #2196, opened 2023-04-18
Add concern about automatic inclusion of dimensions

jsoriano:
Follow up of conversation in https://github.com/elastic/ecs/pull/2172#issuecomment-1512684879.


ruflin (commented):
Based on where the discussion is heading in https://github.com/elastic/ecs/pull/2172, I wonder what we should do about this?

jsoriano:
> Based on where the discussion is heading in #2172, I wonder what we should do about this?

Let's close this, it doesn't seem to be a concern for consumers of ECS, and we can think more about how to support this on the elastic-package implementation.

PR #2201, opened 2023-04-25
[RFC] Stge 0 - New fieldset for volume device

Trinity2019:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? - Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)? - Yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)? - Yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? - N/A
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? - N/A
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. - Yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? - N/A


rfcs/text/0040-removable-media-device.md
```diff
@@ -0,0 +1,133 @@
+# 0000: volume device
```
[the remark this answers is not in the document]
stanek-michal: `storage` or `volume` sound fine to me

rfcs/text/0040-volume-device.md
```diff
@@ -0,0 +1,138 @@
...
+<!--
+As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
+Feel free to remove these comments as you go along.
+-->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+This RFC propose adding the volume device fieldset to describe volume storage devices that are removable disks such as USB, mountable virtual disks such as ISO.
```
ebeahan, line 16: I'm uncertain about using `volume.*` for the field set name if the intended use is specifically removable disks. A storage volume can be removable disk, but there are other types of physical and logical volumes users may want to describe.

ebeahan (commented):
Thanks for the proposal, @Trinity2019.

The motivation of describing alerts and event associated with volumes makes sense as an addition to ECS.

rfcs/text/0040-volume-device.md
Trinity2019: Hi @ebeahan I changed description to also include hard disks, for our requirement could also include non-removable disks.
https://github.com/elastic/ecs/pull/2201/commits/fe9c40c303d679cf9b29718df58c8284042f9a58

rfcs/text/0040-volume-device.md
```diff
@@ -0,0 +1,138 @@
...
+ * volume.writable
+ * volume.default_access
+ * volume.file_system_type
+ * volume.product_id
+ * volume.product_name
+ * volume.vendor_id
+ * volume.vendor_name
+ * volume.serial_number
+ * volume.volume_device_type
+ * volume.action
```
Trinity2019, line 32: `volume.action` will be like "attach", "detach", "mount", "unmount". I'm hesitating to place it here. Seems `event.action` is a better place to save such information. Because these volumes fields will be part of a device event, similar to process events whose `event.action` can be "start"/"end".
thoughts?

ebeahan: I'd also prefer to use the existing `event.action` field over introducing an additional `*.action` field, if we can avoid it.

ebeahan (approved):
LGTM.

Should anyone else give another review before we merge this RFC as stage 0?

stanek-michal:
> LGTM.
> 
> Should anyone else give another review before we merge this RFC as stage 0?

I reviewed this as well, I think we can go ahead with merging

PR #2206, opened 2023-05-01
Adding package.vendor to ECS package mapping

lkuik:
Adding an ECS mapping for package that maps to package.vendor to accompany other fields such as package.name and package.version

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -15,7 +15,7 @@ Thanks, you're awesome :-) -->
 #### Bugfixes
 
 #### Added
-
+* Added `package.vendor` since we already had package version and package name. 
 #### Improvements
 
 #### Deprecated
--- a/schemas/package.yml
+++ b/schemas/package.yml
@@ -25,6 +25,12 @@
   type: group
   fields:
 
+    - name: vendor
+      level: extended
+      type: keyword
+      description: Vendor name
+      example: Google
+
     - name: name
       level: extended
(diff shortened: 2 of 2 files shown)


ebeahan:
@lkuik can you elaborate a bit more about how you plan to use a `package.vendor` field?

My first thought is many packages aren't authored or maintained by a vendor, but instead by an individual, group of maintainers, project, etc. I took a brief look across different package registries and package formats (`npm`, `pypi`, `gem`, `deb`, `rpm`), and I'm not seeing consistent convention.

ebeahan:
The tests are failing because because `make` needs run and the generated files commited.

lkuik:
> @lkuik can you elaborate a bit more about how you plan to use a `package.vendor` field?
> 
> My first thought is many packages aren't authored or maintained by a vendor, but instead by an individual, group of maintainers, project, etc. I took a brief look across different package registries and package formats (`npm`, `pypi`, `gem`, `deb`, `rpm`), and I'm not seeing consistent convention.

That's a great point @ebeahan 
I guess I was thinking more application than package in this context, and perhaps that was my mistake.
I didn't see anything for application ECS mapping so was thinking package was the more appropriate ECS field mapping to use.

I see your point now, will have to think about that, does what I'm saying make sense?

lkuik:
Just following up if you saw my post @ebeahan in regards to where I was coming from, was looking to get your thoughts based on that.

PR #2215, opened 2023-06-06
Add files via upload

SourinPaul:
Strawman draft for **Asset Integrations and Entity Store RFC**

### DIFF
--- a/rfcs/0000-User&HostIntegrationRFC.md
+++ b/rfcs/0000-User&HostIntegrationRFC.md
@@ -1,0 +1,240 @@
+# 0000: Asset Integration
+<!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
+
+- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Date: **TBD** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+
+<!--
+As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
+Feel free to remove these comments as you go along.
+-->
+
+
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+This proposal aims to extend existing ECS fieldset to store inventory metadata for hosts, users from external application repositories. Using ECS to store such fields will improve metadata queryring and retrieval across various use cases.
+
+Within the Security `Entity Analytics` initiatve, we refer to hosts and users as `entities`. We also refer to hosts/ users as `assets` across generic security or obserbality use cases. For simplificaiton, entities, assets will refer to host and user objects.
+
+This proposal includes the following:
+* Additional fields in the `users` object 
+* Introduces a new field set called `assets`
+<!-- * Additional fields in the `host` object --->
+
+This proposal will also facilitate storing of host and user inventory within security solution (aka. the entity store).
+
+### References:
+* [Entity Analytics Plugin Architecture for Host & User Integrations](https://github.com/elastic/beats/issues/33866#issuecomment-1165709940)
+* 0038: Extend event.kind allowed values with asset [Existing RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0038-event-kind-asset.md#0038-extend-eventkind-allowed-values-with-asset)
+* [Creation of an asset `super-schema`](https://github.com/elastic/security-team/issues/2347)
+
+<!--
+Stage 1: If the changes include field additions or modifications, please create a folder titled as the RFC number under rfcs/text/. This will be where proposed schema changes as standalone YAML files or extended example mappings and larger source documents will go as the RFC is iterated upon.
+-->
+
+<!--
+Stage X: Provide a brief explanation of why the proposal is being marked as abandoned. This is useful context for anyone revisiting this proposal or considering similar changes later on.
+-->
+
+## Fields
+
+<!--
+Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+### Proposed New Fields for User object
+
+Field | Type | Example | Description
+--- | --- | --- | ---
+user.profile.id	| keyword	| 1234	| User ID from the identity datasource.
+user.profile.type	| keyword	| Employee	| Type of user account.
+user.profile.status	| keyword |	On board	| Status of the user account.
+user.profile.first_name	| keyword |	First	| First Name of the User.
+user.profile.last_name	| keyword |	Last	| Last Name of the user.
+user.profile.other_identities	| keyword , multifield |	first.last@elk.elastic.co	| Array of additional user identities.
+user.profile.manager	| keyword |	John Doe	| Assigned Manager for the user account.
+user.profile.employee_type	| keyword |	Regular | Further classification type for the user account.
+user.profile.job_family	| keyword |	65-Sales	| Job family associated with the user account.
+user.profile.job_family_group	| keyword |	GTM	| Job family group associated with the user account.
+user.profile.management_level	| keyword |	Individual Contributor	| If the user account is identified as a Manager or Individual contributor.
+user.profile.job_title	| keyword |	Field Sales	| Job title assigned to the user account.
+user.profile.department	| keyword |	x256	| Department name associated with the user account.
+user.profile.organization	| keyword |	Elasticsearch Inc.	| Organization name associated with the account.
+user.profile.location	| keyword |	US - Washington - Distributed	| Assigned location for the user account.
+user.profile.mobilePhone	| keyword |	222-222-2222	
+user.profile.primaryPhone	| keyword |	222-222-2222	
+user.profile.secondEmail	| keyword |	first.l@elastic.co	| Additional email addresses associated with the user account.
+user.profile.sup_org_id	| keyword |	SUP-ORG-75	| Primary organization ID for the user account.
+user.profile.supervisory_Org	| keyword |	Field Sales	| Primary organization name for the user account.
+user.profile.assigned_mdm_id	| keyword |	2950	| Primary host owned by the user. This acts as a correlation to identify the host event.			
+user.account.create_date	| date |	June 5, 2023 @ 18:25:57.000	| Date account was created.
+user.account.activated_date	| date |	June 5, 2023 @ 18:25:57.000	| Date account was activated.
+user.account.change_date	| date |	June 5, 2023 @ 18:25:57.000	| Date user account record was last updated at source
+user.account.status.active	| boolean |	true/ false	| A flag indicating if account is active
+user.account.status.recovery	| boolean |	true/ false	| A flag indicating if account is in recovery
+user.account.status.locked_out	| boolean |	true/ false	| A flag indicating if account is currently locked out
+user.account.status.suspended	| boolean |	true/ false	| A flag indicating if account has been suspended
+user.account.isAdmin	| boolean |	true/ false	| A flag indicating if account is an Admin account
+user.account.isDelegatedAdmin	| boolean |	true/ false	| A flag indicating if account has Delegated Admin rights
+user.account.isPriviledged	| boolean |	true/ false	| A flag indicating if account is a Privileged account
+user.account.status.password_expired	| boolean |	true/ false	| A flag indicating if account password has expired.
+user.account.status.deprovisioned	| boolean |	true/ false	| A flag indicating if account has been deprovisioned
+user.account.password_change_date	| date |	June 5, 2023 @ 18:25:57.000	| Last date/time when account password was updated
+
+### Proposed New Fields for Asset object
+
+Field | Type | Generic Example |	User Entity Example | Host Entity Example | Description
+--- | --- | --- | --- | --- | ---
+asset.category	| keyword |	-	        | Null	                | hardware	                | A further classification of the asset type beyond event.category. For example for host assets {hardware, virtual, container, node}. For user assets {NULL ?}			
+asset.type	    | keyword |	-	        | Null	                | workstation	            | A sub classification of asset. For host assets {workstation, S3, Compute}. For user assets {NULL?}.			
+asset.id	    | keyword |	-	        | 00uhs72c27s6PiK7x1t7	| 2950	                    | A unique ID for the asset. For inventory integrations, it's the id generated from inventory datasource			
+asset.name	    | keyword |	-	        | Sourin Paul	        | Sourin Paul Macbook Pro	| A common name for the asset.			
+asset.vendor	| keyword |	-           |	-	                | Apple	                    | Used primarily for 'Host' entities, the vendor name or brand associated with the asset.			
+asset.product	| keyword |	-           |	-	                | MacBook Pro	            | Used primarily for 'Host' entities, the product name associated with the asset.			
+asset.model	    | keyword |	-           |	-	                |TBD	                    | Used primarily for 'Host' entities, the model name or number associated with this asset.			
+asset.version	| keyword |	-           |	-	                | TBD	                    | Used primarily for 'Host' entities, the version or year associated with the asset.			
+asset.owner	    | keyword |	-           |	-	                | sourin.paul@elastic.co	| The primary user entity who owns the 'Host' asset.			
+asset.priority	| keyword |	Priority 1	| -                     | -                         | A priority classification for the asset obtained from outside this system, such as from some external CMDB or Directory service.			
+asset.criticality	| keyword |	Critical	| - | -                                         | A criticality classification obtained from outside this system, such as from some external CMDB or Directory service.			
+asset.business_unit	| keyword |	Analyst Experience	| - | -                                 | Business Unit associated with the asset (user or host).			
+asset.costCenter	| keyword |	Security - Protections | - | -                              | Cost Center associated with the asset (user or host).			
+asset.cost_center_hierarchy	| keyword |	Engineering	 | - | -                                | Additional cost center information associated with the asset (user or host).			
+asset.status	    | keyword         |	ACTIVE      | - | -                                 | Current status of the asset in the inventory datasource.			
+asset.last_status_change_date	| date |	June 5, 2023 @ 18:25:57.000	| - | -             | The most recent date/time when the asset.status was updated.			
+asset.create_date	            | date |	June 5, 2023 @ 18:25:57.001	| - | -             | For users, it's the hire date, for other assets, it's the in service date.			
+asset.end_date	                | date |	June 5, 2023 @ 18:25:57.002	| - | -             | For users, it's the termination date, for other assets, it's the out-of-service date.			
+asset.first_seen	            | date |	June 5, 2023 @ 18:25:57.003	| - | -             | The earliest date/time at which this asset was observed.  For assets added via config, this would remain empty until the asset was observed.			
+asset.last_seen	                | date |	June 5, 2023 @ 18:25:57.004	| - | -             | The most recent date/time at which this asset was observed.  Would remain empty until the asset was observed.			
+asset.last_updated	            | date |	June 5, 2023 @ 18:25:57.005	| - | -             | The most recent date/time this asset was updated in the inventory datasource.			
+
+**Note:** I will append the asset field sets with additional fields needed to ingest host inventory metadata. I will evaluate if any further changes are needed under the host.* object during this excercise.
+
+
+#### Nesting of existing risk.* fields under asset object
+* We have a set of risk.* fields in ECS which can be further nested under the asset.* object. Reference to [Risk RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0031-risk-fields.md).
+
+<!--
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+
+* As part of Entity Analytics, we are ingesting metadata about [Users](https://github.com/elastic/security-team/issues/4163) and [Hosts](https://github.com/elastic/security-team/issues/4044) from various external vendor applications. We are storing all ingested metadat in ElasticSearch inde. After we map these fields to ECS, we will enrich these ingested events for risk scoring scenarios (e.g. of context enrichments) and detecting advanced analytics (UBA) use cases. 
+
+* This schema will be used to persist `Observed` (queried) entities from ingested security log dataset in an Entity store. This entity store can be further extended to meet broaded overall Asset Mangagement needs.
+
+* We also envision additional enrichment use cases for existing prebuilt detection rules down the line.
+
+
+## Source data
+
+<!--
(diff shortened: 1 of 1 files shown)


rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+<!--
+As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
+Feel free to remove these comments as you go along.
+-->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+This proposal aims to extend existing ECS fieldset to store inventory metadata for hosts, users from external application repositories. Using ECS to store such fields will improve metadata queryring and retrieval across various use cases.
```
ebeahan, line 16:
```suggestion
This proposal aims to extend existing ECS fieldset to store inventory metadata for hosts, users from external application repositories. Using ECS to store such fields will improve metadata querying and retrieval across various use cases.
```

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+* Additional fields in the `users` object 
+* Introduces a new field set called `assets`
+<!-- * Additional fields in the `host` object --->
+
+This proposal will also facilitate storing of host and user inventory within security solution (aka. the entity store).
+
+### References:
+* [Entity Analytics Plugin Architecture for Host & User Integrations](https://github.com/elastic/beats/issues/33866#issuecomment-1165709940)
+* 0038: Extend event.kind allowed values with asset [Existing RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0038-event-kind-asset.md#0038-extend-eventkind-allowed-values-with-asset)
+* [Creation of an asset `super-schema`](https://github.com/elastic/security-team/issues/2347)
```
ebeahan, line 30: Please remove links to non-public repos since this proposal will be public facing. If you want to link the issues together, comment on the private issue referencing the public one.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+user.profile.manager	| keyword |	John Doe	| Assigned Manager for the user account.
+user.profile.employee_type	| keyword |	Regular | Further classification type for the user account.
+user.profile.job_family	| keyword |	65-Sales	| Job family associated with the user account.
+user.profile.job_family_group	| keyword |	GTM	| Job family group associated with the user account.
+user.profile.management_level	| keyword |	Individual Contributor	| If the user account is identified as a Manager or Individual contributor.
+user.profile.job_title	| keyword |	Field Sales	| Job title assigned to the user account.
+user.profile.department	| keyword |	x256	| Department name associated with the user account.
+user.profile.organization	| keyword |	Elasticsearch Inc.	| Organization name associated with the account.
+user.profile.location	| keyword |	US - Washington - Distributed	| Assigned location for the user account.
+user.profile.mobilePhone	| keyword |	222-222-2222
```
ebeahan, line 65:
```suggestion
user.profile.mobile_phone	| keyword |	222-222-2222	
```

ECS fields [conventions](https://www.elastic.co/guide/en/ecs/current/ecs-guidelines.html#_guidelines_for_field_names) use all lower case letters and underscore (`_`) to combine words in field names

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+
+#### Nesting of existing risk.* fields under asset object
+* We have a set of risk.* fields in ECS which can be further nested under the asset.* object. Reference to [Risk RFC](https://github.com/elastic/ecs/blob/main/rfcs/text/0031-risk-fields.md).
+
+
+
+### Proposed New Fields for Host object
+Field | Type | Example | Description
+--- | --- | --- | ---
+host.os.build	| keyword		| 22F66   | Host OS Build information
```
ebeahan, line 125: The current convention for build info is populating into the `host.os.kernel` (e.g. `winlogbeat` populates with `10.0.17763.4377 (WinBuild.160101.0800)` on a Windows Server 2019 Datacenter system). Is the idea here to create another field solely for the build number?

For this change as proposed, the field would go in as `os.build` under `os.*` and populate anywhere `os.*` fields are nested: `host.os.build`, `observer.os.build`, and `user_agent.os.build`.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+host.installed_extensions 	| keyword	  | Nested objects	  | List of installed extensions along with their metadata
```
ebeahan, line 126: What is the motivation for adding `installed_extensions` and `installed_applications` underneath `host.*` instead of the proposed `asset.*` field set?

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+Stage 2: Add or update all remaining field definitions. The list should now be exhaustive. The goal here is to validate the technical details of all remaining fields and to provide a basis for releasing these field definitions as beta in the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+## Usage
+
+<!--
+Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
+-->
+
+* As part of Entity Analytics, we are ingesting metadata about [Users](https://github.com/elastic/security-team/issues/4163) and [Hosts](https://github.com/elastic/security-team/issues/4044) from various external vendor applications. We are storing all ingested metadat in ElasticSearch inde. After we map these fields to ECS, we will enrich these ingested events for risk scoring scenarios (e.g. of context enrichments) and detecting advanced analytics (UBA) use cases.
```
ebeahan, line 141:
```suggestion
* As part of Entity Analytics, we are ingesting metadata about [Users](https://github.com/elastic/security-team/issues/4163) and [Hosts](https://github.com/elastic/security-team/issues/4044) from various external vendor applications. We are storing all ingested metadata in Elasticsearch index. After we map these fields to ECS, we will enrich these ingested events for risk scoring scenarios (e.g. of context enrichments) and detecting advanced analytics (UBA) use cases. 
```

Please remove links to non-public issues.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+## Source data
+
+<!--
+Stage 1: Provide a high-level description of example sources of data. This does not yet need to be a concrete example of a source document, but instead can simply describe a potential source (e.g. nginx access log). This will ultimately be fleshed out to include literal source examples in a future stage. The goal here is to identify practical sources for these fields in the real world. ~1-3 sentences or unordered list.
+-->
+
+There are many sources of asset inventory repositories. In the mid-term we are planning to ingest data from the following applicaiton providers:
+
+### User (Identity) repository sources:
+*  Azure Active Directory
```
ebeahan, line 157:
```suggestion
* Azure Active Directory
```

nit: removing extra spacing

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+* GitHub
+
+### Host repository sources:
+* Azure Active Directory
+* Jamf 
+* Active Directory DS
+* MS Intune
+* ServiceNow Asset CMDB
+
+Mapping of application fields from individual providers are documented here: [User](https://docs.google.com/spreadsheets/d/1ZHGQPGVr35vLZB3DcPnXWxLTkHad_S0YrhtTKO_LL4o/edit#gid=1940810377), [Host-TBD]().
```
ebeahan, line 171: Like any linked GitHub issues, any other linked resources should be publicly viewable for the community.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+Stage 3: Document resolutions for all existing concerns. Any new concerns should be documented along with their resolution. The goal here is to eliminate risk of churn and instability by ensuring all concerns have been addressed.
+-->
+
+## People
+
+The following are the people that consulted on the contents of this RFC.
+
+* @sourinpaul | author
+* @andrewkroh | subject matter expert
+* @jamie Hynds | subject matter expert
```
ebeahan, line 222:
```suggestion
* @jamiehynds | subject matter expert
```

nit: fixing handle

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+
+## References
+
+<!-- Insert any links appropriate to this RFC in this section. -->
+
+### RFC Pull Requests
+
+<!-- An RFC should link to the PRs for each of it stage advancements. -->
+
+* Stage 0: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 249:
```suggestion
* Stage 0: https://github.com/elastic/ecs/pull/2215
```

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+### Proposed New Fields for User object
+
+Field | Type | Example | Description
+--- | --- | --- | ---
+user.profile.id	| keyword	| 1234	| User ID from the identity datasource.
+user.profile.type	| keyword	| Employee	| Type of user account.
+user.profile.status	| keyword |	On board	| Status of the user account.
+user.profile.first_name	| keyword |	First	| First Name of the User.
+user.profile.last_name	| keyword |	Last	| Last Name of the user.
+user.profile.other_identities	| keyword , multifield |	first.last@elk.elastic.co	| Array of additional user identities.
```
ebeahan, line 55: What type of `multifield` would be used here? `.text`?

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,254 @@
...
+
+<!--
+Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+-->
+
+### Proposed New Fields for User object
+
+Field | Type | Example | Description
+--- | --- | --- | ---
+user.profile.id	| keyword	| 1234	| User ID from the identity datasource.
```
ebeahan, line 50: What is the significance for introducing `user.profile.id` vs. using the existing `user.id` field?

ebeahan (commented):
Left some comments based on some observations during my first read through.

lauravoicu (approved):
LGTM!

rfcs/0000-User&HostIntegrationRFC.md
ebeahan: Can that be captured in the doc?

rfcs/0000-User&HostIntegrationRFC.md
ebeahan: > Is the idea here to create another field solely for the build number?

Let's add this under `Concerns`. Adding and populating an `os.build` field isn't breaking, but it is a notably change from the current convention.

rfcs/0000-User&HostIntegrationRFC.md
ebeahan: I'm against about crowding the existing `host.*` field set if the main use for these new fields is tied to asset and entity store use cases. The `host.*` fields are mostly populated by the [add_host_metadata](https://www.elastic.co/guide/en/beats/filebeat/current/add-host-metadata.html) processor for each event, and we wouldn't want to add the list of `installed_extensions` and `installed_applications` to every produced event from the host.

In my own experience, while VM/container instances are tied to an underlying physical host, the instances are also treated as individual compute assets: unique owners, cost center, business owners, admins.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,248 @@
...
+user.profile.type	| keyword	| Employee	| Type of user account.
+user.profile.status	| keyword |	On board	| Status of the user account.
+user.profile.first_name	| keyword |	First	| First Name of the User.
+user.profile.last_name	| keyword |	Last	| Last Name of the user.
+user.profile.other_identities	| keyword , multifield |	first.last@elk.elastic.co	| Array of additional user identities.
+user.profile.manager	| keyword |	John Doe	| Assigned Manager for the user account.
+user.profile.employee_type	| keyword |	Regular | Further classification type for the user account.
```
MSNev, line 53: Capturing this level of user profile data as part of an observability system is extremely problematic due to the PII nature of this data.

The `user.profile.*` listed here is more like the information that would be stored (and protected) by the identity system and only available on the backend and as part of the encrypted details in any token.

i.e. They should not be included with general observability data.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,248 @@
...
+user.profile.job_title	| keyword |	Field Sales	| Job title assigned to the user account.
+user.profile.department	| keyword |	x256	| Department name associated with the user account.
+user.profile.organization	| keyword |	Elasticsearch Inc.	| Organization name associated with the account.
+user.profile.location	| keyword |	US - Washington - Distributed	| Assigned location for the user account.
+user.profile.mobile_phone	| keyword |	222-222-2222	
+user.profile.primaryPhone	| keyword |	222-222-2222	
+user.profile.secondEmail	| keyword |	first.l@elastic.co	| Additional email addresses associated with the user account.
+user.profile.sup_org_id	| keyword |	SUP-ORG-75	| Primary organization ID for the user account.
+user.profile.supervisory_Org	| keyword |	Field Sales	| Primary organization name for the user account.
+user.profile.assigned_mdm_id	| keyword |	2950	| Primary host owned by the user. This acts as a correlation to identify the host event.
```
andrewkroh, line 66: So this correlates to an `asset.id` value? If so, I think the description should explicitly state this.

How do we qualify what is the "primary" device? For example, a user might be associated with a laptop, tablet, and phone.

andrewkroh, line 66: Should there be a way to associate the user to more than once device? I'm not saying there should be, because if they are not explicitly "linked" the association can still be made by querying for `asset.assigned_users: {user.profile.id}`.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,248 @@
...
+asset.end_date	                | date |	June 5, 2023 @ 18:25:57.002	| - | -             | For users, it's the termination date, for other assets, it's the out-of-service date.			
+asset.first_seen	            | date |	June 5, 2023 @ 18:25:57.003	| - | -             | The earliest date/time at which this asset was observed.  For assets added via config, this would remain empty until the asset was observed.			
+asset.last_seen	                | date |	June 5, 2023 @ 18:25:57.004	| - | -             | The most recent date/time at which this asset was observed.  Would remain empty until the asset was observed.			
+asset.last_updated	            | date |	June 5, 2023 @ 18:25:57.005	| - | -             | The most recent date/time this asset was updated in the inventory datasource.		
+asset.serial_number	            | keyword	| C02FG1G1MD6T	| - | -             |		Serial number of the asset.
+asset.tags	                    | keyword	  | watch, mdmaccess		| - | -             |	Tags assigned at the MDM.
+asset.assigned_users	          | keyword	  | user1@email.com, user2@email.com		| - | -             |	List of users assigned to the asset.
```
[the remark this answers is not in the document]
andrewkroh: I think the value of the `asset.owner` should always be included in this list. This will make is simpler to query for devices associated to a user because only one field needs to be searched. If you agree, then can you explicitly state in the description that the `asset.owner` value should always be included.

andrewkroh (commented):
My comments mostly relate to clarifying how user and asset devices are associated. This will help ensure that everyone knows how to correlate between users and assets.

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: > The `host.*` fields are mostly populated by the [add_host_metadata](https://www.elastic.co/guide/en/beats/filebeat/current/add-host-metadata.html) processor for each event, and we wouldn't want to add the list of `installed_extensions` and `installed_applications` to every produced event from the host.

The reference and elaboration are helpful in understanding the complexity. Let's move these two fields under assets.*.

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,248 @@
...
+Feel free to remove these comments as you go along.
+-->
+
+<!--
+Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+This proposal aims to extend the existing ECS fieldset to store inventory metadata for hosts, and users from external application repositories. Using ECS to store such fields will improve metadata querying and retrieval across various use cases.
+
+Within the Security `Entity Analytics` initiative, we refer to hosts and users as `entities`. We refer to hosts/ users as `assets` across generic security or observability use cases. For simplification, entities, and assets will refer to host and user objects.
```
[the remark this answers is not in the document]
SourinPaul: 👋 Andrew, I need clarification, please. Are you suggesting we adopt the 'device' terminology in this RFC scope definition/  ECS schema definition/ for the asset management integration use case (a subset of this RFC)/ all of the above? AFAIK, we have not yet adopted the device terminology anywhere in our solution yet. Thanks.

@lauravoicu curious to hear your thoughts from an operational perspective. We have a mixed bag of vendor terminologies to reconcile from.

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: Addressed this as part of this commit - https://github.com/elastic/ecs/pull/2215/commits/fd289d1eac3c22a64e80d18989d64443ca8a6814

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: I addressed this as part of this commit - https://github.com/elastic/ecs/pull/2215/commits/fd289d1eac3c22a64e80d18989d64443ca8a6814

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: I am just getting this resolved for now. Please lmk, or reopen, if we need to reconsider.

rfcs/0000-User&HostIntegrationRFC.md
MSNev: I would then recommend that a note (at the least) be added to these fields. Identifying there is possible PII issues that sending / storing this data might cause.

e.g. sending / storing this may lead to EUDB violations for example (depending on lots of caveats and how the data is store / whether it can be removed (right to forget) etc.).

 For some people / use cases this might be an acceptable option, but we don't really want to have anyone just including the data because they can.

rfcs/0000-User&HostIntegrationRFC.md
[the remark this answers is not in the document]
SourinPaul: @andrewkroh I'm adding additional clarifications to this section for now.

rfcs/0000-User&HostIntegrationRFC.md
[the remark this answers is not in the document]
SourinPaul: I have incorporated your inputs in lines 52, 93, and 108.

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: It can be the `asset.id` field value. It depends on the user's implementation of directory services and how they map their identifiers.

This ECS scope does not enforce or validate these correlation values. The data query logic will vary by the use case and is abstracted from this ECS definition.

I clarified the same in the field description- line 67.

rfcs/0000-User&HostIntegrationRFC.md
SourinPaul: 👋 @MSNev, thanks for taking the time to review. Great observations.

This subset of the user.* may be most prudent for security use cases and likely not in general observability use cases. Depending on the identity application mappings and definitions, these fields may also contain PII data.

I proposed these to users.* schema changes for three primary reasons:
1. This security use case of ingesting Identity metadata assumes appropriate index-level access within ElasticSearch and by the security solution.
2. ECS does not restrict whether a field should/should not be populated with PII. 
3. Theoretically, if you allow me to generalize, the data mapping definitions are abstracted from how the data is consumed within an application (viewed, queried, etc.)

We have previously discussed general approaches for pseudonymizing data within the ECS framework. Thus, I recommend standardizing these mapping definitions and expanding the data anonymization concepts in parallel. 

Please lmk if these thoughts resonate with you or if further clarifications are needed.

jasonrhodes (commented):
Overall, this has a LOT more fields than we are currently targeting in Observability. We're mostly concerned with:

## Overlapping Fields
* asset.type
* asset.id
* asset.name

## Conflicting Fields
* asset.category vs asset.kind (we use asset.kind as a single string keyword classification for the asset, e.g. "host")

## Missing Fields
* asset.ean - "Elastic Asset Name" string keyword providing an attempt at a reasonably unique ID across *all* asset instances (all types/kinds), uses a "URN-like" structure currently composed of "{asset.kind}:{asset.id}" (where asset.id is the local asset instance's own ID, such as a pod UID, container ID, etc.)
* asset.parents (list of EAN values that are parents of this asset)
* asset.children (list of EAN values that are children of this asset)
* asset.references (list of EAN values that are related to this asset without being parents or children of it)

rfcs/0000-User&HostIntegrationRFC.md
```diff
@@ -0,0 +1,248 @@
...
+user.account.status.deprovisioned	| boolean |	true/ false	| A flag indicating if account has been deprovisioned
+user.account.password_change_date	| date |	June 5, 2023 @ 18:25:57.000	| Last date/time when account password was updated
+
+### Proposed New Fields for Asset object
+
+Field | Type | Generic Example |	User Entity Example | Host Entity Example | Description
+--- | --- | --- | --- | --- | ---
+asset.category	| keyword |	-	        | Null	                | hardware	                | A further classification of the asset type beyond event.category. For example for host assets {hardware, virtual, container, node}. For user assets {NULL ?}			
+asset.type	    | keyword |	-	        | Null	                | workstation	            | A sub classification of asset. For host assets {workstation, S3, Compute}. For user assets {NULL?}.			
+asset.id	    | keyword |	-	        | 00uhs72c27s6PiK7x1t7	| 2950	                    | A unique ID for the asset. For inventory integrations, it's the id generated from inventory datasource
```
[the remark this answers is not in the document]
SourinPaul: We recommend the `asset ID` be unique within a user's environment. For external directory services, this can map to the directory asset id. 

It will largely depend on the source and management of the data (or how a user configures their external asset management or directory service application).

For the o11y use case you mention above, can this map to the asset.ean field and ensure uniqueness?

SourinPaul:
> ## Conflicting Fields
> * asset.category vs asset.kind (we use asset.kind as a single string keyword classification for the asset, e.g. "host")

@jasonrhodes 👋 

I proposed the asset.category mapping considering the ongoing stage [2 RFC for `event.kind](https://github.com/elastic/ecs/blob/main/rfcs/text/0038-event-kind-asset.md).

As you inferred, the asset.category and `asset.type` fields are expected to capture further granularity for infrastructure types. We can default these for asset types that do not have sub-types. 

My mapping hypothesis looked like this (we can evolve this):



event.kind | asset
-- | --
event.category | host
event.type | info
asset.category | hardware
asset.type | workstation

Do you foresee any issues reconciling and reusing these field mappings?


> Missing Fields

Would you like to extend this proposal to these o11y fields? Or would you prefer to wait and add a separate RFC later?

ebeahan (approved):
LGTM from a procedural standpoint.

Met with @SourinPaul to discuss merging so the Stage 1 PR work can get underway. However, there are open conversation threads remaining.

We don't need to conclude all discussion before merging as a Stage 0, but I do recommend we capture any raised concerns in the `Concerns` section for continued discussion in Stage 1+.

rfcs/0000-User&HostIntegrationRFC.md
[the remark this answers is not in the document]
jasonrhodes: What we are currently targeting as a model in observability is to use `asset.id` to represent the "internal" ID of that asset, so for example, for a container asset, that container's "Container ID" would be the "asset.id" value. Within all container assets, we expect that to be unique. Pod assets would list their pod ID there, as well. Hosts might use something like an FQDN hostname value there. In some cases, this value may be reasonably unique within the set of assets of that single type, but not unique across assets of other types. For that reason, we have the `asset.ean` value, which at the moment concatenates `asset.kind` + `asset.id` to help namespace the internal asset ID values. The EAN also allows us to expand the convention, URN-style, to include more information in the overall asset's identifier, while still maintaining that asset's own concept of its own ID within its asset space (e.g. container ID).

PR #2231, opened 2023-06-30
Update csv_generator.py

cicada0007:
## Changes in the `csv_generator.py` file

> 1 change

* Type hints have been added to the code to specify the types of the function parameters and return values. This helps make the code easier to understand and maintain by providing clear information about the expected types of values

> 2 change

*  The file opening mode has been updated to use `'w'` instead of `'wb'` when running Python 3 or above. This ensures compatibility with both Python 2 and Python 3.

> 3 change 

*  Removed unnecessary type casting: The `str.lower()` method is unnecessary when calling `str()` on a boolean value. It has been removed for simplicity.

### DIFF
--- a/scripts/generators/csv_generator.py
+++ b/scripts/generators/csv_generator.py
@@ -1,33 +1,10 @@
-# Licensed to Elasticsearch B.V. under one or more contributor
-# license agreements. See the NOTICE file distributed with
-# this work for additional information regarding copyright
-# ownership. Elasticsearch B.V. licenses this file to you under
-# the Apache License, Version 2.0 (the "License"); you may
-# not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-# 	http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an
-# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-# KIND, either express or implied.  See the License for the
-# specific language governing permissions and limitations
-# under the License.
-
 import _csv
 import csv
 import sys
-from typing import (
-    Dict,
-    List,
-)
-
+from typing import Dict, List
 from os.path import join
+from _types import Field
 from generator import ecs_helpers
-from _types import (
-    Field,
-)
 
 
 def generate(ecs_flat: Dict[str, Field], version: str, out_dir: str) -> None:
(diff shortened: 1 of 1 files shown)


scripts/generators/csv_generator.py
```diff
@@ -1,33 +1,10 @@
-# Licensed to Elasticsearch B.V. under one or more contributor
```
ebeahan, line 1: I suspect this header was removed by your editor, but it would need to remain with any changes.

scripts/generators/csv_generator.py
```diff
@@ -1,33 +1,10 @@
...
-# software distributed under the License is distributed on an
-# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-# KIND, either express or implied.  See the License for the
-# specific language governing permissions and limitations
-# under the License.
-
 import _csv
 import csv
 import sys
-from typing import (
```
ebeahan, line 21: Like with the header, I suspect this changes were done automatically by your editor. If not, can you explain your reasoning for the edits?

scripts/generators/csv_generator.py
```diff
@@ -39,35 +16,29 @@ def generate(ecs_flat: Dict[str, Field], version: str, out_dir: str) -> None:
...
             base_list.append(ecs_flat[field_name])
+
     return base_list + sorted_list
 
 
 def save_csv(file: str, sorted_fields: List[Field], version: str) -> None:
-    open_mode: str = "wb"
-    if sys.version_info >= (3, 0):
-        open_mode: str = "w"
+    open_mode = "w"
```
ebeahan, line 30: I'd suggest simplifying even more:

`with open(file. "w") as csvfile:`

If we do make this change in this generator, I think it'd make sense to do the same across the repo (e.g. [here](https://github.com/elastic/ecs/blob/main/scripts/generators/es_template.py#L283)).

scripts/generators/csv_generator.py
```diff
@@ -39,35 +16,29 @@ def generate(ecs_flat: Dict[str, Field], version: str, out_dir: str) -> None:
...
 
     with open(file, open_mode) as csvfile:
-        schema_writer: _csv._writer = csv.writer(csvfile,
-                                                 delimiter=',',
-                                                 quoting=csv.QUOTE_MINIMAL,
-                                                 lineterminator='\n')
-
+        schema_writer = csv.writer(csvfile, delimiter=',', quoting=csv.QUOTE_MINIMAL, lineterminator='\n')
```
ebeahan, line 33: Why remove the type hint, `_csv._writer`?

scripts/generators/csv_generator.py
```diff
@@ -39,35 +16,29 @@ def generate(ecs_flat: Dict[str, Field], version: str, out_dir: str) -> None:
...
         schema_writer.writerow(["ECS_Version", "Indexed", "Field_Set", "Field",
                                 "Type", "Level", "Normalization", "Example", "Description"])
+
         for field in sorted_fields:
-            key_parts: List[str] = field['flat_name'].split('.')
-            if len(key_parts) == 1:
-                field_set: str = 'base'
-            else:
-                field_set: str = key_parts[0]
+            key_parts = field['flat_name'].split('.')
```
ebeahan, line 38: Again, why remove the type hints?

ebeahan (changes requested):
> Type hints have been added to the code to specify the types of the function parameters and return values. This helps make the code easier to understand and maintain by providing clear information about the expected types of values

I actually see some of the type hinting _removed_ in the changes. Is that intended, and, if so, what's the reasoning? As you stated, the type hints are intended to help.

> The file opening mode has been updated to use 'w' instead of 'wb' when running Python 3 or above. This ensures compatibility with both Python 2 and Python 3.

Python 2.x support ended from the Python community in [2020](https://www.python.org/doc/sunset-python-2/). Since the ECS repo [supports](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md#dev-tools) Python 3.8+, we could simplify by removing the check for Python 2.x vs. 3.x.

> Removed unnecessary type casting: The str.lower() method is unnecessary when calling str() on a boolean value. It has been removed for simplicity.

This is the behavior I see in the REPL. I also don't actually see the `.lower()` method removed.

```python
>>> str(True)
'True'
>>> str(True).lower()
'true'
```

PR #2233, opened 2023-07-12
Stage1 2215 assetintegration rfc

SourinPaul:
Stage1 checklist:
- [ ] .yml files
- [ ] Examples of Source Documents
- [ ] Examples of Mappings
- [ ] New concerns from stage1

### DIFF
--- a/rfcs/text/0041-asset-integration.md
+++ b/rfcs/text/0041-asset-integration.md
@@ -1,7 +1,7 @@
 # 0041: Asset Integration
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (Draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
 - Date: **2023-07-07** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 <!--
@@ -11,6 +11,10 @@ Feel free to remove these comments as you go along.
 
 <!--
 Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
+-->
+
+<!--
+Stage 1: If the changes include field additions or modifications, please create a folder titled as the RFC number under rfcs/text/. This will be where proposed schema changes as standalone YAML files or extended example mappings and larger source documents will go as the RFC is iterated upon.
 -->
 
 This proposal extends the existing ECS field set to store inventory metadata for hosts and users from external application repositories. Using ECS to store such fields will improve metadata querying and retrieval across various use cases.
@@ -24,11 +28,6 @@ This proposal includes the following:
 <!-- * Additional fields in the `host` object --->
 
 This proposal will also facilitate storing host and user inventory within the security solution (the entity store).
-
-
-<!--
-Stage 1: If the changes include field additions or modifications, please create a folder titled as the RFC number under rfcs/text/. This will be where proposed schema changes as standalone YAML files or extended example mappings and larger source documents will go as the RFC is iterated upon.
--->
 
 <!--
 Stage X: Provide a brief explanation of why the proposal is being marked as abandoned. This is useful context for anyone revisiting this proposal or considering similar changes later on.
@@ -135,11 +134,15 @@ Stage 2: Add or update all remaining field definitions. The list should now be e
 Stage 1: Describe at a high-level how these field changes will be used in practice. Real world examples are encouraged. The goal here is to understand how people would leverage these fields to gain insights or solve problems. ~1-3 paragraphs.
 -->
 
-* As part of Entity Analytics, we are ingesting metadata about Users and from various external vendor applications. We are storing all ingested metadata in Elasticsearch. After we map these fields to ECS, we will enrich these ingested events for risk-scoring scenarios (e.g., context enrichments) and detecting advanced analytics (UBA) use cases.
+* As part of Entity Analytics, we are ingesting metadata about Users and from various external vendor applications. We are storing all ingested metadata in Elasticsearch. After we map these fields to ECS, we will enrich these ingested events for risk-scoring scenarios (e.g., context enrichments) and detecting advanced analytics (UEBA) use cases.
+
+### Example of Hosts and Users stored in ES
 
 * This schema will persist `Observed` (queried) entities from the ingested security log dataset in an Entity store. This entity store can be further extended to meet broader Asset Management needs.
 
 * Additional enrichment use cases for existing prebuilt detection rules will leverage these ECS fields.
+
+
 
 
 ## Source data
@@ -165,13 +168,22 @@ There are many sources of asset inventory repositories. In the mid-term, we are 
 * MS Intune
 * ServiceNow Asset CMDB
 
+
+
 <!--
 Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
 -->
 
+
 <!--
 Stage 3: Add more real world example source documents so we have at least 2 total, but ideally 3. Format as described in stage 2.
 -->
+
+### Real-world mapping of an Identity provider 
+
+#### Okta Users
+#### AzureAD Hosts
+
 
 ## Scope of impact
 
@@ -195,7 +207,6 @@ The goal here is to research and understand the impact of these changes on users
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
 -->
 
-* We have a couple of fleet integrations under development. We want them to use these proposed ECS before being released.
 * Schema/ field sets defined here focus on asset inventory data sources. Additional fields may need to be appended (ideally within this RFC lifecycle) to support the entity store needs.
 * Due diligence is needed to avoid the proliferation of field sets and validate business requirements.
 * In stage1, @jasonrhodes identified fields from o11y use cases and a potential conflict: https://github.com/elastic/ecs/pull/2215#pullrequestreview-1498781860
@@ -216,9 +227,8 @@ The following are the people that consulted on the contents of this RFC.
 * @andrewkroh | subject matter expert
 * @jamiehynds | subject matter expert
 * @lauravoicu | subject matter expert
-* @MikePaquette | subject matter expert
+* @MikePaquette | sponsor
 * @sourinpaul | sponsor
-* ?
 
 <!--
 Who will be or has been consulted on the contents of this RFC? Identify authorship and sponsorship, and optionally identify the nature of involvement of others. Link to GitHub aliases where possible. This list will likely change or grow stage after stage.
@@ -242,6 +252,7 @@ e.g.:
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
 * Stage 0: https://github.com/elastic/ecs/pull/2215
+* Stage 1: https://github.com/elastic/ecs/pull/NNN
 
 <!--
 * Stage 1: https://github.com/elastic/ecs/pull/NNN
--- a/rfcs/text/0041/asset.yml
+++ b/rfcs/text/0041/asset.yml
@@ -1,0 +1,26 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+---
+- name: asset
+  fields:
+    - name: asset.name
+      level: extended
+      type: keyword
+      short: 
+      example: 88a1f0ed-5ae5-41ee-af6b-41921c311872
+      description: >
+        Asset name of ....
--- a/rfcs/text/0041/host.yml
+++ b/rfcs/text/0041/host.yml
@@ -1,0 +1,26 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+---
+- name: host
+  fields:
+    - name: host 
+      level: extended
+      type: keyword
+      short: 
+      example: ..
+      description: >
+        Asset name of ....


rfcs/text/0041-asset-integration.md
```diff
@@ -165,14 +166,139 @@ There are many sources of asset inventory repositories. In the mid-term, we are
...
 -->
 
+
 <!--
 Stage 3: Add more real world example source documents so we have at least 2 total, but ideally 3. Format as described in stage 2.
 -->
 
+### Examples of Real-world mapping: 
+
+#### Mapping User object from Okta (partial):
```
ebeahan, line 230: Instead of the pipeline, is it possible to include the final, ECS mapped event?

It's a really helpful aid for understanding how these proposed fields will be used to map a data source.

rfcs/text/0041-asset-integration.md
```diff
@@ -242,6 +364,7 @@ e.g.:
 <!-- An RFC should link to the PRs for each of it stage advancements. -->
 
 * Stage 0: https://github.com/elastic/ecs/pull/2215
+* Stage 1: https://github.com/elastic/ecs/pull/NNN
```
ebeahan, line 367:
```suggestion
* Stage 1: https://github.com/elastic/ecs/pull/2233
```

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,147 @@
+- name: asset.category
+    level: extend
```
ebeahan, line 2:
```suggestion
```

nit: only define `level` once.

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,147 @@
...
+    type: keyword
+    example: Critical
+    description: A criticality classification obtained from outside this system, such as from some external CMDB or Directory service.
+- name: asset.business_unit
+    level: extend
+    type: keyword
+    example: Analyst Experience
+    description: Business Unit associated with the asset (user or host).
+- name: asset.costCenter
```
ebeahan, line 63:
```suggestion
- name: asset.cost_center
```

nit: use snake over camel casing.

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,147 @@
...
+    level: extend
+    type: keyword
+    example: Security - Protections
+    description: Cost Center associated with the asset (user or host).
+- name: asset.cost_center_hierarchy
+    level: extend
+    type: keyword
+    example: Engineering
+    description: Additional cost center information associated with the asset (user or host).
+- name:
```
ebeahan, line 73: Missing entry?

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,147 @@
...
+- name: asset.serial_number
+    level: extend
+    type: keyword
+    example: C02FG1G1MD6T
+    description: Serial number of the asset
+- name: asset.tags
+    level: extend
+    type: keyword
+    example: watch, mdmaccess
+    description: Tags assigned at the MDM
```
ebeahan, line 122:
```suggestion
    description: Tags assigned at the MDM
    normalize:
      - array
```

nit: use `normalize: [ array ]` since we're expecting field hold an array of value, similar to other fields like the top-level [`tags`](https://github.com/elastic/ecs/blob/main/schemas/base.yml#L45).

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    example: US - Washington - Distributed
+    description:  Assigned location for the user account.
+
+- name: user.profile.mobile_phone
+    level: extended
+    type:  keyword 
+    example: 222-222-2222
+    description: 
+
+- name: user.profile.primaryPhone
```
ebeahan, line 99:
```suggestion
- name: user.profile.primary_phone
```

nit: snake case

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    level: extended
+    type:  keyword 
+    example: 222-222-2222
+    description: 
+
+- name: user.profile.secondEmail
```
ebeahan, line 105:
```suggestion
- name: user.profile.second_email
```

nit: snake case

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    example: true/ false
+    description:  A flag indicating if account is currently locked out
+
+- name: user.account.status.suspended
+    level: extended
+    type:  boolean 
+    example: true/ false
+    description:  A flag indicating if account has been suspended
+
+- name: user.account.isAdmin
```
ebeahan, line 165:
```suggestion
- name: user.account.is_admin
```

nit: snake case

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    level: extended
+    type:  boolean 
+    example: true/ false
+    description:  A flag indicating if account is an Admin account
+
+- name: user.account.isDelegatedAdmin
```
ebeahan, line 171:
```suggestion
- name: user.account.is_delegated_admin
```

nit: snake case

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    level: extended
+    type:  boolean 
+    example: true/ false
+    description:  A flag indicating if account has Delegated Admin rights
+
+- name: user.account.isPriviledged
```
ebeahan, line 177:
```suggestion
- name: user.account.is_privileged
```

nit: snake case

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    example: first.l@elastic.co
+    description:  Additional email addresses associated with the user account.
+
+- name: user.profile.sup_org_id
+    level: extended
+    type:  keyword 
+    example: SUP-ORG-75
+    description:  Primary organization ID for the user account.
+
+- name: user.profile.supervisory_Org
```
ebeahan, line 117:
```suggestion
- name: user.profile.supervisory_org
```

nit: casing

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,199 @@
...
+    example: 222-222-2222
+    description: 
+
+- name: user.profile.secondEmail
+    level: extended
+    type:  keyword 
+    example: first.l@elastic.co
+    description:  Additional email addresses associated with the user account.
+
+- name: user.profile.sup_org_id
```
ebeahan, line 111:
```suggestion
- name: user.profile.supervisory_org_id
```

nit: slightly longer naming, but I think it makes sense to detail out `supervisory`. Also better matches the `.supervisory_org` field which uses the full word.

ebeahan (commented):
Left some initial comments on a first review. As a general comment, I suggest reviewing what can be used in the existing schema to avoid adding overlapping fields as much as possible.

For example, is `user.profile.organization` necessary when we have `organization.name` and `organization.id` fields? Reusing existing fields avoids adding more fields, and it also allows users to query across potentially other data sources that also populate the `organization.*` fields.

rfcs/text/0041/asset.yml
SourinPaul: Thanks, removed.

SourinPaul:
> For example, is `user.profile.organization` necessary when we have `organization.name` and `organization.id` fields? Reusing existing fields avoids adding more fields, and it also allows users to query across potentially other data sources that also populate the `organization.*` fields.

@ebeahan thanks and in this case, it makes sense to reuse the existing organization.* fields to reduce overhead under user.profile.*. 

Please. what is the accurate convention to capture this in our schema definition?

```
Assuming the below needs to be appended to organization.yml?

reusable:
    top_level: true
    expected:
      - at: user.profile
        as: user.profile.*
        beta: ?
        <any other attributes?>
``` 

Thanks, Eric!

ebeahan:
> Please. what is the accurate convention to capture this in our schema definition?

My thinking is to use the existing fields, `organization.name` and `organization.id`.

rfcs/text/0041-asset-integration.md
SourinPaul: @ebeahan I have updated the 'Examples of Real-world mapping' section with ECS mappings.

SourinPaul:
I have incorporated all feedback and comments so far.

Unless there are further concerns, this is ready to be merged so we can move to Stage 2. Thanks, @ebeahan

SourinPaul:
> I have incorporated all feedback and comments so far.
> 
> Unless there are further concerns, this is ready to be merged so we can move to Stage 2. Thanks, @ebeahan

A kind bump, @ebeahan. Thanks!

ebeahan:
I've weighed in with feedback, but I urge for at least one other reviewer's approval. Someone who's more familiar with the asset and entity store subject matter and use cases.

My feedback focuses on working best alongside existing schema conventions, field descriptions, ES types, etc., but I lack the context and expertise on the subject to speak if these changes are adequate and appropriate for the type of data we're trying to map.

We have several other folks added as reviewers. Anyone able to review and give some feedback for @SourinPaul's proposal here?

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,143 @@
...
```
tommyers-elastic, line 116: does this have to be as specific as this? tags could have lots of meanings depending on the data source. other examples could be board (DMI/SMBIOS) asset tags, or AWS resource tags.

jasonrhodes (changes requested):
Hey @SourinPaul - I'm back from leave and looking this over. I'd still like to have a discussion about whether security is ok with the concepts of:

- asset.ean - elastic asset name, a unique URI identifier for this instance across all assets (composition subject to further refinement and change)
- asset.parents, asset.children, asset.references: keyword lists of EAN values that map to other assets that this asset is related to, in simple parent/child/reference terms

@tommyers-elastic do we have any other requests?

SourinPaul:
@ebeahan @jasonrhodes @tommyers-elastic @andrewkroh, thanks for capturing your feedback here. I put the RFC on the back burner until now and am ready to move this forward as we approach prototyping the entity store levering this schema. 

I will incorporate some of your feedback that does not require further elaboration as soon as possible. And setting up a time with you for the ones (e.g., [asset.ean usecases](https://github.com/elastic/ecs/pull/2233#pullrequestreview-1636250432)) we need to discuss further.

rfcs/text/0041/asset.yml
SourinPaul: Hi @tommyers-elastic we can broaden the application of `asset.tags` to capture the additional use cases. That will improve the reusability of this field. 

My initial proposal was built with asset repository use cases in mind. 

Given your feedback, I'm considering updating the definitions to the below:

>name: asset.tags
    ....
    example: watch, mdmaccess, SMBIOS, AWStags, etc.
    description: Tags assigned to the asset.

If this looks good to you, no further actions are needed. Else, please provide some suggestions and examples.

SourinPaul: Schema updated.

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,143 @@
+- name: asset.category
+    level: extend
```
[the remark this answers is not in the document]
SourinPaul: Thanks, @andrewkroh. I corrected all linting errors in the yml files.

SourinPaul:
> Hey @SourinPaul - I'm back from leave and looking this over. I'd still like to have a discussion about whether security is ok with the concepts of:
> 
> * asset.ean - elastic asset name, a unique URI identifier for this instance across all assets (composition subject to further refinement and change)
> * asset.parents, asset.children, asset.references: keyword lists of EAN values that map to other assets that this asset is related to, in simple parent/child/reference terms
> 
> @tommyers-elastic do we have any other requests?

@jasonrhodes @chrisdistasio and I met to review the above inputs and agreed on the following as the next steps:

1. Exclude `asset.ean`, `asset.parents`, and `asset.children` from this RFC proposal to move the RFC forward. 
2. Use cases that require us to persist a composite entity identifier (in addition to asset.id) and entity relationship metadata, which are common to both Entity Analytics (in security) and service mapping or inventory (in o11y). We will continue to collaborate on introducing these fields in future RFCs.
3. Once internal research and development work is complete across workstreams and we have confidence in the necessary fields, we will propose a subsequent RFC to extend the ECS schema.

With the above, I'll take the following next steps:
- Capture resolution to the concerns in Stage 1
- Request a review approval from @chrisdistasio for this PR to move RFC to Stage 2.

Please lmk if I'm overlooking anything.

cc: @ebeahan

jasonrhodes:
LGTM 👍

SourinPaul:
I have updated stage 1 RFC artifacts per my prior [comment](https://github.com/elastic/ecs/pull/2233#issuecomment-1917633738).

@jasonrhodes @chrisdistasio please give this PR a formal approval so @ebeahan can help move the RFC to stage 2. Thank you!

SourinPaul:
@ebeahan, resurfacing this so we can move this to Stage 2. Thanks for your help!

rfcs/text/0041-asset-integration.md
```diff
@@ -1,7 +1,7 @@
 # 0041: Asset Integration
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
-- Stage: **0 (strawperson)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
+- Stage: **1 (Draft)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
```
ebeahan, line 4: Are we still targeting stage 1 here @SourinPaul? You mentioned stage 2 in a different conversation.

SourinPaul: @ebeahan, thanks for the ping. We are currently in stage 1 and targeting stage 2. 

Do I need to update this section? Please advise, or feel free to update before merging.

ebeahan:
Before merging, can @elastic/sec-deployment-and-devices and/or @trisch-me also review these changes?

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,174 @@
...
+  group: 2
+  short: Fields for describing an asset
+  beta: |
+    These fields are in beta and are subject to change.
+  description: >
+    Fields to describe an organization asset such as a host, user, or
+    an infrastructure. These fields can be nested under other objects that
+    identifies an asset such as host, user, network, and cloud schemas.
+  reusable:
+    top_level: false
```
ruflin, line 13: Why should asset not be used on the top level?

norrietaylor:
@trisch-me can you please take a look at this from the context or Otel, and if we should be planning a contribution to Semantic Conventions?

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,170 @@
+---
+- name: user.profile.id
```
trisch-me, line 2: we don't create `profile` as  separate namespace because it's tuned and should be used only in `user` namespace?

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,170 @@
...
+  type: keyword
+  example: Last
+  description: Last Name of the user.
+- name: user.profile.other_identities
+  level: extended
+  type: keyword
+  multi_fields:
+    - type: text
+  example: first.last@elk.elastic.co
+  description: Array of additional user identities (usually email addresses).
```
trisch-me, line 33: shouldn't it have 
```
normalize:
        - array
```

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,170 @@
...
+- name: user.profile.manager
+  level: extended
+  type: keyword
+  example: John Doe
+  description: Assigned Manager for the user account.
+- name: user.profile.employee_type
+  level: extended
+  type: keyword
+  example: Regular
+  description: Further classification type for the user account.
```
trisch-me, line 43: how it correlates to just `user.profile.type`
Should we add more details into description to resolve this question for ECS users?

rfcs/text/0041/asset.yml
trisch-me: Probably because they are tied to main object and don't have meaning themselves. But good question

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,170 @@
...
+  level: extended
+  type: keyword
+  example: Elasticsearch Inc.
+  description: Organization name associated with the account.
+- name: user.profile.location
+  level: extended
+  type: keyword
+  example: US - Washington - Distributed
+  description: Assigned location for the user account.
+- name: user.profile.mobile_phone
```
trisch-me, line 79: I thought it was written these 2 fields will be removed?

rfcs/text/0041/user.yml
```diff
@@ -0,0 +1,170 @@
...
+- name: user.account.create_date
+  level: extended
+  type: date
+  example: June 5, 2023 @ 18:25:57.000
+  description: Date account was created.
+- name: user.account.activated_date
+  level: extended
+  type: date
+  example: June 5, 2023 @ 18:25:57.000
+  description: Date account was activated.
```
trisch-me, line 120: The format of the date should be in description

trisch-me:
From otel POV we don't have assets as namespace there and currently is not a good place to start to add a new namespace.
But for the user we can already start to follow our upcoming guidelines: first check Otel and then adopt in ECS because there is an [open PR](https://github.com/open-telemetry/semantic-conventions/pull/731) for the user. it is still in discussions and it might take some time, so it shouldn't prevent us from continue work in ECS but as soon as it's merged we can add `user.profile` there if we think this should be in the next ECS releases and we need it for our product.

For opening PR there we would need a use-case story - why we should have those fields

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,174 @@
...
+    expected:
+      - host
+      - user
+      - network
+      - cloud
+  type: group
+  fields:
+    - name: category
```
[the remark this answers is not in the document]
tinnytintin10: FYI @dimadavid @r4zr32d3k1l, as I belive decisions here will have an impact on our ultimate UX for the asset inventory experience we want to build out

tinnytintin10 (commented):
@tehilashn @eyalkraft @oren-zohar @kfirpeled @romulets, I have taken a first look at some of the suggestions here that I believe will impact some of the experiences we want to build out. See my thoughts and feedback below. I will soon take a second pass-through of the remaining asset and user fields. In the meantime, please review and provide your thoughts.

rfcs/text/0041/asset.yml
jaredburgettelastic: @tinnytintin10 agreed. Because of the open source nature of ECS, we should not limit ourselves to the vision of solution (Security & Observability) use cases only, as any Elastic user may desire to map data into the asset schema that we haven't considered.

rfcs/text/0041-asset-integration.md
```diff
@@ -165,14 +175,540 @@ There are many sources of asset inventory repositories. In the mid-term, we are
...
+  - set:
+      field: user.organization.name
+      copy_from: entityanalytics_okta.user.profile.organization
+      tag: set_user_profile_organization
+      ignore_empty_value: true
+
+```
+
+ 
+#### AzureAD Hosts
```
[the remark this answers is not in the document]
jaredburgettelastic: I would prefer to drop this. In fact, I'd prefer we simply link to the source for these "real world examples", because they could change at any time. In fact...the Okta one has changed since this RFC was written!

In the case of Okta, the link to that source code would be here: https://github.com/elastic/integrations/blob/main/packages/entityanalytics_okta/data_stream/user/elasticsearch/ingest_pipeline/default.yml

rfcs/text/0041/asset.yml
```diff
@@ -0,0 +1,174 @@
...
+      example: hardware
+      description: A further classification of the asset type beyond event.category.
+        Example for host assets {hardware, virtual, container, node}."
+    - name: type
+      level: extended
+      type: keyword
+      example: workstation
+      description: "A sub-classification of assets. Possible values for host assets:
+        workstation, S3,Compute. Possible values for host assets: (NULL/ TBD)"
+    - name: id
```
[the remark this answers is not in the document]
jaredburgettelastic: Experience and external standards suggest that just a singular "ID" field, without context, is not helpful when IDs are exposed beyond some internal bound. With that said, I'd go even further: I don't feel we should only limit ourselves to "native"/"internal" or surrogate/non-surrogate IDs. 

Some concrete examples to make this case:

- Microsoft uses [SIDs](https://learn.microsoft.com/en-us/windows/win32/secauthz/sid-components) for Windows Server, which is made up of not only an identifier, but also what authority assigned that identifier (the authority being the "context" mentioned above).
- AWS EC2 instances have an ARN, but in most AWS APIs you don't use the ARN to identify that EC2 instance. Instead, you use its "Instance ID", which would look something like "i-01e8de571c1ea7903". But [you can compute the ARN ](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-policy-structure.html#EC2_ARN_Format) if you have more information on where that instance is hosted (provided the region and account id). And both of these are useful depending on the context, as you'd need the ARN if you were writing an AWS policy, but the instance ID if you were trying to stop the instance. So ID context is helpful (the identifier type, in this case "instance ID" vs "ARN" being the "context" mentioned above.

Although prior art in ECS doesn't match this (such as the [ECS host.id](https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-id) field), my recommendation would be for `asset.id` to represent something like an array of identifiers, or subfields, having context data, such as (at least) a type. An example document might look like:

```json
{
  "asset": {
    "id": {
      "arn": "arn:aws:sns:us-east-1:123456789012:example-sns-topic-name",
    }
  }
}
```

or even

```json
{
  "asset": {
    "ids": [
      {
        "id_type": "arn",
        "id_value": "arn:aws:sns:us-east-1:123456789012:example-sns-topic-name"
      }
    ] 
  }
}
```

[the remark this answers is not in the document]
romulets: From one side, I like how rich it becomes. From the other side, I wonder how easy it becomes to correlate data.

For example, how easy would be to have a ES|QL query joining documents

tinnytintin10:
@jaredburgettelastic @oatkiller, While working on concrete examples for my ask in [this epic](https://github.com/orgs/elastic/projects/866/views/41?pane=issue&itemId=58588373), I realized there might be an issue with using the term "asset" to model the wide variety of entities present in our logs.

The core concern is that "asset" implies ownership or inherent value to the user, which isn't always the case for some resources in the logs. For example, a network flow log or cloud trail log might reference an external IP address belonging to a third party or even a malicious actor. Modeling these with the "asset" field set could be misleading and cause confusion. In short, while asset equals entity, not all entities are assets.

Instead, I propose that we consistently use the term "entity" across our entire security solution, from the data model being proposed here to the in-product experiences (e.g., ~~asset inventory~~ entity inventory). The term "entity" is more versatile and can accurately represent both user/organization-owned resources and external resources that might appear in logs but aren't necessarily owned by the organization/user.

This change would provide several key benefits:

- **Clarity**: By using "entity" throughout, we avoid the ownership ambiguity of "asset" and present a clearer, more accurate representation of the data being modeled. The inventory experience we build will provide the curated experiences users need to manage the resources that belong to them.
- **Consistency**: As we're committed to the name "entity analytics" for our risk analytics capability, pairing that with an "asset inventory" could be confusing.
- **Alignment**: We've been grappling with the "asset" vs. "entity" question, often starting sentences with "asset/entity." I think a unified term will foster better cross-team collaboration and data correlation.

My ask for you:
- Does my underlying point make sense? (i.e., it could be misleading to model all resources using the term "asset")
- Is swapping the field set name to `entity` feasible?
- If swapping the name isn't feasible/possible at this stage, how would you propose we tackle the issue I highlighted?

jaredburgettelastic:
> @jaredburgettelastic @oatkiller, While working on concrete examples for my ask in [this epic](https://github.com/orgs/elastic/projects/866/views/41?pane=issue&itemId=58588373), I realized there might be an issue with using the term "asset" to model the wide variety of entities present in our logs.
> 
> The core concern is that "asset" implies ownership or inherent value to the user, which isn't always the case for some resources in the logs. For example, a network flow log or cloud trail log might reference an external IP address belonging to a third party or even a malicious actor. Modeling these with the "asset" field set could be misleading and cause confusion. In short, while asset equals entity, not all entities are assets.
> 
> Instead, I propose that we consistently use the term "entity" across our entire security solution, from the data model being proposed here to the in-product experiences (e.g., ~asset inventory~ entity inventory). The term "entity" is more versatile and can accurately represent both user/organization-owned resources and external resources that might appear in logs but aren't necessarily owned by the organization/user.
> 
> This change would provide several key benefits:
> 
> * **Clarity**: By using "entity" throughout, we avoid the ownership ambiguity of "asset" and present a clearer, more accurate representation of the data being modeled. The inventory experience we build will provide the curated experiences users need to manage the resources that belong to them.
> * **Consistency**: As we're committed to the name "entity analytics" for our risk analytics capability, pairing that with an "asset inventory" could be confusing.
> * **Alignment**: We've been grappling with the "asset" vs. "entity" question, often starting sentences with "asset/entity." I think a unified term will foster better cross-team collaboration and data correlation.
> 
> My ask for you:
> 
> * Does my underlying point make sense? (i.e., it could be misleading to model all resources using the term "asset")
> * Is swapping the field set name to `entity` feasible?
> * If swapping the name isn't feasible/possible at this stage, how would you propose we tackle the issue I highlighted?

@tinnytintin10 thank you for the input! 

@tommyers-elastic on the Observability Solution mentioned that they are also shifting away from the term "assets" and toward "entities", so this aligns well from that perspective. However, I'm unsure of:
- whether there are any Observability features or integrations already implemented that align with this RFC in its current state
- whether the future plans for Observability are shifting completely away from fields defined as `asset`. [This schema defined in their most recent implementation](https://github.com/simianhacker/kibana/blob/4a53db85b8b97fcaeb025b42178fc372018fc17f/x-pack/packages/kbn-entities-schema/src/schema/entity.ts) suggests that they are indeed moving away from that term. However, note that that schema does not adhere to this RFC more generally. 

@tommyers-elastic could you provide input from the perspective of Observability?

I know of only one feature in the Security Solution today that adheres to the asset portion of this RFC, which is ["Asset Criticality"](https://www.elastic.co/guide/en/security/current/asset-criticality.html). There are two pieces to this: one is an index that stores asset criticality information, and the other is an enrichment on alert documents that adds the `host/user.asset.criticality` to each alert if one is assigned for that host/user.

The asset criticality feature is currently behind an advanced setting, so there could be wiggle room there in our ability to change the data structures. However, from a nomenclature perspective, "Asset Criticality" explicitly applies to resources that are owned by the customer/Elastic Security user, because they must explicitly assign those classifications. I don't know if our product partners have a desire to change that nomenclature in the Security Solution platform. @MikePaquette could you please provide input from a product perspective on this portion?

cc @oatkiller

oren-zohar:
@tinnytintin10, I completely agree with your underlying point that using the term "asset" to model the wide variety of entities in our logs can be misleading, as it implies ownership or inherent value to the user, which isn't always the case.

However, I suggest keeping "Asset Inventory" as is. The asset inventory integration focuses on managed entities, aligning with your definition of "assets." Additionally, the collected assets could still be indexed into an "entity" ECS, which makes sense in my opinion. This distinction helps maintain clarity and consistency for user/organization-owned resources.

What do you think about this?

lauravoicu:
@tinnytintin10 you are probably aware of this, but to add to @oren-zohar's point from an InfoSec perspective: in InfoSec we are using the term asset, and this comes from the ISO definition of an asset: "An asset is an item, thing or entity that has potential or actual value to an organisation" (although other definitions of an asset would imply ownership, this one does not address ownership explicitly). I acknowledge terminology is extremely important, and there are of course many different inputs to this, hope this helps!

chrisdistasio:
@jaredburgettelastic o11y team has chosen to use "entity" (instead of asset).  i'm not aware of any o11y feature(s) that use `asset`in the context of this rfc.

cc: @tommyers-elastic

rfcs/text/0041/asset.yml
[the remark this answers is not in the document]
jaredburgettelastic: cc @chrisdistasio

defutek-tj:
@MikePaquette @lauravoicu @chrisdistasio - any progress on reviewing this?

PR #2232, opened 2023-07-12
[RFC] Risk field extensions [Stage 0]

rylnd:
👋

PR #2236, opened 2023-07-18
[RFC] Risk Score Extensions - Stage 1

rylnd:
Followup to #2232 .

### DIFF
--- a/rfcs/text/0042-risk-score-extensions.md
+++ b/rfcs/text/0042-risk-score-extensions.md
@@ -45,8 +45,8 @@ Some of the context here was discussed in Stage 0; please read the above for tha
 * `category_2_score`
 * `category_2_count`
 * `category_3_score`
-* `category_3_score`
-* `category_4_count`
+* `category_3_count`
+* `category_4_score`
 * `category_4_count`
 * `category_5_score`
 * `category_5_count`
--- a/rfcs/text/0042-risk-score-extensions.md
+++ b/rfcs/text/0042-risk-score-extensions.md
@@ -54,13 +54,13 @@ Some of the context here was discussed in Stage 0; please read the above for tha
 ### Risk Explainability
 Beyond the per-category explanations, these fields' purpose is to provide more insight/data for the analyst to further investigate the components of the risk score.
 
-* `riskiest_inputs`
+* `risk_inputs`
   * Generally, these objects are meant as a convenience for one investigating risk; they are the "most risky" inputs as determined by the risk engine, and serve as a shortcut to further investigation.
-* `riskiest_inputs.id`, `riskiest_inputs.index`
+* `risk_inputs.id`, `risk_inputs.index`
   * These fields allow one to uniquely identify the original risk input document.
-* `riskiest_inputs.risk_category`, `riskiest_inputs.risk_score`
+* `risk_inputs.risk_category`, `risk_inputs.risk_score`
   * More "convenience" fields that could also be discovered in the original risk input document.
-* `riskiest_inputs.risk_description`
+* `risk_inputs.risk_description`
   * These field is intended to be a precursor to what we are referring to as "Risk Reasons," which seek to aggregate/present multiple risk inputs in an easier to consume/analyze format.
 * `notes`
   * Miscellaneous text field intended to provide more details that cannot be presented in the other fields.
@@ -105,8 +105,8 @@ There are two broad concerns at this stage:
 
 1. Category fields introducing a new "ordered" pattern
   * Rather than having either an array of objects, or an explicit `nested` field type, both of which allow an arbitrary number of items, we're instead opting to add 10 explicit fields (five explicit categories, each with two fields) under the _assumption_ that we won't extend the number of categories further. We have a bit of wiggle room (i.e. six categories, 12 fields wouldn't be out of question), but this is not a scalable solution if we need a large number of categories. However, that is only a potential future issue, and we can likely reevaluate and address it if/when it arises.
-2. Mapping of `riskiest_inputs` as a simple `object`
-  * The biggest motivation for this choice is to avoid the performance/storage/syntax complexities that come with a `nested` mapping, but we also don't have any feature requirements that would currently necessitate `riskiest_inputs` being `nested`.
+2. Mapping of `risk_inputs` as a simple `object`
+  * The biggest motivation for this choice is to avoid the performance/storage/syntax complexities that come with a `nested` mapping, but we also don't have any feature requirements that would currently necessitate `risk_inputs` being `nested`.
 
 <!--
 Stage 1: Identify potential concerns, implementation challenges, or complexity. Spend some time on this. Play devil's advocate. Try to identify the sort of non-obvious challenges that tend to surface later. The goal here is to surface risks early, allow everyone the time to work through them, and ultimately document resolution for posterity's sake.
--- a/rfcs/text/0042-risk-score-extensions.md
+++ b/rfcs/text/0042-risk-score-extensions.md
@@ -125,7 +125,7 @@ Stage 3: Document resolutions for all existing concerns. Any new concerns should
 The following are the people that consulted on the contents of this RFC.
 
 * @rylnd | author
-* @SourinPaul | SME / EA product manager
+* @SourinPaul | sponsor
 * @ebeahan | reviewer
 
 
--- a/rfcs/text/0042-risk-score-extensions.md
+++ b/rfcs/text/0042-risk-score-extensions.md
@@ -9,9 +9,6 @@ As you work on your RFC, use the "Stage N" comments to guide you in what you sho
 Feel free to remove these comments as you go along.
 -->
 
-<!--
-Stage 0: Provide a high level summary of the premise of these changes. Briefly describe the nature, purpose, and impact of the changes. ~2-5 sentences.
--->
 This RFC seeks to extend the [existing risk fields](https://www.elastic.co/guide/en/ecs/current/ecs-risk.html) [(RFC 0031)](https://github.com/elastic/ecs/pull/2048) to support new/extended Risk Score investigation workflows. The workflows that this RFC intends to enable include all those described in 0031, along with the following:
 
 1. Risk Score Explainability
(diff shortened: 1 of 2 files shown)


rfcs/text/0042-risk-score-extensions.md
```diff
@@ -24,16 +21,50 @@ This RFC seeks to extend the [existing risk fields](https://www.elastic.co/guide
...
+
+### Identifier Fields
+These fields are intended to allow future extensibility of our concept of an "identifier." Currently, we leverage two such fields: `host.name` and `user.name`, to group documents for the purposes of risk scoring, but we definitely imagine this being extended in the future to allow multiple such fields for hosts and users, or more simply to allow administrators to configure fields other than those above. The presence of these two fields allows us to audit/explain how a particular risk score was identified.
+
+* `identifier_field`
+* `identifier_value`
+
+
+### Risk Category Fields
+Some of the context here was discussed in Stage 0; please read the above for that. More specifically, these fields seek to provide the category contributions to the score, and the number of risk inputs in that category, across each of the five proposed categories:
```
rylnd, line 41: One other note (which I can add to the RFC, but let's discuss here): The main motivation for having separate fields here per category (as opposed to an `object` like in `riskiest_inputs`) is the ability to search/filter scores on these attributes.

rfcs/text/0042-risk-score-extensions.md
```diff
@@ -24,16 +21,50 @@ This RFC seeks to extend the [existing risk fields](https://www.elastic.co/guide
...
+* `category_3_score`
+* `category_4_count`
+* `category_4_count`
+* `category_5_score`
+* `category_5_count`
+
+### Risk Explainability
+Beyond the per-category explanations, these fields' purpose is to provide more insight/data for the analyst to further investigate the components of the risk score.
+
+* `riskiest_inputs`
```
rylnd, line 57: I think I had originally called this `top_risk_inputs` but decided this was more specific (e.g. what does "top" mean?). However, if we want to change that definition in the future, maybe `top_risk_inputs` or something similar would allow us that flexibility.

rfcs/text/0042-risk-score-extensions.md
```diff
@@ -24,16 +21,50 @@ This RFC seeks to extend the [existing risk fields](https://www.elastic.co/guide
...
+These fields are intended to allow future extensibility of our concept of an "identifier." Currently, we leverage two such fields: `host.name` and `user.name`, to group documents for the purposes of risk scoring, but we definitely imagine this being extended in the future to allow multiple such fields for hosts and users, or more simply to allow administrators to configure fields other than those above. The presence of these two fields allows us to audit/explain how a particular risk score was identified.
+
+* `identifier_field`
+* `identifier_value`
+
+
+### Risk Category Fields
+Some of the context here was discussed in Stage 0; please read the above for that. More specifically, these fields seek to provide the category contributions to the score, and the number of risk inputs in that category, across each of the five proposed categories:
+
+* `category_1_score`
```
rylnd, line 43: @SourinPaul one thought about naming: we could also do alphabetical, instead of numeric (e.g. "Category A", "Category B") 🤷

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,144 @@
...
+      example: host-001.local
+      description: >
+        Identifies which field value was used to group/calculate this risk score.
+        Typically, hosts use values from `host.name` and users use values from `user.name`.
+    - name: category_1_score
+      level: extended
+      type: float
+      example: 33.0
+      description: >
+        The contribution of Category 1 to the overall risk score (`calculated_score`). Category 1 contains Detection Engine Alerts.
```
rylnd, line 38: @SourinPaul please review these descriptions, I feel like they could use some improvement 😄 .

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,144 @@
...
+    These fields are in beta and are subject to change.
+  description: >
+    Fields for describing risk score and risk level of entities such as hosts and users.
+    These fields are not allowed to be nested under `event.*`. Please continue to use
+    `event.risk_score` and `event.risk_score_norm` for event risk.
+  reusable:
+    top_level: false
+    expected:
+      - host
+      - user
```
rylnd, line 16: @SourinPaul We're not currently respecting this fact (that `risk` is present under _either_ `host.risk` or `user.risk`) in the persistence implementation, but I'm about to. Before I do I just wanted to verify this behavior: nesting this under one or the other means that we cannot query/aggregate across users _and_ hosts in a single query, they'll pretty much always be two different sets of data that we have to act on independently; it's effectively normalized. As long as we don't need to return both sets of data in a single spot, I think that's fine 👍

SourinPaul: @rylnd Yes, that's fine. We dont have an explicit search requirement for "Display any entity (host/ user) with risk.level = High".

We have explicit search requirements for each entity type "Display all hosts. all users" in the entity pages. Even if we hit an edge case where we may need to retrieve risk for correlated entities, they can be explicit API calls.

rfcs/text/0042-risk-score-extensions.md
SourinPaul: > maybe `top_risk_inputs` or something similar would allow us that flexibility.

Do you mean 'risk_inputs' will allow us the flexibility? @rylnd

rylnd: No, but `risk_inputs` is certainly more flexible than what I was suggesting!

rylnd: @SourinPaul is that your preference? I have no objections and can make that change.

rfcs/text/0042-risk-score-extensions.md
rylnd: Point being: if there's a better way to do this other than `type: nested` I'm all ears 😄

rfcs/text/0042-risk-score-extensions.md
SourinPaul: I wish to understand your original comment, which I do now. `riskiest_inputs` serves the purpose, although it still begs the question of relative risk. 
 
 The application using this field (in our case, Entity Analytics) can capture the specifics and not the ECS definition.

rylnd: @SourinPaul I'm personally leaning toward your suggestion of the generic `risk_inputs` for this field name, but I was unable to determine your stance from the previous comment. Could you please clarify?

SourinPaul: @rylnd I recommend using `risk_inputs` as the ECS field name.

rfcs/text/0042/risk.yml
SourinPaul: @rylnd reads great. 

A few suggestions. You can ignore any you do not like.

- identifier_field: An ECS field that identifies an entity. Typically, hosts use the field `host.name`, and users use the field `user.name`.
- identifier_value: The entity identifier used to calculate risk. Typically, hosts use values from `host.name` and users use values from `user.name`.
- category_x_score: The risk contribution from a risk category (e.g., Category 'x') to the overall entity risk score. 
Risk Categories logically group risk inputs from various domain use cases - e.g. Alert, Posture Management, Vulnerabilities, and others. 
- category_x_count: The number of risk input documents that contributed to each category score, in this case, category '1'. 
Risk Categories logically group risk inputs from various domain use cases - e.g. Alert, Posture Management, Vulnerabilities, and others.

rylnd: @SourinPaul I added your feedback in [a6df8ea](https://github.com/elastic/ecs/pull/2236/commits/a6df8ea853bb53279066e54a465d18019b99d5d7). However, because we can't describe the fields generally and instead need to define them individually, I left the previous copy about what the particular category includes (for the `_score`s, at least). If you don't want to be that specific and would rather remove that aspect, please let me know.

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,168 @@
...
+        The number of risk input documents that contributed to the Category 5 score.
+
+        Risk Categories logically group risk inputs from various domain use cases - e.g. Alert, Posture Management, Vulnerabilities, and others.
+    - name: risk_inputs
+      level: extended
+      type: object
+      normalize: array
+      description: >
+        An array of the top risk input documents that contributed to the overall risk score.
+    - name: risk_inputs.id
```
ebeahan, line 128: IMO feels redundant to have fields named `risk.risk_inputs.*` (with `risk_` prepended).

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,168 @@
...
+      level: extended
+      type: keyword
+      description: >
+        The identifier (_id) of the risk input document.
+    - name: risk_inputs.index
+      level: extended
+      type: keyword
+      description: >
+        The index (`_index`) of the risk input document.
+    - name: risk_inputs.risk_category
```
ebeahan, line 138: Same question again here, is `risk.risk_inputs.risk_category` naming necessary vs. `risk.inputs.category`?

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,168 @@
...
+      level: extended
+      type: keyword
+      example: category_1
+      description: >
+        The risk category of the risk input document.
+    - name: risk_inputs.risk_description
+      level: extended
+      type: keyword
```
ebeahan, line 146: If you think having full-text indexing on the description may be useful, ECS does support defining multi-fields ([example](https://github.com/elastic/ecs/blob/main/schemas/file.yml#L85))

rfcs/text/0042/risk.yml
```diff
@@ -0,0 +1,168 @@
...
+    These fields are not allowed to be nested under `event.*`. Please continue to use
+    `event.risk_score` and `event.risk_score_norm` for event risk.
+  reusable:
+    top_level: false
+    expected:
+      - host
+      - user
+  type: group
+  fields:
+    - name: identifier_field
```
ebeahan, line 19: ECS steers away from abbreviations, but the schema already uses [`id`](https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-id) if you wanted to shorten to `id_field` and `id_value`.

rfcs/text/0042/risk.yml
SourinPaul: I agree that individual definitions are needed and your correct definitions are adequate. Thanks @rylnd!

rfcs/text/0042/risk.yml
rylnd: Yep, good call 👍

rfcs/text/0042/risk.yml
rylnd: Not necessary, just more explicit/independent of the context of the parent. I'll make those changes, though.

rfcs/text/0042/risk.yml
rylnd: This is an interesting one; while `id` refers to the elasticsearch identifier `_id`, `identifier_*` here have an [entirely different meaning](https://github.com/elastic/ecs/pull/2236/files#diff-40e16fa32db2ffbdeabf56eb2ffeac75dfa12410e662c3328c7c859b426805d2R34).

That's not to say that the unabbreviated name helps at all with that distinction 😅 . @SourinPaul unless you disagree I'm going to change these to `id_field` and `id_value`; I'm into the whole brevity thing.

rfcs/text/0042/risk.yml
rylnd: I moved forward on the assumption that they are equivalent, and then did that nesting. Let me know if that's inappropriate (either for the declaration, or for ECS tooling, etc.)

rylnd:
@ebeahan I've updated the RFC with your feedback, thank you!

ebeahan (approved):
LGTM, but I overlooked before there isn't a [sponsor](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md#responsibilities-in-this-process) listed. Would that be @SourinPaul?

rylnd:
> LGTM, but I overlooked before there isn't a [sponsor](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md#responsibilities-in-this-process) listed. Would that be @SourinPaul?

I believe that's correct; I'll update the PR and consult him. I'm also going to try to solicit another round of feedback from EA folks; if there's additional changes @ebeahan I'll re-request your review. Thanks!

SourinPaul:
> I believe that's correct;

Yes, please. Thanks @rylnd.

rfcs/text/0042-risk-score-extensions.md
```diff
@@ -24,16 +21,50 @@ This RFC seeks to extend the [existing risk fields](https://www.elastic.co/guide
...
+These fields are intended to allow future extensibility of our concept of an "identifier." Currently, we leverage two such fields: `host.name` and `user.name`, to group documents for the purposes of risk scoring, but we definitely imagine this being extended in the future to allow multiple such fields for hosts and users, or more simply to allow administrators to configure fields other than those above. The presence of these two fields allows us to audit/explain how a particular risk score was identified.
+
+* `id_field`
+* `id_value`
+
+
+### Risk Category Fields
+Some of the context here was discussed in Stage 0; please read the above for that. More specifically, these fields seek to provide the category contributions to the score, and the number of risk inputs in that category, across each of the five proposed categories:
+
+* `category_1_score`
```
ajosh0504, line 43: How do users know what Category 1, 2...etc. mean? Would it make sense for these field names to reflect what the category means, for example `alert_category_score`, `posture_category_count` etc?

ajosh0504 (approved):
Great additions! 🚀

rfcs/text/0042-risk-score-extensions.md
susan-shu-c: I was thinking that too, but it seems that similar to `id_field` and whatnot, ideally the content of each numbered category could be changed without changing the fields? e.g. `category_1` could be from alerts one day, and changed to something else in the future 🤔

rylnd: Great question! The more specific `alert_category_score` etc. were actually the original proposition, but we decided to go the more generic route for future-proofing. 

@susan-shu-c got the gist of it, although more accurately we want to allow _extending_ each category's definition as opposed to outright changing it (for backwards compatibility, etc.). For example, `category_1` currently includes alerts, but might be extended in the future to allow more alert-like sources. 

This all was described in a little more detail in [stage 0](https://github.com/elastic/ecs/pull/2232/files#diff-40e16fa32db2ffbdeabf56eb2ffeac75dfa12410e662c3328c7c859b426805d2R20-R25) if you're interested 👍

PR #2219, opened 2023-06-20
Added `container.privileged` field

Tacklebox:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)?
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)?
- If submitting code/script changes, have you verified all tests pass locally using `make test`?
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed.
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)?

PR #2246, opened 2023-08-01
container.security context.privileged boolean

Tacklebox:
My previous PR had the wrong type name.

bool -> boolean

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -15,7 +15,7 @@ Thanks, you're awesome :-) -->
 #### Bugfixes
 
 #### Added
-* Added `container.security_context.privileged` to indicated whether a container was started in privileged mode. #2219, #2225
+* Added `container.security_context.privileged` to indicated whether a container was started in privileged mode. #2219, #2225, #2246
 
 #### Improvements
 
--- a/docs/fields/field-details.asciidoc
+++ b/docs/fields/field-details.asciidoc
@@ -1239,7 +1239,7 @@ example: `docker`
 
 a| Indicates whether the container is running in privileged mode.
 
-type: bool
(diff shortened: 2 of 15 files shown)


kgeller (commented):
@Tacklebox thanks for fixing this!

Would you just add this PR to the changelog ie

`* Added `container.privileged` to indicated whether a container was started in privileged mode. #2219, #2246`

ruflin (approved):
LGTM

@ebeahan Do you know if we have any CI checks that could prevent this from happening as this is not a valid type in the first place?

eyalkoren:
> @ebeahan Do you know if we have any CI checks that could prevent this from happening as this is not a valid type in the first place?

@ruflin Elasticsearch's CI will now [tell us](https://github.com/elastic/elasticsearch/pull/97901/files#diff-5de719b6a576263ffb32a9a7a1754078f91b12abf0d4fe7141faeb7a91e6b66dR289) if such slips in 😉

ebeahan (approved):
LGTM. Can we also reference this fix in the [CHANGELOG.next entry](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md?plain=1#L18)?

ebeahan:
> Do you know if we have any CI checks that could prevent this from happening as this is not a valid type in the first place?

AFAIK, surprisingly not. Filed an issue: https://github.com/elastic/ecs/issues/2249

kgeller (approved):

eyalkoren:
Anything blocking this from being merged?

kgeller:
> Anything blocking this from being merged?

@eyalkoren I assumed it was waiting on an outstanding review from @norrietaylor , but will merge if you all are ready

PR #2267, opened 2023-09-08
Clean up duplicated headers in `CHANGELOG.next`

ebeahan:
Fix dupe'd changelog headers I introduced in #2264


ebeahan:
Will fold fix in #2265 because `gitpython` apparently isn't having a good time 😹

PR #2260, opened 2023-08-27
[RFC] Stage 2 for volume device

Trinity2019:
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)? Yes
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)? Yes
- If submitting code/script changes, have you verified all tests pass locally using `make test`? N/A
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? N/A
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. Yes
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? N/A

PR #2269, opened 2023-09-11
Stage 2 `beta` changes for RFC 0040, `volume.*`

ebeahan:
Implement the changes proposed in [RFC 0040](https://github.com/elastic/ecs/blob/main/rfcs/text/0040/volume.yml) which has advanced to stage 2 (#2260).

These changes add a new `volume.*` field set to describe storage volume devices. The field set will be marked as `beta` since the RFC is currently at RFC stage 2.


ebeahan:
@Trinity2019 will you take a look? This PR is required to land the `volume.*` changes into the schema and publish into the ECS docs.

Trinity2019 (approved):
LGTM. Thanks @ebeahan for moving the proposal into beta stage!

kgeller (approved):

PR #2278, opened 2023-09-20
Respect `reusable.top_level` in Beats generator

### LINKED ISSUE #2277: Incorrect generated/beats/fields.ecs.yml, not accounting on top_level: false
**Description of the issue:**

We are using the following file as a source of ECS mappings for osquery_manager integration:
https://github.com/elastic/ecs/blob/main/generated/beats/fields.ecs.yml

It does have `as` field at the the top level there, which causes our integration validation to fail for 3.0.0 format spec.

As far as I understand generated file should not include `as` field at the top level because it is marked with top_level: false attribute.
https://github.com/elastic/ecs/blob/969aeba5018d92331eda7e4104c4c53225c32fd7/generated/ecs/ecs_nested.yml#L157

Currently it has

Screenshot 2023-09-20 at 10 21 06 AM

and not additional information for us of the field is a top level or not
Screenshot 2023-09-20 at 10 22 00 AM

ebeahan:
Remove field sets with `reusable.top_level: false` from erroneously appearing as valid, top-level fields in `generated/beats/fields.ecs.yml`.

The same check is now performed on both the ES template and Beats fields generators.

Fixes #2277


kgeller (approved):
LGTM

PR #2281, opened 2023-09-27
Remove `expected_values` from *.indicator.name field defs

ebeahan:
The `threat.indicator.name` and `threat.enrichments.indicator.name` are misusing the `expected_value` parameter in their field definitions. This change removes `expected_value`.

I propose the removal is a bug fix vs. a breaking change since the current `expected_values` on these fields is not useful for real-world applications.

### DIFF
--- a/schemas/threat.yml
+++ b/schemas/threat.yml
@@ -111,18 +111,9 @@
       short: Indicator display name
       description: >
         The display name indicator in an UI friendly format
-      expected_values:
-        - 5.2.75.227
-        - 2a02:cf40:add:4002:91f2:a9b2:e09a:6fc6
-        - https://example.com/some/path
-        - example.com
-        - 373d34874d7bc89fd4cefa6272ee80bf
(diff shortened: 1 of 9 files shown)


schemas/threat.yml
```diff
@@ -111,18 +111,9 @@
       short: Indicator display name
       description: >
         The display name indicator in an UI friendly format
-      expected_values:
```
andrewkroh, line 114: This looks like it should have been an example. Does the schema allow this to be converted into an `example` (not sure if a list of individual examples can be conveyed)?

ebeahan: No, the schema does not. I thought about migrating the values to `example` originally but didn't want to infer the fields expected an array of values.

In an earlier revision, I moved the examples into `description`. But after moving, I liked eliminating the specific values in favor of generic terms like URL, port, IP, etc.

andrewkroh (approved):

chrisberkhout (approved):

PR #2285, opened 2023-10-03
8.11 FF changelog updates (#2284)

taylor-swanson:


taylor-swanson:
I accidentally did this out of order, (went into main first, now back porting it instead of forward porting it), but should be fine as I cherry-picked the commit.

kgeller (approved):

PR #2300, opened 2023-11-07
Upgrading pyyaml dependency to fix build failures

kgeller:
This PR does two things:
1) Resolves build failures due to https://github.com/yaml/pyyaml/issues/601 by upgrading the dependency from pyyaml 6.0.0 to 6.0.1
2) Switching a few usages of f-string to string.format to address build formatting failures
<details>
  <summary>Formatting build errors</summary>
  
  ```
      if isinstance(example_value, (list, dict)):
         field_name: str = field['field_details']['name']
-        msg: str = f"Example value for field `{field_name}` contains an object or array which must be quoted to avoid YAML interpretation."
+        msg: str = f"Example value for field `{
+            field_name}` contains an object or array which must be quoted to avoid YAML interpretation."
         strict_warning_handler(msg, strict)
 
     # Examples with arrays must be handled
@@ -295,13 +296,15 @@ def check_example_value(field: Union[List, FieldEntry], strict: Optional[bool] =
         for example_value in example_values:
             match = re.match(pattern, example_value)
             if not match:
-                msg = f"Example value for field `{name}` does not match the regex defined in the pattern attribute: `{pattern}`."
+                msg = f"Example value for field `{
+                    name}` does not match the regex defined in the pattern attribute: `{pattern}`."
                 strict_warning_handler(msg, strict)
 
     if expected_values:
         for example_value in example_values:
             if example_value not in expected_values:
-                msg = f"Example value `{example_value}` for field `{name}` is not one of the values defined in `expected_value`: {expected_values}."
+                msg = f"Example value `{example_value}` for field `{
+                    name}` is not one of the values defined in `expected_value`: {expected_values}."
                 strict_warning_handler(msg, strict)
  ```
</details>

(I don't understand why there was an issue with those 3 specific f-strings, we use them in other places and the build is still happy with them)

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -1,6 +1,6 @@
 pip
 # License: MIT
-PyYAML==6.0
+PyYAML==6.0.1
 # License: BSD
 gitpython==3.1.37
 # License: BSD
--- a/scripts/schema/cleaner.py
+++ b/scripts/schema/cleaner.py
@@ -281,7 +281,8 @@ def check_example_value(field: Union[List, FieldEntry], strict: Optional[bool] =
 
     if isinstance(example_value, (list, dict)):
         field_name: str = field['field_details']['name']
-        msg: str = "Example value for field `{}` contains an object or array which must be quoted to avoid YAML interpretation.".format(field_name)
+        msg: str = "Example value for field `{}` contains an object or array which must be quoted to avoid YAML interpretation.".format(
+            field_name)
         strict_warning_handler(msg, strict)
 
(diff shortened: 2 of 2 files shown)


taylor-swanson (approved):
LGTM

Also not sure why it doesn't like the f-strings. Unless there's something it doesn't like with the backticks 🤷

kgeller:
> Also not sure why it doesn't like the f-strings. Unless there's something it doesn't like with the backticks 🤷

I tried removing them as well, and it still had the same error :(

PR #2328, opened 2024-04-03
Fix broken link for vulnerabilty.id (correct commits)

nicpenning:
The current link for vulnerability.id is: https://cve.mitre.org/about/faqs.html#what_is_cve_id) which is a broken link.
image

This PR makes the link useable as: https://cve.mitre.org/about/faqs.html#what_is_cve_id
image

### DIFF
--- a/CHANGELOG.next.md
+++ b/CHANGELOG.next.md
@@ -13,6 +13,7 @@ Thanks, you're awesome :-) -->
 #### Breaking changes
 
 #### Bugfixes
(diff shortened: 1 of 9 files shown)


mjwolf (approved):
Thanks for finding and fixing this

mjwolf:
I updated the issue title so no one mistakes this for a vulnerability. It's not fixing a vulnerability, it's an update to the vulnerability field

PR #2354, opened 2024-08-13
Update templated fields in RFC0044

mjwolf:
Update some templated fields that were missed before merging the RFC


gogochan (approved):
👍 the PR does what it says on the title.

PR #2377, opened 2024-09-05
add license header like in all other schema yml files 

stbischof:


stbischof:
cla signed

mjwolf (approved):
Thanks for fixing this. 

This does fix all the missing licenses that could be found by running `make check_license_headers`

mjwolf:
I'm also adding a check for license headers, so these shouldn't be missed again https://github.com/elastic/ecs/pull/2378

PR #2379, opened 2024-09-08
trim trailing whitespace in schema

stbischof:


stbischof:
There are more then this. But wanted to be sure that this is also your expected behavier

mjwolf (approved):

PR #2415, opened 2024-12-12
Added documentation and explicit mapping between ECS and OpenTelemetry

AlexanderWert:
## Goal

Have a single source of truth of the relation between ECs fields and OpenTelemetry attributes and metrics

## What this PR adds:

- extended schema files to include explicit mapping from ECS attributes to OTel attributes and metrics
- extended generation tooling to check mapping against OTel semantic conventions repository ensuring the mapping is up to date with semconv changes
- extended generation tooling to generate additional docs:
   - for each ECS field that corresponds to an OTel attribute generate reference to the attribute and document relation type
   - Docs page explaining the relation of ECS and OTel with additional overview pages for the mapping
   
   
image

image

image


scripts/generators/otel.py
```diff
@@ -0,0 +1,241 @@
...
+        raise KeyError(f"Target directory './{target_dir}' not present in git '{git_repo}'!")
+
+
+def get_attributes(
+    model_files: List[OTelModelFile]
+) -> Dict[str, OTelAttribute]:
+    """Retrieves (non-deprecated) OTel attributes from the model files"""
+
+    attributes: Dict[str, OTelAttribute] = {}
+    for mode_file in model_files:
```
trisch-me, line 45: probably
```suggestion
    for model_file in model_files:
```

docs/opentelemetry/otel-fields-mapping.asciidoc
```diff
@@ -0,0 +1,2405 @@
+The following table gives an overview of mappings between individual ECS fields (in ECS version `9.0.0-dev`)
```
trisch-me, line 1: do you think it makes sense to have all the fields from ecs here to make table more comprehensive that one can see how many fields are yet to add to semconv?

docs/opentelemetry/index.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+For this type of relation aliasing approaches (e.g. https://www.elastic.co/guide/en/elasticsearch/reference/current/field-alias.html[Elasticsearch field aliases]) can be used to achieve compatibility between ECS and OTel SemConv.
+
+| image:https://img.shields.io/badge/related-f2f4fb?style=flat[relation,title=related]
+| The ECS field name is different and has related - yet different - semantics as the corresponding SemConv attribute.
+An aliasing approach is not sufficient to resolve compatibility for this type of relation. 
+
+| image:https://img.shields.io/badge/metric-800080?style=flat[relation,title=metric]
+| For this ECS field there is a corresponding metric defined in OTel SemConv.
+
+| image:https://img.shields.io/badge/conflict-910000?style=flat[relation,title=conflict]
```
trisch-me, line 38: should we also have a notation for fields that are never going to be donated to otel? such as `ecs.version`, probably `related.*` fields and maybe some others?

docs/opentelemetry/otel-mapping-summary.asciidoc
```diff
@@ -0,0 +1,969 @@
+
+The following table summarizes the alignment status by namespaces between ECS in version `9.0.0-dev`
+and OpenTelemetry semantic conventions in version `1.29.0`.
```
trisch-me, line 3: do you want to track otel namespaces here to backport them later to ECS? For ECS users new namespaces might be unnecessary

schemas/user.yml
```diff
@@ -65,6 +65,18 @@
       example: S-1-5-21-202424912787-2692429404-2351956786-1000
       description: >
         Unique identifier of the user.
+      otel:
+        - relation: match
+      otel_reuse:
```
trisch-me, line 70: is this an equivalent of embed feature?

docs/opentelemetry/otel-fields-mapping.asciidoc
AlexanderWert: That's why I added that other summary page with the table showing numbers per namespace

docs/opentelemetry/index.asciidoc
AlexanderWert: Good point! I think that would make sense!

docs/opentelemetry/otel-mapping-summary.asciidoc
AlexanderWert: This list shows already a merge of namespaces from both, ECS and SemConv.

For example there's the `Android` namespace, which only exists in OTel but not in ECS that appears in that list

schemas/user.yml
AlexanderWert: yes, for some of the fields that are defined in ECS as reusable (like `process.user.name`) we defined explicit fields in SemConv. Since those do not exist explicitly in the ECS schema we need this additional concept of `otel_reuse` to document the mapping of those explicit attributes with implicit ECS fields

scripts/_types/otel_types.py
```diff
@@ -0,0 +1,39 @@
...
+    stability: str
+    ref: str
+    deprecated: str
+    group_display_name: str
+
+
+class OTelGroup(TypedDict, total=False):
+    id: str
+    type: str
+    prefix: str
```
trisch-me, line 18: prefix is going to be removed in the next release of otel

docs/opentelemetry/otel-fields-mapping.asciidoc
trisch-me: yeah, I think I wanted to have it in one place but we could say that every particular namespace will have it and I can see exact fields there. So I’m ok with current view as well

scripts/generators/otel.py
```diff
@@ -0,0 +1,241 @@
...
+    return repo.head.commit.tree
+
+
+def get_otel_attribute_name(
+    field: Field,
+    otel: OTelMapping
+) -> str:
+    if otel['relation'] == 'match':
+        return field['flat_name']
+    elif 'attribute' in otel:
```
trisch-me, line 115: what if `attribute` is not present

docs/opentelemetry/otel-mapping-summary.asciidoc
trisch-me: yes, that what was my question about :) if you want to include namespaces from both standards and not only from ecs for better tracking

scripts/_types/otel_types.py
AlexanderWert: yes, I know, but it's still being used in some places, so added it as a fallback

docs/opentelemetry/index.asciidoc
AlexanderWert: added, a relation type `na`

scripts/templates/otel_summary.j2
```diff
@@ -0,0 +1,30 @@
...
+and OpenTelemetry semantic conventions in version `{{ semconv_version }}`.
+
+[cols="8*"]
+|====
+h| 
+7+^h| Number of Fields / Attributes
+
+h| Namespace
+h| ECS
+h| OTel
```
felixbarny, line 12:
```suggestion
h| https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html[ECS]
h| https://opentelemetry.io/docs/specs/semconv/attributes-registry[OTel]
```

docs/opentelemetry/otel-mapping-summary.asciidoc
```diff
@@ -0,0 +1,969 @@
+
```
felixbarny, line 1: Maybe add a note to this and the previous file that they're auto-generated.

scripts/templates/otel_summary.j2
felixbarny: Alternatively, maybe add an ECS/OTel icon to the `summary['title']` column and link to the corresponding namespace.

felixbarny (approved):
Fantastic overview!

docs/opentelemetry/otel-mapping-summary.asciidoc
```diff
@@ -0,0 +1,1164 @@
...
+^| ·
+^| ·
+^| ·
+^| ·
+^| ·
+
+
+| User agent
+^| <<ecs-user_agent,4>>
+^| https://opentelemetry.io/docs/specs/semconv/attributes-registry/user_agent[4]
```
felixbarny, line 1071: This is a 404. It should be https://opentelemetry.io/docs/specs/semconv/attributes-registry/user-agent

scripts/generator.py
```diff
@@ -41,6 +42,9 @@
 def main() -> None:
     args = argument_parser()
 
+    if not args.semconv_version:
+        raise KeyError("OTel Semantic Conventions version has not been provided as a config option '--semconv-version'")
```
trisch-me, line 46: should the default be the last one? we can detect it through git and skip manually selecting it
I think for this purpose we need always to adhere to the last available version

AlexanderWert: I'd rather keep that a manual step, or use something like dependabot to update it but not updating it automatically, that has some risks.

trisch-me: we are using there only schema, what risks do you have in mind?
we are adding a lot of new fields into each release but most probably we would need to update mapping anyway and can update the version, so this comment is not blocking.

trisch-me (approved):

docs/opentelemetry/otel-fields-mapping.asciidoc
```diff
@@ -0,0 +1,2568 @@
...
+// ===============================================================
+.2+|
+[[otel-mapping-for-labels]]
+<<field-labels, labels>> 
+
+
+
+^| image:https://img.shields.io/badge/related-efc20d?style=flat[relation,title=related,link=ecs-opentelemetry.html#ecs-opentelemetry-relation]
+
+<| https://opentelemetry.io/docs/specs/semconv/attributes-registry/k8s/#k8s-pod-label[k8s.pod.label] +
```
mjwolf, line 59: Is there any way to link to a specific semconv version in the otel docs? I couldn't find a way. But if there is, it would be nice to link to the same `SEMCONV_VERSION` this is generated from.

mjwolf (commented):
This looks great!

Issue #2416, opened 2024-12-18
Fix all URL encoding in generated docs


mjwolf:
**Description of the issue:**

The URL in `file.mime_type`, and possibly other fields with URLs don't render correctly when generated into markdown (for example, in the integrations documentation).

ECS is using asciidoc format, which doesn't work with markdown. All URLs should be changed to remove formatting schemes, so they work in all generated languages.

It should be fixed, and search for any other cases where its incorrect. There was another previously reported similar issue: https://github.com/elastic/ecs/pull/2328

**Any additional context or examples:**

![Image](https://github.com/user-attachments/assets/3d23c02b-cec8-4264-a4a5-94fde8d6159e)

taylor-swanson:
Here are the other fields that have issues:

- `elf.go_import_hash`
- `macho.go_import_hash`
- `pe.go_import_hash`
- `vulnerability.category`
- `vulnerability.description`
- `vulnerability.id`

I talked with the docs team about it and they are good with using the URL directly in descriptions (the `threat` fields are a good example of this). For reference, the ECS docs site uses asciidoc, the integrations docs site uses the README markdown and parses it to asciidoc (which means the markdown must be valid), and the integrations readme/kibana uses markdown directly.

I did a grep for `http` in the schemas and while we're at it, we should also update http links to https. Namely references to `http://publicsuffix.org`.

taylor-swanson:
Fixed by: #2423

PR #2423, opened 2025-01-15
Fix link rendering issues and usage of http in links

taylor-swanson:
## Description

- Removed link text markup, causes issues since final rendering of docs can be in either markdown or asciidoc
- Replace http with https in links

## Related Issues

- #2416


mjwolf (approved):
LGTM

I checked that there are no other markdown or asciidoc formatted links in the schema files too

PR #2429, opened 2025-01-21
[8.x] Bump version to 8.18.0

mjwolf:
Bump version on 8.x branch to 8.18.0


qcorporation:
@mjwolf, just a question
We are releasing an 8.17 release, but we are bumping the version bump to 8.18-dev - will that cause any confusion?

mjwolf:
> @mjwolf, just a question We are releasing an 8.17 release, but we are bumping the version bump to 8.18-dev - will that cause any confusion?

The `8.17` branch has been created now, which will be used for the release, and has the 8.17 version so it won't cause problems. This PR is to bump the version in preparation for the next pre-release builds and next release

PR #2434, opened 2025-01-29
[RFC] Stage 0: Introduce Entity Field Set into ECS

tinnytintin10:
## Overview 

An entity represents a discrete, identifiable component within an IT environment that can be described by a set of attributes and maintains its identity over time. Entities can be physical (like hosts or devices), logical (like containers or processes), or abstract (like applications or services).

Currently, ECS provides specific field sets for certain categories of entities (e.g., host, user, cloud, orchestrator) to capture their metadata. However, as IT infrastructure continues to evolve, we encounter an increasing number of entity types that don't cleanly fit into existing field sets – for example, storage services like S3, database instances like DynamoDB, or various other cloud services and IT-related infrastructure components (both digital and physical). 

This RFC proposes a new entity fieldset that aims to solve this and several other challenges. Currently at Stage 0 (strawperson), seeking initial feedback on the approach and concept. See `/rfcs/text/0049-entity-fields.md` for more details. 

* Sponsor: @MikePaquette & @YulNaumenko 
* Author: @tinnytintin10  

## PR Guidelines
 
- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? ✅ 
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)? ✅ 
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process] (https://github.com/elastic/ecs/blob/main/rfcs/README.md)? ✅ 
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes? N/A
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed. ✅ 
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)? N/A


tinnytintin10:
@MikePaquette @YulNaumenko, I've drafted the RFC to introduce the entity field set into ECS like we talked about. Before taking it out of draft, I wanted to check with you both to see if there's anything you think should be included or addressed as part of this stage. Lmk 🙏🏾

rfcs/text/0049-entity-fields.md
```diff
@@ -0,0 +1,158 @@
...
+| entity.address | keyword | A URI, URL, or other direct reference to access or locate the entity in its source system. This could be an API endpoint, web console URL, or other addressable location. Format may vary by entity type and source system. |
+| entity.Attributes.* | object |  Entity type-specific attributes using capitalized field names to indicate custom field space. The capital `A` in "Attributes" and the capitalization of all subfields (e.g., `entity.Attributes.StorageClass`, `entity.Attributes.EngineVersion`) distinguishes these as custom entity-type-specific fields that won't be enumerated in the ECS schema.  | 
+| entity.metadata.* | flattened | A flexible container for entity metadata that doesn't fit into other structured fields. This field uses the flattened type to allow arbitrary key-value pairs while maintaining searchability. Useful for provider-specific or non-standardized attributes that don't warrant dedicated fields. |
```
oren-zohar, line 49: @tinnytintin10 would we still need it?

rfcs/text/0049-entity-fields.md
```diff
@@ -0,0 +1,158 @@
...
+3. Enabling the representation of entities that don't fit into existing field sets
+4. Reducing the complexity of the ECS schema by avoiding the proliferation of entity-specific field sets
+
+This approach would allow ECS to accommodate new types of entities without requiring continuous schema expansion through new field sets, while maintaining a consistent structure for entity representation.
+
+## Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| entity.id | keyword | A unique identifier for the entity. This should be a stable, unique value that persists across different observations of the same entity. For entities with dedicated field sets (e.g., host.id, user.id), this value should match the corresponding *.id field. |
```
maxcold, line 42: do we want to solve the problem of different id/name sources for entities? eg. EC2 instance id vs arn?

rfcs/text/0049-entity-fields.md
```diff
@@ -0,0 +1,158 @@
...
+| entity.source | keyword | The module or integration that provided this entity data (similar to event.module). |
+| entity.category | keyword | A standardized high-level classification of the entity type. This provides a normalized way to group similar entities across different providers or systems. Example values: `bucket`, `database`, `container`, `function`, `queue`, `host`, `user`, etc.,. There will be an allowed set of values maintained for this field to ensure consistency. |
```
maxcold, line 44: I'm guessing category and type are chosen to be consistent with `event.*` , but should we be consistent here? I personally always found `type` and `category` confusing, especially when it comes to understanding what is higher in the hierarchy. Wouldn't it be simpler to have `category` and `sub-category` or smth like that?

JordanSh: I've had some offline discussions with tin on this subject, my suggestion is `type` as the higher classification and `subtype` (or anything else) for secondary. This will better align with Observability Inventory which uses `entity.type` to search, group and filter:

image

image
 
And will also align better with our codebase which already mentions and uses `entity.type` quite a bit, having two different fields will require us to "juggle" between them depending on the part of the codebase we are working in.

image

image

rfcs/text/0049-entity-fields.md
tinnytintin10: I've updated the entity.id field description to provide guidance for choosing between multiple identifiers. To be transparent - there will always be some ambiguity in this selection that we can't fully resolve yet as it depends on specific use cases and contexts. However, we can provide these basic criteria for selecting the primary identifier:

1. Persists across the entity's lifecycle
2. Ensures uniqueness within its scope
3. Is commonly used for queries and correlation
4. Is readily available in most observations (events/logs, etc.,)

Alternative identifiers are preserved in entity.raw. Wdyt?

rfcs/text/0049-entity-fields.md
tinnytintin10: I like the suggestion of using more explicitly hierarchical terms. Something like type/sub_type would make the relationship self-evident without needing to know any conventions.

However, I deliberately aligned with ECS's categorization hierarchy (where event.category is higher level than event.type) when coming up with this. I'd love to get input from the ECS team on whether maintaining this consistency is warranted here. While consistency with existing patterns has value, entities serve a different purpose and might benefit from more intuitive naming. 

cc @MikePaquette

tinnytintin10:
Reviewed this with @MikePaquette and are good to go for boarder reviews 🚀 

cc @tehilashn @oren-zohar @YulNaumenko

rfcs/text/0049-entity-fields.md
romulets: The issue with ids is something beyond the scope of this RFC, in my opinion. More important than knowing what type of id is being persisted, is having a consistent id of that entity when ingested from different sources. I don't think adding a `id type` field would ensure consistency, it only gives a context in which I'm not sure it's needed.

We need consistent ids and entity resolution when needed. Neither of these solutions touch the RFC right now, as far as I understand.

rfcs/text/0049-entity-fields.md
```diff
@@ -0,0 +1,158 @@
...
+
+This field set aims to solve several key challenges:
+1. Providing a flexible way to represent different types of entities without requiring new field sets for each category
+2. Supporting a consistent structure for capturing entity metadata across different entity types
+3. Enabling the representation of entities that don't fit into existing field sets
+4. Reducing the complexity of the ECS schema by avoiding the proliferation of entity-specific field sets
+
+This approach would allow ECS to accommodate new types of entities without requiring continuous schema expansion through new field sets, while maintaining a consistent structure for entity representation.
+
+## Fields
```
JordanSh, line 38: shouldn't `tags` and `labels` also be included in `entity`?

mjwolf (commented):
This looks good to me, my comments could also be addessed in future RFC stages. I'll wait for subject matter experts to review and some of the other comments to be addressed before giving it the official approval though

rfcs/text/0049-entity-fields.md
tinnytintin10: @JordanSh @maxcold I've pushed updates to change from category/type to type/sub_type based on our discussion. @romulets before we make this change in the inventory integration, etc., let's see if there is any pushback on this from the broader ECS team. 

@opauloh - we'll absolutely add an `allowed_values` property for `entity.type` similar to what's used for `event.category`. We'll implement this as we move beyond Stage 0 so we can properly define the complete set of entity types with appropriate descriptions.

rfcs/text/0049-entity-fields.md
tinnytintin10: Since labels and tags are [base ECS fields](https://www.elastic.co/guide/en/ecs/current/ecs-base.html), they will be contained at the root of the entity document and not under/nested within the entity object.

romulets (approved):

JordanSh (approved):
🚀

maxcold (approved):

opauloh (approved):

rfcs/text/0049-entity-fields.md
```diff
@@ -0,0 +1,158 @@
...
+
+| Field | Type | Description |
+|-------|------|-------------|
+| entity.id | keyword | A unique identifier for the entity. When multiple identifiers exist, this should be the most stable and commonly used identifier that: 1) persists across the entity's lifecycle, 2) ensures uniqueness within its scope, 3) is commonly used for queries and correlation, and 4) is readily available in most observations (logs/events). For entities with dedicated field sets (e.g., host, user), this value should match the corresponding *.id field. Alternative identifiers (e.g., ARNs values in AWS, URLs) can be preserved in entity.raw. |
```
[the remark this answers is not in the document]
kubasobon: I really like this breakdown, since it answers a lot of questions that keep popping up. Especially stuff like GCP, where there isn't a one defined ARN-like format.

kubasobon (approved):
Left a comment on `entity.id`. Looks good to me!

rfcs/text/0049-entity-fields.md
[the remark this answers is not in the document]
tinnytintin10: Sounds good. I'll incorpate this in subsquent stages of this RFC 👍🏾

mjwolf (approved):

PR #2441, opened 2025-02-03
[RFC] Stage 2: Introducing new fields in file/dll/process fields

AsuNa-jp:
Yes
Yes
Yes
Yes
Yes
Yes

### DIFF
--- a/rfcs/text/0048-fileorigin-fields.md
+++ b/rfcs/text/0048-fileorigin-fields.md
@@ -2,7 +2,7 @@
 <!-- Leave this ID at 0000. The ECS team will assign a unique, contiguous RFC number upon merging the initial stage of this RFC. -->
 
 - Stage: **2 (Candidate)** <!-- Update to reflect target stage. See https://elastic.github.io/ecs/stages.html -->
-- Date: **2024-XX-XX** <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
+- Date: **2025-02-07* <!-- The ECS team sets this date at merge time. This is the date of the latest stage advancement. -->
 
 <!--
 As you work on your RFC, use the "Stage N" comments to guide you in what you should focus on, for the stage you're targeting.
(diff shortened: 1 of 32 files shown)


schemas/dll.yml
```diff
@@ -45,3 +45,17 @@
       type: keyword
       description: Full file path of the library.
       example: C:\Windows\System32\kernel32.dll
+
+    - name: origin_referrer_url
+      level: extended
+      type: keyword
+      ignore_above: 8192
+      description: The URL of the webpage that linked to the dll file.
+      example: http://example.com/article1.html
```
mjwolf, line 54:
```suggestion
      example: http://example.com/article1.html
      beta: This field is beta and subject to change.
```

Please add `beta: This field is beta and subject to change.` here and to all the other new fields that are being added.

schemas/process.yml
```diff
@@ -497,3 +497,17 @@
       type: long
       description: >
         The length of bytes skipped.
+
+    - name: origin_referrer_url
```
mjwolf, line 501: For process, I think it might be better to use `executable.origin_referrer_url` and `executable.origin_url`, since this is a property of the executable file that was used to create the process, not of the process itself.

This is something that's being discussed in otel for a different change: https://github.com/open-telemetry/semantic-conventions/pull/1737. ECS doesn't currently follow this convention to refer to the source executable file all the time (for example with `process.name`, but I think the otel convention is better, and something we should try to do here as well.

rfcs/text/0048-fileorigin-fields.md
```diff
@@ -82,6 +82,44 @@ Example sources of data is shown in the above.
 Stage 2: Included a real world example source document. Ideally this example comes from the source(s) identified in stage 1. If not, it should replace them. The goal here is to validate the utility of these field changes in the context of a real world example. Format with the source name as a ### header and the example document in a GitHub code block with json formatting, or if on the larger side, add them to the corresponding RFC folder.
 -->
 
+The following is the real world example and usage for these fields.
```
mjwolf, line 85: If you have an example of how this can be used on macos, I think it would be good to include here too

mjwolf:
@AsuNa-jp Can you add a changelog entry in `CHANGELOG.next.md`?

schemas/dll.yml
AsuNa-jp: Thank you for the feedback! I've fixed as suggested.

AsuNa-jp:
Hi @mjwolf 
Thank you very much for the feedback. I've added changelog entry to `CHANGELOG.next.md` and added `beta: This field is beta and subject to change.` to all of the newly added fields.

I also added an example of how file origin information is added when downloading a file from a web browser on macOS. Let me know if you need additional examples.

For `process.executable` see my reply at https://github.com/elastic/ecs/pull/2441#discussion_r1942195161

schemas/process.yml
trisch-me: @mjwolf I agree with @AsuNa-jp regarding restriction we currently have with executable being existing leaf [node](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-executable). We have discussed that probably we would need to make adjustments in ECS to accommodate Otel schema. But I think it shouldn’t be a problem for this PR unless you have other ideas for naming

mjwolf: I didn't think of executable already being a keyword, and the problems that would introduce when I made this suggestion. I agree it's best to leave this as it is for now

mjwolf (approved):

PR #2461, opened 2025-03-25
[RFC] Advance Entity Field Set to Stage 1

tinnytintin10:
This PR advances the Entity Field Set RFC (0049) from Stage 0 (strawperson) to Stage 1 (draft).

## Changes Since Stage 0

Since the initial Stage 0 proposal (PR #2434), the following additions have been made:

- Added a "Usage" section highlighting how the entity field set enables normalized entity data querying and its role in the upcoming security solution inventory experience

- Added "Source data" section explaining how the field set's taxonomy allows entity modeling from any data source

- Added "Concerns" section addressing potential challenges (To Do)

- Added subject matter experts to the "People" section

- Created YAML schema definition in the `rfcs/text/0049/` directory

## Next Steps

After advancing to Stage 1, we plan to:

1. Implement experimental field definitions in the ECS schema
2. Gather feedback from early adopters
3. Refine the field definitions based on practical usage
4. Begin work toward Stage 2 criteria


tinnytintin10:
@mjwolf, to what level of detail are we supposed to document usage and source data sections in a stage 1 RFC?  Does the current level of detail I provide suffice? Also, for the concerns section, are we supposed to update that during the PR review process or upfront (I guess a mix of both but wanted to clarify)? Thanks!

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+      A unique identifier for the entity. When multiple identifiers exist, this should be
+      the most stable and commonly used identifier that: 1) persists across the entity's
+      lifecycle, 2) ensures uniqueness within its scope, 3) is commonly used for queries
+      and correlation, and 4) is readily available in most observations (logs/events).
+      For entities with dedicated field sets (e.g., host, user), this value should match
+      the corresponding *.id field. Alternative identifiers (e.g., ARNs values in AWS, URLs)
+      can be preserved in entity.raw.
+    example: i-04ff5d36be3d6896c, arn:aws:s3:::my-bucket, projects/123456789/locations/us-central1/instances/my-db
+
+  - name: entity.source
```
maxcold, line 31: what about `provider`? I couldn't find any example of `*.source` field

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+      categorization.
+    example: aws_s3_bucket, gcp_cloud_storage_bucket, azure_blob_container, aws_lambda_function
+
+  - name: entity.name
+    level: core
+    type: keyword
+    multi_fields:
+      - name: text
+        type: text
+    short: The human-readable name of the entity.
```
maxcold, line 68: it's not very human readeable in the examples. Do we need additional `entity.title` to capture the human readable title of the entity? as in EC2 instance `maxcold-ec2` with title `EC2 instance for testing`

rfcs/text/0049-entity-fields.md
```diff
@@ -68,91 +45,38 @@ This approach ensures backward compatibility, maintains existing ECS patterns, a
...
 
 ### RFC Pull Requests
 
-<!-- An RFC should link to the PRs for each of it stage advancements. -->
-
-<!--
-* Stage 1: https://github.com/elastic/ecs/pull/NNN
-...
--->
 * Stage 0: https://github.com/elastic/ecs/pull/2434
```
mjwolf, line 81: You can leave this in, for historical reference

rfcs/text/0049-entity-fields.md
```diff
@@ -68,91 +45,38 @@ This approach ensures backward compatibility, maintains existing ECS patterns, a
...
-* @Yasmina | author
-* @Monique | sponsor
-* @EunJung | subject matter expert
-* @JaneDoe | grammar, spelling, prose
-* @Mariana
--->
-
 The following are the people that consulted on the contents of this RFC.
 
-* Author: @tinnytintin10
```
mjwolf, line 139: You can leave this info from stage 0 in, along with the below sections. The RFCs stages are intended to build on each other, so you can keep relevant data from past stages.

rfcs/text/0049-entity-fields.md
```diff
@@ -68,91 +45,38 @@ This approach ensures backward compatibility, maintains existing ECS patterns, a
...
-
-Out Of Scope for Stage 0 (based on template)
+This approach will enable security analysts to view all the entities discovered inside of their environment, whether from logs or other data sources. The entity field set will then begin powering all parts of our security solution experience like alerts, where we can now represent more entities beyond just users and hosts.
 
-<!--
-Stage 1: Provide a high-level description of example sources of data. This does not yet need to be a concrete example of a source document, but instead can simply describe a potential source (e.g. nginx access log). This will ultimately be fleshed out to include literal source examples in a future stage. The goal here is to identify practical sources for these fields in the real world. ~1-3 sentences or unordered list.
--->
+Essentially, this field set gives us a standard way to represent any entity's metadata, regardless of its type or source, and provides customers with the same ability to standardize that information across their environments.
 
-<!--
```
mjwolf, line 85: It's not necessarily required, but if you leave in the comments for future stages, it'll be easier for you to update this doc when you get to those stages.

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+      For entities with dedicated field sets (e.g., host, user), this value should match
+      the corresponding *.id field. Alternative identifiers (e.g., ARNs values in AWS, URLs)
+      can be preserved in entity.raw.
+    example: i-04ff5d36be3d6896c, arn:aws:s3:::my-bucket, projects/123456789/locations/us-central1/instances/my-db
+
+  - name: entity.source
+    level: core
+    type: keyword
+    short: Source module or integration that provided the entity data.
+    description: >
```
mjwolf, line 35: I think the document source can usually be determined from `event.dataset` or the index its in. Do you know of any cases where the existing data isn't enough to determine the source?

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
+---
+# Entity field set schema definition
+name: entity
+title: Entity
+group: 2
+type: group
+short: Fields to describe various types of entities across IT environments.
```
romulets, line 7: Nitpick: What's IT environments? We might have term to describe de space in which entities can be found.

IT Environments has the common usage in inside the company environments, meanwhile this field should capture entities from services and external references too

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+    type: keyword
+    short: Unique identifier for the entity.
+    description: >
+      A unique identifier for the entity. When multiple identifiers exist, this should be
+      the most stable and commonly used identifier that: 1) persists across the entity's
+      lifecycle, 2) ensures uniqueness within its scope, 3) is commonly used for queries
+      and correlation, and 4) is readily available in most observations (logs/events).
+      For entities with dedicated field sets (e.g., host, user), this value should match
+      the corresponding *.id field. Alternative identifiers (e.g., ARNs values in AWS, URLs)
+      can be preserved in entity.raw.
```
romulets, line 28: > this value should match the corresponding *.id field. Alternative identifiers (e.g., ARNs values in AWS, URLs) can be preserved in entity.raw

Specially the ARN values in AWS seems to contradict what was previously suggested as guidance to entity ids. Am I misreading something?

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+  - name: entity.sub_type
+    level: extended
+    type: keyword
+    short: The specific type designation for the entity as defined by its provider or system.
+    description: >
+      The specific type designation for the entity as defined by its provider or system. This field
+      provides more granular classification than entity.type. While entity.type provides a normalized
+      classification across different systems, entity.sub_type preserves the provider-specific
+      categorization.
+    example: aws_s3_bucket, gcp_cloud_storage_bucket, azure_blob_container, aws_lambda_function
```
orouz, line 60: maybe we should favor the original asset type string as it's defined by the vendor and only create one if there isn't any? 

this list would be:
| name                     | vendor name                                                 |
| ------------------------ | ----------------------------------------------------------- |
| aws_s3_bucket            | `AWS::S3::Bucket`                                           |
| gcp_cloud_storage_bucket | `storage.googleapis.com/Bucket`                             |
| azure_blob_container     | `Microsoft.Storage/storageAccounts/blobServices/containers` |
| aws_lambda_function      | `AWS::Lambda::Function`                                     |


- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html
- https://learn.microsoft.com/en-us/azure/governance/resource-graph/reference/supported-tables-resources
- https://cloud.google.com/asset-inventory/docs/asset-types

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: text
+        type: text
+    short: The human-readable name of the entity.
+    description: >
+      The human-readable name of the entity. The keyword field enables exact matches for filtering
+      and aggregations, while the text field enables full-text search. For entities with dedicated
+      field sets (e.g., host), this field should mirror the corresponding *.name value.
+    example: my-production-database, web-server-01, payment-processing-queue
+
+  - name: entity.url
```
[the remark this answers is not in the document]
tinnytintin10: Good point. `URL` is too specific, and `reference` is used in other fieldsets for the same or similar purposes. I'll make this update.

rfcs/text/0049/entity.yml
tinnytintin10: I recommend keeping "source" rather than switching to "provider" for the entity.source field. Since "provider" already has established usage in ECS (like cloud.provider) with a specific meaning related to service providers, using "source" avoids confusion and better captures that this field identifies the data origin (CloudTrail, CMDB, asset scanners, etc.) rather than service ownership. I'll update the examples to clarify this distinction.

That being said, there's an open question of how we capture multiple sources that have contributed to our understanding of an entity. For example, we might obtain some metadata about an entity from CloudTrail logs and other details from a source like our cloud asset discovery integration. The current approach with a single source field doesn't enable us to capture this complexity. This is a question we'll need to consider for the future, especially as our product requirements evolve and we need to represent more complex entity discovery and enrichment patterns.

rfcs/text/0049/entity.yml
tinnytintin10: Nice catch in nuance. For now I'll drop the "human readable" part of the description so we can keep things simple. Entity.name is meant to represent the given name of the entity, human-readable or otherwise.

tinnytintin10: https://github.com/elastic/ecs/pull/2461/commits/374ca057f6e1aeff33f5094412a1db0e408750a7

rfcs/text/0049/entity.yml
tinnytintin10: For entity-centric documents whose metadata we captured from multiple source documents, the idea is that this entity.source field will capture multiple values (a multivalue keyword field). Related to my comment here https://github.com/elastic/ecs/pull/2461#discussion_r2051854253

I propose we retain this for Stage 1 and revisit the decision in subsequent RFCs after further testing and requirements gathering.

rfcs/text/0049/entity.yml
tinnytintin10: > What's IT environments?

While there is no standardized definition, I'm using the term to broadly refer to the entire ecosystem of technology infrastructure, systems, and platforms that support an organization's computing and data processing needs. This includes:

1. Hardware infrastructure (servers, networks, storage, endpoints)
2. Software systems (operating systems, applications, databases)
3. "Cloud" resources (IaaS, PaaS, SaaS)
4. Virtualized environments (VMs, containers)
5. External services and APIs
6. Hybrid architectures

It can encompass both internal systems managed directly by an organization and external resources provided by third parties. They can be on-premises, cloud-based, or distributed across multiple locations. 

 @romulets while I'm okay with the current description (I agree its a nit ;)), feel free to suggest an alternative copy that you think captures this better.

rfcs/text/0049/entity.yml
tinnytintin10: Nothing has changed there, I'll just make the last sentence more explicit, something like

> If the entity has multiple identifiers, these alternative identifiers can be preserved in entity.raw.*

tinnytintin10: https://github.com/elastic/ecs/pull/2461/commits/f71cc784e02ca10bfb80eebe9d3a67edf5bd695f

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+    type: flattened
+    short: Original, unmodified fields from the source system.
+    description: >
+      Original, unmodified fields from the source system stored in a flattened format that maintains
+      basic searchability. While entity.attributes should be used for normalized fields requiring
+      advanced queries, this field preserves all source metadata with basic search capabilities.
+      Supports existence queries, exact value matches, and simple aggregations.
+    dynamic: true
+
+  - name: entity.risk
```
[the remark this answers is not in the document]
tinnytintin10: I reviewed the host and user fieldset schemas and discovered that we should not track `entity.risk` in this PR. Instead, we should address it in a separate pull request for the risk fieldset. In that PR, we will indicate that the risk fieldset can be reused under `entity.*`

I'll track that work separately & after this RFC is merged. For now, I'll remove the entity.risk field.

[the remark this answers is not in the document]
tinnytintin10: https://github.com/elastic/ecs/pull/2461/commits/4de79b0d743f41cdcae66bcae40ab9997548f74e

rfcs/text/0049/entity.yml
tinnytintin10: Let's discuss the advantages and disadvantages of this approach in the upcoming integration sync. However, regarding this RFC, it is out of scope as this field does not determine the format for sub_type values.

rfcs/text/0049/entity.yml
```diff
@@ -0,0 +1,115 @@
...
+    example: aws, azure, gcp, kubernetes, cloudflare
+
+  - name: entity.type
+    level: core
+    type: keyword
+    short: Standardized high-level classification of the entity.
+    description: >
+      A standardized high-level classification of the entity. This provides a normalized way
+      to group similar entities across different providers or systems. There will be an
+      allowed set of values maintained for this field to ensure consistency.
```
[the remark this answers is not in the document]
tinnytintin10: Why do you see this as contradictory?

I see trade-offs in either direction when it comes to maintaining an allowed list of values for entity.type. I raised the concern above about manageability because I agree that a rigid, overly restrictive list can limit utility, especially as new technologies emerge. Tracking and updating such a list over time is tedious, and the maintenance overhead isn’t trivial. That said, this is exactly the kind of evolution the ECS RFC process (especially with community involvement) is designed to support... giving us a structured and transparent way to manage changes like these.

On the flip side, not having an allowed list also introduces real downsides... particularly when it comes to building consistent experiences across the product. For example, features like inventory views or flyouts that adapt based on whether you’re looking at a user, host, or storage entity rely on normalized, predictable values. Without that structure, different teams, users, etc., might populate arbitrary values, which can lead to fragmentation and a degraded UX. cc @opauloh would love to hear your thoughts here. 

So I’m trying to balance flexibility and governance (and drawing inspiration from how ECS has handled categorization fields historically). In this case, I think having a well-defined set of allowed values for entity.type (even if it’s relatively short to start) is worthwhile. It gives us a foundation to build on while allowing refinement through future proposals.

Regarding the "there will be" lanaguage, I agree. I’ll adjust that language for now while it’s in beta and I'm still refining what exactly should be on that list.

JordanSh (approved):
lgtm 👍

rfcs/text/0049/entity.yml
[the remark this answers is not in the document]
romulets: I see your reasoning, I said it contradicted the wording in this document - and you updated the description, so all good!

romulets (approved):

maxcold (approved):

orouz (approved):

opauloh (approved):

mjwolf (approved):

PR #2477, opened 2025-05-22
Allow use of the dash character in the data_stream.namespace field

### LINKED ISSUE #2388: Allow use of dash character (`-`) in `data_stream.namespace`
**Summary**

Propose ECS removes the restriction of using the dash character, `-`, in the `data_stream.namespace` field.

**Motivation**:

[RFC 0009 - Data stream fields](https://github.com/elastic/ecs/blob/main/rfcs/text/0009-data_stream-fields.md#restrictions-on-values) includes this restrictions for the `data_stream.namespace` field:

```
No longer than 100 chars
```

The [current description](https://github.com/elastic/ecs/blob/main/schemas/data_stream.yml#L69) for `data_stream.namespace` also restricts using the dash character:

```yaml
    - name: namespace
      level: extended
      type: constant_keyword
      example: production
      description: >
        A user defined namespace. Namespaces are useful to allow grouping of data.

        Many users already organize their indices this way, and the data stream naming scheme now provides this
        best practice as a default. Many users will populate this field with `default`. If no value is used, it falls back to `default`.

        Beyond the Elasticsearch index naming criteria noted above, `namespace` value has the additional restrictions:
          * Must not contain `-` <= Not present in RFC
          * No longer than 100 characters
      short: A user defined namespace. Namespaces are useful to allow grouping of data.
```

Suspect these restrictions may have been mistakenly copied from the same restriction on `data_stream.type` and `data_stream.dataset`.

I would not consider the change breaking. It makes `data_stream.namespace` more permissive and the `data_stream.*` field set is still considered beta.

taylor-swanson:
- Update the documentation to allow the use of the dash character in the data_stream.namespace field.

## Related

Closes #2388


leehinman:
Not a blocker.  But we probably need something to communicate this change.

https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme

which introduced the data_stream naming scheme specifically says

```
 In all three parts, the “-” character is not allowed. 
```

That blog post is often linked to when explaining "how data_streams work".  Maybe we can update the blog post?

taylor-swanson:
Good catch. Do we know who manages blog posts?

As an aside, do we know of any mechanisms elsewhere that may do validation on this field? ECS doesn't appear to, but Kibana, Elasticsearch, or something else might?

cc @qcorporation @ruflin

leehinman:
> As an aside, do we know of any mechanisms elsewhere that may do validation on this field? ECS doesn't appear to, but Kibana, Elasticsearch, or something else might?

I know that we have had multiple support issues with the "Custom Logs" integrations where people pick data_stream names with uppercase and with dashes.

qcorporation:
> Good catch. Do we know who manages blog posts?
> 
> As an aside, do we know of any mechanisms elsewhere that may do validation on this field? ECS doesn't appear to, but Kibana, Elasticsearch, or something else might?
> 
> cc @qcorporation @ruflin

Hi @taylor-swanson thanks ccing me.
I asked @nick-alayil about this as he does these things, I'll share the intake process offline

mjwolf (approved):

taylor-swanson:
This is going to require a bit more coordination than I have time for at the moment. There are other places in our product line that make assumptions on extra dashes being not being present in this field that need to be accounted for (namely elasticsearch), and this needs to be coordinated with them. We would also need to coordinate a change to the blog post that Lee mentioned.

PR #2475, opened 2025-05-20
[RFC 0050] Stage 1: Introducing GenAI fields

susan-shu-c:
## Overview

This RFC proposes GenAI fields, `gen_ai*`, with the increase of Generative AI and LLM logging. This will benefit our customers and users, allowing them to monitor and protect their LLM/Generative AI deployments.

Stage 0:
* https://github.com/elastic/ecs/pull/2337

Stage 1:
* https://github.com/elastic/ecs/pull/2465

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md)?
- For proposing substantial changes or additions to the schema, have you reviewed the [RFC process](https://github.com/elastic/ecs/blob/main/rfcs/README.md)?
- If submitting code/script changes, have you verified all tests pass locally using `make test`?
- If submitting schema/fields updates, have you generated new artifacts by running `make` and committed those changes?
- Is your pull request against main? Unless there is a good reason otherwise, we prefer pull requests against main and will backport as needed.
- Have you added an entry to the [CHANGELOG.next.md](https://github.com/elastic/ecs/blob/main/CHANGELOG.next.md)?

PR #2489, opened 2025-06-11
[RFC 0050] [Update] Stage 2: Introducing GenAI fields

susan-shu-c:
## Overview

This PR proposes changes on top of
* https://github.com/elastic/ecs/pull/2475

based on discussion with @felixbarny @eyalkoren in this PR:
* https://github.com/elastic/elasticsearch/pull/129122

namely, updating `gen_ai.request.model` and `gen_ai.agent.description` from `text` to `keyword`.

The rationale:
* `gen_ai.request.model`: doesn't require text analyzing
* `gen_ai.agent.description`: phrase queries may not be used often for this field, and may be expensive from storage and indexing - via @felixbarny 

> While it has a limit on how many characters it can store in [doc_values](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/doc-values) it falls back to storing the value in a stored field if [ignore_above](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/ignore-above) is set, which seems like a good compromise here.

~~I'm looking for feedback if this is covered, as it's part of the generated files and I'm not completely sure which changes are expected for `ignore_above` to work.~~

## Checklist


eyalkoren:
LGTM

> While it has a limit on how many characters it can store in [doc_values](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/doc-values) it falls back to storing the value in a stored field if [ignore_above](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/ignore-above) is set, which seems like a good compromise here.
>> I'm looking for feedback if this is covered , as it's part of the generated files and I'm not completely sure which changes are expected for `ignore_above` to work.

I am not sure what you mean. Most (or all) `keyword` fields in ECS are defined with `ignore_above: 1024`.
In addition, this is the setting we use in [`ecs@mappings`](https://github.com/elastic/elasticsearch/blob/0702e429f045ccd9ad09cd0ae58c04beaa0c500c/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json#L216) by default for all string fields that are not mapped otherwise.

schemas/gen_ai.yml
```diff
@@ -27,7 +27,7 @@
   type: group
   fields:
     - name: agent.description
-      type: text
+      type: keyword
```
felixbarny, line 30: If we're not expecting to filter based on the exact value of the description, we could disable indexing the field to reduce disk space and ingest time overhead. I think it will be more likely/sensible to do filtering on the `request.model` rather than the description, with the description just being additional context when looking at individual docs.

```suggestion
      type: keyword
      index: false
```

susan-shu-c:
Thanks @eyalkoren this is clear enough to me! I just wanted to make sure I wasn't missing something

> Most (or all) keyword fields in ECS are defined with ignore_above: 1024.

schemas/gen_ai.yml
susan-shu-c: Thanks for the suggestion, it makes sense as other fields have the coverage and `gen_ai.agent.name` can also be directly used.

susan-shu-c:
cc @trisch-me @mjwolf regarding updating `gen_ai.request.model` and `gen_ai.agent.description` from text to keyword, as the former doesn't require text analyzing and and latter is not often used for direct phrase queries; `gen_ai.agent.name` can be used instead. 

This came up as part of:
[Update ecs@mappings.json](https://github.com/elastic/elasticsearch/pull/129122)

susan-shu-c:
Thank you, @trisch-me !

experimental/generated/beats/fields.ecs.yml
```diff
@@ -3465,9 +3465,10 @@
     - name: agent.description
       level: extended
       type: keyword
-      ignore_above: 1024
       description: Free-form description of the GenAI agent provided by the application.
       example: Helps with math problems; Generates fiction stories
+      index: false
+      doc_values: false
```
felixbarny, line 3471: It's curious that doc_values is also false now but just setting `index: false` in the schema.

Is it possible to explicitly set `doc_values: true` in here to retain it in the mappings (and to keep `ignore_above`)?

https://github.com/elastic/ecs/blob/a1237a30337e855fe8dd97355783855e4e96044a/schemas/gen_ai.yml#L29-L37

susan-shu-c: Thanks for pointing this out! - To summarize what we discussed on the [Elasticsearch issue](https://github.com/elastic/elasticsearch/pull/129122) we'll merge first and then continue to adjust in the upcoming weeks.

mjwolf: @felixbarny It's possible to explicitly set `doc_values: true`, and it will be applied in the generated files.

I've created an ECS issue to update the generator: https://github.com/elastic/ecs/issues/2492. 

Should `ignore_above: 1024` be set as the default value for all keywords? I can add that to the issue to update the generator too.

PR #2493, opened 2025-06-18
[RFC] Stage 0: Add `user.is_privileged` boolean field

hop-dev:
This RFC proposes adding a new boolean field, `user.is_privileged`. It will explicitly flag when a user has elevated or administrative rights.

### DIFF
--- a/rfcs/text/0000-user-is-privileged.md
+++ b/rfcs/text/0000-user-is-privileged.md
@@ -1,0 +1,81 @@
+# 0000: Add user.is_privileged boolean field
+
+- Stage: **0 (strawperson)** 
+- Date: **TBD** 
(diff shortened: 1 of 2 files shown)


mjwolf (approved):
I think this is good for RFC stage 0.

I've set this RFC to 0051.

For future RFC stages, there are a few more things to consider.  

* What is a privileged user exactly? For example, it may not be just root, it can be a user with `wheel` or other similar groups.
* In Linux, you can already use `user.effective.id` to see if the user is or has assumed root privilages, which holds similar info to `user.is_privileged`  
* In Windows, is this set for all actions with an administrator account, or only actions with the elevated UAC permissions?

PR #2502, opened 2025-07-25
[RFC] Stage 0: Add user.is_privileged boolean field

hop-dev:
Moved from https://github.com/elastic/ecs/pull/2493

This RFC proposes adding a new boolean field, `user.is_privileged`. It will explicitly flag when a user has elevated or administrative rights.


mjwolf:
This is merged with #2493

PR #2513, opened 2025-08-15
Update Entity Fields RFC for 9.2 Stack Release

MikePaquette:
#### 1. What does this PR do?
Adds newly proposed entity.* fields to the Entity Fields RFC
Adds currently used `entity.*` fields to this RFC
Proposes a nested location in the naming hierarchy for entity.* fields populated by ECS producers
Proposes a root namespace location for entity.* fields when used by ECS consumers and entity-related data stores

<!-- List all new or modified ECS fields.
Entire list of entity.* fields has been updated
New fields added, new examples provided
No fields deleted.

See [RFC document](https://github.com/elastic/ecs/blob/main/rfcs/text/0049-entity-fields.md).

See [RFC document](https://github.com/elastic/ecs/blob/main/rfcs/text/0049-entity-fields.md).

See [RFC document](https://github.com/elastic/ecs/blob/main/rfcs/text/0049-entity-fields.md).

 NO

NO

No.  See [RFC document](https://github.com/elastic/ecs/blob/main/rfcs/text/0049-entity-fields.md).

---

#### Commit Message
Proposed updates to entity.* fields for 9.2 Stack Release.

### DIFF
--- a/rfcs/text/0049-entity-fields.md
+++ b/rfcs/text/0049-entity-fields.md
@@ -2,6 +2,7 @@
 
 - Stage: **1 (draft)**
 - Date: 2025-04-23
+- Updated: 2025-08-15
 
 An entity represents a discrete, identifiable component within an IT environment that can be described by a set of attributes and maintains its identity over time. Entities can be physical (like hosts or devices), logical (like containers or processes), or abstract (like applications or services).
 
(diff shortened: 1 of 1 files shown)


eyalkraft:
Hi Mike, Thanks for the proposal!

Is there a reason for having the new `entity.schema_version` instead of using the standard [`ecs.version`](https://www.elastic.co/docs/reference/ecs/ecs-ecs#field-ecs-version)?

MikePaquette:
> Is there a reason for having the new entity.schema_version instead of using the standard [ecs.version](https://www.elastic.co/docs/reference/ecs/ecs-ecs#field-ecs-version)?

Good question. I don't know when this was added, or how it is used, but I found it in the existing mappings in 8.19.1.

```
"entity": {
          "properties": {
            "definition_id": {
              "type": "keyword",
              "ignore_above": 1024
            },
            "definition_version": {
              "type": "keyword",
              "ignore_above": 1024
            },
            "display_name": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 1024
                }
              }
            },
            "id": {
              "type": "keyword",
              "ignore_above": 1024
            },
            "identity_fields": {
              "type": "keyword"
            },
            "last_seen_timestamp": {
              "type": "date"
            },
            "name": {
              "type": "keyword"
            },
            "schema_version": {
              "type": "keyword",
              "ignore_above": 1024
            },
            "source": {
              "type": "keyword"
            },
            "type": {
              "type": "keyword",
              "ignore_above": 1024
            }
          }
        },
```

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
 
 | Field | Type | Description |
 |-------|------|-------------|
+| entity.definition_id | keyword | Used Elastic solutions (e.g., Security, Observability) to denote the ID of the entity definition which is used to extract entity details from ingested logs, events, intelligence, and other data types. Use of this value is reserved, and ECS producers, including data ingestion pipelines, must not populate this field|
+| entity.definition_id | keyword | Used by Elastic solutions (e.g., Security, Observability) to denote the version of the entity definition which is used to extract entity details from ingested logs, events, intelligence, and other data types. Use of this value is reserved, and ECS producers, including data ingestion pipelines, must not populate this field|
```
mjwolf: `entity.definition_id` is repeated here

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
 
 | Field | Type | Description |
 |-------|------|-------------|
+| entity.definition_id | keyword | Used Elastic solutions (e.g., Security, Observability) to denote the ID of the entity definition which is used to extract entity details from ingested logs, events, intelligence, and other data types. Use of this value is reserved, and ECS producers, including data ingestion pipelines, must not populate this field|
```
mjwolf: I'm not too sure about this field being reserved. There are no other fields in ECS that are reserved for use by Elastic, and I don't really know if it makes sense to have them. Since the intention of ECS is a common schema that's shared with others, I don't if this would make sense to include.

Are there alternatives to adding this, such as using a custom field that's not defined in ECS?

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+| entity.attributes.* | object | A set of static or semi-static attributes of the entity. Usually boolean or keyword field data types. Examples include: `entity.attributes.Storage_class`, `entity.attributes.Mfa_enabled` , `entity.attributes.Privileged` , `entity.attributes.Granted_permissions` , `entity.attributes.Known_redirect` , `entity.attributes.Asset` , `entity.attributes.Managed` ,`entity.attrbitues.Os_current` , `entity.attibutes.Os_patch_current` , `entity.attributes.Oauth_consent_restriction`). Use this field set when you need to track static or semi-static characterstics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern for Examples indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```
mjwolf:
```suggestion
| entity.attributes | object | A set of static or semi-static attributes of the entity. Usually boolean or keyword field data types. Examples include: `entity.attributes.Storage_class`, `entity.attributes.Mfa_enabled` , `entity.attributes.Privileged` , `entity.attributes.Granted_permissions` , `entity.attributes.Known_redirect` , `entity.attributes.Asset` , `entity.attributes.Managed` ,`entity.attrbitues.Os_current` , `entity.attibutes.Os_patch_current` , `entity.attributes.Oauth_consent_restriction`). Use this field set when you need to track static or semi-static characterstics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern for Examples indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```

The existing flattened fields in ECS don't use `.*` in the name. I think it makes sense to remove here too, to be consistent

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
 
-## Usage
+## Field Re-use
 
-The entity field set enables us to normalize entity data in such a way where we can easily query key attributes in a standardized way regardless of the type and source of the entity. This will be how we'll normalize all entity data in the upcoming inventory experience that we're planning for the security solution.
+### ECS Producers
+
+For ECS producers, such as Beats, Elastic Agent integrations, ingest pipelines, and other methods for shipping data to Elastic, the `entity.*` fields are expected to be nested as follows:
+- If the entity type is one of host, user, service, cloud, orchestrator), then the entity fields should be nested under the respecitve root field set, for example `host.entity.*` , `user.entity.*`, etc.
+- If the entity type is not one of the above, then that `entity.*` fields should be nested under a new root-level object, called `generic`, as `generic.entity.*`
```
mjwolf: Do you think `generic` will have any other fields apart from `generic.entity.*`

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+
+Special note for use with entity fields that use `target` fields.
+If the entity type includes a provision for `target.*` fields, such as `user.target.*`, then the `entity.*` fields should be nested under the `*.target.*` field set, for example, `user.target.entity.*`.
+
+### ECS Consumers or Data Stores
+
+For ECS consumers, such as the Elastic Security Solution entity store indices, the `entity.*` fields should be used directly at the root of the events.
```
mjwolf: Could you expand on what this means, I don't think I understand it. What do you mean by "used". Why can a data store only use `entity.*` if the producers can write to other top level fieldsets?

I think this is also an implementation specific detail that doesn't need to be part of ECS, so maybe it can be removed.

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+| entity.raw.* | object | Original, unmodified fields from the source system stored in a flattened format that maintains basic searchability. While `entity.attributes` should be used for normalized fields requiring advanced queries, this field preserves all source metadata with basic search capabilities. Supports existence queries, exact value matches, and simple aggregations. |
+| entity.metrics.* | object | Field set for any fields containing numeric entity metrics. These use dymanic field data type mapping. Any fields under entity.metrics.* will automatically get the appropriate numeric mapping without needing explicit definition.
 
-When representing entities that correspond to existing ECS field sets (e.g., hosts, users, services, containers), the relevant ECS field set should be used to capture detailed metadata about that entity. For example:
+When representing entities that have existing ECS entiy field sets (e.g., hosts, users, services, containers), the relevant ECS field set should continue to be used to capture a copy of the detailed metadata about that entity. For example:
```
uri-weisman:
```suggestion
When representing entities that have existing ECS entity field sets (e.g., hosts, users, services, containers), the relevant ECS field set should continue to be used to capture a copy of the detailed metadata about that entity. For example:
```

rfcs/text/0049-entity-fields.md
romulets:
```suggestion
| entity.attributes | object | A set of static or semi-static attributes of the entity. Usually boolean or keyword field data types. Examples include: `entity.attributes.Storage_class`, `entity.attributes.Mfa_enabled` , `entity.attributes.Privileged` , `entity.attributes.Granted_permissions` , `entity.attributes.Known_redirect` , `entity.attributes.Asset` , `entity.attributes.Managed` ,`entity.attributes.Os_current` , `entity.attributes.Os_patch_current` , `entity.attributes.Oauth_consent_restriction`). Use this field set when you need to track static or semi-static characterstics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern for Examples indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```

Corrected two typos

MikePaquette: Thanks @mjwolf what do you mean by "flattened" here?  As proposed, the object will contain keyword or boolean leaf fields, and not the flattened field datatype.  No problem with removing the `.*` though, thanks.

rfcs/text/0049-entity-fields.md
MikePaquette: Not foreseen.  Can anyone else think of a reason why we'd use the root field set `generic.*` for another purpose?

rfcs/text/0049-entity-fields.md
MikePaquette: I agree we can remove this from the ECS spec, but the idea is that the entity store schema will include entity.* at the root level in each document.  this will allow us and users to query the entity store for entity ID's (and other attributes) w/o burdening them with knowing under which entity class {host, user, service, generic, etc.} it might be stored.

rfcs/text/0049-entity-fields.md
MikePaquette: Yes, I agree, it's an internal-only field - no need to have it defined in the ECS spec.  We already intend to publish the entity store schema in our docs.  Any problem with using `entity.Definition_id` as a custom field? (nesting a custom leaf field under an ECS-defined root object `entity.*`

romulets: @MikePaquette as we speak we are storing this information under `entity.Metadata.EngineType`. What do you think about keeping it under metadata?

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
 
 | Field | Type | Description |
 |-------|------|-------------|
+| entity.definition_id | keyword | Used Elastic solutions (e.g., Security, Observability) to denote the ID of the entity definition which is used to extract entity details from ingested logs, events, intelligence, and other data types. Use of this value is reserved, and ECS producers, including data ingestion pipelines, must not populate this field|
+| entity.definition_id | keyword | Used by Elastic solutions (e.g., Security, Observability) to denote the version of the entity definition which is used to extract entity details from ingested logs, events, intelligence, and other data types. Use of this value is reserved, and ECS producers, including data ingestion pipelines, must not populate this field|
+| entity.schema_version | keyword | Denotes the version of the entity schema,as published in Elastic Security documentation, to which this entity information conforms. Usually conforms to the Elastic Stack version.
```
romulets: Is this a new guideline? Or is it something that already happens? I'm not aware of an entity schema published in Elastic Security Docs that usually conforms to elastic stack version

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
-| entity.sub_type | keyword | The specific type designation for the entity as defined by its provider or system. This field provides more granular classification than entity.type. Examples: `aws_s3_bucket`, `gcp_cloud_storage_bucket`, `azure_blob_container` would all map to type `bucket`. |
+| entity.type | keyword | A standardized high-level classification of the entity. This provides a normalized way to group similar entities across different providers or systems. Example values: `bucket`, `database`, `container`, `function`, `queue`, `host`, `user`, `application`, `session`, etc. |
+| entity.sub_type | keyword | The specific type designation for the entity as defined by its provider or system. This field provides more granular classification than entity.type. Examples: `aws_s3_bucket`, `gcp_cloud_storage_bucket`, `azure_blob_container` would all map to type `bucket`.  `hardware` , `virtual` , `container` , `node` , `cloud_instance` would all map to entity type `host'.|
```
romulets:
```suggestion
| entity.sub_type | keyword | The specific type designation for the entity as defined by its provider or system. This field provides more granular classification than entity.type. Examples: `aws_s3_bucket`, `gcp_cloud_storage_bucket`, `azure_blob_container` would all map to entity type `bucket`.  `hardware` , `virtual` , `container` , `node` , `cloud_instance` would all map to entity type `host'.|
```

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+| entity.attributes | object | A set of static or semi-static attributes of the entity. Usually boolean or keyword field data types. Examples include: `entity.attributes.Storage_class`, `entity.attributes.Mfa_enabled` , `entity.attributes.Privileged` , `entity.attributes.Granted_permissions` , `entity.attributes.Known_redirect` , `entity.attributes.Asset` , `entity.attributes.Managed` ,`entity.attributes.Os_current` , `entity.attributes.Os_patch_current` , `entity.attributes.Oauth_consent_restriction`). Use this field set when you need to track static or semi-static characterstics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern for Examples indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```
romulets: The current pattern of snake case with capital first letter isn't friendly for most coding tools and linters, since it's not a standard pattern.

Could we adopt a pattern such as `PascalCase`. Reading the [documentation over capitizaion for non ecs fields ](https://www.elastic.co/guide/en/ecs/8.17/ecs-custom-fields-in-ecs.html#_capitalization) I see precedent for it, since it's mentioned both in HAProxy and NGINX examples

romulets: I understand what you mean "static or semi-static attributes" as opposed to `lifecycle` and `behaviour` fields, but maybe we have a better way of describing what do we mean by that? By the `static or semi-static` definition, `first_seen` and `issued_at` would also fit under it, wouldn't it?

As I see attributes should be used to describe non temporal entity properties that could not be expressed in other parts of the entity field set.

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+| entity.behavior.* | object | A set of ephemeral characteristics of the entity, derived from observed behaviors during a specific time period. Behaviors are usually captured in event logs under fields such as `event.action` and other fields, but this field set captures "attributified" behavior indicators, using semantics like "this behavior was seen one or more times during this time period." Sytems using this field set may need to force a "reset" of these behavioral indicators at the end of their current period. Usually boolean field data type. Examples include: `entity.behavior.Used_usb_device`, `entity.behavior.Brute_force_victim` , `entity.behavior.New_country_login` ,etc. ). Use this field set when you need to capture and track ephemeral characterstics of an entity for advanced searching, correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```
romulets:
```suggestion
| entity.behavior.* | object | A set of ephemeral characteristics of the entity, derived from observed behaviors during a specific time period. Behaviors are usually captured in event logs under fields such as `event.action` and other fields, but this field set captures "attributified" behavior indicators, using semantics like "this behavior was seen one or more times during this time period." Systems using this field set may need to force a "reset" of these behavioral indicators at the end of their current period. Usually boolean field data type. Examples include: `entity.behavior.Used_usb_device`, `entity.behavior.Brute_force_victim` , `entity.behavior.New_country_login` ,etc. ). Use this field set when you need to capture and track ephemeral characteristics of an entity for advanced searching, correlation of normalized values across different providers/sources and entity types. Note the initial capitalization pattern indicates that any such fields are custom entity-specific fields that won't be enumerated in the ECS schema, and won't collide with any fields that may be defined by ECS in the future. |
```

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+| entity.raw.* | object | Original, unmodified fields from the source system stored in a flattened format that maintains basic searchability. While `entity.attributes` should be used for normalized fields requiring advanced queries, this field preserves all source metadata with basic search capabilities. Supports existence queries, exact value matches, and simple aggregations. |
```
romulets: Should be of type `flattened` instead of `object`, no?

rfcs/text/0049-entity-fields.md
romulets: I can't think of any other reason

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+This approach ensures backward compatibility, maintains existing ECS patterns, and preserves the rich metadata capabilities of established field sets while allowing the entity fields to provide a consistent way to identify and categorize *all* entity types. The entity fields serve as a complementary layer that enables unified entity representation, particularly for entity types that don't have dedicated field sets.
 
-## Usage
+## Field Re-use
 
-The entity field set enables us to normalize entity data in such a way where we can easily query key attributes in a standardized way regardless of the type and source of the entity. This will be how we'll normalize all entity data in the upcoming inventory experience that we're planning for the security solution.
+### ECS Producers
+
+For ECS producers, such as Beats, Elastic Agent integrations, ingest pipelines, and other methods for shipping data to Elastic, the `entity.*` fields are expected to be nested as follows:
+- If the entity type is one of host, user, service, cloud, orchestrator), then the entity fields should be nested under the respecitve root field set, for example `host.entity.*` , `user.entity.*`, etc.
```
romulets:
```suggestion
- If the entity type is one of host, user, service, cloud, orchestrator, then the entity fields should be nested under the respecitve root field set, for example `host.entity.*` , `user.entity.*`, etc.
```

rfcs/text/0049-entity-fields.md
```diff
@@ -20,31 +21,55 @@ This approach would allow ECS to accommodate new types of entities without requi
...
+- User entities should also leverage the `user.*` field set for user-specific attributes
+- Cloud resource entities should also use the `cloud.*` field set for cloud provider details
+- Service entities should also employ the `service.*` field set for service-specific metadata
 
-This approach ensures backward compatibility, maintains existing ECS patterns, and preserves the rich metadata capabilities of established field sets while allowing the entity fields to provide a consistent way to identify and categorize all entity types. The entity fields serve as a complementary layer that enables unified entity representation, particularly for entity types that don't have dedicated field sets.
+This approach ensures backward compatibility, maintains existing ECS patterns, and preserves the rich metadata capabilities of established field sets while allowing the entity fields to provide a consistent way to identify and categorize *all* entity types. The entity fields serve as a complementary layer that enables unified entity representation, particularly for entity types that don't have dedicated field sets.
 
-## Usage
+## Field Re-use
```
romulets, line 47: I believe `Field Usage` it's still a good section name since we are also talking about `ECS consumers` using `entity.*` as root

rfcs/text/0049-entity-fields.md
MikePaquette: No, I found this in the existing index mappings 8.19.1, and was not aware of it.  Included here in case it was already in use, but it does not appear to be used, so we can remove it from this RFC.

rfcs/text/0049-entity-fields.md
MikePaquette: @mjwolf do you have a position on this topic?

MikePaquette: I see your point, perhaps we can add "non-temporal" to the definition of the `entity.attributes.*` fields?  Would that help?

rfcs/text/0049-entity-fields.md
MikePaquette: I think so for `entity.raw` .  We've agreed to remove the `.*` from the objects, and this would be the same.

rfcs/text/0049-entity-fields.md
romulets: I think so!

mjwolf: ECS uses snake case for multiword fields, so I think it makes the most sense to keep using that. OTel semantic conventions also uses snake case.

I think for these examples in documentation it should keep using snake case. For the actual implementations, we don't require it, so other cases are allowed (as long as it starts with a capital).

MikePaquette: for ECS, this is irrelevant because we should not provide examples using custom fields.

PR #2518, opened 2025-08-23
Add required `--semconv-version` to TLDR example

smortex:
The "ECS Tooling Usage" guide has a "TLDR example" to get started.  It needs to be adjusted by the user to generate templates, but since 009caf435b0a5330ef0b935c81e7146d3ea8a4e4, the user also have to provide a value for the `--semconv-version` required parameter.  Unfortunately, nothing is told about this parameter in that file, leading the user with a cryptic:

```
  File "./ecs/scripts/generator.py", line 144, in <module>
    main()
  File "./ecs/scripts/generator.py", line 46, in main
    raise KeyError("OTel Semantic Conventions version has not been provided as a config option '--semconv-version'")
KeyError: "OTel Semantic Conventions version has not been provided as a config option '--semconv-version'"
```

Adjust the example to make bootstraping a custom config easier.

Only the first example is adjusted in this PR, as it makes examples more verbose and I assumed the user would find how to quickly fix the issue when they follow the guide bottom-down.  On request, I can adjust the PR to fix the other examples for correctness, or adjust the script so that it loads the version from that file if not provided.


mjwolf (approved):
Looks good, thanks for the contribution!

Issue #2522, opened 2025-09-03
New `event.category` for device events


ricardo-estc:
**Summary**

Elastic Defend is expanding its data collection capabilities to include peripheral device connection events. Currently existing values for the event.category field don't categorize device-related events such as USB devices, external storage, input devices, and other peripherals being connected or disconnected from hosts. Adding a dedicated category will provide clear distinction for device related events.

**Motivation**:

We suggest adding a new value for `event.category`: `device`. The new value would be used to categorize events related to peripheral devices connecting to, disconnecting from, or being managed by the host operating system. This enables security teams to monitor unauthorized device usage, track hardware inventory, enforce device policies, and detect potential security threats from malicious devices or data exfiltration attempts through removable storage.

**Detailed Design**:

Provide additional details around the design of the proposed changes.

* Field names
* Example values for the fields
`event.category`: `device`
`event.type` : `start`, `info`, `allowed`, `denied` (existing values)
* Suggested appropriate datatypes
* Any example events that map to the proposed use case(s)

ricardo-estc:
Related https://github.com/elastic/ecs/pull/2521

ricardo-estc:
Closing issue. It was agreed to have a new `event.type` instead of `event.category`

PR #2524, opened 2025-09-04
Adding new value for `event.type`:`device`

ricardo-estc:
#### 1. What does this PR do?

This PR adds a new value for ~`event.category:device`~ `event.type`:`device`. The new value would be used to categorize events related to peripheral devices connecting to, disconnecting from, or being managed by the host operating system. This enables security teams to monitor unauthorized device usage, track hardware inventory, enforce device policies, and detect potential security threats from malicious devices or data exfiltration attempts through removable storage.

Related issue: https://github.com/elastic/ecs/issues/2522

### DIFF
--- a/docs/reference/ecs-allowed-values-event-type.md
+++ b/docs/reference/ecs-allowed-values-event-type.md
@@ -25,6 +25,7 @@ This field is an array. This will allow proper categorization of some events tha
 * [creation](#ecs-event-type-creation)
 * [deletion](#ecs-event-type-deletion)
 * [denied](#ecs-event-type-denied)
+* [device](#ecs-event-type-device)
 * [end](#ecs-event-type-end)
 * [error](#ecs-event-type-error)
 * [group](#ecs-event-type-group)


qcorporation:
> Have you considered proposing a new value for `event.type` instead? It too is an array, and should accommodate the proposed "device" allowed value.
> 
> I have two concerns with adding `device` as a new value for event.category.
> 
> 1. For ECS users in general, it may cause confusion and diversion from `event.category:host` which is the preferred category for all host events.  (FYI, when ECS was first defined, both "system" and "device" were considered before deciding on "host".)
> 2. It may be difficult to identify a device action when the `event.category` (array) contains both `host` and `device`, with an `event.type` value such as `info`.  (i.e., Which is it - host info or device info?)
> 
> If instead we extend `event.type` to include a new allowed value of `device`, then an event that had: `event.kind: event` `event.category:host` `event.type: [device], [denied]` `event.action: usb_storage_device_blocked`
> 
> would be very clear that it took place on a host, and an attempt to install a device was blocked by the EDR.

@MikePaquette this makes a lot of sense

@ricardo-estc do you think you can make the necessary changes based upon Mike's recommendations?

ricardo-estc:
@MikePaquette @qcorporation done, please review the updated changes

qcorporation (approved):
LGTM

taylor-swanson (approved):

mjwolf (approved):

MikePaquette (approved):
Please close PR #2522 
Please fix docs issue where "device" is not added to the list at the top
image

Otherwise, LGTM.

ricardo-estc:
image


Updated docs @MikePaquette, thank you!

PR #2521, opened 2025-09-03
[RFC] New device fields - Stage 0

ricardo-estc:
#### 1. What does this PR do?

This PR proposes adding new fields to the existing device field set in ECS to provide more comprehensive device identification and classification capabilities. It enhances device tracking, inventory management, and security monitoring by adding vendor, product, and device type information.

#### 2. Which ECS fields are affected/introduced?

* device.vendor_name (keyword)
* device.vendor_id (keyword)
* device.product_name (keyword)
* device.product_id (keyword)
* device.device_type (keyword)

#### 4. Have you added/updated documentation?

YES / NO / N/A

#### 5. Have you built ECS and committed any newly generated files?

YES / NO

#### 6. Have you run the ECS validation tests locally?

YES / NO

#### 7. Anything else for the reviewers?

---

PR #2530, opened 2025-09-12
[RFC 0053][Update] Stage2: New device fields 

ricardo-estc:
This PR proposes changes on top of:
* https://github.com/elastic/ecs/pull/2521

Added fields based on previous PR:
* device.vendor.name 
* device.vendor.id
* device.product.name 
* device.product.id 
* device.type


schemas/device.yml
```diff
@@ -75,3 +75,43 @@
...
       example: DJGAQS4CW5
       beta: This field is beta and subject to change.
+    - name: product.id
+      level: extended
+      type: keyword
+      short: ProductID of the device
+      description: >
+        ProductID of the device.
+      example: 0xABCD
+      beta: This field is beta and subject to change.
```
ricardo-estc, line 89: @trisch-me should I remove these beta entries?

trisch-me: no, beta fields should stay in.

schemas/device.yml
```diff
@@ -75,3 +75,43 @@
         The unique serial number serves as a distinct identifier for each device, aiding in inventory management and device authentication.
       example: DJGAQS4CW5
       beta: This field is beta and subject to change.
+    - name: product.id
+      level: extended
+      type: keyword
+      short: ProductID of the device
+      description: >
```
trisch-me, line 82: you should add also links to what exactly product ID is (as in phase 0 PR)
You should treat each description in a way that everyone, who wants to use this field, should understand how the field is generated (or what constraints it has)

trisch-me: so while short might be the same as in phase 0, description should elaborate more details on it

ricardo-estc: @trisch-me added proper description for each field.
Is it fine to include links to the microsoft documentation about vid and pid?
Their documentation is very well written

trisch-me: yes!

schemas/device.yml
```diff
@@ -75,3 +75,59 @@
...
+      example: Extreme V2 SSD
+      beta: This field is beta and subject to change.
+    - name: type
+      level: extended
+      type: keyword
+      short: Device type classification
+      description: >
+        A classification of the device based on its primary function or device class. This field categorizes devices into functional groups to enable policy enforcement and monitoring based on device capabilities.
+
+        The classification should follow standard device class definitions where possible, such as "Storage Device", "Human Interface Device", "Audio", "Video", "Network", "Communication", etc. This allows for consistent categorization across different device types and manufacturers.
```
trisch-me, line 107: should this field be open or an enum?

schemas/device.yml
```diff
@@ -75,3 +75,59 @@
...
+    - name: product.id
+      level: extended
+      type: keyword
+      short: ProductID of the device
+      description: >
+        A unique identifier assigned by the vendor to distinguish different product models. This is typically a hexadecimal value that, combined with the vendor ID, creates a globally unique device identifier.
+
+        The product ID is assigned by the device manufacturer and should remain consistent across all instances of the same product model. For hardware devices, this often corresponds to the Product ID (PID) in device descriptors.
+
+        See https://learn.microsoft.com/en-us/windows-hardware/drivers/install/standard-usb-identifiers for more details on product identification standards.
```
trisch-me, line 87: the same here - does it make sense to make it an enum?

schemas/device.yml
ricardo-estc: In my opinion it should be open, there are numerous device types and integration can leverage it.
For example, some integrations might report Virtual devices and there is an infinite number of types that can be associated with this field

schemas/device.yml
ricardo-estc: `product.id` is a number. Example here for Apple (vendor.id == 05AC)

https://devicehunt.com/view/type/usb/vendor/05AC

trisch-me (approved):

trisch-me:
@mjwolf I will wait for your review before merge

PR #2537, opened 2025-09-24
Advance RFC 0049 (entity store) to stage 2

mjwolf:
#### 1. What does this PR do?

This advanced RFC 0049 to stage 2, and adds entity fields to the ECS schema as beta fields.

This adds new `entity` fields as beta, and modifies host, user, cloud and service fieldset by adding entity fields to them through re-use.

This advanced RFC 0049 through the RFC process, and makes the entity fields part of the ECS schema, which will unlock use cases using these fields.

YES

YES

YES

These changes do not completely implement the RFC requirements in the schema yet. There are tooling changes needed to allow `user.target.entity.*` reuse to work. `user.target` is itself a reused field, and added a reuse field to another reuse field is not working properly.

---

#### Commit Message

Advance RFC 0049 (entity store) to stage 2


kgeller:
> #### 7. Anything else for the reviewers?
> These changes do not completely implement the RFC requirements in the schema yet. There are tooling changes needed to allow `user.target.entity.*` reuse to work. `user.target` is itself a reused field, and adding a reuse field to another reuse field is not working properly.

@mjwolf did you try altering the subsets file to allow this? I'm pretty sure that's how we allow the complex nested re-use ie client.user.group. I think you just need to add entity to the user.target section https://github.com/elastic/ecs/blob/main/schemas/subsets/main.yml#L600

PR #2547, opened 2025-10-01
Change entity multi_field text type to match_only_text

mjwolf:
#### 1. What does this PR do?

Change the multi_field field type in the entity schema to `match_only_text`.

`entity.name` and `entity.display_name`, and the other fieldsets where entity is re-used are affected.

`match_only_text` is the dynamic mapping type that matches these fields, and is what is used for other similar fields, so changing the type will make this consistent with others.

YES

YES

YES

---

#### Commit Message

Change the field type for multi_field values in entity from text to match_only_text.

This change will match the multi_field type used for other similar fields in other schema files, and will match the expected dynamic value in ecs@mappings.


flash1293 (approved):
LGTM, this looks like it would fix the problem explained here: https://buildkite.com/elastic/elasticsearch-ecs-dynamic-templates-tests/builds/766#0199b51a-cf46-46fa-a72a-f589ff822206

kgeller (approved):

PR #2556, opened 2025-10-22
Add entity to the top level namespaces in which it's being used

### LINKED ISSUE #2555: Entity subdomain doesn’t exist in top level domains
After recent [changes](https://github.com/elastic/ecs/pull/2537/) (adding entity fieldset to the ECS ) we have following inconsistency for all affected groups (`cloud`, `host`, `service`, `user`). Let discuss check only `user` as an example:

RFC says that 
```
If the entity type includes a provision for `target.*` fields, such as `user.target.*`, then the `entity.*` fields should be nested under the `*.target.*` field set, for example, `user.target.entity.*`. 
```
PR above adds `entity` sub namespace to the `target` but not to the namespace itself i.e. `user.target` has type of `user` namespace [itself](https://www.elastic.co/docs/reference/ecs/ecs-user#ecs-user-nestings). So essentially it means `target` is of type `user` therefore we adding `entity` to the `user` namespace.

### Changes required

Update yaml files to include `entity` to the top level namespace - `user.entity`, `host.entity` etc

cc @nick-alayil @MikePaquette @mjwolf

mjwolf:
#### 1. What does this PR do?

Add entity to the top level namespace of which it's used in sub-types.

For example, we want `user.target.entity` fields to be present. `user.entity` itself is a `user` type, so `entity` must be reused on the user type.

This also removes custom generator code which was being used to workaround the fact that this wasn't true before.

Fixes #2555 

YES

YES

YES

---

#### Commit Message

Add entity to the top level namespace of which it's used in sub-types.

For example, we want `user.target.entity` fields to be present. `user.entity` itself is a `user` type, so `entity` must be reused on the user type.

This also removes custom generator code which was being used to workaround the fact that this wasn't true before.

### DIFF
--- a/scripts/schema/finalizer.py
+++ b/scripts/schema/finalizer.py
@@ -21,7 +21,8 @@ import re
 from schema import visitor
 
 # This script takes the fleshed out deeply nested fields dictionary as emitted by
-# cleaner.py, and performs field reuse in two phases.
+# cleaner.py, and performs field reuse in two phases, repeated for each reuse order, from highest
+# priority to lowest.
 #
 # Phase 1 performs field reuse across field sets. E.g. `group` fields should also be under `user`.
 # This type of reuse is then carried around if the receiving field set is also reused.
@@ -76,7 +77,6 @@ def perform_reuse(fields):
             for schema_name, reuse_entries in foreign_reuses[order].items():
                 schema = fields[schema_name]
                 for reuse_entry in reuse_entries:
-                    # print(order, "FOREIGN: {} => {}".format(schema_name, reuse_entry['full']))
                     nest_as = reuse_entry['as']
                     destination_schema_name = reuse_entry['full'].split('.')[0]
                     destination_schema = fields[destination_schema_name]
@@ -104,7 +104,6 @@ def perform_reuse(fields):
                 reused_fields = copy.deepcopy(schema['fields'])
                 set_original_fieldset(reused_fields, schema_name)
                 for reuse_entry in reuse_entries:
-                    # print(order, "SELF-NESTING: {} => {}".format(schema_name, reuse_entry['full']))
                     nest_as = reuse_entry['as']
                     new_field_details = copy.deepcopy(schema['field_details'])
                     new_field_details['name'] = nest_as


docs/reference/ecs-entity.md
```diff
@@ -35,7 +35,11 @@ The `entity` fields are expected to be nested at:
```
trisch-me, line 35: `cloud.entity` is missing

docs/reference/ecs-entity.md
```diff
@@ -35,7 +35,11 @@ The `entity` fields are expected to be nested at:
 
 * `cloud.target.entity`
 * `host.entity`
+* `service.entity`
 * `service.target.entity`
+* `user.entity`
 * `user.target.entity`
 
-Note also that the `entity` fields are not expected to be used directly at the root of the events.
+Note also that the `entity` fields may be used directly at the root of the events.
```
trisch-me: is it correct though? 
they can’t be used directly at the root as of now. they are always attached to some parent, i.e. `host`, `user.target` etc

mjwolf: yes, it shouldn't be a top-level field. I've updated it

docs/reference/ecs-entity.md
mjwolf: Added

schemas/subsets/main.yml
```diff
@@ -657,8 +657,8 @@ fields:
         fields:
           domain: {}
           email: {}
-          entity:
-            fields: "*"
+#          entity:
```
kgeller: should we remove this all together, or add a comment with our intention if you think there's value in keeping it commented out ?

schemas/entity.yml
```diff
@@ -13,21 +13,21 @@
     maintains its identity over time.
   root: false
   reusable:
-    top_level: true
+    top_level: false
```
kgeller, line 16: this change feels unrelated to the description of the PR... is it intentional?

schemas/subsets/main.yml
mjwolf: I've removed it now

schemas/entity.yml
mjwolf: This is intentional, it's a requirement from the RFC that entity is only nested under another object type

generated/csv/fields.csv
```diff
@@ -95,6 +95,21 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 9.3.0-dev,true,cloud,cloud.target.provider,keyword,extended,,aws,Name of the cloud provider.
 9.3.0-dev,true,cloud,cloud.target.region,keyword,extended,,us-east-1,"Region in which this host, resource, or service is located."
 9.3.0-dev,true,cloud,cloud.target.service.name,keyword,extended,,lambda,The cloud service name.
+9.3.0-dev,true,cloud,cloud.target.target.entity.attributes,object,extended,,,A set of static or semi-static attributes of the entity.
```
kgeller: Looks like we are getting target duplicated in the output

`cloud.target.target.entity.attributes`

generated/csv/fields.csv
```diff
@@ -1828,21 +1843,6 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 9.3.0-dev,true,user,user.roles,keyword,extended,array,"[""kibana_admin"", ""reporting_user""]",Array of user roles at the time of the event.
 9.3.0-dev,true,user,user.target.domain,keyword,extended,,,Name of the directory the user is a member of.
 9.3.0-dev,true,user,user.target.email,keyword,extended,,,User email address.
-9.3.0-dev,true,user,user.target.entity.attributes,object,extended,,,A set of static or semi-static attributes of the entity.
```
kgeller, line 1831: Looks like all the user.* entity related fields are removed?

mjwolf:
This is a summary of the changes I made in this PR now:
* Removed entity as a top-level field
* Added entity as a field on top-level objects:
  * cloud.entity.* 
  * user.entity.*
  * service.entity.*
  * host.entity.*
  * orchestrator.entity.* (this is new. It's requested in the RFC but was missed before)
* Added with self-nested objects:
  * cloud.origin.entity.*
  * cloud.target.entity.*
  * user.changes.entity.*
  * user.effective.entity.*
  * user.target.entity.*
  * service.origin.entity.*
  * service.target.entity.*
 * Reverted the changes I made to the generator scripts previously for entity, which enabled the incorrect reuse that was fixed with this
 * Modified finalizer.py, to consider order level first, and process foreign and self-nested reuse at the same time for each level.  This was needed to allow both self-nested and foreign reuse to work with entity.


@nick-alayil, can you confirm if these changes are correct and what you want for entity?  entity is added to some other objects as a consequence of being added to the top-level field (i.e. `user.effective` is a user type, and since `entity` is now a field on user, it's also added at `user.effective.entity`). It could be changed, if you don't want entity on these fields

nick-alayil:
> @nick-alayil, can you confirm if these changes are correct and what you want for entity?

Thanks @mjwolf for making these changes! From [this ticket’s](https://github.com/elastic/ecs/issues/2555#issuecomment-3430619596) context, everything looks good. I was mainly aiming to get `user.entity.*` and `service.entity.*` added.

There are a few other fields missing from a Graph Viz perspective, but I think those should be handled separately from this PR:

* `host.target.entity.*` needed for Graph Viz, but the host FieldSet [doesn’t](https://www.elastic.co/docs/reference/ecs/ecs-host#ecs-host-nestings) currently have a target nested under host.
* `generic.entity.*` and `generic.target.entity.*` also needed for Graph Viz, though the generic FieldSet doesn’t exist yet. This would introduce a new root-level object and likely requires a separate discussion/PR.

>  It could be changed, if you don't want entity on these fields

Not required (`user.changes.entity.*`, `user.effective.entity.*`) from the Graph Viz perspective. Similarly, `*.origin.entity.*` fields (like `cloud.origin.entity.*` and `service.origin.entity.*`) are also not needed.

generated/csv/fields.csv
```diff
@@ -1795,10 +1870,31 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
...
 9.3.0-dev,true,user,user.changes.name.text,match_only_text,core,,a.einstein,Short name or login of the user.
+9.3.0-dev,true,user,user.changes.risk.calculated_level,keyword,extended,,High,A risk classification level calculated by an internal system as part of entity analytics and entity risk scoring.
+9.3.0-dev,true,user,user.changes.risk.calculated_score,float,extended,,880.73,A risk classification score calculated by an internal system as part of entity analytics and entity risk scoring.
+9.3.0-dev,true,user,user.changes.risk.calculated_score_norm,float,extended,,88.73,A normalized risk score calculated by an internal system.
+9.3.0-dev,true,user,user.changes.risk.static_level,keyword,extended,,High,"A risk classification level obtained from outside the system, such as from some external Threat Intelligence Platform."
+9.3.0-dev,true,user,user.changes.risk.static_score,float,extended,,830.0,"A risk classification score obtained from outside the system, such as from some external Threat Intelligence Platform."
+9.3.0-dev,true,user,user.changes.risk.static_score_norm,float,extended,,83.0,A normalized risk score calculated by an external system.
```
kgeller, line 1878: did we mean for `user.changes.risk.*` to be added?

generated/csv/fields.csv
```diff
@@ -1808,8 +1904,29 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
...
 9.3.0-dev,true,user,user.effective.name.text,match_only_text,core,,a.einstein,Short name or login of the user.
+9.3.0-dev,true,user,user.effective.risk.calculated_level,keyword,extended,,High,A risk classification level calculated by an internal system as part of entity analytics and entity risk scoring.
+9.3.0-dev,true,user,user.effective.risk.calculated_score,float,extended,,880.73,A risk classification score calculated by an internal system as part of entity analytics and entity risk scoring.
+9.3.0-dev,true,user,user.effective.risk.calculated_score_norm,float,extended,,88.73,A normalized risk score calculated by an internal system.
+9.3.0-dev,true,user,user.effective.risk.static_level,keyword,extended,,High,"A risk classification level obtained from outside the system, such as from some external Threat Intelligence Platform."
+9.3.0-dev,true,user,user.effective.risk.static_score,float,extended,,830.0,"A risk classification score obtained from outside the system, such as from some external Threat Intelligence Platform."
+9.3.0-dev,true,user,user.effective.risk.static_score_norm,float,extended,,83.0,A normalized risk score calculated by an external system.
```
kgeller, line 1912: same with `user.effective.risk.*`?

experimental/generated/csv/fields.csv
```diff
@@ -1859,6 +1976,12 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 9.3.0-dev+exp,true,user,user.target.id,keyword,core,,S-1-5-21-202424912787-2692429404-2351956786-1000,Unique identifier of the user.
 9.3.0-dev+exp,true,user,user.target.name,keyword,core,,a.einstein,Short name or login of the user.
 9.3.0-dev+exp,true,user,user.target.name.text,match_only_text,core,,a.einstein,Short name or login of the user.
+9.3.0-dev+exp,true,user,user.target.risk.calculated_level,keyword,extended,,High,A risk classification level calculated by an internal system as part of entity analytics and entity risk scoring.
```
mjwolf, line 1979: These new risk fields should have existed previously. `risk` is a field in user type, and since `user.target` is a user type, `user.target.risk` should have exist. This is fixed by the change to finalizer ordering that fixed entity fields in the same situation.

kgeller (approved):

PR #2564, opened 2025-11-03
Rename the _types module to ecs_types.

mjwolf:
#### 1. What does this PR do?

Modules with leading underscore are reserved for internal use, and in newer version of python there is an internal _types module which conflicts with the ecs _types. The ECS module is renamed to ecs_types, to remove the conflict with the internal _types.

None

This fixed a build error when using Python 3.14.0+

N/A

YES

YES

---

#### Commit Message

Modules with leading underscore are reserved for internal use, and in newer version of python there is an internal _types module which conflicts with the ecs _types. The ECS module is renamed to ecs_types, to remove the conflict with the internal _types.


trisch-me:
This PR could be closed after your commits being merged through my PR

PR #2567, opened 2025-11-11
Update entity field set

uri-weisman:
#### 1. What does this PR do?
This PR updates the Entity field set documentation to add two new entity types (cloud and orchestrator), enable usage at the root level of events, and introduce self-nesting via `entity.target`.

Added cloud and orchestrator as new allowed values for `entity.type`
Changed guidance to allow entity fields at the root of events (previously not expected)
Added `entity.target` to allow entity fields to be nested under themselves.

Clarify the nested target use.
Adds new supported types to `entity.type`

YES

YES

YES

---

#### Commit Message
Added cloud and orchestrator as new allowed values for `entity.type`
Changed guidance to allow entity fields at the root of events (previously not expected)
Added `entity.target` to allow entity fields to be nested under themselves.


docs/reference/ecs-entity.md
```diff
@@ -27,16 +27,22 @@ The entity fields provide a standardized way to represent and categorize differe
...
 The `entity` fields are expected to be nested at:
 
 * `cloud.entity`
+* `entity.target`
 * `host.entity`
 * `orchestrator.entity`
 * `service.entity`
 * `user.entity`
 
 Note also that the `entity` fields are not expected to be used directly at the root of the events.
```
trisch-me: this needs to be removed

schemas/entity.yml
```diff
@@ -11,6 +11,13 @@
...
+  footnote: >
+    The entity fields may be self-nested under entity.target.* to describe
+    the target entity in the context of an action or event. The fieldset
+    entity.target.* must not be confused with the root entity fieldset that
+    is used to describe the primary entity under observation. The fieldset
+    entity.target.* may only be used to describe the targeted entity of an
+    action taken.
   root: false
   reusable:
     top_level: false
```
trisch-me: also this should be true

schemas/entity.yml
```diff
@@ -62,7 +72,7 @@
       short: Standardized high-level classification of the entity.
       description: >
         A standardized high-level classification of the entity. This provides a normalized way
-        to group similar entities across different providers or systems. Example values: `bucket`, `database`, `container`, `function`, `queue`, `host`, `user`, `application`, `session`, etc.
+        to group similar entities across different providers or systems. Example values: `bucket`, `database`, `container`, `function`, `queue`, `host`, `user`, `application`, `session`, `cloud`, `orchestrator`, etc.
```
trisch-me, line 75: I would also add here something similar to 
If an entity is nested under a top-level namespace such as `host` or `cloud`, its `type` array should include the corresponding value - for example, `host` or `cloud`.

uri-weisman:
@trisch-me - feels like a lot of auto generated files, let me know if that's expected.

schemas/entity.yml
```diff
@@ -62,7 +72,8 @@
...
+        If an entity is nested under a top-level namespace like `host` or `cloud`, its type array should include the matching value — for example, `host` or `cloud`.
```
trisch-me:
```suggestion
        If an entity is nested under a top-level namespace like `host` or `cloud`, or similar, its type array should include the matching value — for example, `host` or `cloud`.
```

trisch-me (approved):
lgtm except small refinement

PR #2577, opened 2026-01-08
Extend entity schema with relationship and risk

uri-weisman:
#### 1. What does this PR do?
- Adds a new `entity.relationship` field (beta) to track relationship attributes (already part of entity store schema as of 9.2)
- Enables risk fields to be nested under entity schema (adds `entity` to risk schema's reusable expected locations)

Change type to `entity.source` to become an array, to be able to hold multiple data sources that construct an entity.

**New fields:**
- `entity.relationship` (extended, object, beta) - A set of relationship attributes that can vary between entity types. Similar to `entity.attributes`, this field uses object type to allow flexible schema definitions.

**Field reuse enabled:**
- `entity.risk.*` - Risk fields can now be nested under entity (e.g., `entity.risk.calculated_score`, `entity.risk.calculated_level`, etc.)

**Purpose:**
- `entity.relationship` enables tracking relationship characteristics of entities for advanced searching and correlation across different providers/sources and entity types
- `entity.risk.*` allows risk scoring for any entity type, not just hosts and users

**Entity relationship field:**
- Enables better entity correlation and relationship tracking in security and observability use cases
- Already part of the entity store schema.
- We plan to extract relationship data from relevant integration logs, some might be inferred by entity analytics.

**Entity risk field reuse:**
- Already part of the schema.

YES

YES

YES

- https://github.com/elastic/security-team/issues/15021
#### Commit Message
Add entity.relationship field and enable risk field reuse for entity


schemas/entity.yml
```diff
@@ -163,6 +163,14 @@
         A set of static or semi-static attributes of the entity. Usually boolean or keyword field data types. Use this field set when you need to track static or semi-static characteristics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types.
       beta: This field is beta and subject to change.
 
+    - name: relationship
+      level: extended
+      type: object
+      short: A set of relationship attributes that can vary between entity types.
+      description: >
+        The structure and content of this field set may differ depending on the entity type. Use this field set when you need to track relationship characteristics of an entity for advanced searching and correlation of normalized values across different providers/sources and entity types.
+      beta: This field is beta and subject to change.
```
trisch-me, line 178: can you add examples?

trisch-me: how this field should be used for external user? It’s not clear what is inside. is it `any` object or there is some structure.

schemas/risk.yml
```diff
@@ -30,6 +30,7 @@
     expected:
       - host
       - user
+      - entity
```
trisch-me, line 33: how this work if entity is of type of host? Will be there doubled information for risk?

schemas/entity.yml
uri-weisman: Hey! I've added an example. let me know if it's clear enough.

schemas/risk.yml
uri-weisman: Even though host and user have risk scores, we decided at the offsite to only update `entity.risk` and stay agnostic to entity type.
We won’t keep risk data in two places, it’ll live only under `entity.*`.

schemas/entity.yml
trisch-me: I think we could try to improve it and provide different options inside `relationship`, there are for sure more than just `owns`?. 

Is it actually `relationship` or `relationships`? 

Also how we capture multiple devices? 

For this kind of changes I propose to follow ECS RFC process and first create a straw project where all these discussions are happening including events/logs examples and after defined names and types next step with implementation should follow

PR #2583, opened 2026-02-11
Documentation correction of wrong event name in source and destination fields table

Ayoub-Rami:
This is only a documentation correction of the wrong event name in the first table that should be "DNS query request 2" instead of "DNS answer request 2".


trisch-me:
Closed as duplicate of https://github.com/elastic/ecs/pull/2582

PR #2594, opened 2026-03-27
docs: add guidance on object and flattened field types

### LINKED ISSUE #2595: Restrict use of open `object` typed fields in ECS schema
## Summary

ECS currently defines several fields using `type: object` with zero defined children. These fields create namespaces where any producer can define arbitrary subfields with any type, bypassing ECS's core guarantee: that field names, types, and semantics are coordinated across all data sources. We need two changes:

1. The project should give guidance on when open-ended `object` fields are acceptable and be clear that other submissions will be rejected.
2. The entity fields need to be refined before moving them beyond beta status.

## Motivation

The `entity` field set is the primary example. It introduced 5 open object containers, each replicated across 15 reuse locations. Before the entity field set, ECS had only 2 intentionally open object fields (`labels` and `container.labels`).

- `entity.attributes`
- `entity.behavior`
- `entity.lifecycle`
- `entity.metrics`
- `entity.raw`

This problem was [flagged during review](https://github.com/elastic/ecs/pull/2577#discussion_r2829543634) but we still proceeded.

If we allow this pattern to continue, creating consistent data producers and consumers will be harder than it needs to be.

### 1. Field type conflicts across integrations

If one integration defines `entity.attributes.managed` as `boolean` and another defines it as `keyword`, documents from both landing in the same data view will produce mapping conflicts. There is no schema enforcement because the subfields were never defined in ECS.

### 2. Bypasses the ECS review process

Fields like `entity.attributes.mfa_enabled`, `entity.attributes.granted_permissions`, and `entity.relationships.supervises` carry the authority of the ECS namespace but were never proposed, reviewed, or ratified as ECS fields. Individual integration teams define them independently, which is exactly the coordination problem ECS exists to solve.

### 1. Document the policy and update the RFC template

Add guidance to the schema documentation and the RFC template clarifying when `object` and `flattened` field types are acceptable, and when explicit leaf fields must be used instead. See https://github.com/elastic/ecs/pull/2594.

### 2. Address existing beta entity fields

Since `entity.attributes`, `entity.behavior`, `entity.lifecycle`, and `entity.metrics` are all still **beta**, the subfields that integrations need (e.g., `entity.attributes.mfa_enabled`, `entity.lifecycle.first_seen`, `entity.attributes.granted_permissions`) should be proposed as explicit leaf fields through the RFC process. This converts the parent objects from empty containers into structured field sets with defined children, types, and descriptions. This is the same pattern used successfully throughout the rest of ECS (e.g., `user.group` with children `.id`, `.name`, `.domain`).

The stated purpose of these fields is "advanced searching and correlation of normalized values across different providers/sources." That purpose is only achievable when the field names and types are coordinated. An open object with no defined children provides a namespace but no coordination. It shifts the type-safety problem from "undefined" to "defined independently by each integration," which is worse because it creates the appearance of compatibility without the substance.

`entity.raw` should remain as-is. It is explicitly documented as source-specific opaque data and is the correct use of an open object.

andrewkroh:
## Summary

Add documentation clarifying when `object` and `flattened` field types are
acceptable in ECS, and when explicit leaf fields must be used instead.

## Motivation

The `entity` field set introduced 5 open object containers
(`entity.attributes`, `entity.behavior`, `entity.lifecycle`,
`entity.metrics`, `entity.raw`) with zero defined children. This allows
integrations to independently define subfields with conflicting types and
semantics, bypassing the coordination that ECS exists to provide.

The pattern is being replicated (see #2577), and the concern was raised
during review but there is no documented policy to point to.

## What this establishes

An `object` with no defined children should only be used for opaque,
source-specific data with a homogeneous shape (like `labels`). A `flattened`
field should be used when arbitrary keys would cause mapping explosion (like
HTTP headers). Neither is appropriate when the subfields carry semantic
meaning. In those cases, explicit leaf fields with specified types must be
defined in the schema.

The RFC template now prompts contributors to justify any use of `object` or
`flattened` without defined children.


kgeller (approved):

rfcs/0000-rfc-template.md
```diff
@@ -25,6 +25,15 @@ Stage X: Provide a brief explanation of why the proposal is being marked as aban
 
 <!--
 Stage 1: Describe at a high level how this change affects fields. Include new or updated yml field definitions for all of the essential fields in this draft. While not exhaustive, the fields documented here should be comprehensive enough to deeply evaluate the technical considerations of this change. The goal here is to validate the technical details for all essential fields and to provide a basis for adding experimental field definitions to the schema. Use GitHub code blocks with yml syntax formatting, and add them to the corresponding RFC folder.
+
```
trisch-me, line 28: nit; it seems your change is having specific width (80?), can you make it similar to existing text style?

schemas/README.md
```diff
@@ -171,6 +171,31 @@ Supported keys to describe fields
...
+aggregation.
+
+Use of `object` with no defined children is acceptable only when the field
+represents **opaque, source-specific data** with a **homogeneous and
+well-defined shape** — for example, string keys with string values only (like
+`labels` and `container.labels`). Use of `flattened` is acceptable when the key
+names are arbitrary and unbounded, making `object` impractical due to mapping
+explosion.
+
+Neither type is appropriate when the subfields carry semantic meaning (e.g.,
```
trisch-me: I think examples should be either more generic or removed, because if someone just reads it without context of entity, the question rises - why these fields should not be mapped in the first place

schemas/README.md
```diff
@@ -171,6 +171,31 @@ Supported keys to describe fields
...
+
+A `flattened` type field indexes all leaf values as keywords regardless of their
+actual type. This avoids the mapping explosion that `object` causes when the key
+names are arbitrary (e.g., HTTP headers), but it sacrifices typed querying and
+aggregation.
+
+Use of `object` with no defined children is acceptable only when the field
+represents **opaque, source-specific data** with a **homogeneous and
+well-defined shape** — for example, string keys with string values only (like
+`labels` and `container.labels`). Use of `flattened` is acceptable when the key
```
trisch-me: we could add here also example of raw events such as `entity.raw`

schemas/README.md
```diff
@@ -171,6 +171,31 @@ Supported keys to describe fields
   - array: the content of the field should be an array (even when there's only one value).
 - beta (optional): Adds a beta marker for the field to the description. The text provided in this attribute is used as content of the beta marker in the documentation. Note that when a whole field set is marked as beta, it is not necessary nor recommended to mark all fields in the field set as beta. Beta notices should not have newlines.
 
+#### Guidance on `object` and `flattened` field types
+
+An `object` type field with no defined children creates a namespace where any
```
trisch-me: similar nit comment here about width to make it homogeneous

trisch-me:
closes #2595 
@andrewkroh do you want to have more restrictions i.e. via CI or this guidance is sufficient?

rfcs/0000-rfc-template.md
andrewkroh: Fixed in 28104e7917616eb2a895579ec204b7521daccd61

schemas/README.md
andrewkroh: Fixed in 28104e7917616eb2a895579ec204b7521daccd61

schemas/README.md
andrewkroh: 223cd79

schemas/README.md
andrewkroh: 223cd79

andrewkroh:
> do you want to have more restrictions i.e. via CI or this guidance is sufficient?

Let's start with this guidance only.

trisch-me (approved):

Issue #2597, opened 2026-03-31
RFC: extend entity schema with additional attributes, lifecycle and relationships fields, and enable entity.risk reuse


uri-weisman:
## Summary

Propose a focused 9.4 extension to the ECS `entity.*` schema.

This follows the direction established in [#2513](https://github.com/elastic/ecs/pull/2513) and consolidates with the ongoing discussion in [#2577](https://github.com/elastic/ecs/pull/2577).

The proposal has two parts:

1. Add a small set of concrete `entity.attributes.*`, `entity.lifecycle.*`, and `entity.relationships.*` leaves.
2. Enable `entity.risk.*` field reuse so risk can be represented consistently for any entity type.

### New leaves

| Field | Type | Description |
|-------|------|-------------|
| `entity.attributes.storage_class` | keyword | The storage tier or class assigned to an object storage resource. |
| `entity.attributes.mfa_enabled` | boolean | Indicates whether multi-factor authentication is enabled for this entity. |
| `entity.attributes.granted_permissions` | keyword | The set of permissions explicitly granted to this entity. These are individual action-level permissions, not roles or groups — e.g., Read, Write, Delete, Execute, Reset_Password. |
| `entity.attributes.known_redirect` | keyword | A known redirect URI or URL associated with this entity, commonly for OAuth applications or services. |
| `entity.attributes.managed` | boolean | Indicates whether the entity is managed by an external administration or control system. Applies primarily to host entity. |
| `entity.attributes.oauth_consent_restriction` | keyword | Indicates any restriction applied to OAuth consent for this entity (e.g., admin_only, verified_only, unrestricted). |
| `entity.lifecycle.last_activity` | date | The timestamp of the most recent action performed by or attributed to this entity. Distinct from `last_seen`: this implies the entity was active, not just observed. |
| `entity.relationships.owns` | keyword | Captures asset ownership — e.g., a user owning a host, email address, Windows SID, or employee ID. We want the raw data on what each integration vendor provides, on the Entity store side we will figure out how we can relate it to our own entity ids. |
| `entity.relationships.depends_on` | keyword | Identifiers of entities that this entity depends on to function. |
| `entity.relationships.supervises` | keyword | Identifiers of entities that this entity supervises, manages, or is responsible for. |
| `entity.relationships.administrators` | keyword | The list of identities (usernames, emails, or IDs) that administer this entity.|

### Field reuse enabled

- `entity.risk.*`

[...]

PR #2598, opened 2026-03-31
Add RFC for extended entity fields

uri-weisman:
## Summary
- add a 9.4 strawperson RFC proposing new `entity.attributes.*`, `entity.lifecycle.*`, and `entity.relationships.*` leaves
- document `entity.risk.*` reuse under `entity` for normalized entity-level risk
- align the RFC field descriptions with the wording captured in issue #2597


rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
+# 0000: Extend entity fields with additional attributes, lifecycle, relationships, and risk reuse
+
+
+
+- Stage: **0 (strawperson)** 
+- Date: **TBD** 
+
+
+
+This RFC proposes a focused extension to the ECS `entity.`* schema by adding a small set of concrete `entity.attributes.*`, `entity.lifecycle.*`, and `entity.relationships.*` leaves, and by enabling `entity.risk.*` reuse under `entity`. The goal is to make normalized entity data more useful for security, asset, and graph-oriented use cases without introducing a broad or underspecified object model. This proposal follows the direction established in the existing entity RFC and is intended to consolidate the ongoing discussion around relationship naming and risk placement.
```
chemamartinez: Is it going to be `relationship` or `relationships`? I see it is singular in the PR https://github.com/elastic/ecs/pull/2577

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+        "last_activity": "2026-03-31T11:42:00Z"
+      },
+      "risk": {
+        "calculated_level": "Moderate",
+        "calculated_score": 47.3,
+        "calculated_score_norm": 47.3
+      },
+      "relationships": {
+        "owns": ["host:workstation-123"],
+        "supervises": ["user:erik@okta"]
```
chemamartinez: I don't think it should be just an array of simple keywords, from most of data sources we are getting array of objects as follows

```
[
    {
        "email": "supervised1@test.com",
        "id": 1,
        "name": "user1"
    },
    {
        "email": "supervised2@test.com",
        "id": 2,
        "name": "user2"
    },
    {
        "email": "supervised3@test.com",
        "id": 3,
        "name": "user3"
    }
]
```

Same for all `relationships` fields.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| entity.attributes.oauth_consent_restriction | keyword | Indicates any restriction applied to OAuth consent for this entity (e.g., `admin_only`, `verified_only`, `unrestricted`). |
+| entity.lifecycle.last_activity              | date    | The timestamp of the most recent action performed by or attributed to this entity. Distinct from `last_seen`: this implies the entity was active, not just observed. |
+| entity.relationships.owns                   | keyword | Captures asset ownership, such as a user owning a host, email address, Windows SID, or employee ID.                       |
+| entity.relationships.depends_on             | keyword | Identifiers of entities that this entity depends on to function.                                                           |
+| entity.relationships.supervises             | keyword | Identifiers of entities that this entity supervises, manages, or is responsible for.                                       |
+| entity.relationships.administrators         | keyword | The list of identities (usernames, emails, or IDs) that administer this entity.                                            |
```
andrewkroh: Three relationship fields use outgoing verbs (`owns`, `depends_on`, `supervises`) while `administrators` is an incoming noun, making the directionality ambiguous. Either all fields should be outgoing ("this entity administers X" -> `administers`) or the naming should make direction explicit (like `administered_by`).

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+At a high level, this proposal should be additive. Producers that already emit `host.entity.*`, `user.entity.*`, `service.entity.*`, or root `entity.*` would be able to populate these leaves directly. Consumers such as entity stores, graph enrichment systems, and investigation workflows would gain a more concrete and queryable schema for ownership, dependencies, supervision, access posture, and entity-level risk.
+
+## Concerns
+
+
+
+- Should `entity.relationships.administrators` remain a multi-valued `keyword` list of identifiers, or should ECS eventually define a richer identity structure for administrator references?
+- Should `entity.attributes.known_redirect` remain singular in name while allowing multiple values, or should ECS prefer a pluralized field name?
```
andrewkroh: Per ECS guidelines ([ref](https://github.com/elastic/ecs/blob/6e0fa49259efc80377271b6e3f93dbfa17836f4c/docs/reference/ecs-guidelines.md?plain=1#L43)):

> Use singular and plural names properly to reflect the field content.

OAuth apps very often have multiple redirects, so this should be `known_redirects` with `normalize: [array]`.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Fields
+
+
+
+| Field                                       | Type    | Description                                                                                                                |
+| ------------------------------------------- | ------- | -------------------------------------------------------------------------------------------------------------------------- |
+| entity.attributes.storage_class             | keyword | The storage tier or class assigned to an object storage resource.                                                          |
+| entity.attributes.mfa_enabled               | boolean | Indicates whether multi-factor authentication is enabled for this entity.                                                  |
+| entity.attributes.granted_permissions       | keyword | The set of permissions explicitly granted to this entity. These are individual action-level permissions, not roles or groups. |
```
andrewkroh: Why the "granted" and not just "permissions"? To me the name implies a distinction from inherited or effective permissions. Do we intend this? Are `inherited_permissions` or `effective_permissions` are planned? If not, plain `permissions` is simpler.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+
+This RFC proposes a focused extension to the ECS `entity.`* schema by adding a small set of concrete `entity.attributes.*`, `entity.lifecycle.*`, and `entity.relationships.*` leaves, and by enabling `entity.risk.*` reuse under `entity`. The goal is to make normalized entity data more useful for security, asset, and graph-oriented use cases without introducing a broad or underspecified object model. This proposal follows the direction established in the existing entity RFC and is intended to consolidate the ongoing discussion around relationship naming and risk placement.
+
+## Fields
+
+
+
+| Field                                       | Type    | Description                                                                                                                |
```
andrewkroh: Can you add a column to indicate which are arrays (will have `normalize: [array]` in the schema).

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+      "lifecycle": {
+        "last_activity": "2026-03-31T11:42:00Z"
+      },
+      "risk": {
+        "calculated_level": "Moderate",
+        "calculated_score": 47.3,
+        "calculated_score_norm": 47.3
+      },
+      "relationships": {
+        "owns": ["host:workstation-123"],
```
andrewkroh: This example implies a `type:identifier` format.

- Is this format required or just illustrative?
- What constitutes the identifier -- `entity.id`, `entity.name`, a provider-specific ID?
- How should the Okta integration represent a supervision relationship? `"user:00u1234567890"`? `"user:erik@okta.com"`? Just `"00u1234567890"`?

Without a defined format, every integration will invent its own convention, defeating the purpose of normalization. This needs a documented convention, similar to how `related.ip` and `related.user` define what goes into the array.

rfcs/text/0054-extend-entity-fields.md
andrewkroh: The proposed `entity.attributes.*` fields span very different entity types:

- `mfa_enabled`  primarily users
- `storage_class`  cloud storage resources
- `managed`  hosts/devices
- `known_redirect`, `oauth_consent_restriction`  OAuth applications

When these fields are reused under `host.entity.attributes.*` and `user.entity.attributes.*`, most fields will be empty/irrelevant for any given entity type. I think the RFC should document which fields are expected to be populated for which entity types. A mapping matrix (entity type by field) would help integration developers.

And potentially this could be used to limit what fields get added into the generated schema. For example, exclude `user.entity.attributes.storage_class` from the code generation.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+At a high level, this proposal should be additive. Producers that already emit `host.entity.*`, `user.entity.*`, `service.entity.*`, or root `entity.*` would be able to populate these leaves directly. Consumers such as entity stores, graph enrichment systems, and investigation workflows would gain a more concrete and queryable schema for ownership, dependencies, supervision, access posture, and entity-level risk.
+
+## Concerns
+
+
+
+- Should `entity.relationships.administrators` remain a multi-valued `keyword` list of identifiers, or should ECS eventually define a richer identity structure for administrator references?
+- Should `entity.attributes.known_redirect` remain singular in name while allowing multiple values, or should ECS prefer a pluralized field name?
+- Once `entity.risk.*` is enabled under `entity`, should ECS explicitly recommend avoiding duplicate risk storage under entity-type-specific paths when the same score describes the normalized entity?
```
andrewkroh: Users need clear guidance on how to populate and query the fields.

The existing `risk.*` field set is reused under `host.risk.*` and `user.risk.*`. So now adding it under `entity.*` means a host entity document could have risk scores at both `host.risk.calculated_score` AND `host.entity.risk.calculated_score`.

If the entity store produces the score, does it go under `entity.risk.*`? If a host risk engine produces it, does it go under `host.risk.*`? Can they coexist? Etc...

rfcs/text/0054-extend-entity-fields.md
uri-weisman: @chemamartinez `relationships`.
cc @erikh-elastic

rfcs/text/0054-extend-entity-fields.md
uri-weisman: I'm fine with either, @erikh-elastic let's make a decision here.

rfcs/text/0054-extend-entity-fields.md
uri-weisman: You’re right. The catch is that ESQL (our solution is based on it) doesn’t support [nested mappings](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/nested), so we can’t persist that structure in the entity store indices.
What we can do is instruct users and integrations to emit this structure, and in our entity-building queries apply the prioritization we already use today: prefer email, then id, then name.
The field in the entity store will be stored as a keyword.
cc @romulets

rfcs/text/0054-extend-entity-fields.md
uri-weisman: @erikh-elastic `entity.attributes.known_redirects` it is?

rfcs/text/0054-extend-entity-fields.md
uri-weisman: I'm fine with `entity.attributes.permissions`.
cc @erikh-elastic

rfcs/text/0054-extend-entity-fields.md
uri-weisman: We create a unique identification for entities (EUID), those IDs do not exist in the source indices, of course.
It's internal logic that we prefer not to spread, as it may change in the future, therefore we created a translation layer in Kibana that will do this normalization for our stakeholders.

We should change it from a keyword to an object (as @chemamartinez suggested: https://github.com/elastic/ecs/pull/2598#discussion_r3021081054) and provide guidelines to populate as many identification fields as the user can (email, id, name), and it will help us to make this correlation ourselves rather than exposing this logic.

@romulets @erikh-elastic, what are your thoughts?

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Good question. The entity store does not compute a risk score on its own.
We use entity maintainers to enrich entities that already exist, the risk engine is one of those maintainers. We deliberately limited updates to `entity.risk` to keep the maintainer entity type agnostic
We do not mirror risk into `user.*`, `host.*`, or `service.*`, it lives under `entity.*` only.
Will add those guidelines to this proposition.

rfcs/text/0054-extend-entity-fields.md
erikh-elastic: Sounds good to me.

rfcs/text/0054-extend-entity-fields.md
erikh-elastic: Just permissions seems better. Agreed.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+This RFC proposes a focused extension to the ECS `entity.`* schema by adding a small set of concrete `entity.attributes.*`, `entity.lifecycle.*`, and `entity.relationships.*` leaves, and by enabling `entity.risk.*` reuse under `entity`. The goal is to make normalized entity data more useful for security, asset, and graph-oriented use cases without introducing a broad or underspecified object model. This proposal follows the direction established in the existing entity RFC and is intended to consolidate the ongoing discussion around relationship naming and risk placement.
+
+## Fields
+
+
+
+| Field                                       | Type    | Description                                                                                                                |
+| ------------------------------------------- | ------- | -------------------------------------------------------------------------------------------------------------------------- |
+| entity.attributes.storage_class             | keyword | The storage tier or class assigned to an object storage resource.                                                          |
```
trisch-me: will it be a enum i.e. list of possible values or free text?
Description doesn’t provide detailed information

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| entity.attributes.mfa_enabled               | boolean | Indicates whether multi-factor authentication is enabled for this entity.                                                  |
+| entity.attributes.granted_permissions       | keyword | The set of permissions explicitly granted to this entity. These are individual action-level permissions, not roles or groups. |
+| entity.attributes.known_redirect            | keyword | A known redirect URI or URL associated with this entity, commonly for OAuth applications or services.                      |
+| entity.attributes.managed                   | boolean | Indicates whether the entity is managed by an external administration or control system. Applies primarily to host entity. |
+| entity.attributes.oauth_consent_restriction | keyword | Indicates any restriction applied to OAuth consent for this entity (e.g., `admin_only`, `verified_only`, `unrestricted`). |
+| entity.lifecycle.last_activity              | date    | The timestamp of the most recent action performed by or attributed to this entity. Distinct from `last_seen`: this implies the entity was active, not just observed. |
```
trisch-me: we don’t define last_seen don’t we?

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| entity.relationships.owns                   | keyword | Captures asset ownership, such as a user owning a host, email address, Windows SID, or employee ID.                       |
```
trisch-me: what is the value, is it a string? email for example is clear, but what owing a host means, what should be written here?

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| entity.relationships.depends_on             | keyword | Identifiers of entities that this entity depends on to function.                                                           |
```
trisch-me: it’s not clear from description what means “depends on to function"

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| entity.relationships.supervises             | keyword | Identifiers of entities that this entity supervises, manages, or is responsible for.                                       |
+| entity.relationships.administrators         | keyword | The list of identities (usernames, emails, or IDs) that administer this entity.                                            |
+
+
+This RFC also proposes enabling `entity.risk.*` field reuse so the existing ECS `risk.*` fields can be nested under `entity` when the risk score describes the normalized entity rather than a specific entity-type field set.
```
trisch-me: can we extract it into it’s own issue/PR?

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+
+At a high level, this proposal should be additive. Producers that already emit `host.entity.*`, `user.entity.*`, `service.entity.*`, or root `entity.*` would be able to populate these leaves directly. Consumers such as entity stores, graph enrichment systems, and investigation workflows would gain a more concrete and queryable schema for ownership, dependencies, supervision, access posture, and entity-level risk.
+
+## Concerns
+
+
+
+- Should `entity.relationships.administrators` remain a multi-valued `keyword` list of identifiers, or should ECS eventually define a richer identity structure for administrator references?
```
trisch-me: You can define it yourself in this proposal if it’s something you need. But I don’t think we need to define administrators as separate namespace if that’s what you asked for

rfcs/text/0054-extend-entity-fields.md
uri-weisman: I discussed it with @romulets, ESQL limited nested support is blocking us from querying a list of objects.
But for `entity.relationships.*` you can publish a single object with a bunch of keyword array fields: `entity_ids`, `host_ids`, `user_ids`, `emails`, `hostnames`, `usernames`, etc.
So we stay flat and don’t need nested for ESQL.
Downside is we’re not pairing “this email goes with this id”, it’s more of a bag of identifiers we can match on. The Upside is that integrations can throw in whatever identifiers they have and we can still extend the list over time.

Example:
```
{
  "entity_id": ["entity:abc123"],
  "host_id": ["host-uuid-123", "host-uuid-456"],
  "user_id": ["00u123"],
  "email": ["supervisor@corp.example"],
  "hostname": [],
  "username": ["jsmith"]
}
```

chemamartinez: I don't have a problem from the integration side to deliver the data that way if that's the only way you can query later. It should be well documented though so developers of new integrations and people that populate these fields in the future are aware of this.

@andrewkroh what do you think about this way of storing the data without pairing? is this already happening for any other fields in ECS?

trisch-me: why would you do hostname and host_id separately? why not using host.hostname and host.id etc?

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Will be calculated by entity store log extraction process, no need to be extracted from integrations data.

rfcs/text/0054-extend-entity-fields.md
uri-weisman: No reason, will update.

chemamartinez: Another question I have with `relationships` fields, are they going to store predefined fields inside? such as `id`, `email`, `username`?

In case they don't, we might have the risk to have same data with different names in different integrations (e.g., username and name both storing an entity username). In addition, in case their content is undefined, should them be categorised as flattened?

trisch-me: The goal of this PR is actually to define all fields so there will be no `object` anymore, but detailed data structure. this should resolve problem for different integrations to understand the schema.

rfcs/text/0054-extend-entity-fields.md
trisch-me: I agree with the object here, but the object should be defined.

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Yes the object will consist of the following identifiers:
```
{
  "entity.id": ["entity:abc123"],
  "host.id": ["host-uuid-123", "host-uuid-456"],
  "user.id": ["00u123"],
  "email": ["supervisor@corp.example"],
  "host.name": [],
  "user.name": ["jsmith"]
}
```

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Same as here: https://github.com/elastic/ecs/pull/2598#discussion_r3073860027

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,146 @@
...
+| entity.attributes.managed | boolean | Host, Service | Indicates whether the entity is managed by an external administration or control system. |
+| entity.attributes.oauth_consent_restriction | keyword | Service | Restriction applied to OAuth consent for this entity (e.g. `admin_only`, `verified_only`, `unrestricted`). **Values:** integration-defined keywords unless standardized later. |
+| entity.lifecycle.last_activity | date | User, Host, Service | Timestamp of the most recent action performed by or attributed to this entity (active use). Distinct from **`entity.last_seen_timestamp`**, which records when the entity was last observed in data; `last_activity` implies the entity was active, not only seen. |
+| entity.relationships.owns | object | all | Identifiers of assets or identities this entity owns. Structure is a **flat bag of keyword arrays** (see [Relationship identifier structure](#relationship-identifier-structure)), not an array of nested objects, so storage and querying remain compatible with ESQL limitations on nested mappings. |
```
trisch-me, line 26: below there is just an example of such event. Is this example binding? If so it should be defined in this PR accordingly with names, types, examples etc. This applies to every object in this table

Also we have updated and streamlined the RFC process
https://github.com/elastic/ecs/tree/main/rfcs

Currently there is only 1 stage and requirement to create also yaml file with fields if RFC describes them

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Yes, removed from this one.
Will create a dedicated one for `entity.risk`.

rfcs/text/0054-extend-entity-fields.md
uri-weisman: Yes, it's binding as this this the structure we're expecting in the logs extraction process.
Hopes it's better defined now.
I saw the RFC process has been updated, do we expect this RFC to be updated accordingly?

chemamartinez (approved):
LGTM! My review is mostly focused on conversation's I've been involved and from an integration perspective so please wait for the rest of reviewer's approval.

erikh-elastic (approved):
LGTM

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,154 @@
...
+| entity.lifecycle.last_activity | date | User, Host, Service | Timestamp of the most recent action performed by or attributed to this entity (active use). Distinct from **`entity.last_seen_timestamp`**, which records when the entity was last observed in data; `last_activity` implies the entity was active, not only seen. |
+| entity.relationships.owns | object | all | Identifiers of assets or identities this entity owns. Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). Same rules apply to every `entity.relationships.*` in this table. |
+| entity.relationships.depends_on | object | all | Identifiers of entities this entity **requires for operation**—for example: a service depending on a database or upstream API, an application depending on an identity provider, or a workload depending on a host or cluster. Same object shape and **allowed keys** as `owns`. |
```
trisch-me:
```suggestion
| entity.relationships.depends_on | object | all | Identifiers of entities this entity **requires for operation**—for example: a service depending on a database or upstream API, an application depending on an identity provider, or a workload depending on a host or cluster. Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
```

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,154 @@
...
+| entity.attributes.managed | boolean | Host, Service | Indicates whether the entity is managed by an external administration or control system. |
+| entity.attributes.oauth_consent_restriction | keyword | Service | Restriction applied to OAuth consent for this entity (e.g. `admin_only`, `verified_only`, `unrestricted`). **Values:** integration-defined keywords unless standardized later. |
+| entity.lifecycle.last_activity | date | User, Host, Service | Timestamp of the most recent action performed by or attributed to this entity (active use). Distinct from **`entity.last_seen_timestamp`**, which records when the entity was last observed in data; `last_activity` implies the entity was active, not only seen. |
+| entity.relationships.owns | object | all | Identifiers of assets or identities this entity owns. Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). Same rules apply to every `entity.relationships.*` in this table. |
```
trisch-me:
```suggestion
| entity.relationships.owns | object | all | Identifiers of assets or identities this entity owns. Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
```

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,154 @@
...
+| entity.relationships.depends_on | object | all | Identifiers of entities this entity **requires for operation**—for example: a service depending on a database or upstream API, an application depending on an identity provider, or a workload depending on a host or cluster. Same object shape and **allowed keys** as `owns`. |
+| entity.relationships.supervises | object | all | Identifiers of entities this entity supervises, manages, or is responsible for (e.g. manager–reporting-line or org hierarchy). Same object shape and **allowed keys** as `owns`. |
```
trisch-me:
```suggestion
| entity.relationships.supervises | object | all | Identifiers of entities this entity supervises, manages, or is responsible for (e.g. manager–reporting-line or org hierarchy). Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
```

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,154 @@
...
+| entity.relationships.administered_by | object | all | Identifiers of identities that administer this entity (incoming relationship: “who administers this entity”). Same object shape and **allowed keys** as `owns`. |
```
trisch-me:
```suggestion
| entity.relationships.administered_by | object | all | Identifiers of identities that administer this entity (incoming relationship: “who administers this entity”). Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
```

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,154 @@
...
+
+_Indicative only; specific integrations may justify exceptions. **all** under Applied Entity Type means User, Host, or Service whenever the relationship applies to that normalized entity._
```
trisch-me, line 36: as of now we don’t have a built-in option to specify that some of the parameters applies only for specific entity types. This should be added verbatim to every field as additional note

trisch-me (approved):
lgtm in general with some minor feedback

rfcs/text/0054-extend-entity-fields.md
trisch-me: `risk.*` fields are not a part of this PR anynore

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,156 @@
...
+
+| Field | Type | Applied Entity Type | Description |
+| ----- | ---- | ------------------- | ----------- |
+| entity.attributes.storage_class | keyword | Service | The storage tier or class assigned to an object storage resource (e.g. S3/GCS/Azure object tiers). **Values:** free-text or provider-specific labels unless/until ECS defines `allowed_values`; common examples include `STANDARD`, `STANDARD_IA`, `GLACIER`, `COLDLINE` (not an exhaustive enum in this RFC). |
+| entity.attributes.mfa_enabled | boolean | User | Indicates whether multi-factor authentication is enabled for this entity. |
+| entity.attributes.permissions | keyword (list) | User, Host, Service | Action-level permissions associated with this entity (not roles or groups). Prefer this name over `granted_permissions` unless a future RFC introduces parallel fields for inherited or effective permissions. |
+| entity.attributes.known_redirects | keyword (list) | Service | Known redirect URIs or URLs associated with this entity, commonly for OAuth applications or services. Plural name reflects ECS guidance to use singular/plural names to reflect field content. |
```
trisch-me:
```suggestion
| entity.attributes.known_redirects | keyword (list) | Service | Known redirect URIs or URLs associated with this entity, commonly for OAuth applications or services. |
```

why do we need this sentence?

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -31,7 +31,7 @@ Where a field is multi-valued in ECS, the type is written as **`keyword (list)`*
...
 | entity.relationships.supervises | object | all | Identifiers of entities this entity supervises, manages, or is responsible for (e.g. manager–reporting-line or org hierarchy). Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
-| entity.relationships.administered_by | object | all | Identifiers of identities that administer this entity (incoming relationship: “who administers this entity”). Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
+| entity.relationships.administers | object | all | Identifiers of entities that this entity administers (for example managed hosts, delegated accounts, or administered resources). Value **must** be an object whose keys are those defined in [Allowed keys on `entity.relationships.*` objects](#allowed-keys-on-entityrelationships-objects). |
```
trisch-me, line 35: I don’t quite understand how this differs from `supervises`?

uri-weisman: **Supervises** applies to user entities and captures management hierarchy: who reports to whom as people - “who is whose manager in the org?” It’s about org chain of command, not technical privileges.

**Administers** means this user is an administrator of something technical, for example Entra ID, Microsoft 365, or a GCP organization, not merely “uses” those systems with normal permissions.


@erikh-elastic — does this match your understanding, or should we adjust anything?

trisch-me: from my pov it’s very subtle difference and it’s not clear from description. Does it mean first is only applicable to user entities and second to all? i can’t tell the difference from the field name and description doesn’t help much. After your explanation it’s a bit better but still not 100% clear. Can we improve it?

chemamartinez: I just want to highlight that changing the direction of the field (from `identities that administer this entity` to `entities that this entity administers`) at this point, it also affects the changes that we already made to the integrations to support this field.

They have not been merged yet, but it will require extra effort to support a new field, that could also lead to adding changes to the Filebeat input and wait for a new release to get it working.

erikh-elastic: Yes, thank you. @uri-weisman! That is exactly it.

From a security product perspective "Supervises" versus "Administrator" are completely different things.

Supervises is a manager to employee type relationship versus administrator is a user to a host or service or etc type relationship.

uri-weisman: @trisch-me
I’ll tighten up the description and will add examples that highlight supervises as a user→user relationship and administers is user→service/host, but I’d rather keep the description a bit open-ended regarding types so we’re not tied only to those.

@chemamartinez Totally agree, and thanks for surfacing it. We want this data to be valuable for customers. Thinking it through again, the passive direction (“identities that administer this entity”) makes less sense, so If reversing the direction turns into a heavy lift, we’re good deferring that to the next iteration.

rfcs/text/0054-extend-entity-fields.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Allowed keys on `entity.relationships.*` objects
+
+Values in each array must follow the same conventions as the corresponding ECS field (e.g. `user.email` values are email-shaped strings; `user.domain` values follow ECS directory or AD/LDAP domain name semantics; `host.id` values follow host id semantics from ECS).
+
+| Property key (JSON) | ECS type | `normalize: [array]` | Semantics |
+| ------------------- | -------- | --------------------- | ----------- |
+| `entity.id` | keyword | yes | Identifiers of referenced entities, using the same meaning as root `entity.id` (stable id for correlation within scope). |
+| `host.id` | keyword | yes | Referenced host ids. |
+| `host.name` | keyword | yes | Referenced host names. |
```
chemamartinez, line 52: Does it make sense to add here host.domain as well?

uri-weisman: We can add, for the user broader context, but we don't utilize `host.domain` in the entity store.
Identity ranking for hosts is `host.id` → `host.name` → `host.hostname` only.
`host.domain` is not in that ranking.

uri-weisman:
@andrewkroh can I please ask you to have a second look?

uri-weisman:
We decided to wait with follow up PRs that utilize the fields defined in this RFC until it's merged

Can I get another reviewer with write access to be able to merge this one? @trisch-me asking for your assistance.

trisch-me:
@taylor-swanson @andrewkroh could you check this out?

bhapas (approved):
LGTM

taylor-swanson (approved):

uri-weisman:
Thanks everyone, @trisch-me I need your assistance with pushing this one?

trisch-me:
@uri-weisman can you create a follow up PR with fields put into yaml files? Or do you need help with it?

uri-weisman:
> @uri-weisman can you create a follow up PR with fields put into yaml files? Or do you need help with it?

Will create one, thanks!

narph:
> > @uri-weisman can you create a follow up PR with fields put into yaml files? Or do you need help with it?
> 
> Will create one, thanks!

@uri-weisman Is the PR already in? We want to make sure the ECS fields are added in before we make changes to our integrations, so we can reduce future friction or potential breaking changes for users.

uri-weisman:
Hey @narph, IIUC @trisch-me is working on introducing the follow up PR - [ticket](https://github.com/elastic/ecs/issues/2645).

trisch-me:
hey @narph - the PR is there, but actual fields are not yet in ECS, I was going to introduce them, but there is a technical question I need to solve first, I hope it will be ready soon

taylor-swanson:
> We want to make sure the ECS fields are added in before we make changes to our integrations, so we can reduce future friction or potential breaking changes for users.

@narph, just as a heads up, the dynamic ECS templates in elasticsearch won't be updated until the next stack release, so any new fields that don't already match the dynamic template won't map properly if relying on that method. Looking at the list of fields, `entity.lifecycle.last_activity (date)` is one of those fields. The field can be mapped statically through `external: ecs` as long as the ECS reference in the integration is updated to the next ECS release that includes these fields.

narph:
> > We want to make sure the ECS fields are added in before we make changes to our integrations, so we can reduce future friction or potential breaking changes for users.
> 
> @narph, just as a heads up, the dynamic ECS templates in elasticsearch won't be updated until the next stack release, so any new fields that don't already match the dynamic template won't map properly if relying on that method. Looking at the list of fields, `entity.lifecycle.last_activity (date)` is one of those fields. The field can be mapped statically through `external: ecs` as long as the ECS reference in the integration is updated to the next ECS release that includes these fields.

cc @chemamartinez

PR #2599, opened 2026-04-03
Add  field maturity and remove experimental pipeline

kgeller:
## Summary

- Add `alpha` as a first-class schema attribute mirroring `beta` (validation, reuse propagation, doc rendering), with mutual exclusion enforced
- Remove the `experimental/` build pipeline, directory, and unused `cgroup.*` fields — zero adoption, `container.*` already covers the use case
- Document field stability definitions (alpha, beta, GA) and promotion criteria in `ecs-principles-design.md`
- Update authoring docs in `schemas/README.md`, `USAGE.md`, and internal developer docs

This is the first of three PRs: (1) alpha maturity + remove experimental, (2) [streamline RFC process](https://github.com/elastic/ecs/pull/2600), (3) [annotate in-flight RFCs](https://github.com/elastic/ecs/pull/2601).

## Test plan

- [x] `make test` — 153 tests pass
- [x] `make generate` — clean build, no diff
- [x] No stale `experimental` references in `.py`, `.yml`, `.j2`, or `Makefile`
- [x] Beta fieldset (`device`) renders correctly — no regression
- [x] Manually added `alpha` to `agent.version`, confirmed italic marker renders in generated docs, then reverted

### DIFF
--- a/.github/scripts/ecs_release/changelog.py
+++ b/.github/scripts/ecs_release/changelog.py
@@ -1,0 +1,159 @@
+"""Changelog manipulation for ECS releases.
+
+Operates on ``CHANGELOG.next.md`` using only the Python standard library.
+"""
+
+import re
+from pathlib import Path
+
+NEXT = "CHANGELOG.next.md"
+
+_TEMPLATES_DIR = Path(__file__).parent / "templates"
+
+
+def _load_template(name: str) -> str:
+    return (_TEMPLATES_DIR / name).read_text()
+
+
+_PR_RE = re.compile(r"#(\d+)\s*$")
+
+
+def _pr_sort_key(line: str) -> int:
+    m = _PR_RE.search(line)
+    return int(m.group(1)) if m else 0
+
+
+# ── Internal helpers ───────────────────────────────────────────────
+
+
+def _has_bullet_entries(lines: list[str]) -> bool:
+    """Return True if any line in *lines* is a ``*`` bullet entry."""
+    return any(line.strip().startswith("*") for line in lines)
+
+
+def _remove_empty_sections(text: str) -> str:
+    """Remove ``###`` / ``####`` sections that contain no bullet entries.
+
+    Parses the markdown into (heading, body) pairs, then reassembles
+    only those sections whose body contains at least one ``*`` entry.
+    ``##`` headings are always preserved.
+    """
+    sections: list[tuple[str, list[str]]] = []
+    current_heading: str | None = None
+    current_body: list[str] = []
+
+    for line in text.split("\n"):
+        if line.startswith("## ") or line.startswith("### ") or line.startswith("#### "):
+            if current_heading is not None or current_body:
+                sections.append((current_heading or "", current_body))
+            current_heading = line
+            current_body = []
+        else:
+            current_body.append(line)
+
+    sections.append((current_heading or "", current_body))
+
+    result: list[str] = []
+    # Track whether the parent ### had any kept #### children so we
+    # can suppress empty ### groups too.
+    pending_h3: str | None = None
+    h3_has_content = False
+
+    for heading, body in sections:
+        if heading.startswith("## ") and not heading.startswith("### "):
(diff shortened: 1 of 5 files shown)


docs/reference/ecs-principles-design.md
```diff
@@ -37,6 +37,24 @@ Examples:
...
+
+**Beta.** A beta field has reached a reasonable level of stability and is expected to become part of the stable schema, but has not yet received a final commitment. The field name and type are unlikely to change, though adjustments to allowed values, descriptions, or minor semantic refinements may still occur. Beta fields are suitable for broader adoption, including dashboards, detection rules, and integrations, with the understanding that minor migration effort may occasionally be required. Breaking changes to beta fields are not expected but may still occur.
+
+**GA (stable).** The default. GA fields are backwards-compatible; breaking changes are not made outside of major versions.
+
+### Criteria for each level [_field_stability_criteria]
+
+These criteria give schema authors and reviewers a shared rubric for assigning maturity:
+
+* **Alpha when:** field definitions are reasonably complete but may still evolve; usage patterns are not yet well understood; field definitions have not yet been validated against real-world data; open questions remain.
```
trisch-me:
```suggestion
* **Alpha when:** field definitions are reasonably complete but may still evolve; usage patterns are not yet well understood; field definitions have not yet been completely validated against real-world data; open questions remain.
```

I think we should expect it to be validated but without guarantees

schemas/README.md
```diff
@@ -169,7 +176,8 @@ Supported keys to describe fields
...
 - normalize: Normalization steps that should be applied at ingestion time. Supported values:
   - array: the content of the field should be an array (even when there's only one value).
-- beta (optional): Adds a beta marker for the field to the description. The text provided in this attribute is used as content of the beta marker in the documentation. Note that when a whole field set is marked as beta, it is not necessary nor recommended to mark all fields in the field set as beta. Beta notices should not have newlines.
+- alpha (optional): Adds an alpha marker for the field. The text provided in this attribute is used as content of the alpha marker in the documentation. Alpha notices should not have newlines. A field cannot have both `alpha` and `beta`.
+- beta (optional): Adds a beta marker for the field. The text provided in this attribute is used as content of the beta marker in the documentation. Note that when a whole field set is marked as beta, it is not necessary nor recommended to mark all fields in the field set as beta. Beta notices should not have newlines. A field cannot have both `alpha` and `beta`.
```
trisch-me, line 180: should we have similar remark regarding fieldset for alpha?

kgeller: @trisch-me which remark? I they have all the same warnings?

trisch-me:
one thing that came to my mind - in our docs we don’t have a prominent beta (or alpha) sign, we only have `This field is beta and subject to change. `

Maybe we should add a tag similar to what we do for otel such as related? and put it near the name so that person see at once the maturity status. Or even use a new column for maturity. wdyt?

kgeller:
> one thing that came to my mind - in our docs we don’t have a prominent beta (or alpha) sign, we only have `This field is beta and subject to change. `
> 
> Maybe we should add a tag similar to what we do for otel such as related? and put it near the name so that person see at once the maturity status. Or even use a new column for maturity. wdyt?

@trisch-me we can consult with the docs team on this as a followup? It definitely sounds nice to make it clearer, but I think it would be great to get their take on it to make sure we can match existing / define the convention

trisch-me:
sure, do you know who is contact person for this? :)

schemas/README.md
trisch-me: `Note that when a whole field set is marked as beta, it is not necessary nor recommended to mark all fields in the field set as beta.`
this one

kgeller:
> sure, do you know who is contact person for this? :)

I'm not sure, my plan was to post in their slack channel 😄

schemas/README.md
kgeller: ahh good catch, updated

PR #2623, opened 2026-04-28
feat: add ECS MCP server

taylor-swanson:
- Add Go code to repo under `go-ecs`
- Add new ECS MCP server


go-ecs/cmd/ecs-mcp/main.go
```diff
@@ -0,0 +1,296 @@
...
+	modVer, vcsRef := getVersion()
+	mcpSrv := mcp.NewServer(&mcp.Implementation{
+		Name:    "ecs-mcp",
+		Version: modVer + "(" + vcsRef + ")",
+	}, nil)
+	ecsmcp.AddTools(mcpSrv, store.DDL, db)
+	ecsmcp.AddPrompts(mcpSrv)
+
+	if listen != "" {
+		var handler http.Handler = mcp.NewStreamableHTTPHandler(func(r *http.Request) *mcp.Server { return mcpSrv }, nil)
```
andrewkroh: Do you need stateful sessions? Otherwise, I'd recommend passing in `&mcp.StreamableHTTPOptions{Stateless: true}` instead of `nil`.

taylor-swanson: Not that I'm aware of, good catch.

go-ecs/internal/store/queries.go
```diff
@@ -0,0 +1,216 @@
...
+	if limit <= 0 {
+		limit = 25
+	}
+
+	//Normalize the query for FTS5 matching:
+	//1. Sanitize (replace dots with spaces)
+	//2. Split camelCase tokens (e.g. "ProcessTTYAttached" → "Process TTY Attached")
+	//3. Join plain terms with OR for additive discovery ranking
+	sanitized := sanitizeFTSQuery(query)
+	sanitized = splitCamelCase(query)
```
qcorporation, line 91: @taylor-swanson should you be feeding in the `sanitized` result back into the `splitCamelCase`

Similarly for the line below `implicitOR(sanitized)`

go-ecs/internal/store/queries.go
```diff
@@ -0,0 +1,216 @@
...
+	var results []SearchResult
+	for rows.Next() {
+		var result SearchResult
+		if err = rows.Scan(&result.Name, &result.Type, &result.Description, &result.IsArray); err != nil {
+			return nil, err
+		}
+		results = append(results, result)
+	}
+	if rows.Err() != nil {
+		return nil, err
```
qcorporation, line 116: should it be 
```
if err = rows.Err(); err != nil {
    return nil, err
}
``` 

?

qcorporation (commented):
Two fundamental questions:
- should we have any CI tests for future contributions?
- ecs predominately is generally written in python - this is a new toolchain. I know this is a bit late but do we want to introduce this toolchain into the ecs repo?

- Should we be adding changelog

go-ecs/internal/store/queries.go
taylor-swanson: Good catch, I'll fix that.

taylor-swanson:
> * should we have any CI tests for future contributions?

Yeah this was an oversight on my part... can certainly add tests and getting that running in CI is trivial.

> * Should we be adding changelog

I suppose this is tooling, but it's not used in the process of ECS itself... so it sits somewhere else I feel. More on this below.

> * ecs predominately is generally written in python - this is a new toolchain. I know this is a bit late but do we want to introduce this toolchain into the ecs repo?

This isn't related to the tools in `scripts/`, as those are specifically for building ECS. I'm not opposed to having a separate repo, as I get that we don't want to mix usages here (and I suppose toolchains, although I would argue that Go is probably one of the most lightweight toolchains out there). 

Would something like an `ecs-tools` or `go-ecs` repo make more sense? I have other things I want to add that are already written in Go.

taylor-swanson:
`go-ecs` is already taken, so `ecs-tools` it is. I'll move the go code over to a new public elastic repo called `elastic/ecs-tools` and close this PR. I'll apply the fixes for the above issues and add some tests where appropriate.

I think having it separate is a better idea anyway, since while the tool itself consumes ECS, it is not ECS nor does it build ECS (if that makes sense).

taylor-swanson:
New PR here in the elastic/ecs-tools repo: https://github.com/elastic/ecs-tools/pull/1

PR #2625, opened 2026-04-30
[Agents] Add PR triage agent workflow

bhapas:
This PR adds a github actions workflow when a new PR is created / reopened then a agentic triage workflow runs and adds a triage comment related to the PR.


trisch-me (approved):

.github/workflows/pr-triage.yml
```diff
@@ -0,0 +1,334 @@
...
+          RUNTIME="$GITHUB_WORKSPACE/pr-triage-runtime"
+          PROMPT_FILE="$RUNTIME/prompt.md"
+          REPORT_FILE="$RUNTIME/pr-triage-report.md"
+          TRANSCRIPT_FILE="$RUNTIME/agent-transcript.md"
+          export HOME="$RUNTIME/home"
+          mkdir -p "$RUNTIME" "$HOME"
+
+          {
+            cat <<EOF
+          # ECS PR Triage (GitHub Actions)
```
kgeller, line 133: should we extract this to a skill in the repo 1) so this file can be slightly more compact and 2) so contributors could run it themselves if they wanted to?

bhapas: This is a prompt to the triage agent , not a skill though.. But this prompt has the variables , so not sure extracting this would help either..

kgeller: that makes sense, thanks for explaining

kgeller (approved):

PR #2584, opened 2026-02-17
Increase total fields limit from 2000 to 2500

Mpdreamz:
# ECS 9.3.0 Specification Analysis

## Total Field Count

| Version | Fields | Multi-fields | Total |
|---------|--------|--------------|-------|
| v9.3.0 | 2,141 | 123 | **2,264** |
| v9.0.0 | 1,848 | 91 | **1,939** |
| **Delta** | **+293** | **+32** | **+325 (16.8%)** |

## Special Field Types

| Type | v9.0.0 | v9.3.0 | Change |
|------|--------|--------|--------|
| `object` (container/grouping) | 15 | 90 | +75 (+500%) | | `nested` (array-type) | 25 | 28 | +3 |
| Fields with `multi_fields` | 91 | 123 | +32 |

## New Field Sets in v9.3.0

| Field Set | Fields |
|-----------|--------|
| `entity` | 26 |
| `gen_ai` | 26 |
| **Total** | **52** |

## Field Sets with Significant Growth

| Field Set | v9.0.0 | v9.3.0 | Growth | % Change | 
|-----------|--------|--------|--------|----------| 
| `user` | 46 | 116 | +70 | +152% |
| `host` | 42 | 110 | +68 | +162% |
| `cloud` | 33 | 72 | +39 | +118% |
| `service` | 33 | 72 | +39 | +118% |
| `orchestrator` | 15 | 28 | +13 | +87% |
| `device` | 5 | 10 | +5 | +100% |
| `process` | 389 | 391 | +2 | +0.5% |
| `threat` | 435 | 437 | +2 | +0.5% |

## Largest Field Sets (v9.3.0)

| Field Set | Fields | % of Total |
|-----------|--------|------------|
| `threat` | 437 | 19.3% |
| `process` | 391 | 17.3% |
| `file` | 144 | 6.4% |
| `user` | 116 | 5.1% |
| `host` | 110 | 4.9% |
| `tls` | 77 | 3.4% |
| `cloud` | 72 | 3.2% |
| `service` | 72 | 3.2% |
| `dll` | 46 | 2.0% |
| `observer` | 41 | 1.8% |

## Summary

- **54 total field sets**: 2 new, 6 with growth, 46 unchanged
- **Growth drivers**: new field sets (`entity`, `gen_ai`) account for 52 fields; existing field set expansions (`user`, `host`, `cloud`, `service`, `orchestrator`, `device`) account for 241 fields; 32 new multi-field variants
[...]

PR #2648, opened 2026-05-11
added new fieldset and new fields to entity set as per RFC 0054

trisch-me:
#### 1. What does this PR do?

Fields addition as per https://github.com/elastic/ecs/pull/2598/

---

#### Commit Message
Add `entity.attributes.*`, `entity.lifecycle.last_activity`, and `entity.relationships.*` fields from RFC 0054. Introduce reusable `entity_reference` field set for shared relationship identifier keys.

FYI @uri-weisman @taylor-swanson


trisch-me:
run docs-build

scripts/templates/fieldset.j2
```diff
@@ -16,7 +18,7 @@ applies_to:
 {# Fieldset label alpha header -#}
 {% if fieldset['alpha'] -%}
 
-::::{caution}
+::::{warning}
```
taylor-swanson, line 21: Was `caution` an invalid symbol to use here?

taylor-swanson (commented):
Change itself seems fine to me, but this does introduce another ~500+ fields all said and done. Do we need to adjust the composable field limit that was [recently changed again?](https://github.com/elastic/ecs/pull/2584)

(EDIT: it's 615 fields added, `main` has 1996 entries in ecs_flat.yml, this change has 2611)

scripts/templates/fieldset.j2
shainaraskas: [yes](https://elastic.github.io/docs-builder/syntax/admonitions/) - we don't support a `caution` type

trisch-me: yes, it’s not supported anymore (there was a lint warning)

taylor-swanson (approved):

PR #2601, opened 2026-04-03
Annotate 34 in-flight RFCs with process transition status notes

kgeller:
## Summary

- Add a process transition status note to each of the 34 RFCs that were still in-flight when the multi-stage RFC process was retired
- Each note reflects the actual schema state: merged as **GA**, merged as **beta**, partially merged, actively being implemented, or never merged
- 15 finished (stage 3) and 4 already-abandoned (stage X) RFCs are left untouched
- Stage lines are preserved as-is in every RFC for historical record

This is the third of three PRs: (1) [alpha maturity + remove experimental](https://github.com/elastic/ecs/pull/2599), (2) [streamline RFC process](https://github.com/elastic/ecs/pull/2600), (3) annotate in-flight RFCs.

PR #2655, opened 2026-05-14
[RFC 0055] Propose vulnerability.status as a constrained keyword

kcreddy:
#### 1. What does this PR do?

Adds a new Proposal RFC (0055) for `vulnerability.status`, a constrained keyword field with four lifecycle values: `open`, `fixed`, `reopened`, `unknown`. This is a fresh submission under the current single-stage Proposal process, replacing the inactive RFC 0045 which was retired along with the multi-stage process (#2601).

Ref: elastic/security-team#17210

- `vulnerability.status` (new) — Lifecycle state of a vulnerability finding on an asset.
  - Constrained to four values: `open` (present), `fixed` (remediated), `reopened` (reappeared after fix), `unknown` (cannot determine).
  - Target maturity: **beta**.

The CDR Vulnerability Findings model assumes every document in `vulnerability_latest` is an open finding. When integrations consume incremental/delta APIs (e.g. Microsoft Defender's `SoftwareVulnerabilityChangesByMachine`), remediated records flow into the index and appear as open vulnerabilities on the Findings page.

This field lets the Findings page default to `NOT vulnerability.status: fixed`, matching how `result.evaluation` (`passed`/`failed`/`unknown`) works for misconfiguration findings.

The four values are well-supported across vendors:

- `open`: MS Defender New/Updated, Qualys New/Active, Tenable OPEN, CrowdStrike open, Wiz OPEN
- `fixed`: MS Defender Fixed, Qualys Fixed, Tenable FIXED, CrowdStrike closed, Wiz RESOLVED
- `reopened`: Qualys Re-Opened, Tenable REOPENED, CrowdStrike reopen
- `unknown`: fallback for full-snapshot integrations with no lifecycle

Triage dispositions (false_positive, risk_accepted, etc.) are intentionally excluded — they describe human decisions, not whether the vulnerability is present. CrowdStrike already separates these via `suppression_info.reason`.

**How full-snapshot integrations work:** Full-snapshot APIs return the complete list of currently open vulnerabilities on each poll. Fixed vulnerabilities simply stop appearing in subsequent snapshots — the `vulnerability_latest` transform's 90-day retention policy eventually evicts the stale record. These integrations never produce `fixed` records, so the field is set to `open`. If a legacy integration hasn't been updated and omits the field entirely, the Findings page filter (`NOT vulnerability.status: fixed`) still displays the record correctly since the field is absent rather than set to `fixed`.

N/A — documentation is within the RFC itself and the schema YAML (`rfcs/text/0055/vulnerability.yml`).

NO — this PR adds only an RFC and proposed schema YAML, no changes to `schemas/` or `generated/`.

NO — RFC-only change, no schema or code modifications.

- This replaces the inactive RFC 0045, which proposed 20+ vulnerability fields and was never advanced. This proposal is narrowly scoped to the one field with immediate demand from CDR. The remaining fields from RFC 0045 can be submitted as separate proposals if needed.
- Target maturity is **beta** rather than alpha. The field design is stable — four values derived from five major vendors, mirroring the proven `result.evaluation` pattern — and will be used immediately on the Kibana Vulnerability Findings page. 

---

#### Commit Message

```
[RFC 0055] Propose vulnerability.status as a constrained keyword

Add a new Proposal RFC for vulnerability.status with four lifecycle
values: open, fixed, reopened, unknown. This is a fresh submission
under the current single-stage Proposal process, replacing the
inactive RFC 0045 which was retired with the multi-stage process
(#2601).

The field enables the CDR Vulnerability Findings page to filter out
remediated findings by default, matching how result.evaluation works
for misconfiguration findings. The four values are supported across
Microsoft Defender, Qualys, Tenable, CrowdStrike, and Wiz.

Target maturity is beta — the design is stable and the field will be
used on the Findings page immediately upon adoption.

Ref: elastic/security-team#17210
```


rfcs/text/0055/vulnerability.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: vulnerability
+  fields:
```
taylor-swanson, line 3: Please add a list of `allowed_values`, similar to how `event.outcome` works:
https://github.com/elastic/ecs/blob/main/schemas/event.yml#L431

kcreddy: Updated in [e5c7964](https://github.com/elastic/ecs/pull/2655/commits/e5c79647d3d59bf0f74c7ab39a7285983d0537ea)

rfcs/text/0055-vulnerability-status.md
```diff
@@ -0,0 +1,147 @@
+# 0055: Vulnerability Status Field
+
+- Stage: **Proposal**
+- Date: **TBD**
+- Target maturity: **alpha**
```
kcreddy: @taylor-swanson, 
From the new [single-stage process](https://github.com/elastic/ecs/blob/main/rfcs/PROCESS.md) it seems that the contributor can decide if the field can be proposed at maturity `alpha/beta`. 
> The contributor specifies a target maturity of alpha or beta for the proposed fields. See [Field stability](https://github.com/elastic/ecs/blob/main/docs/reference/ecs-principles-design.md#_field_stability) for definitions.

The field requirement is: https://github.com/elastic/security-team/issues/17210, and is going to be used inside [`Vulnerability Findings`](https://www.elastic.co/docs/solutions/security/cloud/findings-page-3) page. Do you have any concern if we directly adopt it at level `beta` instead of `alpha`? This is because as per [Field stability](https://github.com/elastic/ecs/blob/main/docs/reference/ecs-principles-design.md#_field_stability) guide the `alpha` fields are suggested not to use for production workloads, and long-live dashboards.

If it can only be released at `alpha`, then how long can we expect it to be moved into `beta`, and what would it take (more adoption across integrations?)

----

@maxcold, do you have any concern with adopting the field directly at `beta`? I'm trying to avoid bad UX for the users until the field is moved from `alpha` into `beta`, and only then it can be used inside the Findings page.

trisch-me: hey @kcreddy, both alpha and beta fields are not production ready. Being alpha is also not a prerequisite for becoming a beta. Some fields if they are well defined or maybe already have been used in production can be added directly as beta fields.

The main reason for alpha fields is to allow folks to have some exploration with the fields already defined in schema, but they could be changed or deleted later if initial idea and design weren't good enough.

so if in your opinion this field is mature enough (used already for some time) you can go with beta

rfcs/text/0055-vulnerability-status.md
```diff
@@ -0,0 +1,147 @@
...
+
+The CDR Vulnerability Findings model assumes every document in the `vulnerability_latest` index is an open finding. When integrations consume incremental/delta APIs, remediated records flow into the index and appear as open. This field lets the Kibana Findings page filter them out by default.
+
+## Usage
+
+The primary consumer is the Kibana Vulnerability Findings page. With this field the page can default to `NOT vulnerability.status: fixed`, hiding remediated findings. This mirrors how `result.evaluation` (`passed`/`failed`/`unknown`) is used on the Misconfiguration Findings page.
+
+**Incremental APIs** (e.g. Microsoft Defender for Endpoint [`SoftwareVulnerabilityChangesByMachine`](https://learn.microsoft.com/en-us/defender-endpoint/api/get-assessment-software-vulnerabilities)) provide change events with an explicit status. The integration maps vendor values to `vulnerability.status` in the ingest pipeline.
+
+**Full-snapshot APIs** produce a complete list of open vulnerabilities on each poll. The field defaults to `open` or is omitted — existing behaviour is preserved.
```
trisch-me: I don’t understand how it suppose to work. If every snapshot defines only open vulnerabilities, who will set them to fixed? And why one should omit it?

rfcs/text/0055-vulnerability-status.md
kcreddy: Thanks @trisch-me for the clarification. I'd like to move this to **beta**.

Since the field design isn't exploratory — the four values are directly derived from what five major vendors already ship (Microsoft Defender, Qualys, Tenable, CrowdStrike, Wiz), and the pattern mirrors `result.evaluation` which has been stable in production for misconfiguration findings for a long time.

The immediate consumer is the Kibana Vulnerability Findings page, which needs to filter on this field by default. 

The field hasn't been used in integrations yet, but that's because we're blocked on it being defined first (elastic/security-team#17210). The design confidence is high; I don't expect the values to change.

I'll update the PR to target `beta`.

rfcs/text/0055-vulnerability-status.md
kcreddy: Good question. For full-snapshot integrations:

- Each poll cycle returns the **complete list** of currently open vulnerabilities. Vulnerabilities that have been fixed simply stop appearing in subsequent snapshots.
- We then rely on `vulnerability_latest` transform which uses a 90-day retention policy — when a vulnerability stops appearing, its record eventually ages out of the destination index.
- These integrations don't produce `fixed` records at all. Setting the field to `open` is the natural fit. "Omit" is acceptable for backward compatibility — existing integrations that haven't been updated yet will have no `vulnerability.status` field, and the Findings page filter (`NOT vulnerability.status: fixed`) still shows those records correctly since the field is absent, not set to `fixed`.

In short: for full-snapshot sources the lifecycle is implicit (present = open, absent = gone). The field matters most for incremental/delta APIs where `fixed` records explicitly flow into the index and need to be filtered out.

I'll clarify this in the RFC text.

kcreddy: Done: [e3ef42b](https://github.com/elastic/ecs/pull/2655/commits/e3ef42b66c04b59f2895643232fed62b1be38793)

rfcs/text/0055-vulnerability-status.md
kcreddy: Updated to `beta` in the commit: [e3ef42b](https://github.com/elastic/ecs/pull/2655/commits/e3ef42b66c04b59f2895643232fed62b1be38793) and also updated the PR description/commit message

maxcold: I have no concerns with the field going directly to beta

kcreddy:
@trisch-me @taylor-swanson can you please review this and let me know if you have any other concerns merging it?

taylor-swanson (approved):
Changes look good to me.

rfcs/text/0055-vulnerability-status.md
taylor-swanson: I'll defer this to @trisch-me

trisch-me (approved):

rfcs/text/0055-vulnerability-status.md
trisch-me: > We then rely on vulnerability_latest transform which uses a 90-day retention policy — when a vulnerability stops appearing, its record eventually ages out of the destination index.

does it mean we see already closed vulnerability for 90 days?

kcreddy: > > We then rely on vulnerability_latest transform which uses a 90-day retention policy — when a vulnerability stops appearing, its record eventually ages out of the destination index.

> does it mean we see already closed vulnerability for 90 days?

yes, there is a 90-day staleness window. That existing records then ages out after 90 days via the retention policy.

Issue #2340, opened 2024-05-07
[Discuss] Add `agent.group` and `host.group` field 


adrianchen-es:
Summary

Add `agent.group`,`host.group` field to facilitate searching for a group of agents/hosts based on a user-configurable arbitrary grouping.

Motivation:
These arbitrary user-configurable fields will improve the search and analysis experience for Security and Observability. 

Security: 
- Find alerts and events based on `agent.group` or `host.group`
- Create custom alerts to target a group of hosts

Observability:
- Find alerts and metrtics based on `agent.group` or `host.group`
- Create custom alerts to target a group of hosts

ABAC:
- Grant access to a user group based on `agent.group` or `host.group`

Once added as an ECS field, it should be included in the [Elastic Security fields](https://www.elastic.co/guide/en/security/current/siem-field-reference.html).

Currently, users need to create a custom field to facilitate this type of group. 

Detailed Design:
The `agent.group` or `host.group` field could treated in a manner similar 

Setting a field type of wildcard would match the url.full and url.original fields. A .text multi-field could be added.
* Field names
- `agent.group`
- `host.group`

* Example values for the fields
-`APP-A_PROD`
-`APP-A_DEV`
-`APP-B_DEV`

* Suggested appropriate datatypes
Type: keyword

* Any example events that map to the proposed use case(s)

Event A from host001 in the APP-A_PROD group
```
{
  "_index": ".ds-metrics-system.process.summary-default-2024.05.03-000042",
  "_id": "FVwmT7YoA8jJjD7HAAABj0PMPl0",
  "_version": 1,
  "_score": 0,
  "_source": {
    "@timestamp": "2024-05-04T13:30:09.885Z",
    "agent": {
      "ephemeral_id": "85174b4b-4415-46c4-89ce-160e27a56bc5",
      "id": "a8435320-1d60-49f7-b1e7-25ed5d5eb59c",
      "name": "host001",
      "group": "APP-A_PROD",
      "type": "metricbeat",
      "version": "8.13.3"
    },
    "data_stream": {
      "dataset": "system.process.summary",
      "namespace": "default",
      "type": "metrics"
    },
    "ecs": {
      "version": "8.0.0"
    },
    "elastic_agent": {
      "id": "a8435320-1d60-49f7-b1e7-25ed5d5eb59c",
      "snapshot": false,
      "version": "8.13.3"
    },
    "event": {
      "agent_id_status": "verified",
      "dataset": "system.process.summary",
      "duration": 16289873,
[...]

PR #2670, opened 2026-07-14
[RFC 0056] Propose host.tags as a free-form keyword field

adrianchen-es:
#### 1. What does this PR do?

Adds a new Proposal RFC (0056) for `host.tags`, a free-form keyword array field for tagging hosts with one or more user-configurable labels, used for search, alerting, and access control.

Addresses #2340.

**Update:** Reworked per review feedback. The field was originally proposed as `host.group`, a single scalar value. It's now `host.tags`, a keyword array, and the parallel `agent.group` field from the original issue has been dropped entirely. See the Concerns section of the RFC for the full reasoning.

- `host.tags` (new): user-configurable tags for classifying a host. Level: extended. Type: keyword array. Target maturity: alpha.

`group` collided in name with ECS's existing reusable `group` field set (`user.group.*`, `process.group.*`), and the field is free-form and user-defined rather than a formal membership construct, so `tags` fits better. It's modeled as an array rather than a scalar since a host commonly needs more than one classification at once (an environment tag, a team tag, and so on).

`agent.tags` is not proposed. In the common case an agent's tags and its host's tags would always be identical, and the one case where they'd genuinely differ (an agent collecting for hosts other than itself) is already `observer.*`'s job, not `agent`'s. Details in the RFC's Concerns section.

Teams have no standardized field today for tagging hosts by application, environment, or business unit, and fall back to ad hoc custom fields with inconsistent naming across integrations. `host.tags` gives Security and Observability a single, consistent field to filter, alert, and grant access on by tag.

N/A. Documentation is within the RFC itself and the schema YAML (`rfcs/text/0056/host.yml`).

NO. This PR adds only an RFC and proposed schema YAML, no changes to `schemas/` or `generated/`.

NO. RFC-only change, no schema or code modifications.

- Target maturity is alpha. Nothing currently populates the field, and Fleet would need a new policy-level setting to assign tags at enrollment.
- Elastic Defend does not inherit generic Fleet Agent Policy field injection the way Beats-based integrations do, so populating `host.tags` for Defend-originated data needs separate confirmation with the Endpoint team. Flagged in Scope of impact.
- Worth a maintainer read: the RFC calls out that `host.tags` being multi-valued is a real tradeoff for the ABAC use case, since matching "at least one tag present" is weaker than matching a single canonical value. If per-host access control needs an authoritative single value, that may still be worth its own field later.
- Also worth confirming: whether an entity-scoped keyword-array tags field is a pattern the team wants to establish, since the closest existing precedent (`container.labels`) is a key/value object, not an array.

---

#### Commit Message

```
[RFC 0056] Propose host.tags as a free-form keyword array field

Add a new Proposal RFC for host.tags, letting users attach one or
more arbitrary tags to a host for search, alerting, and access
control. Reworked from an earlier host.group scalar field per review
feedback: group collided with ECS's existing reusable group field
set, and the free-form, user-defined nature of the field fits tags
better. Drops the agent.tags/agent.group field proposed in the
originating issue since agent and host would carry the same value in
the common case, and ECS already scopes asset-level classification
fields to host/user rather than agent.
```


rfcs/text/0056-host-group.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Does `group` collide with the existing reusable `group` field set?** Not in the schema, but the name overlaps in a confusing way. ECS already has a reusable `group` field set (`schemas/group.yml`) for OS/directory-level groups: an object with `id`, `name`, and `domain`, reused today as `user.group.*`, `process.group.*`, `process.real_group.*`, and others. `host.group` is a different concept: a flat, user-assigned deployment label, not OS group membership. `host` isn't in that field set's `expected` list, so the paths never collide and there's no mapping conflict. What's left is a naming question: someone searching for "group" fields might expect the `id`/`name`/`domain` shape instead of a plain string. `group` matches the term used in the originating issue, and the two concepts sit on separate paths, but a more specific name like `host.deployment_group` may be worth weighing against that familiarity.
```
trisch-me: this is something where we should come with another solution. Group is an existing namespace so it will be indistinguishable from the look and despite ES currently supports both scalar and object with the same name, we can’t mix them in the standard. If we would even accept it - we would never be able to reuse group as a namespace anymore.

~~Have you weighted if we can reuse existing `group` namespace? From the examples it looks likes it just a string, do we have options where it might be an object with id?~~

actually scratch that, I don’t think `group` namespace would be a good solution

adrianchen-es: What if we made it more specific - e.g. in the quoted section from my RFC? 
Where the new field is `host.deployment_group`? host.group was shorter.

trisch-me: is it really a `delpoyment_group`? Based on description it feels like it could be some generic group, not specifically related to deployment

adrianchen-es: It wouldn't be specific to deployment groups. But that way it would avoid using a term already present and causing confusion. Unless I understood the group namespace aspect?

taylor-swanson: Would something like `host.tags` or `host.labels` be better? I agree with @trisch-me that `host.group` would clash with the group `fieldset` (even though it's not used on `host`). It seems like this is a free-form, user-defined field, so I feel `tags` or `labels` would be a better fit. Otherwise, if we call it something like `deployment_groups`, it has to refer to that.

adrianchen-es: I can rejig the RFC on host.tags ?

Fits the keyword nature in my mind

taylor-swanson: That sounds good to me, how about you, @trisch-me?

adrianchen-es: Update to be `host.tags`

rfcs/text/0056-host-tags.md
```diff
@@ -0,0 +1,219 @@
...
+
+- Stage: **Proposal**
+- Date: **TBD**
+- Target maturity: **alpha**
+
+## Summary
+
+Add `host.tags`: a free-form keyword field that lets users attach one or more arbitrary tags to a host (e.g. `APP-A_PROD`, `team-platform`) for search, alerting, and access control across Security and Observability. It's the host-scoped counterpart to the existing root `tags` field: where `tags` describes the event, `host.tags` describes the host asset itself, and it propagates anywhere `host` is reused (e.g. `host.target.tags`).
+
+An earlier draft of this RFC proposed `host.group` as a single scalar value, and the originating issue also proposed a parallel `agent.group`. Both are dropped here. Reviewers pointed out that `group` collides in name with ECS's existing reusable `group` field set, and that a free-form, user-defined value fits the existing `tags` naming better than an implied membership construct. See [Concerns](#concerns) for the full reasoning, including whether this is redundant with the root `tags` field.
```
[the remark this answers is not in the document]
adrianchen-es: Noted, updating.

rfcs/text/0056/host.yml
```diff
@@ -0,0 +1,21 @@
...
+        a team or application tag.
+
+        These values are expected to be assigned by the entity that
+        manages the host's monitoring configuration (e.g. a Fleet
+        policy or configuration management system), not derived from
+        data on the host itself.
+      example: '["APP-A_PROD", "team-platform"]'
+      normalize:
+        - array
+      alpha: This field is alpha and subject to change.
```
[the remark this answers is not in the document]
adrianchen-es: Fixed in a1bd9f04. Added `synthetic_source_keep: "none"` to match the root `tags` field and other unordered keyword-array fields in ECS.

rfcs/text/0056-host-tags.md
```diff
@@ -0,0 +1,219 @@
...
+        a team or application tag.
+
+        These values are expected to be assigned by the entity that
+        manages the host's monitoring configuration (e.g. a Fleet
+        policy or configuration management system), not derived from
+        data on the host itself.
+      example: '["APP-A_PROD", "team-platform"]'
+      normalize:
+        - array
+      alpha: This field is alpha and subject to change.
```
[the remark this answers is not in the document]
adrianchen-es: Fixed in a1bd9f04. Synced the embedded snippet with rfcs/text/0056/host.yml.

adrianchen-es:
@trisch-me  or @taylor-swanson any feedback?

taylor-swanson (approved):
This looks good to me.

@qcorporation, @trisch-me is out, could you take a look at this?

trisch-me (approved):

PR #2671, opened 2026-07-23
feat: Detect type conflicts when adding intermediate fields

taylor-swanson:
#### 1. What does this PR do?

- When loading schemas, the scripts can now detect situations where either an intermediate field overwrites another field of a different type, or when a leaf field overwrites an intermediate field of a different type.

- Not applicable

- Catches situations where type conflicts will cause an invalid mapping exception to occur in Elasticsearch.

- N/A, internal tooling improvement. Exception produced is self-documenting.

```
ValueError: Type conflict detected when adding intermediate field 'domain' (adding: object, existing: keyword)

ValueError: Type conflict detected when adding leaf field 'domain' (adding: keyword, existing: object)
```

- No changes to schema

- New tests added, all passing.

- N/A

---

#### Commit Message

- When loading schemas, the scripts can now detect situations where either an intermediate field overwrites another field of a different type, or when a leaf field overwrites an intermediate field.

---

## Related

- Closes elastic/security-team#17913


scripts/schema/loader.py
```diff
@@ -189,17 +189,23 @@ def nest_fields(field_array: List[Field]) -> Dict[str, Dict[str, FieldEntry]]:
             # Respect explicitly defined object fields
             if 'type' in field_details and field_details['type'] in ['object', 'nested']:
                 field_details.setdefault('intermediate', False)
-            else:
+            elif 'type' not in field_details:
                 field_details.setdefault('type', 'object')
                 field_details.setdefault('name', '.'.join(parent_fields[:idx + 1]))
                 field_details.setdefault('intermediate', True)
+            else:
+                msg = f"Type conflict detected when adding intermediate field '{level}' (adding: object, existing: {field_details['type']})"
```
trisch-me, line 197: so this basically triggers when `'type' in field_details` but type is not object?

taylor-swanson: Sorry, missed this comment somehow.

That's the idea. It's to prevent situations where something like this is defined:

```yaml
- name: foo.bar
  type: keyword
- name: foo
  type: keyword
```

`foo` gets created as an "intermediate" object field but then the second `foo` gets added as a keyword.

trisch-me: Yes, I wanted to unwind the if structure
```
if 'type' in field_details and field_details['type'] in ['object', 'nested’]:
elif 'type' not in field_details:
else: <-- type IS in field_details but it’s not object
```

trisch-me (approved):

PR #2680, opened 2026-08-04
[RFC 0057] Propose service.namespace field

adrianchen-es:
#### 1. What does this PR do?

Adds a new Proposal RFC (0057) for `service.namespace`, a keyword field that groups related `service.name` values under a common owner, such as a team, product line, or business unit.

- `service.namespace` (new): groups services by owning team, product, or business unit. Level: extended. Type: keyword. Target maturity: beta.

Mirrors OpenTelemetry's stable `service.namespace` resource attribute. ECS already has equivalents for the other two members of OTel's service identity triplet (`service.name`, and `service.node.name` for `service.instance.id`), so this closes the remaining gap. Because it's added to the base `service` group, it's also automatically available at `service.origin.namespace` and `service.target.namespace`.

Teams have no standard way to group services by ownership today. The closest existing field, `service.type`, describes what kind of service something is (`elasticsearch`, `nginx`), not who owns it, and one namespace commonly contains several different `service.type` values. Some teams already repurpose `service.type` for grouping as a workaround when it isn't otherwise in use, but that breaks the moment they also need to distinguish actual service types. `service.namespace` gives the grouping use case, and OTel/EDOT producers that already emit this resource attribute, a proper field to land in.

N/A. Documentation is within the RFC itself and the schema YAML (`rfcs/text/0057/service.yml`).

NO. This PR adds only an RFC and proposed schema YAML, no changes to `schemas/` or `generated/`.

NO. RFC-only change, no schema or code modifications.

- Target maturity is beta. The field's shape and meaning are a direct match to a stable OTel attribute, not a new concept, so the open question is integration coverage rather than definition. Same reasoning RFC 0055 used for `vulnerability.status`.
- The Concerns section addresses the `service.type` workaround directly, plus why `group.*` doesn't fit (scoped to user/process, not service), why this doesn't overlap with `service.environment`, and the naming similarity to (but non-conflict with) `data_stream.namespace`.
- No existing Elastic integration populates this field yet. Source data examples show three realistic ways it could be populated: OTel/EDOT resource attributes directly, a Kubernetes namespace copied over by a pipeline processor, and an AWS ECS task tag.

---

#### Commit Message

```
[RFC 0057] Propose service.namespace field

Adds a Proposal RFC for service.namespace, a keyword field grouping
related service.name values by team, product line, or business unit.
Mirrors OpenTelemetry's stable service.namespace resource attribute,
closing the gap left by ECS already having equivalents for the other
two identity attributes (service.name, and service.node.name for
service.instance.id).

Addresses the recurring ask to group services meaningfully without
repurposing service.type, which describes what kind of service
something is rather than who owns it.
```

### DIFF
--- a/rfcs/text/0057/service.yml
+++ b/rfcs/text/0057/service.yml
@@ -6,7 +6,7 @@
       type: keyword
       short: Namespace grouping service.name values by team, product, or business unit.
       beta: This field is beta and subject to change.
-      example: shop
+      example: elastiflix
       description: >
         A namespace for `service.name`.
 
@@ -18,10 +18,21 @@
         than by individual name, and disambiguates `service.name` values
         that are reused across different groups.
 
-        `service.namespace` is independent of `service.environment`.
-        Namespace identifies who owns a service; environment identifies
-        which deployment stage a running instance is in. The two can be
-        combined, for example the same namespace running in both
-        `production` and `staging`.
+        `service.namespace` is different from `service.type`. Take the
+        Elasticsearch cluster behind a streaming application called
+        Elastiflix: `service.namespace` is `elastiflix`, `service.name`
+        is `elastiflix-es`, and `service.type` is `elasticsearch`. The
+        API and web frontend that make up the rest of Elastiflix share
+        that same `service.namespace`, but carry their own `service.name`
+        and `service.type` values. Namespace says which application or
+        team a service belongs to, type says what the service is built
+        on, and one `service.type` such as `elasticsearch` can appear
+        under many different namespaces.
+
+        `service.namespace` is also independent of `service.environment`.
+        The Elastiflix Elasticsearch cluster above could have
+        `service.environment` set to `development`, and the same
+        `service.namespace` and `service.name` would apply just as well
+        if that cluster were running in `production`.
       otel:
         - relation: match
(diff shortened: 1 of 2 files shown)


trisch-me (approved):
Thank you for adding it, LGTM

rfcs/text/0057/service.yml
```diff
@@ -0,0 +1,27 @@
...
+
+        Namespace distinguishes a group of related services, for example
+        the team, product line, or business unit that owns them. Unlike
+        `service.name`, which identifies one service, `service.namespace`
+        identifies the group it belongs to. This lets services be
+        filtered, dashboarded, or access-controlled by owning group rather
+        than by individual name, and disambiguates `service.name` values
+        that are reused across different groups.
+
+        `service.namespace` is independent of `service.environment`.
```
trisch-me: could you please add also difference to the `service.type` with examples?

adrianchen-es: Added a worked example using Elastiflix. On the Elasticsearch cluster behind it: `service.namespace: elastiflix`, `service.name: elastiflix-es`, `service.type: elasticsearch`, `service.environment: development`. The API and web frontend share the same `service.namespace` but get their own `service.name`/`service.type`, which is what separates namespace (which application/team) from type (what the service is built on). Updated in both the field description and the Concerns section.

adrianchen-es:
@trisch-me , anything else you wanted addressed?

trisch-me:
@taylor-swanson could you please check this PR

taylor-swanson (approved):

